PR #2, opened 2016-06-14
mio: Retry poll if interrupted

kamalmarhubi:
Apparently on Linux, epoll_wait can return EINTR if a ptracer
attaches. The SIGSTOP is suppressed, but the call is still interrupted.
This retries the call to make it easier to do things like observe what's
going on with strace.


alexcrichton:
ugh EINTR

PR #5, opened 2016-07-01
First stab at mio integration for alloc-free futures

aturon:
Here's a first, quick attempt at mio reintegration.

It mostly matches what we discussed, in terms of using global lock-free token allocation, and a hashmap for dispatch.

The interesting new piece is trying to tease out the readiness vs completion story. I introduced a notion of "ReadinessFuture", which is one where basically `poll` is meaningless but `schedule` waits for readiness. Then there's a completion layer that lets you couple a readiness future with a custom poll which will attempt completion.

In principle, this should allow us to expose readiness futures at the lowest levels, and then pair them up with completions to give you various higher-level APIs (without being locked into those).

This should also be useful for things like writing a proxy server, where you might want to *join* a pair of readiness futures (one for reading from one socket, one for writing to another), and then complete with an action that uses both sockets.

Curious what you think of this factoring!

### DIFF
--- a/mio/src/lib.rs
+++ b/mio/src/lib.rs
@@ -16,7 +16,57 @@ use fnv::FnvHasher;
 
 use futures::{Future, Tokens, promise, Complete, Wake, PollError, PollResult};
 
-pub type IoFuture<T> = Future<Item=T, Error=io::Error>;
+pub struct MioEvent {
+    events: mio::EventSet,
+    source: Source,
+    loop_handle: LoopHandle,
+}
+
+impl MioEvent {
+    fn into_future<C: PollCompletion>(self, completion: C) -> MioFuture<C> {
+        MioFuture {
+            event: self,
+            completion: completion,
+        }
+    }
+}
+
+pub trait PollCompletion {
+    type Item: Send + 'static;
+    type Error: Send + 'static;
+
+    fn poll_completion(&mut self) -> Option<PollResult<Self::Item, Self::Error>>;
+}
+
+pub struct MioFuture<C> {
+    event: MioEvent,
+    completion: C,
+}
+
+impl<C> Future for MioFuture<C> where C: PollCompletion + Send + 'static {
+    type Item = C::Item;
+    type Error = C::Error;
+
+    fn poll(&mut self, tokens: &Tokens) -> Option<PollResult<C::Item, C::Error>> {
+        // TODO: filter by tokens
+        self.completion.poll_completion()
+    }
+
+    fn schedule(&mut self, wake: Arc<Wake>) {
+        // TODO: record token
+        self.event.loop_handle.schedule(Interest {
+            waiter: wake,
+            source: self.event.source.clone(),
+            events: self.event.events,
+            first_time: false, // assume already registered
+        });
+    }
+
+    fn tailcall(&mut self) -> Option<Box<Future<Item=Self::Item, Error=Self::Error>>> {
+        None
+    }
+}
+
 
 #[derive(Clone)]
 pub struct TcpListener {
@@ -77,40 +127,6 @@ impl Future for TcpListenerAccept {
 pub struct TcpStream {
     tcp: Arc<mio::tcp::TcpStream>,
     loop_handle: LoopHandle,
-}
-
-pub struct TcpStreamReady {
-    events: mio::EventSet,
-    inner: TcpStream,
-}
-
-impl Future for TcpStreamReady {
-    type Item = ();
-    type Error = io::Error;
-
-    fn poll(&mut self, tokens: &Tokens) -> Option<PollResult<(), io::Error>> {
-        Some(Ok(()))
-    }
-
-    fn schedule(&mut self, wake: Arc<Wake>) {
-        // TODO: record token
-        self.inner.loop_handle.schedule(Interest {
-            waiter: wake,
-            source: self.inner.tcp.clone(),
-            events: self.events,
-            first_time: false,
-        });
-    }
-
-    fn tailcall(&mut self) -> Option<Box<Future<Item=Self::Item, Error=Self::Error>>> {
-        None
-    }
-}
-
-impl ReadinessFuture for TcpStreamReady {
-    fn should_poll(&mut self, tokens: &Tokens) -> bool {
-        true // TODO
-    }
 }
 
 pub struct ReadCompletion {
@@ -188,41 +204,36 @@ impl TcpStream {
         self.tcp.peer_addr()
     }
 
-    pub fn ready_to_read(&self) -> TcpStreamReady {
-        TcpStreamReady {
+    pub fn ready_to_read(&self) -> MioEvent {
+        MioEvent {
             events: mio::EventSet::readable(),
-            inner: self.clone(),
+            source: self.tcp.clone(),
+            loop_handle: self.loop_handle.clone(),
         }
     }
 
-    pub fn ready_to_write(&self) -> TcpStreamReady {
-        TcpStreamReady {
+    pub fn ready_to_write(&self) -> MioEvent {
+        MioEvent {
             events: mio::EventSet::writable(),
-            inner: self.clone(),
+            source: self.tcp.clone(),
+            loop_handle: self.loop_handle.clone(),
         }
     }
 
     // TODO: wrap in newtype
-    pub fn read(&self, mut into: Vec<u8>) -> IntoCompletion<TcpStreamReady, ReadCompletion> {
-        IntoCompletion {
-            readiness: self.ready_to_read(),
-            completion: ReadCompletion {
-                tcp: self.tcp.clone(),
-                into: Some(into)
-            }
-        }
+    pub fn read(&self, mut into: Vec<u8>) -> MioFuture<ReadCompletion> {
+        self.ready_to_read().into_future(ReadCompletion {
+            tcp: self.tcp.clone(),
+            into: Some(into)
+        })
     }
 
-    pub fn write(&self, offset: usize, data: Vec<u8>)
-                 -> IntoCompletion<TcpStreamReady, WriteCompletion> {
-        IntoCompletion {
-            readiness: self.ready_to_write(),
-            completion: WriteCompletion {
-                tcp: self.tcp.clone(),
-                offset: offset,
-                data: Some(data),
-            }
-        }
+    pub fn write(&self, offset: usize, data: Vec<u8>) -> MioFuture<WriteCompletion> {
+        self.ready_to_write().into_future(WriteCompletion {
+            tcp: self.tcp.clone(),
+            offset: offset,
+            data: Some(data),
+        })
     }
 }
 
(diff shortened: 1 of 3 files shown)


aturon:
cc @carllerche

aturon:
Oh, and I should say that if we do want the whole readiness/completion concepts, obviously it should move into the futures library proper.

mio/src/lib.rs
```diff
@@ -1,111 +1,129 @@
...
-    done: HashMap<usize, Complete<(), io::Error>>,
-}
-
-enum Message {
-    Wait(Complete<(), io::Error>, mio::EventSet, Arc<mio::Evented + Send + Sync>),
-    Register(Arc<mio::Evented + Send + Sync>),
-}
+pub type IoFuture<T> = Future<Item=T, Error=io::Error>;
 
+#[derive(Clone)]
```
alexcrichton: In the long run we probably don't want `impl Clone for TcpListener` I think, it seems like it could get into some uncomfortably weird situations if you can silently clone an I/O object. (the actual impl details here look good to me though)

mio/src/lib.rs
```diff
@@ -1,111 +1,129 @@
...
 
-    pub fn accept(&self) -> Box<IoFuture<(TcpStream, SocketAddr)>> {
-        match self.tcp.accept() {
-            Err(e) => return futures::failed(e).boxed(),
+    pub fn accept(&self) -> TcpListenerAccept {
+        TcpListenerAccept { inner: self.clone() }
+    }
+}
+
+impl Future for TcpListenerAccept {
```
alexcrichton: You know I wonder if we could get this to work:

```rust
impl IoFuture<(TcpStream, SocketAddr)> for TcpListenerAccet {

   // lack of `type Foo = ...;`

   fn poll(..) 
}
```

aturon:
So, thinking more, I don't think that readiness actually makes sense to expose directly as a `Future`. Once you start thinking about how you'd want to use the combinators to build something useful on top of a future exposing readiness, it goes off the rails a bit -- basically, you *have* to introduce some new, readiness-specific combinator to make it work. And many of the normal combinators (like `and_then`) just don't make sense for pure readiness.

I think we should instead consider exposing a first-class readiness abstraction, which supports `schedule` and provides at least two combinators: `select` and `join`. Given one of these readiness things (whatever we call it), you can construct a full-fledged future by pairing it with a `poll` implementation -- but you have to be careful to ensure that the events of interest are perfectly matched between the `schedule` and `poll`, to avoid lost wakeups.

Note that `join` is particularly useful for things like building a proxy, where you want a readiness signal when both a read and a write event are ready -- and *then* you try to complete. If you fail to complete, you want to loop back again, using the `schedule` from the read and write events.

Right now, the code sort of contains these abstractions, but they're awkwardly built *on top* of Futures. I'm increasingly feeling like the right thing to do is actually expose a readiness abstraction *under* Futures -- possibly via a `Schedule` super-trait?

aturon:
OK, I pushed an update to this that cleans things up quite a bit.

As I was arguing above, I'm moving away from using `Future`s to directly signal readiness. But instead of trying to introduce a *completely* generic readiness abstraction right now, I just introduced `MioEvent` as an abstraction for readiness around leaf mio events -- i.e., Evented things.

Given a `MioEvent`, you can get a `MioFuture` by pairing it with a method that attempts to complete (basically like `poll`, but without any tokens).

This all works like a charm, and makes it possible to wrap the base readiness events in custom leaf futures (that expose various buffering strategies, etc).

The thing that's missing here is any notion of compositionality at the `MioEvent` level (or perhaps an even more general version of the abstraction). That *might* be something we want eventually, but we can cross that bridge later.

aturon:
I want to explore using TLS to talk to the event loop fairly soon -- that's an essential enough piece, with enough unknowns, that I'd like to get it squared away early.

mio/src/lib.rs
```diff
@@ -198,39 +283,15 @@ impl Loop {
...
-            Ok(e) => Ok(e),
-            Err(PollError::Other(e)) => Err(e),
-            Err(PollError::Panicked(p)) => panic::resume_unwind(p),
-            Err(PollError::Canceled) => panic!("canceled"),
-        }
-    }
-
-    fn _await(&mut self, done: &mut FnMut() -> bool) {
-        while !done() {
+    pub fn run(&mut self) {
```
alexcrichton: It may be worth getting `cargo test` to work in this directory perhaps to see how this hooks up with driving a future

mio/src/lib.rs
```diff
@@ -115,77 +149,128 @@ impl TcpStream {
...
+}
+
+type Waiter = Arc<Wake>;
+type Source = Arc<mio::Evented + Send + Sync>;
+
+pub struct Loop {
+    io: mio::Poll,
+    tx: mio::channel::Sender<Message>,
+    rx: mio::channel::Receiver<Message>,
+    dispatch: HashMap<usize, Waiter, BuildHasherDefault<FnvHasher>>,
```
alexcrichton: we probably actually want to go with a hasher here that looks like:

```rust
struct DumbHasher(usize);

impl Hasher for DumbHasher {
    fn write(&mut self, bytes: &[u8]) { panic!() }
    fn write_usize(&mut self, u: usize) { self.0 = u; }
    fn finish(&self) -> u64 { self.0 as u64 }
}
```

That should be the "fastest possible hash" for this where the keys are literally just indexes.

mio/src/lib.rs
aturon: Yep. Had limited time so wanted to get a draft in front of you first, but once you're OK with the general strategy here I'll wrap up the final bits.

mio/src/lib.rs
```diff
@@ -1,111 +1,145 @@
...
+    events: mio::EventSet,
+    source: Source,
+    loop_handle: LoopHandle,
 }
 
-enum Message {
-    Wait(Complete<(), io::Error>, mio::EventSet, Arc<mio::Evented + Send + Sync>),
-    Register(Arc<mio::Evented + Send + Sync>),
+impl MioEvent {
+    fn into_future<C: PollCompletion>(self, completion: C) -> MioFuture<C> {
```
alexcrichton: This seems somewhat similar to:

```rust
impl MioEvent {
    fn as_future<F, T, E>(self, f: F) -> MioFuture<F> 
        where F: FnMut(&Tokens) -> Option<PollResult<T, E>>
    {
        // ...
    }
}
```

That is, I'm not sure if the `PollCompletion` trait is buying us much over `FnMut`?

mio/src/lib.rs
```diff
@@ -1,111 +1,145 @@
...
+    type Item: Send + 'static;
+    type Error: Send + 'static;
+
+    fn poll_completion(&mut self) -> Option<PollResult<Self::Item, Self::Error>>;
 }
 
+pub struct MioFuture<C> {
+    event: MioEvent,
+    completion: C,
+}
```
alexcrichton: I think it definitely makes sense to have a central bundle of:

* What we're waiting on
* What we're waiting for
* How we know when we're done

and all that packaged up is a future

mio/src/lib.rs
```diff
@@ -1,111 +1,145 @@
...
 impl TcpListener {
     pub fn local_addr(&self) -> io::Result<SocketAddr> {
         self.tcp.local_addr()
     }
 
-    pub fn accept(&self) -> Box<IoFuture<(TcpStream, SocketAddr)>> {
-        match self.tcp.accept() {
-            Err(e) => return futures::failed(e).boxed(),
+    pub fn accept(&self) -> TcpListenerAccept {
+        TcpListenerAccept { inner: self.clone() }
```
alexcrichton: Shouldn't this return a `MioFuture` as well?

mio/src/lib.rs
alexcrichton: lol we can't name closures nevermidn

mio/src/lib.rs
```diff
@@ -115,77 +149,128 @@ impl TcpStream {
...
-            }
-            Err(mio::channel::SendError::Io(e)) => {
-                return futures::failed(Error::new(e, into)).boxed()
-            }
-            Err(mio::channel::SendError::Disconnected(..)) => panic!("closed channel"),
-        }
     }
+}
+
+pub struct WriteCompletion {
```
alexcrichton: You probably feel the same way, but as many of these as possible should probably be private

mio/src/lib.rs
```diff
@@ -115,77 +149,128 @@ impl TcpStream {
...
+impl TcpStream {
+    pub fn local_addr(&self) -> io::Result<SocketAddr> {
+        self.tcp.local_addr()
+    }
+
+    pub fn peer_addr(&self) -> io::Result<SocketAddr> {
+        self.tcp.peer_addr()
+    }
+
+    pub fn ready_to_read(&self) -> MioEvent {
```
alexcrichton: I've wrestled in the past if we want an interface like this, and I think we may in the long run but we may want to keep it internal for awhile. Without a `Read` and `Write` implementation I'm not sure these'll be too useful though in isolation :(

mio/src/lib.rs
aturon: No -- I think you only want that split in the case that there's more than one `Future` interface you want to provide. So for `TcpStream`, exposing `MioEvent`s means that we can easily build up a lot of different `Future` abstractions over the core event, e.g. for different buffering strategies. But that seems much less useful for listeners -- the existing `Future` interface (or, ultimately, a stream interface) should be enough to build anything else you might want.

mio/src/lib.rs
```diff
@@ -246,49 +307,66 @@ impl Loop {
...
+    fn register_(&mut self, source: Source) {
+        self.io.register(&*source, mio::Token(0), mio::EventSet::none(), mio::PollOpt::empty());
+    }
+
+    fn schedule_(&mut self, token: usize, interest: Interest) {
+        let Interest { waiter, source, events, first_time } = interest;
+        let old = self.dispatch.insert(token, waiter);
+        debug_assert!(old.is_none());
+
+        // TODO handle failure
```
alexcrichton: I think the way we'll end up handling failure here will look like:

* See the error
* Store the error with the token somewhere
* Call the `waiter` callback

Communicating to the original future efficiently that it should check the storage may be... interesting. Just need to figure out how to get it off the fast path though, I don't think this will come up much in practice.

mio/src/lib.rs
```diff
@@ -316,21 +394,66 @@ impl Loop {
...
+        next
+    }
+}
+
+// TODO: use TLS to avoid sending messages
+impl LoopHandle {
+    fn register(&self, source: Source) {
+        self.tx.send(Message::Register(source))
+            .map_err(|_| ())
+            .expect("failed to send register message") // todo: handle failure
```
alexcrichton: This probably wants to get punted upwards immediately, and that way the futures themselves can squirrel away this error, immediately call the wake callback, and then return the error on the next call to `poll`

mio/src/lib.rs
```diff
@@ -316,21 +394,66 @@ impl Loop {
...
+    fn schedule(&self, interest: Interest) -> usize {
+        let token = self.tok.next_token();
+        self.tx.send(Message::Schedule(token, interest))
+            .map_err(|_| ())
+            .expect("failed to send schedule message"); // TODO: handle failure?
+        token
+    }
+
+    fn deschedule(&self, token: usize) {
+        unimplemented!()
```
alexcrichton: may be worth prototyping this out sooner rather than later

mio/src/lib.rs
alexcrichton: Yeah everything's looking good to me, what I expected!

mio/src/lib.rs
alexcrichton: I guess at least in terms of an implementation detail though it should use a `MioFuture`?

aturon: Ah yeah, you're right, we can factor a bit more code out that way.

aturon:
Closing this out. Will open a fresh PR with the new version.

PR #6, opened 2016-07-11
Mio reintegration

aturon:
Here's the second stab at mio integration.

- Talks to event loop via TLS when possible.
- Uses synchronization with the event loop (via futures) to acquire tokens, therefore allowing us to use a simple slab instead of a hashmap.
- Generic `ReadinessStream` abstraction: a stream of `()` values to signify readiness, which in general can include spurious notifications.
- A `TcpListener`, which wraps up a readiness stream with the ability to accept incoming connections.
- `tcp_connect`, which acquires a `TcpStream`

I've also updated the `main.rs` to use these features. Right now, that means building directly on top of readiness streams (and hence using `read` directly). The plan from here is to build a simple tokio-like implementation of the /plaintext test, which should point the way toward using additional stream combinators etc. internally to structure things. Note, however, that you always have the option of continuing to work in "manual mode", basically pulling together readiness streams and IO actions into your own custom scheduler/dispatch within the futures framework. That means we have a great deal of flexibility in finding the right factoring.

Finally, adds a few stream combinators along the way.


src/stream/mod.rs
```diff
@@ -133,4 +146,18 @@ pub trait Stream: Send + 'static {
...
     }
+
+    fn skip_while<P>(self, pred: P) -> SkipWhile<Self, P>
+        where P: FnMut(&Self::Item) -> Result<bool, Self::Error> + Send + 'static,
+              Self: Sized,
+    {
+        skip_while::new(self, pred)
+    }
+
+    fn for_each<F>(self, f: F) -> ForEach<Self, F>
```
alexcrichton: I wonder if perhaps this would be best done as a `fold` for now?

src/stream/mod.rs
```diff
@@ -133,4 +146,18 @@ pub trait Stream: Send + 'static {
     {
         flat_map::new(self)
     }
+
+    fn skip_while<P>(self, pred: P) -> SkipWhile<Self, P>
+        where P: FnMut(&Self::Item) -> Result<bool, Self::Error> + Send + 'static,
```
alexcrichton: Oh I might actually expect this to return `bool` rather than `Result<bool, Self::Error>`

mio/src/tcp.rs
```diff
@@ -0,0 +1,82 @@
...
+                            Err(e) => {
+                                if e.kind() == ErrorKind::WouldBlock {
+                                    Ok(true)
+                                } else {
+                                    Err(e)
+                                }
+                            }
+                        }
+                    });
+                    let connected = connected.into_future();
```
alexcrichton: ok that's super cool

src/stream/mod.rs
aturon: Yeah, so we need to talk about the interaction between the combinators and `Result`...

alexcrichton:
Ok I haven't taken a peek at event_loop and the TCP stuff *too* closely yet, but I'm pretty confident that it's better than what's there now!

I want to start working towards removing `PollError` which'll just cause a bunch of conflicts with this, so I'm gonna merge. I'm sure I'll have more comments about this, though.

mio/src/event_loop.rs
```diff
@@ -0,0 +1,342 @@
...
+use std::io::{self, ErrorKind};
+use std::sync::Arc;
+use std::sync::atomic::{AtomicUsize, ATOMIC_USIZE_INIT, Ordering};
+use std::sync::mpsc;
+
+use mio;
+use slab::Slab;
+use futures::{Future, Tokens, Wake, PollResult};
+
+pub type Waiter = Arc<Wake>;
```
alexcrichton: `"Waiter".len()` vs `"Arc<Wake>".len()` 

hm...

:)

mio/src/event_loop.rs
```diff
@@ -0,0 +1,342 @@
...
+    }
+
+    pub fn handle(&self) -> LoopHandle {
+        LoopHandle {
+            id: self.id,
+            tx: self.tx.clone(),
+        }
+    }
+
+    pub fn run<F: Future>(self, f: F) -> Result<F::Item, F::Error> {
```
alexcrichton: s/self/&mut self/

or at least that's what I'd expect

alexcrichton: (maybe &self)

mio/src/event_loop.rs
```diff
@@ -0,0 +1,342 @@
...
+            // attaching strace, or similar.
+            loop {
+                match self.io.borrow_mut().poll(None) {
+                    Ok(a) => {
+                        amt = a;
+                        break;
+                    }
+                    Err(ref e) if e.kind() == ErrorKind::Interrupted => {}
+                    err @ Err(_) => {
+                        err.unwrap();
```
alexcrichton: s/unwrap/panic!()/ with a message

mio/src/event_loop.rs
aturon: That's what I had at first, but I got uneasy about the leftover state persisting across invocations. (Shutdown can happen with arbitrary stuff potentially in flight). Maybe it's ok, though!

mio/src/event_loop.rs
```diff
@@ -0,0 +1,342 @@
...
+            id: self.id,
+            tx: self.tx.clone(),
+        }
+    }
+
+    pub fn run<F: Future>(self, f: F) -> Result<F::Item, F::Error> {
+        let (tx_res, rx_res) = mpsc::channel();
+        let handle = self.handle();
+        f.then(move |res| {
+            handle.shutdown();
```
alexcrichton: oh lol I see `self` now, may be a bit aggressive though!

mio/src/event_loop.rs
```diff
@@ -0,0 +1,342 @@
...
+    }
+
+    fn add_source(&self, source: Source) -> usize {
+        let sched = Scheduled {
+            source: source,
+            reader: None,
+            writer: None,
+        };
+        let mut dispatch = self.dispatch.borrow_mut();
+        // TODO: handle out of space
```
alexcrichton: This... is a very good question, and one I haven't thought much about

mio/src/event_loop.rs
```diff
@@ -0,0 +1,342 @@
...
+            Message::Schedule(tok, dir, wake) => self.schedule(tok, dir, wake),
+            Message::Deschedule(tok, dir) => self.deschedule(tok, dir),
+            Message::Shutdown => self.active.set(false),
+        }
+    }
+}
+
+impl LoopHandle {
+    fn send(&self, msg: Message) {
+        let mut msg_dance = Some(msg);
```
alexcrichton: 💃

mio/src/event_loop.rs
```diff
@@ -0,0 +1,342 @@
...
+
+        if CURRENT_LOOP.is_set() {
+            CURRENT_LOOP.with(|lp| {
+                if lp.id == self.id {
+                    // Need to execute all existing requests first, to ensure
+                    // that our message is processed "in order"
```
alexcrichton: tricksy

mio/src/event_loop.rs
```diff
@@ -0,0 +1,342 @@
...
+                    return Some(Ok(id));
+                }
+            }
+        }
+
+        None
+    }
+
+    fn schedule(&mut self, wake: Arc<Wake>) {
+        if self.scheduled { return; }
```
alexcrichton: I think our contract is that we need to call the most recent `wake`, right?

mio/src/event_loop.rs
aturon: Slabs can expand, should be ok on that front.

mio/src/event_loop.rs
```diff
@@ -0,0 +1,342 @@
...
+        self.send(Message::Shutdown);
+    }
+}
+
+const ADD_SOURCE_TOKEN: usize = 0;
+
+pub struct AddSource {
+    loop_handle: LoopHandle,
+    source: Option<Source>,
+    id: Arc<AtomicUsize>,
```
alexcrichton: Hm this `Arc` is kinda painful...

mio/src/event_loop.rs
aturon: Argh, good point. That's... going to be a bit tricky...

Need to think a bit more overall about what's going on here. There may be a better way to factor this that allows us to sidestep this question.

PR #7, opened 2016-07-12
Transmit AddSource errors and handle schedule()

alexcrichton:
* Handle errors as part of `register` in the `add_source` method on the event
  loop
* Also ensure that multiple calls to `schedule` will always call the *last*
  `wake` callback scheduled.

Re-use the `Slot` abstraction from the futures crate to coordinate inserting
values and arranging callbacks.

### DIFF
--- a/mio/src/event_loop.rs
+++ b/mio/src/event_loop.rs
@@ -8,6 +8,8 @@ use mio;
 use slab::Slab;
 use futures::{Future, Tokens, Wake};
 
+use slot::{self, Slot};
+
 pub type Source = Arc<mio::Evented + Send + Sync>;
 
 static NEXT_LOOP_ID: AtomicUsize = ATOMIC_USIZE_INIT;
@@ -68,7 +70,7 @@ impl Scheduled {
 }
 
 enum Message {
-    AddSource(Source, Arc<AtomicUsize>, Arc<Wake>),
+    AddSource(Source, Arc<Slot<io::Result<usize>>>),
     DropSource(usize),
     Schedule(usize, Direction, Arc<Wake>),
     Deschedule(usize, Direction),
@@ -193,7 +195,7 @@ impl Loop {
         rx_res.recv().unwrap()
     }
 
-    fn add_source(&self, source: Source) -> usize {
+    fn add_source(&self, source: Source) -> io::Result<usize> {
         let sched = Scheduled {
             source: source,
             reader: None,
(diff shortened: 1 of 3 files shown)


alexcrichton:
r? @aturon 

Curious what you think of the strategy here, I'm not sure if it's readily applicable to the transmission of the error on `schedule`, `deschedule`, and `drop_source`, but I'd like to remove those `unwrap` calls before we start benchmarking as well. Transmission of these errors seems like they may have a bit of an impact on the "happy path".

mio/src/event_loop.rs
```diff
@@ -363,43 +364,47 @@ const ADD_SOURCE_TOKEN: usize = 0;
...
-
-        None
     }
 
     fn schedule(&mut self, wake: Arc<Wake>) {
-        if self.scheduled { return; }
-        self.scheduled = true;
-        self.loop_handle.add_source_(self.source.take().unwrap(), self.id.clone(), wake);
+        if let Some((ref result, ref mut token)) = self.result {
+            result.cancel(*token);
```
aturon: What about the case where the previous callback is invoked just before we cancel it? The wake we're about to register will then never be invoked, meaning we may never wake up to realize the id is ready.

alexcrichton: That should be fine because the `on_full` callback fires immediately if the slot is full. If the slot's been emptied then this is pseudo-undefined-behavior because you're using a future after it's done.

Does that make sense?

(slot is a bit of a wonky abstraction)

aturon:
OK, so yes I was missing that `on_full` doesn't consume the value. Nice!

re: the error transmission strategy: I would prefer to avoid this expense on every registration if at all possible. I can see some ways to do that. When we add a source, we have an opportunity to create some arbitrary shared state between the event loop and the IO futures. We also know that at most one future is using each "direction" of a source. So we can stash the error in some atomic (perhaps even just stashing the raw OS error suffices?), and immediately invoke the associated Wake. On the other side, when polling readiness, we can check the associated error slot and propagate.

If that sounds plausible to you, I'm happy to implement.

alexcrichton:
Yeah I was thinking that's basically what we'd have to do if we do anything at all. It's a good point that we only have two futures at most looking for readiness, means we can control the outcomes a little better.

It's a bit of a shame to do so many allocations for each I/O object created, but I think we can at least consolidate them all down to one and that's probably not too bad in the long run.

You feel good about this general strategy though? Basically having a slot which is filled in and then the future is responsible for communicating that error back at some point. I'd want to implement all of it (communicating errors) and do some benchmarking to ensure we're not losing too much perf, the overhead here may be slightly worrisome.

aturon:
SGTM

PR #23, opened 2016-08-09
Fix some typos in tutorial

killercup:


killercup:
I have not actually finished reading this, so I might add more stuff later :)

killercup:
Okay, I'm done. Awesome tutorial! I was a bit confused what the relationships were between the crates, but this cleared it up pretty quickly and the explanations of the internals are really good. I would love to see a few more examples on how to work with streams, but this will probably be better put in the docs of Stream itself or an `examples/` directory.

alexcrichton:
Oh man, thanks so much @killercup! Sorry about the coveralls comments, forgot to turn those off...

I still feel like there's a lot of ground to cover and improve here, this "tutorial" turned into more of a brain dump at the end. I like the suggestion though of working with streams more, I'll add an example or two.

PR #37, opened 2016-08-11
Fix links to poll method docs incorrectly going to Poll type

SimonSapin:
It looks like link names in Markdown are case-insensitive?


alexcrichton:
Interesting! I had no idea...

PR #38, opened 2016-08-11
2 small tests for fuse

MarkusJais:
I played with fuse and I hope I understand it correctly. Hope this tests are useful.


alexcrichton:
Thanks!

PR #39, opened 2016-08-11
Implement IntoFuture for tuples of IntoFutures

Stebalien:
This way, instead of having to call the correct `join$n` function, one can just call `(a, b, c, ...).into_future()` (or just pass the tuple to a function expecting an `IntoFuture`).

Note: There should probably also be an implementation on `(F,)` for completeness but that's not very useful outside of macros.

I understand if you want to postpone stuff like this till the library stabilizes but I thought I'd propose it just in case.

### DIFF
--- a/src/join.rs
+++ b/src/join.rs
@@ -2,7 +2,7 @@
 
 use std::mem;
 
-use {Future, Task, Poll};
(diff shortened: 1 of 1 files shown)


alexcrichton:
Whoa, that's a pretty slick trick!

PR #40, opened 2016-08-11
Update doc-comment refering to poll() returning None

SimonSapin:
A assume that in a previous revision the return type was `Option`?

By the way, please consider using semantic line breaks:
http://rhodesmill.org/brandon/2012/one-sentence-per-line/


alexcrichton:
Thanks!

PR #42, opened 2016-08-11
Fix links to promise function vs Promise type

SimonSapin:


alexcrichton:
Gee thanks markdown. 

Thanks though :)

SimonSapin:
These are only the issues I happened to notice while reading the tutorial, I haven’t done more of an audit than that…

PR #45, opened 2016-08-11
Make stream::Fuse return None after finishing, not NotReady

Stebalien:
And add a test. The test could probably be better but it works (and fails without this patch).


alexcrichton:
Hm yeah this differs from `Futures::fuse` but I think it otherwise makes more sense because it mirrors iterator's fuse method. Thanks!

PR #48, opened 2016-08-12
Fix typo in tutorial

Wyverald:


alexcrichton:
Thanks for the PR! This was actually intended though to show off how you can add trait bounds to the `Item` type of a future, rather than specify a particular type which the `Error =` syntax is doing. I wonder though if perhaps it's confusing to bound the type by the `Future` trait which is also being discussed? Maybe something like `Clone` would be a bit more ubiquitous and less confusing?

Wyverald:
Ah, indeed! I suppose my change wouldn't actually compile, would it :P

I agree that using a different trait would be less confusing. But then again, I wonder how important it is to show off this particular aspect of the Rust grammar, as it's not specific to the futures-rs crate.

In any case, I'll leave this decision to you guys. Great work, by the way!

PR #55, opened 2016-08-12
fixed some clippy warnings

llogiq:


src/select_all.rs
```diff
@@ -72,14 +72,14 @@ impl<A> Future for SelectAll<A>
     }
 
     fn schedule(&mut self, task: &mut Task) {
-        for f in self.inner.iter_mut() {
+        for f in &mut self.inner {
```
alexcrichton: I personally prefer `.iter_mut()` instead of `&mut foo`, could this stay the same?

llogiq:
Sure, can revert those once I had some sleep.

alexcrichton:
Thanks!

llogiq:
You're welcome. If I may ask: What's your rationale for preferring the explicit call instead of using `IntoIterator`?

alexcrichton:
I personally find `for foo in bar.iter()` much easier to read than `for foo in &bar`. If you're new to Rust then the first is pretty clear what's happening where the latter just has this sigil `&` that seems odd.

PR #56, opened 2016-08-13
Fix typo in join example

pravic:
Full example to check:

```rust
extern crate futures;
extern crate futures_mio;

fn main() {
	use futures::*;

	let a = finished::<u32, u32>(1);
	let b = finished::<u32, u32>(2);
	let pair = a.join(b);

	let res = pair.map(|(a, b)| {
	    assert_eq!(a, 1);
	    assert_eq!(b, 1); // <-- must be 2
	});

	let mut lp = futures_mio::Loop::new().unwrap();
	lp.run(res).unwrap();
}
```

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -680,7 +680,7 @@ pub trait Future: 'static {
     ///
     /// pair.map(|(a, b)| {
     ///     assert_eq!(a, 1);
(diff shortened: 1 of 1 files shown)


alexcrichton:
Thanks!

PR #64, opened 2016-08-15
Clarify the effect of chaining and_then/for_each

aidanhs:
The sentence I've added felt quite implicit to me unless

1. you make an educated guess given the surrounding text or
2. you fully understand the implications of the "This function can be used to run a unit of work when the next successful value on a stream is ready. The closure provided will be yielded a value when ready, and the returned future will then be run to completion to produce the next value on this stream." quote in the documentation for `and_then`


alexcrichton:
Sounds good to me, thanks!

PR #68, opened 2016-08-16
clarification about Stream::into_future()

dwrensha:
I had been confused about why the `Stream::into_future()` used `Option`. This patch adds a sentence to the documentation to clarify.


alexcrichton:
Thanks!

PR #72, opened 2016-08-17
Implement a Peekable stream adaptor

josephDunne:
This pull request adds a `Peekable` stream adaptor which exposes a `peek` method to any stream.

The peek method is the same as a poll except it returns a reference to the `Stream::Item` (if available). I found this useful in my project but I have some concerns so I thought I would open a pull request to discuss.

The implementation of `peek` calls poll and keeps a record of the returned value in a `Option` field. `Poll::NotReady`, `Poll::Ok(None)` and `Poll::Err(_)` are passed trough to the caller. Susequent calls to `peek` will not call `poll` if a item has already been peeked at.

My concern is that if the end user calls `peek` but then does nothing with the value i.e. doesn't call `poll` then its possible the stream will not make any progress. I'm not sure if this is a problem in any real use case but it could lead to hard to find stalls. Perhaps there is a better way to implement this?

### DIFF
--- a/src/stream/peek.rs
+++ b/src/stream/peek.rs
@@ -1,0 +1,62 @@
+use {Task, Poll};
+use stream::{Stream, Fuse};
+
+/// A `Stream` that implements a `peek` method.
+///
+/// The `peek` method can be used to retrieve a reference
+/// to the next `Stream::Item` if available. A subsequent
+/// call to `poll` will return the owned item.
+pub struct Peekable<S: Stream> {
+    stream: Fuse<S>,
+    peeked: Option<S::Item>,
+}
+
+
+pub fn new<S: Stream>(stream: S) -> Peekable<S> {
+    Peekable {
+        stream: stream.fuse(),
+        peeked: None
+    }
+}
+
+
+impl<S: Stream> Stream for Peekable<S> {
+    type Item = S::Item;
+    type Error = S::Error;
+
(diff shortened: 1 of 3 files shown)


src/stream/peek.rs
```diff
@@ -0,0 +1,63 @@
...
+        if self.peeked.is_some() {
+            return Poll::Ok(self.peeked.as_ref());
+        }
+        match self.poll(task) {
+            Poll::NotReady => Poll::NotReady,
+            Poll::Err(e) => Poll::Err(e),
+            Poll::Ok(None) => Poll::Ok(None),
+            Poll::Ok(Some(item)) => {
+                self.peeked = Some(item);
+                task.notify();
```
josephDunne: I suspect this extra notify is spurious..

I just don't know how to ensure a stream containing a peekable makes progress if the end user does not remember to call poll after every peek (or do something else which makes progress)

alexcrichton:
Thanks for the PR! Implementation looks good to me, just two comments:

* Yeah it should be safe to remove that call to `notify`, right now that primarily just happens during `schedule`.
* I think you may want to `.fuse()` the stream to ensure that we don't poll it too much. Right now if `peek()` discovered the stream is done a later `poll()` may call `poll()` again on the underlying stream, which generally leads to bad behavior.

josephDunne:
I have removed the notify and used a fused stream. Thanks for the review

alexcrichton:
Thanks!

PR #74, opened 2016-08-18
added test for filter_map function of streams

MarkusJais:
This adds a unit test for filter_map of streams.

Right now in the adapters function.

After this I propose to split the adapters test function into many smaller functions, each covering on single function to test. That makes it somewhat easier to read and one could easily add more than one test case for each function without making the adapters function huge.

What do you think? I am happy to do it.


MarkusJais:
I am confused by the coverage check failure. I added a test but it still says that coverage decreased.

alexcrichton:
Thanks! Yeah I'd be totally fine splitting up the tests, feel free to do so!

MarkusJais:
Ok, will do. Expect a PR in a few days.

Issue #32, opened 2016-08-11
Never vs. PhantomData


cramertj:
[An RFC](https://github.com/rust-lang/rfcs/blob/773cbfb1a8f17835173f47a76f6aee72095775a8/text/1216-bang-type.md) was recently approved for the `!` (`Never`) type, which represents the type of a value that can never exist.

Several structs in of `futures-rs` use `PhantomData` to mark the type of values that can't exist, such as [the `Error` type of `Finished`](https://github.com/alexcrichton/futures-rs/blob/d99a7e20b4a3d1f6ddd52c416549987ae1dd23a5/src/finished.rs#L10), [the `Item` type of `Failed`](https://github.com/alexcrichton/futures-rs/blob/d99a7e20b4a3d1f6ddd52c416549987ae1dd23a5/src/failed.rs#L9), and [the `Item` and `Error` types of `Empty`](https://github.com/alexcrichton/futures-rs/blob/d99a7e20b4a3d1f6ddd52c416549987ae1dd23a5/src/empty.rs#L12). I believe that, once [this PR](https://github.com/rust-lang/rust/issues/35121) lands (and then makes its way to stable), these structs should be rewritten to use the `Never` type. `Empty`, for example, could be written as follows:
```rust
pub struct Empty {}

pub fn empty() -> Empty { // Is this even necessary any more?
    Empty {}
}

impl Future for Empty {
    type Item = !;
    type Error = !;

    fn poll(&mut self, _: &mut Task) -> Poll<Self::Item, Self::Error> {
        Poll::NotReady
    }

    fn schedule(&mut self, task: &mut Task) {
        drop(task);
    }
}
```

Unfortunately, this feature has yet to land on nightly, let alone stable. Therefore, a transition to `Never` is likely a ways off. As this library is already getting lots of attention, I believe there should be a plan for how and if to structure the library so that a future (hehe 😃 ) transition to `Never` is possible.

In order to be more consistent with terminology, I also think that `Empty` should be renamed to `Never`. `!` is called `Never` rather than `Empty` or `Void` as a result of the discussion [at the end of the `Never` PR](https://github.com/rust-lang/rfcs/pull/1216). The gist of it is that `Empty` isn't a future for an `Empty` value, it's a future for a value that can `Never` exist (i.e. it `Never` completes).

P.S. Great work on this library, @alexcrichton & company! I'm excited to use this in my future Rust projects.

alexcrichton:
[...]

PR #75, opened 2016-08-19
[WIP] Use ! for Never

cramertj:
Part of #32 

I gave this a shot now that the never_type feature landed in nightly, and it works. However, after using it in person, I'm not sure this approach is better than the existing solution using `PhantomData`. For one thing, ! can only coerce at top-level, so converting `Never {}` to a `Future<Item = I, Error = E>` requires mapping both the error and the success case, which (I think) will be non-obvious to end-users and cause unnecessary confusion. Fortunately, the pain introduced by requiring `map` and `map_err` can be easily handled by the existing `never` function. Unfortunately, specifying an appropriate return type for that function (at first glance) requires `impl Trait`, which is _yet another_ nightly (and not soon to be stable) feature.

On the other hand, I believe this approach to be more "correct" from a type-theory perspective and, when also applied to `Finished` and `Failed`, would allow users to generalize computations on futures without having to specify a bogus `Error` type. 

@alexcrichton Thoughts?

### DIFF
--- a/tests/stream.rs
+++ b/tests/stream.rs
@@ -60,6 +60,13 @@ fn adapters() {
     assert_done(|| list().fold(0, |a, b| finished::<i32, u32>(a + b)), Ok(6));
     assert_done(|| err_list().fold(0, |a, b| finished::<i32, u32>(a + b)), Err(3));
     assert_done(|| list().filter(|a| *a % 2 == 0).collect(), Ok(vec![2]));
+    assert_done(|| list().filter_map(|x| {
+        if x % 2 == 0 {
+            Some(x + 10)
+        } else {
+            None
+        }
+    }).collect(), Ok(vec![12]));
     assert_done(|| list().and_then(|a| Ok(a + 1)).collect(), Ok(vec![2, 3, 4]));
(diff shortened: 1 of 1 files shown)


alexcrichton:
Hm yeah the type annotations here are unfortunate, maybe not always necessary though? I'd be fine having this as an off-by-default feature, but we'd probably want to keep the `empty` name for now so we can phase in the `never` name slowly over time.

cramertj:
Okay. I'll finish converting `Finished` and `Failed` to use `!`, and I'll see how many of the type annotations can be removed. Can you clarify your last point about keeping the `empty` name? Should I add `never` and just keep around but deprecate `empty`?

cramertj:
Addendum: one nice thing about leaving the tests as-is (with the type annotations) is that they all work both on stable and on nightly.

alexcrichton:
For now yeah we could have the `never` feature enable the `never` function/type. We'd leave `empty` as-is and then we could one day deprecate it if it pans out.

cramertj:
Closed to reopen with different organization.

PR #79, opened 2016-08-19
cleaned up streams tests and split adapters test into several methods

MarkusJais:
as discussed I've removed out commented code for stream tests and split the adapters method into individual test methods for each method tested.


alexcrichton:
Thanks!

Issue #70, opened 2016-08-17
Panic when "forgetting" ReadTask.


Antti:
When experimenting with futures and task, I got the weird panic:
```
thread 'main' panicked at 'assertion failed: `(left == right)` (left: `4567965872`, right: `4567966032`)', /Users/admytrenko/.cargo/git/checkouts/futures-rs-a4f11d094efefb0a/master/src/task.rs:202
...
  7:        0x10fc46e01 - futures::task::Task::get::h11c78cea13602c1a
   8:        0x10fc44056 - _<futures_io..task..TaskIoTake<'a, 'b, T>>::new::h733b3b84a5e2dd45
   9:        0x10fc460b6 - _<futures_io..task..TaskIoRead<T> as futures_io..ReadTask>::read::h8e19612329de43fb
  10:        0x10fc4a451 - _<futures_io..read_exact..ReadExact<A, T> as futures..Future>::poll::ha3f0ac7f0692a919
```

I understand, that "forgetting" task here probably does not make much sense, but "safe" rust should not panic unless I'm calling some functions that may panic (unwrap, etc).

futures-rs version: https://github.com/alexcrichton/futures-rs/commit/f6729636520f313c0e1f420329c1e2f19aba826c  (latest master at this time).
The smallest example I came up with, which crashes when you try to connect to the server:

```rust
extern crate futures;
extern crate futures_io;
extern crate futures_mio;

use std::env;
use std::net::SocketAddr;

use futures::Future;
use futures_io::{copy, TaskIo, read_exact, write_all};
use futures::stream::Stream;

fn main() {
    let addr = env::args().nth(1).unwrap_or("127.0.0.1:8080".to_string());
    let addr = addr.parse::<SocketAddr>().unwrap();
    let mut l = futures_mio::Loop::new().unwrap();
    let server = l.handle().tcp_listen(&addr);
    let done = server.and_then(move |socket| {
        // Once we've got the TCP listener, inform that we have it
        println!("Listening on: {}", addr);
        socket.incoming().for_each(|(socket, addr)| {
            let io = TaskIo::new(socket);
            let pair = io.map(|io| io.split());
            let amt = pair.map(|(reader, writer)| {
                read_exact(reader, vec![0; 1]).forget()
            }).forget();
            Ok(())
        })
    });
    l.run(done).unwrap();
}
```

alexcrichton:
[...]

PR #80, opened 2016-08-20
futures-mio: update comment and simplify echo example

dwrensha:
Removes an obsolete comment about `Arc` and simplifies some code.

### DIFF
--- a/futures-mio/src/bin/echo.rs
+++ b/futures-mio/src/bin/echo.rs
@@ -26,16 +26,13 @@ fn main() {
         println!("Listening on: {}", addr);
 
         // Pull out the stream of incoming connections and then for each new
-        // one spin up a new task copying data. We put the `socket` into a
-        // `Arc` structure which then allows us to share it across the
-        // read/write halves with a small shim.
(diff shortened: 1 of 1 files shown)


alexcrichton:
Thanks! I think the `lazy` is still needed though to ensure that the `TaskData` is associated with the correct task though? (we're spawning a new task for each connection).

dwrensha:
Yeah, I'm confused about that! It seems to work fine without the `lazy` on OSX.  But just now I've just tested it on Linux and there it hits the panic described in #70.

alexcrichton:
Interesting! This definitely shouldn't encounter any form of platform-specific behavior. I ran it locally on OSX and it seems to panic, though. I wonder if may be the OSX/Linux versions you're testing may accidentally be different?

dwrensha:
Ah, no, sorry. My problem was apparently that I misunderstood how cargo workspaces work. Now I'm seeing the error in both places.

dwrensha:
Ok, now, this PR just updates the comment about `Arc`.

Issue #83, opened 2016-08-20
Return A on futures_io::read_to_end


jeandudey:
When using read_to_end on streams like `UnixStream`, the `A` type parameter of read_to_end could be reused.

PR #84, opened 2016-08-20
Return A type parameter on ReadToEnd

jeandudey:
**Note:** i've used an enum because the compiler was complaining me about `self` being borrowed.
Solves #83


alexcrichton:
Seems reasonable to me, thanks! Looks like some of the other tests need to be updated though?

jeandudey:
I've updated the tests

alexcrichton:
Thanks!

PR #86, opened 2016-08-21
Correct link to Hello World loop's run() in TUTORIAL.md

nelsonjchen:
The namespacing of URL aliases in Markdown is global, and thus, `run` was redefined further ahead around the Task section which overrode the `run` defined around the initial tutorial.


alexcrichton:
Thanks!

PR #92, opened 2016-08-22
Remove confusing bits in the tutorial

nbaksalyar:
It seems that the `store` future [has been removed](https://github.com/alexcrichton/futures-rs/commit/cb863674fc6b8148649484191bc5b7f84d07a726#diff-d477b91ecd9addc3f0a733d3ebd5d43fL1) along with several TaskData methods and now several links in the tutorial lead to 404 pages.

### DIFF
--- a/TUTORIAL.md
+++ b/TUTORIAL.md
@@ -1017,31 +1017,6 @@ between futures:
   naively store data in an `Arc`/`Rc` or worse, in an `Arc<Mutex>` if we wanted
   to mutate it.
 
-But both of these solutions are relatively heavyweight, so let's see if we
-can do better!
-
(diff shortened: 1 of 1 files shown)


alexcrichton:
Thanks for the PR! I'd actually prefer to keep this section rather than delete most of it, but I'll send an update shortly.

alexcrichton:
Ok, I tweaked in https://github.com/alexcrichton/futures-rs/commit/308a7fdbaa0fd7d7d97cf01117ebdde504816b60, but let me know if that's still unclear!

nbaksalyar:
Now it's great, thank you!

PR #93, opened 2016-08-22
Fix some bitrot in the first example

jonathandturner:


alexcrichton:
Thanks! I think though this reflects a [recent update](https://github.com/alexcrichton/futures-rs/pull/84), so I think the local copy may just need to be updated

Issue #58, opened 2016-08-14
DNS resolution


sbstp:
This is a feature that's missing from `mio` and `futures-rs`. All the APIs take a `SocketAddr`, and do not offer a way of resolving host names. The `lookup_host` method from the standard library is only available on nightly, ~~making it "technically" impossible to resolve host names on stable~~. I found a hack around this using `TcpStream::connect` followed by a call to `peer_addr()` but it's far from ideal and it's not asynchronous.

I think a really simple implementation could be offered using a thread pool and a call to `getaddrinfo`. Eventually there could be  a better solution such as bindings to the `getdns` library, or a home made solution.

alexcrichton:
Yeah I would love to have asynchronous DNS resolution, it's definitely a missing feature! Right now you can actually do this on stable Rust via `to_socket_addrs` and the SOCKSv5 proxy [has an example of this](https://github.com/alexcrichton/futures-rs/blob/e9381c2dd020f95604c7eec17764ba7f85040e7c/futures-socks5/src/main.rs#L719-L746). That example also does the work on a thread pool.

In general the pieces should be there in the sense of:

* The `futures-cpupool` crate provides a thread pool to perform resolution on
* The standard library provides `to_socket_addrs` to perform resolution

In the future though I'd love to integrate with something like [rotor-dns](https://crates.io/crates/rotor-dns) or [c-ares](https://crates.io/crates/c-ares), as they'd slot in very nicely!

sbstp:
Gee, I feel stupid, I totally forgot about the ToSocketAddr trait. I haven't been using it because it may return addresses that don't work, and you have to try many addresses and [figure out which one works](https://github.com/rust-lang/rust/blob/master/src/libstd/net/tcp.rs#L98). It would be nice if that process could be implemented to the futures crate too.

alexcrichton:
It's true yeah, although I'm not actually sure what the best practice is for a DNS query that returns multiple addresses. If that happens should we issue a bunch of parallel TCP connects and select the first one that returns? Or should we try each one in series with a particular timeout?

sbstp:
Sequential could be slow (although it is the current behavior of std::net), and parallel could "spam" the server if all the addresses end up working. Perhaps the loop could have a setting describing how it should try to connect, and the user could pick the appropriate one for their use. Perhaps use sequential as a default, and if the users need performance, they can choose parallel.

MarkusJais:
[...]

PR #94, opened 2016-08-22
Create futures-dns crate (#58)

sbstp:
I also made a [branch](https://github.com/alexcrichton/futures-rs/compare/master...sbstp:dns-mio) where `futures-dns` is integrated into `futures-mio` as a sort of POC, but it isn't part of the PR at the moment since Alex doesn't it that way for now.

cc @alexcrichton  @carllerche

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,6 +27,7 @@ script:
   - cargo test --manifest-path futures-iobuf/Cargo.toml
   - cargo test --manifest-path futures-cpupool/Cargo.toml
   - cargo test --manifest-path futures-mio/Cargo.toml
+  - cargo test --manifest-path futures-dns/Cargo.toml
   - cargo test --manifest-path futures-tls/Cargo.toml
   - cargo test --manifest-path futures-tls/Cargo.toml --features force-openssl
   - cargo test --manifest-path futures-minihttp/Cargo.toml
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,6 +27,7 @@ members = [
   "futures-socks5",
   "futures-curl",
   "futures-uds",
+  "futures-dns"
 ]
 
 [dependencies]
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -18,6 +18,7 @@ test_script:
   - cargo test --manifest-path futures-iobuf/Cargo.toml
   - cargo test --manifest-path futures-cpupool/Cargo.toml
   - cargo test --manifest-path futures-mio/Cargo.toml
+  - cargo test --manifest-path futures-dns/Cargo.toml
   - cargo test --manifest-path futures-tls/Cargo.toml
   - cargo test --manifest-path futures-minihttp/Cargo.toml
   - cargo test --manifest-path futures-socks5/Cargo.toml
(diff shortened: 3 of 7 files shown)


alexcrichton:
Thanks @sbstp! Could you also add some tests to this crate as well?

I think it may also be worth fleshing out what `TcpStream::connect` might look like if it were to take a `T: ToEndpoint`. Perhaps some helper functions could be added to this crate?

sbstp:
Sure thing. There already are 2 tests for `ToEndpoint`, could add more to test IPv4/IPv6 parsing for sure. But I don't think it would be a good idea to test if the CpuResolver works as the test would fail when the computer running the tests has no internet. Also checking if a domain resolves to a particular IP is probably not a good idea as it might change.

When you say `TcpStream::connect`, do you mean `LoopHandle::tcp_connect`? I can try creating an extension trait but it will not give the full experience because reusing the thread pool requires to keep a `CpuPool` around, in the `LoopHandle` for instance.

alexcrichton:
Ah yeah basically just have an extension trait or two for the I/O objects to use hostnames instead of socket addresses

sbstp:
Alright I wrote something that uses the resolver. The `tcp_connect` method will try all the IP addresses it got, sequentially, using a future chain. Not sure how to properly test this code though. Anyway, API looks like this:
```rust
let mut lp = Loop::new().unwrap();
let c = Connector::new(lp.handle(), 4); // 4 is the number of threads in the resolver's pool
let f = c.tcp_connect("google.ca:80").and_then(|sock| {
    println!("connect successful {:?}", sock.peer_addr().unwrap());
    Ok(())
});
lp.run(f).unwrap();
```

sbstp:
This is what I feared, the localhost/resolve test failed on travis. I'll see what I can do.

alexcrichton:
Looks pretty neat! Maybe the parallel approach could also be prototype with `select_all`? (just as a proof of concept)

sbstp:
Thanks! I've looked quickly at `select_all` and I'm not sure I can use it. The future returned by `select_all` resolves when any of its given futures resolves to `Ok` or `Err`. I need a future that resolves on the first `Ok` it sees, but it must not resolve on an `Err`.

That future might be useful in other cases as well.

sbstp:
Ran into an issue with duplicate addresses. [It's been fixed](https://github.com/rust-lang/rust/commit/0314d179aa07905cc3bd4ec3206be28ff11696f9) but hasn't made its way into stable yet.

sbstp:
@alexcrichton if you have the time, please check if the implementation of `select_all_ok` is sensible.

alexcrichton:
@sbstp the implementation looks good to me, and we chatted briefly on IRC, but I'm going to close this for now.

With the extraction of the I/O crates from this repository into the tokio-rs organization this doesn't necessarily belong here any more, and I think we'll want to prototype a `tokio-dns` crate perhaps. While it's under development though I wonder if you'd be willing to host the crate?

sbstp:
Yeah, I will. I was just waiting for the refactoring to settle.

sbstp:
https://github.com/sbstp/tokio-dns

alexcrichton:
Thanks @sbstp!

PR #95, opened 2016-08-23
WIP: overhaul task/executor story

aturon:
This PR does most of the work toward clarifying our task/executor story.

## Basic executors

The basic idea is that futures themselves do not know or control how/where they are executed. In order to run a future, you need to create a task, and at that point you bind the future/task to an *executor*. In this PR there are three executors:

- `futures::Inline`: the "inline" executor just runs a future immediately whenever/wherever it is signaled for readiness. This is what we had as the default behavior previously. Requires `Send + 'static`

- `futures_mio::LoopHandle`: executes the task on the event loop, to which it is pinned. In other words, *all* polling of the task happens on the event loop; signaling readiness to the task just enqueues a notification to the event loop. Requires the future to be Send (since we are moving it onto the loop). But not 'static -- see below.

- `futures_mio::LoopPin`: like `LoopHandle`, but does *not* require the future to be `Send`, because the future is guaranteed to already be located on the event loop thread.

In terms of relaxing `'static`, the event loop now takes a lifetime, tying it to a particular stack frame whose data you can access within loop futures. This will allow you to share `&` references among futures running within the event loop.

## Unpark/Executor revamp

To make the above work, I've revamped the representation such that blocked tasks are represented by an `Unpark` trait. This is basically where the core executor code lives -- it says what to do when you want to *continue* executing a task.

Executors themselves only provide a way to *spawn* tasks, at which point they'll construct an `Arc<Unpark>` to be used to continue poking the task in the future.

## Spawning

To actually use one of the executors, there are *spawn* methods on `Future` and `Stream`:

- `my_future.spawn(executor)`: spawns the future on the given executor, returning a *new* future that proxies its result. By default, the task is "daemonized": it will continue even if the proxy future is dropped. I believe this is the least surprising behavior. There's also a `cancel` method on the proxy, however.

- `my_stream.spawn(executor)`: spawns the stream on the given executor, returning a new stream that proxies its results. The subtask progresses only *on demand*: until more items are requested from the proxy stream, the subtask will idle.

## Yielding

There's now a `task::yield_now` method which can be used to signal that you want to poll the current task again even on `NotReady`. Useful for cooperative execution.

## Task-local data

I've renamed `TaskData` to `TaskRc`, to better reflect what it's doing. I'm planning to add a new `TaskData` that uses something like [typemap](https://crates.io/crates/typemap) to provide something more like thread-local storage.


aturon:
A big TODO item is pushing this through curl, etc.

carllerche:
Thanks for this, I think that the executor has been much improved by this change. I do have one bike shed that I will raise for discussion.

While `Future::spawn(executor)` is a much better API than calling `.forget()`, I believe that it has a similar readability problem.

Take the following code:

```rust
my_future.map(…)
         .and_then(…)
	   .map_err(…)
	   .spawn(…)
         .select(…)
         .and_then(…)
         .map(…)
```

The `spawn` function has much heavier runtime implications than the other combinators in the chain, yet visually it blends in making it easy to read over. Reading the code snippet, it is not obvious (to me at least) that the implication of `spawn` is that all combinators defined above it will be executed on the specified executor where as all combinators below it get executed on the current executor.

Given the significant runtime implications of the `spawn` combinator, my proposal is to match something closer to `thread::spawn`, where the logic to be executed on the given executor is passed as an argument. Something like:

```rust
let fut = my_executor.spawn(
	my_future.map(…)
               .and_then(…)
	         .map_err(…));

fut.select(…)
   .and_then(…)
   .map(…)
```

This API pushes the user to split up the future chain according to where the futures end up being executed.

ipetkov:
> By default, the task is "daemonized": it will continue even if the proxy future is dropped. I believe this is the least surprising behavior.

I would advocate against "daemonizing" by default because it goes against several core Rust principals: it is neither explicit in its behavior, nor zero-cost (more like mail-in-rebate); resources are tied up, unless the programmer opts-out. It reminds me too much of Java-land, and of worrying about not doing the wrong thing by accident, instead of focusing on doing the right thing on purpose.

Here is a silly example to illustrate both points:

```rust
// Implicit footgun
fn do_stuff() {
	let burn_cpu = IntoFuture::into_future(|| { loop{} });
	let future = burn_cpu.spawn(executor);

	if i_forgot_to_get_milk {
		panic!("oh no!"); // Oops, future dropped... is it getting warm in here?
	}

	if i_am_warm {
		future.cancel(); // Whew, that was close!
	}
}
```

```rust
// Explicit opt-in
fn do_stuff() {
	let burn_cpu = IntoFuture::into_future(|| { loop{} });
	let future = burn_cpu.spawn(executor);

	if i_am_cold {
		future.daemonize(); // Feeling toasty!
 	} else {
 		// cpu free to do other work
 	}
}
```

Getting the explicit daemonization call wrong is easy to spot and fix (e.g. failing tests) because the behavior is wrong. Getting explicit opt-out of implicit behavior 100% correct everywhere is hard to spot and debug (e.g. why is so much CPU being consumed).

ksf:
I'm worried that the out-of-band signalling that Task::yield() does is going to effectively disable loop optimisations.

To recap: Compilers mess about in various ways with loops -- they fuse them to do multiple passes over the same data in lockstep, but they also fission loops operating over multiple arrays to increase locality of reference. They also do things like replace "set state variable, jump to loop beginning, switch on state variable" with direct jumps, "spaghetting up trampolines". They (including LLVM) are *very* good at this kind of thing.

I'd be very surprised indeed if the compiler can still even detect that it is dealing with a loop through all those indirections: Signalling via TLS, getting the actual loop body out of (roughly)  `Arc<UnsafeCell<Option<Box<_>>>>`. If `LoopTask` sees that there should be a repoll, it should just do that and not jump through confusing (for the compiler) hoops, including invoking threading primitives.

---

As per the docs, `poll()` should finish very quickly. "Quickly" of course is relative, however "I did the minimum amount of work possible to make progress so now I should yield" is surely a valid interpretation. For a fold, that would be consuming one upstream element -- and what else should it do? Looping until upstream returns `NotReady` would not need cooperative yields in the first place, and anything in between is complete guesswork.

I'd argue that this "minimum amount of work possible" should be the actual design guideline: Both because it is not up to interpretation and scheduling should not be the concern of leaf `poll()`s. It should be the concern of the driver loop, at most special-purpose combinators... better yet, compiler magic, which is able to make these maximally-granular yields performant: If and only if we harness the loop optimiser, which means having a discernible top-level loop so the compiler has something to fission and spaghetti.

And that means using in-band signalling, to be precise: `Poll::Yield` (or whatever you want to call it). API-wise it wouldn't be a big change: Combinators should largely treat it just like `Poll::NotReady`, which is usually passing it on. The only code that actually *needs* to distinguish them is the top-level driver loop, though there probably should also be a driver combinator in case the loop optimiser needs some prodding, semantically it should be a no-op.

I'd also argue that it's the cleaner API as "I don't *need* to make progress right now" and "I *can't* make progress right now" are two fundamentally different things that should be clearly distinct in the enum, not smeared all over the place.

---

And, yes, I might be eyeing streams/futures as iterator replacements which explains my angle, here. If nothing else, it seems silly to use different libraries depending on whether you're interleaving arbitrary IO or just iterating over a random `Vec`.

ipetkov:
Really liking @ksf's idea of having a `Poll::Yield` variant. I can think of situations where a `Future` can neither make progress nor know when it *could* make progress (i.e. can't or doesn't know when to `unpark` its `Task`): for example, waiting on a child process to exit or some external event that isn't directly/immediately observable.

In such situations we could have a nice distinction between `Poll::NotReady` ("I'm not yet done, please poll me as soon as you can") and `Poll::Yield` ("I'm not done, but I don't know when I might be. Feel free to prioritize some other work and check back with me when you can") so the top-level loop can optimize execution scheduling.

aturon:
@ipetkov To be clear, this PR already contains such a mechanism, but it's communicated through a "side-channel": by setting a flag on the task itself, and then returning `NotReady`. The rationale is that this makes things significantly easier when working with the `poll` method, because the logic for `NotReady` and `Yield` is basically identical until you get back out to the task.

ipetkov:
@aturon Thanks for pointing that out! Hadn't looked too closely at the changes, but it looks like you are referring to `yield_now()` and it covers the case I had in mind.

As a side note, `no_std` crates who depend on the core `Future` definitions may not have side-channels like `Task` to communicate this yield behavior. A `Poll::Yield` variant may be useful there, however, I'm not even sure if they will care about making this distinction themselves in a more limited environment.

aturon:
Closed in favor of https://github.com/alexcrichton/futures-rs/pull/112

PR #111, opened 2016-08-29
[WIP] Add a History adaptor for streams

josephDunne:
This is an attempt to add an adapter which records the history of all
the items that pass through it. This allows the stream to act as a
type of Publisher.

There are two types of History streams: Full or Partial. A Full History will return a full list of items to a new "subscriber". A partial history will only return those items created since subscription. The history stream can limit the amount of history by an optional size.

The current implementation is not very clever about using locks. I think there are better data structures to use than a simple `Mutex<Vec<Item>>`.
Its a WIP :-)

I am excited about something like the History adaptor because it opens upon dataflow type algorithms so it would be could to get feedback on what the best way to do this is.

Thanks for the awesome work

```Rust
    let mut history = list().history(None);
    let mut sub1 = history.clone();
   
    assert_eq!(history.poll(), Poll::Ok(Some(1)));

    let mut sub2 = history.clone();

    assert_eq!(history.poll(), Poll::Ok(Some(2)));

    assert_eq!(sub1.poll(), Poll::Ok(Some(1)));
    assert_eq!(sub1.poll(), Poll::Ok(Some(2)));

    assert_eq!(history.poll(), Poll::Ok(Some(3)));

    assert_eq!(sub2.poll(), Poll::Ok(Some(2)));
    assert_eq!(sub2.poll(), Poll::Ok(Some(3)));

    assert_eq!(history.poll(), Poll::Ok(None));

    assert_eq!(sub1.poll(), Poll::Ok(Some(3)));
    assert_eq!(sub1.poll(), Poll::Ok(None));
    assert_eq!(sub2.poll(), Poll::Ok(None));
}
```

full history:
```Rust
    let mut history = list().history(None);
    let mut sub1 = history.clone();
   
    assert_eq!(history.poll(), Poll::Ok(Some(1)));
    assert_eq!(history.poll(), Poll::Ok(Some(2)));
    assert_eq!(history.poll(), Poll::Ok(Some(3)));

    assert_eq!(sub1.poll(), Poll::Ok(Some(1)));
    assert_eq!(sub1.poll(), Poll::Ok(Some(2)));
    assert_eq!(sub1.poll(), Poll::Ok(Some(3)));

    let mut sub2 = history.clone();

    assert_eq!(sub2.poll(), Poll::Ok(Some(1)));
    assert_eq!(sub2.poll(), Poll::Ok(Some(2)));
    assert_eq!(sub2.poll(), Poll::Ok(Some(3)));

    assert_eq!(history.poll(), Poll::Ok(None));
    assert_eq!(sub1.poll(), Poll::Ok(None));
    assert_eq!(sub2.poll(), Poll::Ok(None));
}
```

### DIFF
--- a/src/stream/history.rs
+++ b/src/stream/history.rs
@@ -1,0 +1,202 @@
+use poll::Poll;
+use stream::{Stream, Fuse};
+use std::prelude::v1::*;
+use std::sync::{Arc, Mutex};
+
+/// A stream which records all of the values of a stream as a history.
+///
+/// A history of type `HistoryType::Full` will keep a full history
+/// and any future consumers that create a clone of the stream
+/// will recieve all items produced so far
+///
+/// A history of type `HistoryType::Partial` will only keep a history
+/// of items of items starting from the first clone. Subsequent clones
+/// will not recieve any history from before the clone. This is similiar
+/// to a Pub/sub model.
+///
+/// This stream is created by the `Stream::history` or `Stream::full_history`
+/// method.
+pub struct History<S: Stream>
+    where <S as Stream>::Item: Clone {
+    cursor: (usize, usize),
+    history: Vec<S::Item>,
+    inner: Arc<Mutex<ProtectedHistory<S>>>
+}
+
+pub enum HistoryType {
+    Full(Option<usize>),
+    Partial(Option<usize>),
+}
+
+
+pub fn new<S: Stream>(stream: S, type_of: HistoryType) -> History<S>
+    where <S as Stream>::Item: Clone {
+    let mut v = Vec::new();
+    v.push(0);
+    History {
+        cursor: (0,0),
+        history: Vec::new(),
+        inner: Arc::new(Mutex::new(ProtectedHistory {
+            inner: stream.fuse(),
+            history: history::<S>(type_of),
+            index: 0,
+            cursors: v
+        }))
+    }
+}
+
+struct ProtectedHistory<S: Stream>
+    where <S as Stream>::Item: Clone
+{
+    inner: Fuse<S>,
+    history: HistoryHelper<S::Item>,
+    index: usize,
+    cursors: Vec<usize>
+}
+
+impl<S: Stream> ProtectedHistory<S>
+    where <S as Stream>::Item: Clone
+{
+    fn poll(&mut self) -> Poll<Option<S::Item>, S::Error> {
+        let res = self.inner.poll();
+        match res {
+            Poll::Ok(Some(item)) => {
+                self.history.push(item.clone());
+                self.index = self.index.wrapping_add(1);
+                Poll::Ok(Some(item))
+            }
+            otherwise => otherwise
+        }
+    }
+}
+
+enum HistoryHelper<I: Clone> {
+    Full(Vec<I>, Option<usize>),
+    Partial(Vec<I>, Option<usize>),
+}
+
+impl<I: Clone> HistoryHelper<I> {
+    fn is_full(&self) -> bool {
+        match *self {
+            HistoryHelper::Full(_,_) => true,
+            _ => false
+        }
+    }
+
+    fn clone_vec(&self) -> Vec<I> {
+        match self {
+            &HistoryHelper::Full(ref v,_) => v.clone(),
+            &HistoryHelper::Partial(ref v,_) => v.clone(),
+        }
+    }
+
+    fn push(&mut self, item: I) {
+        match *self {
+            HistoryHelper::Full(ref mut v, _limit) => {
+                v.push(item)
+            },
+            HistoryHelper::Partial(ref mut v,_limit) => {
+                v.push(item);
+                // maybe flush items
+            },
+        }
+    }
+
+    fn get_history(&self, cursor: (usize, usize)) -> Option<Vec<I>> {
+        match self {
+            &HistoryHelper::Full(ref v,_) => {
+                if v.len() > cursor.0 {
+                    let mut ret = Vec::new();
+                    let slice = &v[cursor.0..];
+                    ret.reserve_exact(slice.len());
+                    unsafe {
+                        ret.set_len(slice.len());
+                    }
+                    ret.clone_from_slice(slice);
+                    return Some(ret)
+                }
+                None
+            }
+            &HistoryHelper::Partial(ref v,_) => {
+                if v.len() > cursor.0 {
+                    let mut ret = Vec::new();
+                    let slice = &v[cursor.0..];
+                    ret.reserve_exact(slice.len());
+                    unsafe {
+                        ret.set_len(slice.len());
+                    }
+                    ret.clone_from_slice(slice);
+                    return Some(ret)
+                }
+                None
+            }
+        }
+    }
+}
+
+fn history<S: Stream>(type_of: HistoryType) -> HistoryHelper<S::Item>
+    where <S as Stream>::Item: Clone
+{
+    match type_of {
+        HistoryType::Full(limit) => HistoryHelper::Full(Vec::new(), limit),
+        HistoryType::Partial(limit) => HistoryHelper::Partial(Vec::new(), limit),
+    }
+}
+
+impl<S> Stream for History<S>
+    where S: Stream,
+         <S as Stream>::Item: Clone
+{
+    type Item = S::Item;
+    type Error = S::Error;
+
+    fn poll(&mut self) -> Poll<Option<Self::Item>, Self::Error> {
+        if let Some(item) = self.history.pop() {
+            return Poll::Ok(Some(item))
+        }
+        let mut protected = self.inner.lock().unwrap();
+        // first try fetch some history
+        if let Some(mut history) = protected.history.get_history(self.cursor) {
+            self.cursor.0 += history.len();
+            history.reverse();
+            let item = history.pop().unwrap();
+            self.history = history;
+            return Poll::Ok(Some(item))            
+        }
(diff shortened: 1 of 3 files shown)


josephDunne:
I am going to need some helping understanding what is in prelude...The build is failing because of unknown `vec![]` macro but it compiles for me (rustc 1.13.0-nightly (e9bc1bac8 2016-08-24))

carllerche:
Thanks for the contrib, definitely interesting. I skimmed through the impl and have a few thoughts & questions.

* The cloned history streams could live on different tasks. As such, I believe that the implementation is going to need to have a way to "fan out" readiness notifications. I haven't thought of a good way to do this off the top of my head.

* I didn't quite follow what happened when the history stream is cloned and it is a partial history. What happens when a new item overflows another history stream? Are values silently dropped? How will the other stream be notified that values were dropped?

* Given that this combinator is going to be pretty heavy weight and probably require thread locals (point one), I would suggest to keep the `history` combinator off of the trait. Instead, make it a module function, something like `futures::history::full(stream)`

* *Bikeshed* What about `Replay` as a name? `History` was a little vague to me initially.

But this seems like a good start. I'm curious what thoughts others will have as well.

josephDunne:
> "fan out" readiness notifications

Yeah that is going to be necessary I don't have any feedback on this yet.

> I didn't quite follow what happened when the history stream is cloned and it is a partial history

This isn't complete yet. 

>How will the other stream be notified that values were dropped?

yeah not sure. Is this a udp or tcp style problem. Perhaps needs different apis for either.
My thoughts where that this would never drop items. If it hit the limit to the number of replay items it should back-pressure by not calling `poll` on the upstream. 

> I would suggest to keep the history combinator off of the trait.

Its possible this type of abstraction should live in a layer above futures. I type of futures pub/sub layer.

> What about Replay as a name?

works for me

killercup:
> I am going to need some helping understanding what is and isn't in prelude...I don't know how to fix the integration build failures :-/ Tests pass locally (rustc 1.13.0-nightly (e9bc1bac8 2016-08-24))

Parts of this crate can work without `std`, and the tests fail because you didn't annotate that you need `std`. Since the feature that enables `std` is active by default, you probably didn't notice this locally :)

Basically, add `#[cfg(feature = "use_std")]` to your trait methods and move your mod/use items into the `if_std!` block.

alexcrichton:
Thanks for the PR @josephDunne! This seems like a pretty slick combinator, and it'd be nice to have! In addition to [the points](https://github.com/alexcrichton/futures-rs/pull/111#issuecomment-243331128) raised by @carllerche I'd also be curious to explore the backpressure story here a bit. One point of `Future` and `Stream` is that they naturally allow for avoidance of unnecessary buffering, which also prevents problems of tasks running away with resources if they're working too quickly. It ends up being that a whole computation naturally reaches a steady state the system can process.

Here, though, I wonder how that works? I guess that's part of the size you pass in perhaps? Up to N elements are buffered and once that limit is hit... "something" happens? Along those lines, I've got a few questions:

* Do you have in mind some use cases for full history vs partial history? It looks like having to store the entire history for a channel permanently may be kinda unfortunate? Maybe there's a compelling use case though!
* Perhaps we could avoid buffering altogether? You could imagine having a bunch of streams move in lockstep along the lines of:
   * At any point in time there are N active "subscriptions"
   * When a new item is requested from any subscription for the first time globally, it pulls from the underlying stream
   * When the next item is requested, it's not pulled out until *all* streams have made progress.
* If we do have buffering, maybe we could always require a limit? vs having to potentially store an infinite number of items.


Also yeah I believe @killercup [is right](https://github.com/alexcrichton/futures-rs/pull/111#issuecomment-243552416) about getting tests working, but if that doesn't work out just let me know!

josephDunne:
@killercup is right I missed the #[if_std] part in the mod includes.

> Do you have in mind some use cases for full history vs partial history?

Honestly not good ones! However at the time I was reading the GLL Parser paper [here](http://dotat.at/tmp/gll.pdf) and essentially the paper solves left hidden recursion for LR parsers by inserting a kind of 'async' step. Instead of immediately trying to consume a (possibly infinite recursive) production rule , it spawns a stream of future parse results - of course I lie it actually it pushes a continuation onto an explicit stack but you can see where I was going with this. But to make progress it has to remember all previous results for this continuation and give those results to any consumers that need it. So doing parsing with futures is probably a silly thing to do but perhaps there are other dataflow problems out there that could use a full history *shrug*

> You could imagine having a bunch of streams move in lockstep

I think this is a much nicer solution and fits in with the current philosophy you are going for. I can't say what implications this would have though. From a systems perspective it means my system as a whole is only as fast as my slowest consumer. Of course from a systems perspective back-pressure is essential...so yes either limit the buffer by a small amount for small bursts of work but force some sort of lock step. Alternate semantics could be built whereby slow consumers miss items because they are dropped when the buffer hits N - the system designer should choose.

josephDunne:
I have a question regarding @carllerche point about this implementation requiring fan out of ready notifications. With the removal of the `schedule()` function which part of the code is responsible for unparking the `Task`? If I have a stream A polling Stream B where is the code that calls unpark? Is it the executer? code somewhere in Stream B.poll() or somewhere in Stream A.poll()? I assume its Stream B. If I am correct then am I also correct in saying a fan out mechanism would be to just store a set of TaskHandles inside Stream B? Stream can then iterate and unpark its "subscribers" when its complete?

I feel I might be missing something about how the futures system makes progress.

tilal6991:
BTW a lot of the concepts mentioned here are already formalised in the Rx terminology - that is it is up to the application to decide what the backpressure strategy is (see http://reactivex.io/RxJava/2.x/javadoc/io/reactivex/BackpressureStrategy.html)

josephDunne:
@tilal6991 Of course. I need to get familiar with the other observer/stream patterns out there. 

A difference I see from a glance with reactivex is that it is push vs poll. So it needs a back pressure strategy. The streams in this library have back pressure baked in to the implementation and the fact that this adaptor raised those concerned is probably because it's a bad fit.

I think the issue with this pull adaptor is that it is mixing too many concerns. I think it should be split into an adaptor called 'Tap' (as in tapped) (to distinguish it from pub/sub which is a push metaphor). 

A tapped stream will only make progress once all the taps have made progress. In addition you could have a notion of an owned or borrowed tap. If the stream item is 'Copy' then you can get a full owned copy of the item. If not then each tap can get a reference to the latest item. Once all reference taps have had a chance to make progress the final owned item can be delivered to the sink. 

I think the history/replay part of this proposal should perhaps be implemented outside of the stream model. For example you can stream results to a persistent sink. If a process needs the history it can ask the sink for a replay.

Still not sure how to fan out the notifications though

tilal6991:
As mentioned in https://github.com/alexcrichton/futures-rs/issues/110 Rx (at least in the newer versions) along with Reactive Streams uses a push-pull model: i.e. the Observer registers interest in the number of elements it is able to process and the Observable sends that many items. Think of it like a super-charged poll method where instead of just 1 element, you can request n elements.

Changing to this model introduces a lot of complexity and while I think it makes sense for Java where allocations are expensive so you do not want to constantly request only 1 element, with Rust's zero cost abstractions, a single element at a time may be just as efficient and significantly easier to reason about and implement ideas for.

In any case, before anyone implements things like this PR, I think it's well worth looking to see how Rx has done it to get a feel for how the problem might be solved. For example, my prototype for publish which uses the lockstep idea which was propsed earlier in this thread and it came from an earlier version of Rx (which got abandoned I think because of the push-pull model usage which made it inefficient). I will attempt to fix up my change (it was implemented when schedule was still a thing) and send a PR.

alexcrichton:
Another thought I had last night. I wonder if we could express this with an interface like:

```rust
fn split(self, f: F) -> (impl Stream<T>, impl Stream<U>)
    where F: FnMut(Self::Item) -> (T, U)
```

That is, we could generalize `Clone` here as well as the synchronization to perhaps just two streams. This allows you to take any stream and split it into two streams which progress in lockstep. 

One downside would maybe be that it's not quite as optimized in the case where you clone a stream many times, but in theory a history stream is just continual application of:

```rust
stream.split(|x| (x.clone(), x))
```

@josephDunne 

For your point about fanning out notifications and such, when a task needs to block you can call `task::park` and then when it needs to unblock you can call `.unpark()` on the returned handle.

drewhk:
> Changing to this model introduces a lot of complexity and while I think it makes sense for Java where allocations are expensive so you do not want to constantly request only 1 element, with Rust's zero cost abstractions, a single element at a time may be just as efficient and significantly easier to reason about and implement ideas for.

What we did in Akka Streams is that we added a programming model where you can only request up to 1 element (exactly because of the complexities you mention), but this request can be actually served from a buffer that you as a user do not see. Once a certain number of elements have been consumed, the framework will request the next batch automatically for you to keep the buffer filled. I.e. you can get the simplicity of the simple element request, and the performance of batching.

As for the cost, it is not the allocations, many implementations have no allocations at all on the `request()` path as it usually turns into something like (getAndAdd is an atomic operation here):

```
if (upstream.demand.getAndAdd(demand) == 0) // We might unblock/schedule the producer now as it had 0 demand before
```

The reason why we do batching is to minimize atomics and inter-thread communications and also to minimize wakeups (scheduling the task to a pool), allocations usually don't enter the picture.

josephDunne:
Hi @alexcrichton, I would like to close this ticket. I believe the history storage requirement should be handled by a separate library. I am still keen to see a subscriber/split model in his stream model. I think this would be better served with a separate issue.

PR #112, opened 2016-08-30
Executor revamp take 2

aturon:
This is another stab at the executor model revamp. As before, the main
idea is that a "task" is a future paired with an executor. However, the 
implementation is now factored much more cleanly, and the executors
themselves implement a direct `spawn` method. Other changes:

- Centralizes around the `Task` type, which replaces `TaskHandle`.

- Provides two task execution models: via an `Executor` (which is
  generally one or more threads together with a queue of tasks to
  execute) or by blocking on the current thread (via `ThreadTask`).

- Removes the `forget` method. Instead, futures will always be run using
  one of the above execution models. The core futures library provides
  only the `wait` executor, which executes the future to completion on
  the current thread via `ThreadTask`.

- Renames `TaskData` to `TaskRc`, to more directly reflect its purpose.

- Add task-local data using [`TypeMap`](https://github.com/reem/rust-typemap).

- Introduces the notion of an "unpark event", which is a way to pass
  information from an event source back to the task upon unpark. The
  information is passed by inserting IDs into concurrent sets prior to
  wakeup. This works via a scoped API, `with_unpark_event`, which
  decorates the current task so that `park` will include the desired
  unpark events.

- A new `catch_unwind` combinator.


aturon:
Note: the task-local API is currently only exposed via `task::with_local_data` (and `_mut`), which exposes the entire table. We'll probably want to provide future combinators to access the local data, but the core APIs here are enough to build those out of tree.

I also don't have tests for the task-local data yet; ran out of time.

aturon:
As far as rolling this out in `tokio-core`, I'm not going to have time to do that, but @alexcrichton I think it's quite straightforward if you aren't trying to add a lifetime to the event loop. Given that my previous experiment showed that this was possible, I think we should punt for now and just make the minimal change needed to switch the event loop over to this executor model.

aturon:
Final note: I didn't bother with the yield stuff on this take, because we should consider it with the potential larger revamp of `Poll`.

aturon:
Update: added `TaskRc` back.

src/catch_unwind.rs
```diff
@@ -0,0 +1,45 @@
+use {Future, Poll};
+
+use std::prelude::v1::*;
+use std::any::Any;
+use std::panic::{catch_unwind, UnwindSafe, AssertUnwindSafe};
+
+/// Future for the `catch_unwind` combinator.
+///
+/// This is created by this `Future::catch_uwnind` method.
```
hexsel: uwnind->unwind

PR #114, opened 2016-08-31
Implement Stream for Empty

### LINKED ISSUE #113: Implement Stream for Empty
Tokio has a use case for Empty with Stream:

https://github.com/tokio-rs/tokio-line/blob/master/src/empty.rs

nabijaczleweli:
Closes #113

### DIFF
--- a/src/empty.rs
+++ b/src/empty.rs
@@ -1,6 +1,7 @@
 use core::marker;
 
 use {Future, Poll};
+use stream::Stream;
 
 /// A future which is never resolved.
 ///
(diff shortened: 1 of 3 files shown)


alexcrichton:
Thanks for the PR! It seems reasonable to me to reuse the type here, but to reduce the surprise factor could you also add this to `src/stream/mod.rs`?

```rust
pub use empty::{Empty, empty};
```

That is, we in theory also want `futures::stream::{empty, Empty}`, but it's fine if it ends up returning the same underlying type itself.

alexcrichton:
Although conversely this would be a bit of a departure from the existing conventions of the futures/streams section being pretty separated. One example issue that could arise is that now `empty().poll()` requires UFCS to disambiguate because it wouldn't be sure if you're calling `Future::poll` or `Stream::poll`. 

Perhaps this could add a new `src/stream/empty.rs` instead in that case?

nabijaczleweli:
I've made it so `stream::Empty` is a `Stream` (with impl as suggested in #113) and `stream::empty()` to create it.

alexcrichton:
Thanks!

PR #122, opened 2016-09-02
[WIP] Add publish combinator

tilal6991:
This is an initial attempt at adding a publish combinator - one which mutlicasts a stream from one upstream to several downstream. The whole combinator proceeds as fast as the slowest consumer.

There are several things which are not optimal and would appreciate ideas about:
1. There is no unsubscription from publish. I feel this should be implemented by adding a Drop implementation to Publish but this requires other combinators (e.g. take) to also drop upstreams when they no longer intend to poll the upstream.
2. The system relies on only polling once for every unpark - I'm not sure this is a valid assumption but it seemed to work when I wrote the test.
3. Locking for every access to the inner - this should be changed for sure as it is highly inefficient to have a single point of contention. I'll be attempting to change this to an atomic based implementation at some point.

As I say would appreciate feedback especially as I'm fairly new to writing Rust :)

### DIFF
--- a/tests/stream.rs
+++ b/tests/stream.rs
@@ -199,3 +199,43 @@ fn wait() {
     assert_eq!(list().wait().collect::<Result<Vec<_>, _>>(),
                Ok(vec![1, 2, 3]));
 }
+
+#[test]
+fn publish() {
+    let (tx, rx) = channel::<_, u32>();
+    tx.send(Ok(1))
+        .and_then(|tx| tx.send(Ok(2)))
+        .and_then(|tx| tx.send(Ok(3)))
+        .forget();
+
+    let first = rx.publish();
+    let second = first.clone();
+    let third = first.clone();
+
+    let mut first = first.wait();
+    let mut second = second.wait();
+    let mut third = third.wait();
+
+    // First register the original one.
+    assert_eq!(first.next(), Some(Ok(1)));
+
+    // The second one now comes in - it should have been dropped
+    // for the first but will recieve the second.
+    assert_eq!(second.next(), Some(Ok(2)));
+
+    // The third one also comes in and should receive the second.
+    assert_eq!(third.next(), Some(Ok(2)));
+
+    // The first returns to get the second.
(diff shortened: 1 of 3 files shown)


tests/stream.rs
```diff
@@ -199,3 +199,43 @@ fn wait() {
...
+    let mut first = first.wait();
+    let mut second = second.wait();
+    let mut third = third.wait();
+
+    // First register the original one.
+    assert_eq!(first.next(), Some(Ok(1)));
+
+    // The second one now comes in - it should have been dropped
+    // for the first but will recieve the second.
+    assert_eq!(second.next(), Some(Ok(2)));
```
josephDunne: What if I want to subscribe to every message even the very first one?

alexcrichton:
Thanks for the PR! It seems like this may be quite similar to https://github.com/alexcrichton/futures-rs/pull/111, right?

I've looked a bit at the implementation as well, and I believe the guarantee we provide is that a `Future` will always wake up the *last* task to call `poll`, but it looks like here it wakes up the first?

tests/stream.rs
tilal6991: With the current design it would not be possible. We would need something similar to `connect()` (http://reactivex.io/documentation/operators/connect.html) from Rx to allow an arbitary number of consumers to hook into the system before the first poll is carried out upstream.

tilal6991:
@alexcrichton This combinator provides the guarantee that _all_ tasks which previously polled are woken up rather than just the first or just the last.

alexcrichton:
@tilal6991 oh but it looks like `inner.add_task(task.clone());` is only called if `self.task` is none, which is false after the first call to `poll`. That is, if this stream migrates amongst tasks, it'd only wake up the first one?

tilal6991:
Oh I see what you mean now. If the outer publish stream moves between tasks yes that is a problem. I have an idea to fix that - will do when I get some time.

alexcrichton:
Closing due to inactivity, but feel free to resubmit with a rebase!

PR #134, opened 2016-09-08
[RFC] Add an unordered buffer stream adapter.

Stebalien:
Unlike the current buffered stream, this one does not preserve order. This is primarily useful when handling a stream of requests. Example use case:

```rust
my_requests.map(|(client, req)| {
    log_request_begin(req);
    do_some_work.then(move |reply| {
        client.send_response(reply)
    })
}).unordered(connection_pool_size).for_each(|result| {
    log_request_end(result);
})
```

### DIFF
--- a/src/stream/buffer_unordered.rs
+++ b/src/stream/buffer_unordered.rs
@@ -1,0 +1,85 @@
+use std::prelude::v1::*;
+
+use {Async, IntoFuture, Poll, Future};
+use stream::{Stream, Fuse};
+
+/// An adaptor for a stream of futures to execute the futures concurrently, if
+/// possible, delivering results as they become available.
+///
+/// This adaptor will buffer up a list of pending futures, and then return their
+/// results in the order that they complete. This is created by the
+/// `Stream::buffer_unordered` method.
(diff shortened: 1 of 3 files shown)


Stebalien:
This is in response to the removal of `forget` and the resulting conversation (https://github.com/alexcrichton/futures-rs/issues/33#issuecomment-244386180). Basically, I need a way to drive outstanding requests without caring about order (long running requests should not hold up quick requests). Initially, I thought I wanted an unbounded buffer but quickly realized that that would be a very bad idea :tm:. Instead, hung requests should just timeout.

The PR as written appears to work but:

1. I haven't tested it thoroughly.
2. "unorderd" is not a good name for this.

alexcrichton:
Nice! This looks good to me, the only question I'd have is around the name. To mirror `buffered` perhaps we could get the word "buffer" in the name somehow as well? Maybe `buffer_unordered`?

Stebalien:
Done. I think the name could still be better (dispatch? drive?) but at least it (mostly) parallels buffered now.

alexcrichton:
Ok cool, we may end up doing a pass for naming at a later date, but this seems great for now.

PR #137, opened 2016-09-09
Registering unpark is the responsibility of the callee

aidanhs:
Previously it was ambiguous whether the caller has a responsibility to do some registration (interpreting it like this is particularly confusing given `poll()` is meant to do the registration).

In practice, unpark is put in the Task `unpark` struct field which is called whenever `Task::unpark` is called, i.e. it has been registered for you.


alexcrichton:
Thanks!

PR #149, opened 2016-09-13
Describe oneshot return values in the right order.

jimblandy:
Correctly describe the first element of the tuple returned as the
`Complete`, and the second as the `Oneshot`. Spell out that the value is
passed to the `Complete`, and then the `Oneshot` resolves to it.


alexcrichton:
Thanks!

PR #150, opened 2016-09-13
Fix docs for Future::or_else.

jimblandy:
The prior docs seem to be copied from `and_then`. They incorrectly
suggest that the method operates on two pre-existing futures, and that
both will run. They don't point out `or_else`'s usefulness in
establishing fallbacks.


alexcrichton:
Thanks!

PR #172, opened 2016-09-22
Update futures::stream::iter docs to reflect implementation.

tmiasko:
Previously implementation of `futures::stream::iter` required items with
`Result` type from iterator. `Ok` has been mapped into a stream item,
and `Err` into a stream error.

Now `futures::stream::iter` is generic over stream error type, because
implementation never returns one. Iterator items are mapped directly
into stream items. This behaviour matches the existing documentation for
`futures::stream::iter` function.

### DIFF
--- a/src/stream/iter.rs
+++ b/src/stream/iter.rs
@@ -9,13 +9,11 @@ pub struct IterStream<I> {
     iter: I,
 }
 
-/// Converts an `Iterator` into a `Stream` which is always ready to yield the
-/// next value.
+/// Converts an `Iterator` over `Result`s into a `Stream` which is always ready
+/// to yield the next value.
 ///
 /// Iterators in Rust don't express the ability to block, so this adapter simply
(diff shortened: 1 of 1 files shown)


alexcrichton:
Thanks for the PR! This was intended to be an analog to `futures::done`, though, where you can express either errors or `Ok` with it. Perhaps this could be added as a separate adapter?

tmiasko:
Oh, so documentation is wrong, not implementation? 

> Additionally, the error type is generic here as it will never be returned,
> instead the type of the iterator will always be returned upwards as a
> successful value.

If so, then feel free to close this PR. Probably doesn't matter too much which
way it is fixed. You can easily get one functionality from the other either by
sticking map before `stream::iter`, or `and_then` after it.

alexcrichton:
Oh bah that's a typo holdover from a previous iteration, want to update this to fix the docs?

tmiasko:
Rebased and updated the docs.

alexcrichton:
Thanks!

PR #176, opened 2016-09-26
Return inner stream from `ForEach` when done.

dwrensha:
This allows resources to be reclaimed from the stream, similar to how `StreamFuture` returns its inner stream when the end is reached.

After this change, `cargo test` continues to pass for `tokio-core`, but there are some new "unused result which must be used" warnings.

### DIFF
--- a/src/stream/for_each.rs
+++ b/src/stream/for_each.rs
@@ -7,7 +7,7 @@ use stream::Stream;
 /// This structure is returned by the `Stream::for_each` method.
 #[must_use = "streams do nothing unless polled"]
 pub struct ForEach<S, F> {
-    stream: S,
+    stream: Option<S>,
     f: F,
 }
 
@@ -16,7 +16,7 @@ pub fn new<S, F>(s: S, f: F) -> ForEach<S, F>
           F: FnMut(S::Item) -> Result<(), S::Error>,
 {
     ForEach {
-        stream: s,
+        stream: Some(s),
         f: f,
     }
 }
@@ -25,15 +25,24 @@ impl<S, F> Future for ForEach<S, F>
     where S: Stream,
           F: FnMut(S::Item) -> Result<(), S::Error>,
 {
-    type Item = ();
+    type Item = S;
     type Error = S::Error;
 
-    fn poll(&mut self) -> Poll<(), S::Error> {
-        loop {
-            match try_ready!(self.stream.poll()) {
-                Some(e) => try!((self.f)(e)),
-                None => return Ok(Async::Ready(())),
+    fn poll(&mut self) -> Poll<S, S::Error> {
+        match self.stream {
+            Some(ref mut s) => {
+                loop {
+                    let r = try_ready!(s.poll());
+                    match r {
+                        Some(e) => try!((self.f)(e)),
+                        None => break,
+                    }
+                }
+
             }
+            None => panic!("poll a ForEach after it's done"),
         }
+
+        Ok(Async::Ready(self.stream.take().unwrap()))
     }
 }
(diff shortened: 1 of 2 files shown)


dwrensha:
Hm. Perhaps it would make sense to define a wrapper type `DrainedStream<S>` for this case and the `StreamFuture` case, in order to avoid spurious "unused result" warnings.

alexcrichton:
Hm yeah I feel like we haven't quite figure out the best story here in terms of ownership of streams and what combinators should do to be as flexible as possible. On one hand consumption in a context like this is costly due to the loss of potentially valuable resources, but we can also take a hit to ergonomics in some cases.

In this case though could the future just resolve to `S` itself instead of `Drained<S>`?

dwrensha:
> In this case though could the future just resolve to S itself instead of Drained<S>?

That's what my first commit does. I added `Drained` in the second commit because it seemed awkward to return a stream that's going to give a warning like "must use -- streams do nothing unless polled" when in fact it would be an error to `poll()` the returned stream.

> On one hand consumption in a context like this is costly due to the loss of potentially valuable resources, but we can also take a hit to ergonomics in some cases.

As far as I can tell, the ergonomic downside to my proposed changes is that you might occasionally need to add`.map(|_| ())`.

In the case I have in mind, my proposed changes have significant ergonomic upside. I want to read from a socket until EOF and then write something to the socket.  I want to do the reads through a `ForEach` and then get the socket back to do the write. If `ForEach` consumes its stream, then I need to split the socket or figure out some other way to keep it around for the write.

I admit that it's possible that the case I have in mind is unusual. Regardless, I think "don't unnecessarily drop things" is a pretty solid principle.

alexcrichton:
> That's what my first commit does. I added Drained in the second commit because it seemed awkward to return a stream that's going to give a warning like "must use -- streams do nothing unless polled" when in fact it would be an error to poll() the returned stream.

Oh good point. I think though that as arguments to closures and such it's fine to ignore, so I don't think the warning would crop up here, right?

---

I think you've sold me on these semantics, seems good to me!

dwrensha:
If I don't add the `Drained<S>` wrapper, the warning does crop up in the tokio-core examples:

```
warning: unused result which must be used: streams do nothing unless polled, #[warn(unused_must_use)] on by default
  --> examples/sink.rs:35:5
   |
35 |     l.run(server).unwrap();
   |     ^^^^^^^^^^^^^^^^^^^^^^^

warning: unused result which must be used: streams do nothing unless polled, #[warn(unused_must_use)] on by default
  --> examples/hello.rs:24:5
   |
24 |     core.run(server).unwrap();
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^
```

Moreover, I like that `type Item = Drained<S>;` succinctly conveys the semantics, while I think that `type Item = S;` would possibly be confusing.

alexcrichton:
Hm that seems reasonable to me actually, I wouldn't mind adding some calls to `drop`. I'd prefer to not expand the surface area with `Drained` as it seems like it may be an otherwise unnecessary wrapper.

dwrensha:
OK. Please feel free to only pull in the first commit (https://github.com/alexcrichton/futures-rs/pull/176/commits/82afb2e62761a954041ff5313cc90450b04f193d) if you think that would be better.

carllerche:
I actually think that this can be done without any loss of ergonomics by adding inherent fns to individual future structs.

For example, `ForEach` could add:

```rust
impl<T> ForEach<T> {
    pub fn with_inner(self) -> WithInner<Self> {
        WithInner::new(self)
    }
}
```

such that `WithInner` polls the future (`ForEach` in this case) but then returns the final value plus `ForEach::into_inner`.

rphmeier:
@carllerche great idea! although I would recommend avoiding using `with` in the name to avoid confusion with `Sink::with`. maybe `reclaim_inner`?

carllerche:
I don't have a ton of great ideas around how to name this combinator :(

alexcrichton:
Ok now merged into 0.2!

PR #178, opened 2016-09-26
Annotate `CpuFuture` with `#[must_use]`.

tmiasko:
When dropped `CpuFuture` will try to cancel the computation,
thus not using it is most likely an error.


alexcrichton:
Thanks!

Issue #166, opened 2016-09-20
Lifecycle hooks for cpupool


tailhook:
Hi,

To get some useful statistics on thread pools I need to keep track which thread ids belong to which CpuPool. Probably it can provide lifecycle hooks to threads in a pool. I.e. run some user code after spawning a new thread and before it exists.

I think this might require turning CpuPool constructor to a builder/fluent.

What do you think? Will such PR be accepted? Any alternate ideas?

alexcrichton:
In general `futures-cpupool` was meant to be a simple proof-of-concept of what thread-pool-based futures could look like. If fine-grained management of a thread pool is required, however, then it may be relatively easy to hand-roll your own. The crate is quite small after all!

That being said, if the changes are relatively unobtrusive, I wouldn't mind looking them over!

alexcrichton:
Added in #180

PR #180, opened 2016-09-26
A `Builder` structure for CpuPool

tailhook:
Here is an attempt to solve #166. It's without tests just to show you the API.

From my perspective:

1. I'd remove `CpuPool::new` entirely, because it's as easy as `Builder::new().create()` in most cases. But I'm okay with keeping it.
2. As we have a `Builder` I would add `stack_size` and `name` from the `std::thread::Builder` (I'm not sure is it fine to keep all threads in pool with the same name?)

While @alexcrichton says this crate is just an example implementation I think this is going to be default pool implementation, and it will be hard to convince anyone (i.e. an author of a library that uses thread pool) that they needs to change pool implementation just to make it easier to monitor an application using the library.

So should I continue this code?

### DIFF
--- a/futures-cpupool/src/lib.rs
+++ b/futures-cpupool/src/lib.rs
@@ -64,8 +64,23 @@ use futures::task::{self, Run, Executor};
 ///
 /// Currently `CpuPool` implements `Clone` which just clones a new reference to
 /// the underlying thread pool.
+///
+/// **Note:** if you use CpuPool inside a library it's better accept a
+/// `Builder` object for thread configuration rather than configuring just
+/// pool size.  This not only future proof for other settings but also allows
+/// user to attach monitoring tools to lifecycle hooks.
 pub struct CpuPool {
     inner: Arc<Inner>,
+}
+
+/// Thread pool configuration object
+///
+/// Builder starts with a number of workers equal to the number
+/// of CPUs on the host. But you can change it until you call `create()`.
+pub struct Builder {
+    pool_size: usize,
+    after_start: Option<Arc<Fn() + Send + Sync>>,
+    before_stop: Option<Arc<Fn() + Send + Sync>>,
 }
 
 struct Sender<F, T> {
@@ -84,6 +99,8 @@ struct Inner {
     queue: MsQueue<Message>,
     cnt: AtomicUsize,
     size: usize,
+    after_start: Option<Arc<Fn() + Send + Sync>>,
+    before_stop: Option<Arc<Fn() + Send + Sync>>,
 }
 
 /// The type of future returned from the `CpuPool::spawn` function, which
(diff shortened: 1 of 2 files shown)


tmiasko:
Another option would be to provide a custom thread factory, i.e., spawn
function. This would give control over stack size, thread names, and pre-post
hooks for free.

tailhook:
@tmiasko while functionally it's equivalent, it's still more complex interface. Let's imagine a static serving crate, that need CpuPool, in one case it's configured as:

```rust
mount.mount("/js/", Static::new(Path::new("public/js/"), cpupool::Builder::new()));
```

In other case, `Static::new` constructor needs to accept both the pool size and a function with a signature of (untested):
```
fn new<F, J>(path: &path, size: usize, fun: Option<F>)
  where F: Fn(J), J: Fn();
```

While it might accept ready to use CpuPoll there are [precedents of not doing that](https://github.com/sbstp/tokio-dns/blob/master/src/resolver.rs#L28) for some reason.

tmiasko:
Whether builder pattern is used or not is orthogonal to providing spawn as
customization point, and I didn't intend to support one way or over another.
What I wanted to indicate, was possibility of replacing most of those
customization points (stack size, after start hook, before stop hook, thread
name?) with a single one. The spawn function also has additional benefit that
it nicely aligns with providing sequence of different thread names like
"cpu-thread-N".

I think you also raise important question with regards how this CpuPool should
be used in other crates - whether it should be already constructed, or not. In
general, I would rather expect that pool would come from the outside, unless
it has to satisfy some very specific requirements. Potentially, a some new
trait would be nice to avoid tightly coupling everything with CpuPool,
especially that interface surface area is actually very small.

futures-cpupool/src/lib.rs
```diff
@@ -188,12 +202,14 @@ impl CpuPool {
 }
 
 fn work(inner: &Inner) {
+    inner.after_start.as_ref().map(|fun| fun());
```
liran-ringel: I think "fun" should be renamed to "func".
It's only 1 more character, but much more clear.

alexcrichton:
Ok, looks good to me! I'd be fine exploring this and seeing what we can add here. Were there more pieces you'd like to add?

tailhook:
Added a test. This should be good enough for the start. Other things may be added later.

futures-cpupool/src/lib.rs
```diff
@@ -252,3 +268,61 @@ impl<F: Future> Future for Sender<F, Result<F::Item, F::Error>> {
...
+    pub fn new() -> Builder {
+        Builder {
+            pool_size: num_cpus::get(),
+            after_start: None,
+            before_stop: None,
+        }
+    }
+
+    /// Set size of a future CpuPool
+    pub fn pool_size(mut self, size: usize) -> Self {
```
alexcrichton: Could these functions take `&mut self` and return `&mut Self`?

futures-cpupool/src/lib.rs
```diff
@@ -252,3 +268,61 @@ impl<F: Future> Future for Sender<F, Result<F::Item, F::Error>> {
...
+    /// This is initially intended for bookkeeping and monitoring uses
+    pub fn before_stop<F>(mut self, fun: F) -> Self
+        where F: Fn() + Send + Sync + 'static
+    {
+        self.before_stop = Some(Arc::new(fun));
+        self
+    }
+
+    /// Create CpuPool with configured parameters
+    pub fn create(&self) -> CpuPool {
```
alexcrichton: To be maximally flexible we may want this method to take `&mut self`

futures-cpupool/src/lib.rs
```diff
@@ -252,3 +268,61 @@ impl<F: Future> Future for Sender<F, Result<F::Item, F::Error>> {
...
+    /// of CPUs on the host.
+    pub fn new() -> Builder {
+        Builder {
+            pool_size: num_cpus::get(),
+            after_start: None,
+            before_stop: None,
+        }
+    }
+
+    /// Set size of a future CpuPool
```
alexcrichton: Perhaps this could mention that "size" is the number of threads that will be spawned?

futures-cpupool/src/lib.rs
```diff
@@ -252,3 +268,61 @@ impl<F: Future> Future for Sender<F, Result<F::Item, F::Error>> {
...
+    pub fn pool_size(mut self, size: usize) -> Self {
+        self.pool_size = size;
+        self
+    }
+
+    /// Execute function `fun` before each worker thread
```
alexcrichton: Could this clarify that "before" is "right after the thread is spawned but before it does work"?

Also, typically closure variables are called things like `f` or `g` rather than longer names like "fun" I've seen at least

alexcrichton:
Looks good to me, thanks!

PR #189, opened 2016-10-03
Allow public access to the channel Sender error type

proman21:
Just a quick change to allow users of this crate to access the `SendError` type that `FutureSender` uses.


alexcrichton:
Oops, thanks!

PR #193, opened 2016-10-05
Move oneshot away from `Slot`

alexcrichton:
* Reduces allocations
* Simplifies implementation
* Easier to read/write
* `Slot` is going away for channel anyway, no need to share implementation


alexcrichton:
cc @aturon, you might be interested in the implementation here

PR #196, opened 2016-10-08
simplify the docs example for fold()

oconnor663:
The previous example included a channel stream and some and_then()
recursion, which was pretty cool, but it took me an hour to understand
what was happening :-D


alexcrichton:
Thanks!

Issue #188, opened 2016-10-02
Stream::take and Stream::skip mismatch


tmiasko:
`Stream::take` includes errors in the number of elements it will take from the
stream, on the other hand `Stream::skip` does not include errors in the number
of elements skipped.

In both cases I think it would be better to apply those limits only to
`Stream::Item`s and just pass all errors through without counting them in.

alexcrichton:
Oh right, now I'm remembering why `take` didn't include errors originally.

I'm of two minds about this. On one hand it seems conceptually correct to consider an error part of take/skip, but on the other hand it seems bad to *knowingly* drop errors on the ground.

@aturon @carllerche, thoughts about what the most appropriate behavior here is?

NeoLegends:
Since most of the combinators only work on the `Item`s and not on the `Error`s, I would make `Stream::skip` and `Stream::take` behave the same.

tmiasko:
Let me expand a little bit on rationale for suggested behaviour, because I
didn't include it in initial comment. 

In general I would expect `Stream::take(n)` and `Stream::skip(n)`, for given
`n`, to be complementary with respect to produced results. They could be
complementary with respect to both stream items and stream errors, but this has
unfortunate effect of ignoring some of the errors in skip as mentioned by Alex.

Thus, given special treatment errors receive in stream combinators overall, the
second best thing to do, would be making them complementary with respect to
items produced while passing all stream errors as is.

alexcrichton:
Fixed by #199

PR #199, opened 2016-10-09
Pass through all errors in both Stream::take and Stream::skip until stream is done.

tmiasko:
Fixes issue #188.


alexcrichton:
Thanks @tmiasko!

PR #201, opened 2016-10-09
fix some copy/paste errors in docs

dwrensha:
It looks like some of the `Stream` docstrings were copied from those for `Future`, but not all instances of "future" were updated to stream.


alexcrichton:
Heh, thanks!

PR #207, opened 2016-10-14
add select_any impl

bluejekyll:
@alexcrichton this is the select_any as discussed.

I got rid of the index from the result, b/c it didn't seem worth it to track for the last error. It was an issue due to the skipping of errors, which result in those being removed from the inner Vec.

I added the loop { ... } so this should be good. Otherwise, still pretty close to the select_all interface.

### DIFF
--- a/src/select_any.rs
+++ b/src/select_any.rs
@@ -1,4 +1,4 @@
-rm use std::mem;
+use std::mem;
 use std::prelude::v1::*;
 
 use {Future, IntoFuture, Poll, Async};
--- a/src/select_any.rs
+++ b/src/select_any.rs
@@ -1,0 +1,77 @@
+use std::mem;
+use std::prelude::v1::*;
+
+use {Future, IntoFuture, Poll, Async};
+
+/// Future for the `select_any` combinator, waiting for one of any of a list of
+/// futures to succesfully complete. unlike `select_all`, this future ignores all
+/// but the last error, if there are any.
+///
+/// This is created by this `select_any` function.
+#[must_use = "futures do nothing unless polled"]
+pub struct SelectAny<A> where A: Future {
+    inner: Vec<A>,
+}
+
+/// Creates a new future which will select the first successful future over a list of futures.
+///
+/// The returned future will wait for any future within `list` to be ready and Ok. Unlike
+/// select_all, this will only return the first successful completion, or the last
+/// failure. This is useful in contexts where any success is desired and failures
+/// are ignored, unless all the futures fail.
+///
+/// # Panics
+///
+/// This function will panic if the iterator specified contains no items.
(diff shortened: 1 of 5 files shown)


src/select_any.rs
```diff
@@ -0,0 +1,77 @@
...
+    inner: iter.into_iter()
+    .map(|a| a.into_future())
+    .collect(),
+  };
+  assert!(ret.inner.len() > 0);
+  ret
+}
+
+impl<A> Future for SelectAny<A> where A: Future {
+  type Item = (A::Item, Vec<A>);
```
alexcrichton: Mind bumping these up to 4-space tabs?

alexcrichton: Ah I personally avoid this in repos for now, but when rustfmt comes along hopefully that'll help!

alexcrichton:
Ok, my only remaining qualm here is the name. I feel like `select_any` is a little too similar to `select_all` to move forward with this naming unfortunately. That is, it's kind of unclear why `any` might wait for all futures (currently) where `all` will only ever wait for at most one future.

Maybe something like `select_ok`, though, to emphasize that it explicitly tries to select an `Ok` variant?

bluejekyll:
Yeah, i've been thinking about the naming for a little bit. The reasoning behind the `any` was derived from `any` in Iterator. It's basically meant to say, select any ok result, otherwise it is err. This is the same definition of `any` in Iterator, that is it is true if anything in the iterator is true.

Now the complicated thing is that in this context `select_all`, means select across all futures in a vector and return the first that completes, which matches the definition of `select`. `all`, though, does not necessarily fit this context, at least not as defined on Iterator, which requires that all elements must be successful. For that in futures-rs, there is `collect`, but `collect` does not function like `select`, and also is implemented as requiring each `Future` to complete in sequence, not in parallel like `select`. This leads me to think that `select_all` is possibly misnamed as well :). It also makes me think that `join_all` is possibly missing as well, though the location I would possibly use it is actually better implemented with a `select_all` with a loop since I need to process all results, doing something with errors and successes...

Anyway, long story short, I'm happy to change the name, perhaps dropping `select` is the correct option, because given current semantics, that's the one that `select_any` matches least. What do you think of just a bare `any`? or `any_ok`

carllerche:
`select_ok` maybe?

alexcrichton:
How about `first_ok`? I tried to keep the name "select" with `select_all` to be similar to `Future::select`, but it's true that `collect` should actually be `join_all` if that's the case. I don't mind keeping the "breakaway" though from the standard names on `Future` to get a more invocative name.

bluejekyll:
I'm leaning towards `first_ok` but `select_ok` is good too.

It's your library, you decide ;)

alexcrichton:
Ok, let's go with `select_ok` as I'd probably like to rename `collect` to `join_all` to keep a similar pattern there.

alexcrichton:
Thanks!

Issue #210, opened 2016-10-16
Debounce / sample stream combinators


NeoLegends:
Having a sample and debounce stream combinator would be really nice-to-have, especially in scenaria with user interaction. This would bring the library closer to other FRP-implementations.

Debounce works by ignoring values of a stream until a certain amount of time has passed without any new events. It is used by e.g. web developers when processing user input and ignoring the changes while the user is typing.
Sample works in a similar fashion: instead of ignoring all values the stream produces, it passes on values from the input stream at specified intervals. It can be used to ignore most intermediate values, yet provide the user with some feedback from time to time.

I don't know whether the suggestion is in the reach of this library since the implementations of both combinators are probably going to require a timer and an event loop, but this can be figured out, I guess.

alexcrichton:
Sounds plausible to me! If it can't be implemented here then these may perhaps be best implementable at the tokio-core layer which has more access to scheduling and timers and such.

NeoLegends:
Great! Since most FRP libraries feature a ton of combinators (_way_ more than futures-rs) and I'm sure not all of those combinators are applicable in Rust or don't really fit well into the zero-cost model, I suggest taking a look at http://reactivex.io/rxjs/class/es6/Observable.js~Observable.html or https://msdn.microsoft.com/en-us/library/system.reactive.linq.observable(v=vs.103).aspx to get an idea of what is possible and which combinators could be ported.

NeoLegends:
When the combinators to be ported are figured out, I could definetely start implementing some of them.

alexcrichton:
Oh I at least personally don't mind implementing some more heavyweight combinators. Just because the core `Future` trait is zero-cost doesn't mean that everything around it needs to be as well! 

To that end I'd be down for basically implementing any useful combinator you've seen from other libraries, in my mind they're all welcome!

alexcrichton:
Yeah if we want to implement these at the tokio-core layer we'd want an extension trait (or perhaps just free function constructors). I'd be fine exploring though to see if we can implement them at the futures layer as well. With `std::time::Instant` going all the way to tokio-core may not be necessary? Unsure though.

NeoLegends:
I already thought about implementing the combinators only with `std::time` primitives, but the problem with that is that e.g. debounce needs to fire its last result _after_ the timeout has passed (asynchronously). This means we cannot rely on the underlying stream to notify the task of new items, which is why we're going to need a timer that `unpark()`s the task at the right time.

[...]

PR #211, opened 2016-10-18
Implement buffer combinator

NeoLegends:
This is the first combinator of #210 I'd like to have added to futures-rs. I'm kinda unsure about the name as it might be confusing in regards to `buffered` and `buffer_unordered` but since buffering is what it does, I just went with the name.

### DIFF
--- a/src/stream/buffer.rs
+++ b/src/stream/buffer.rs
@@ -21,6 +21,8 @@ pub struct Buffer<S>
 pub fn new<S>(s: S, capacity: usize) -> Buffer<S>
     where S: Stream
 {
+    assert!(capacity > 0);
+    
     Buffer {
         capacity: capacity,
         items: Vec::with_capacity(capacity),
--- a/src/stream/buffer.rs
+++ b/src/stream/buffer.rs
@@ -35,27 +35,25 @@ impl<S> Stream for Buffer<S>
     type Error = <S as Stream>::Error;
 
     fn poll(&mut self) -> Poll<Option<Self::Item>, Self::Error> {
-        let maybe_next = try_ready!(self.stream.poll());
-
-        if let Some(item) = maybe_next {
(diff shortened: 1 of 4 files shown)


NeoLegends:
Hmm, seems like sometimes the test works and sometimes it doesn't.

src/stream/buffer.rs
```diff
@@ -0,0 +1,62 @@
...
+    fn poll(&mut self) -> Poll<Option<Self::Item>, Self::Error> {
+        let maybe_next = try_ready!(self.stream.poll());
+
+        if let Some(item) = maybe_next {
+            // Push the item into the buffer and check whether it is
+            // full. If so, replace our buffer with a new and empty one
+            // and return the full one.
+            self.items.push(item);
+            if self.items.len() < self.capacity {
+                Ok(Async::NotReady)
```
tmiasko: When NotReady is returned from poll, the Stream implementation is obliged to
notify the task once a new item may be ready. As no notification of any kind is
send there, a task will wait and wait, without calling poll ever again.

What should be done here instead: the Buffer::poll should continue calling the
pull of underlying stream until NotRead is returned. Effectively offloading the
responsibility for notification to the underlying stream.

NeoLegends:
Of course! Forgot about that.

src/stream/buffer.rs
NeoLegends: Of course! Forgot about that.

alexcrichton:
Nice! 

Perhaps this could be called `chunks` as I think it's performing similar functionality to `std::slice::chunks`? Also, could you add a test or two for the various edge cases here?

NeoLegends:
I added an assert and a test for a capacity of zero. Which other edge cases come to your mind that are not already covered in the tests?

alexcrichton:
Looks good to me, thanks!

PR #216, opened 2016-10-28
[WIP] Scalable buffer_unordered

tailhook:
This is an alternative to https://github.com/tokio-rs/tokio-core/pull/60

While this is much faster, I'm not sure it's as scalable, here are some benchmarks.

So I'm testing a [redis server implementation](https://github.com/tailhook/tokio-tutorial/blob/05acb04df8e155312f6b13ffde6896a79a97f84a/redis_bu/src/main.rs#L61-L65).

The method is the following. I'm running a bunch of redis-benchmark processes with `-I` flag:

```
redis-benchmark -c 20000 -I -p 7001 -h 127.0.0.$i
```

It creates inactive connections (we need to use `127.0.0.x` IPs to avoid the limit for ephemeral ports) 

And then benchmark GET,SET on usually 50 connections:

```
redis-benchmark -c 1000 -p 7001 -n 100000 GET,SET
```
## Benchmark Results

| Test | RPS | Idle | Active |
| --- | --- | --- | --- |
| This patch | 129836.41 | 0 | 50 |
| This patch | 62770.70 | 40k | 50 |
| This patch | 46593.98 | 60k | 50 |
| This patch | 36387.45 | 80k | 50 |
| original | 3418.80 | 0 | 50 |
| original | 42698.55 | 0 | 1000 |
| original | 2477.70 | 20k | 50 |
| original | 36697.25 | 20k | 1000 |
| tokio-core | 133280.03 | 0 | 50 |
| tokio-core | 127877.23 | 80k | 50 |

As you can see with the patch applied performance matches the performance of tokio core. Still it doesn't scale well with increasing the number of idle connections.

Other interesting observations is that before this patch performance increases with the number of active connections (well, It's understandable in the retrospective).
## Implementation

While the implementation is far from being perfect, it passes the tests and the only inefficiency I can see is [here](https://github.com/alexcrichton/futures-rs/compare/master...tailhook:fast_buffer_unordered?expand=1#diff-236773cc553adb1ea6bfcc851419c09dR88). But it would only influence benchmarks if they were closing connections often. These benchmarks keep keep-alive connection for the whole duration of the test.

Another issue is that `buffer_unordered` is probably not very good combinator for this use case because we don't want to return any result at all (and non-returning a value allows us to fix the bug above more efficiently). So should it be something like `consume()` ?

### DIFF
--- a/src/stream/buffer_unordered.rs
+++ b/src/stream/buffer_unordered.rs
@@ -1,7 +1,12 @@
 use std::prelude::v1::*;
+use std::sync::Arc;
+use std::collections::VecDeque;
+
+use task::{self, EventSet, UnparkEvent};
 
 use {Async, IntoFuture, Poll, Future};
 use stream::{Stream, Fuse};
+use stack::Stack;
 
 /// An adaptor for a stream of futures to execute the futures concurrently, if
 /// possible, delivering results as they become available.
@@ -17,6 +22,7 @@ pub struct BufferUnordered<S>
     stream: Fuse<S>,
     futures: Vec<Option<<S::Item as IntoFuture>::Future>>,
     cur: usize,
+    stack: Arc<Stack<usize>>,
 }
 
 pub fn new<S>(s: S, amt: usize) -> BufferUnordered<S>
@@ -27,6 +33,7 @@ pub fn new<S>(s: S, amt: usize) -> BufferUnordered<S>
         stream: super::fuse::new(s),
         futures: (0..amt).map(|_| None).collect(),
         cur: 0,
+        stack: Arc::new(Stack::new()),
     }
 }
 
@@ -39,11 +46,12 @@ impl<S> Stream for BufferUnordered<S>
 
     fn poll(&mut self) -> Poll<Option<Self::Item>, Self::Error> {
         // First, try to fill in all the futures
-        for future in &mut self.futures {
+        for (idx, future) in self.futures.iter_mut().enumerate() {
             if future.is_none() {
                 match try!(self.stream.poll()) {
                     Async::Ready(Some(s)) => {
                         *future = Some(s.into_future());
+                        self.stack.push(idx);
                     }
                     Async::Ready(None) => break,
                     Async::NotReady => break,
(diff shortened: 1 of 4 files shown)


src/stream/buffer_unordered.rs
```diff
@@ -39,11 +46,12 @@ impl<S> Stream for BufferUnordered<S>
 
     fn poll(&mut self) -> Poll<Option<Self::Item>, Self::Error> {
         // First, try to fill in all the futures
-        for future in &mut self.futures {
+        for (idx, future) in self.futures.iter_mut().enumerate() {
```
alexcrichton: I wonder if this is perhaps why the perf isn't scaling? If this is full of idle connections we'll always fly over this array looking for more but we won't find any. Perhaps this could be converted to a `Slab`-style where we can check if a slot is open in O(1) time?

src/stream/buffer_unordered.rs
```diff
@@ -54,25 +62,31 @@ impl<S> Stream for BufferUnordered<S>
...
+                        Err(e) => Err(e),
+                    }
                 },
                 None => continue,
             };
-            self.cur = i + 1;
             *future = None;
+            // push unresolved futures back
+            for idx in drained {
+                self.stack.push(idx);
```
alexcrichton: This may also be related to the scalability problem, this means that if there are N ready connections that we'll do N^2 operations to actually process them I think. It should be possible to implement an O(1) append I think with the stack here, though?

tailhook:
Well, I'm feeling stupid (my excuse is that there were no coffee at hack place :) )

Added `Slab` to the `buffer_unordered` and got `141k - 146k` on the same machine with difference being not significant in all tests.

Basically, its now faster that using `spawn` from tokio for some reason (I have few guesses, but..)

So questions left are:

1. What to do with slab? Can we depend on it? Should this combinator be in a separate crate? Should we embed a stripped down implementation of a slab?
2. Should we have a `consume` variant (without a return value)?
3. We need Box for `stack`. How can we use it in stable rust with `no_std`? (I'm not good with `no_std` so maybe I'm missing something, but `alloc` crate, which originally contains box, is unstable)

tailhook:
And the memory usage:

|Mem (kb) | Conns    |                                 |
|---------|----------|---------------------------------|
|403124   | 20k      | Buffer unordered                |
|318536   | 20k -> 0 | After all connections are closed|
|500612   | 40k      | Buffer unordered                |
|122144   | 20k      | Spawned                         |
| 23180   | 20k -> 0 | After all connections are closed|
|223548   | 40k      | Spawned                         |

While overall memory usage is disappointing (~5k per idle connection), this patch doesn't change the per-connection memory usage too much (I need more tests to measure real difference). The constant part + 300Mb is probably because of a slab for 1M connection we have configured in example.

alexcrichton:
> Basically, its now faster that using spawn from tokio for some reason (I have few guesses, but..)

Whoa! That's unexpected... I wonder if it's because there's only one virtual dispatch (to the one stream) vs N virtual dispatches (one per stream). In any case, promising numbers!

> What to do with slab? Can we depend on it? Should this combinator be in a separate crate? Should we embed a stripped down implementation of a slab?

Let's vendor a stripped down version for now and see how far we can get with that.

> Should we have a consume variant (without a return value)?

Hm I'm not sure I follow, can you remind me what this is?

> We need Box for stack. How can we use it in stable rust with no_std? (I'm not good with no_std so maybe I'm missing something, but alloc crate, which originally contains box, is unstable)

Ah many combinators just aren't available in `no_std`, so this combinator can just be turned off in the `no_std` case.

---

Also I had an idea for dealing with leftover nodes in the stack that we drained if a stream resolves. We should just have `pending: Option<Drain<T>>` on the stream itself. That way we never need to re-enqueue and it should remain just as efficient in theory!

alexcrichton:
Ok! I've cleaned up a few pieces here and there and otherwise landed this now with small vendorings of `Stack` and `Slab`. I added a test as well and covered a few other use cases I believe. Let me know if it doesn't work out though!

tailhook:
Great. Thanks!

(it's not like I didn't want to finish it, I'm on vacation all this time)

alexcrichton:
No worries! Was just doing some futures work this weekend and figured I could help land it :)

Issue #214, opened 2016-10-20
[channels] Provide a MPSC channel


carllerche:
Currently, channel is SPSC, which is fairly limiting. It would be quite helpful to provide a more versatile channel.

This relates to #213

alexcrichton:
Closing in favor of https://github.com/alexcrichton/futures-rs/issues/229 (which has a possible solution)

PR #218, opened 2016-10-31
feat(mpsc): add an mpsc channel that implements back pressure

compressed:
I needed an mpsc channel and saw that #214 was opened, so I tried to implement this myself. I'm still quite new to the futures library, but have done a lot of testing around this implementation, checking for race conditions, etc and am hopeful it is correct now. Hopefully my comments help explain how it works!

I pulled in the crossbeam crate to use the `AtomicOption` and `MsQueue` data structures to make the implementation a bit easier.

I also re-used your tests from the spsc channel and added one more. I made a modification though to the sequence test case, moving those first two `sassert_empty(&mut rx);` checks their own new test case since it caused some dead tasks to be stored on the `tx_task` list which then didn't allow the test to proceed anymore (deadlock because the Receiver couldn't wake up the Sender).

Anyways, not sure if this it what you were looking for, but figured I would submit it. I'm open to any feedback. Thanks!

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,6 +16,10 @@ composability, and iterator-like interfaces.
 [dependencies]
 log = { version = "0.3", default-features = false }
 
+[dependencies.crossbeam]
+optional = true
+version = "0.2.10"
+
 [features]
+default = ["use_std", "crossbeam"]
 use_std = []
-default = ["use_std"]
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -207,6 +207,8 @@ pub use select::{Select, SelectNext};
 pub use then::Then;
 
 if_std! {
+    extern crate crossbeam;
+
     mod lock;
     pub mod task;
 
--- a/src/stream/mod.rs
+++ b/src/stream/mod.rs
@@ -64,6 +64,8 @@ if_std! {
     mod buffer_unordered;
     mod catch_unwind;
     mod channel;
+    /// a mpsc channel
+    pub mod mpsc;
     mod chunks;
     mod collect;
     mod wait;
--- a/src/stream/mpsc.rs
+++ b/src/stream/mpsc.rs
@@ -1,0 +1,411 @@
+use std::any::Any;
+use std::error::Error;
+use std::fmt;
+use crossbeam::sync::{AtomicOption, MsQueue};
+use std::sync::Arc;
+use {Async, Future, Poll};
+use std::sync::atomic::AtomicUsize;
+use std::sync::atomic::Ordering::SeqCst;
+use stream::Stream;
+use task::{self, Task};
+
+const EMPTY: usize = 0;
+const DATA: usize = 1;
+const GONE: usize = 2;
+const UNPARKED: usize = 3;
+const PARKED: usize = 4;
+
+/// `Sender` of the channel. `Sender` can be cloned.
+pub struct Sender<T, E> {
+    /// `Inner`
+    inner: Arc<Inner<T, E>>,
+    /// Count of senders, needed for determining when the last sender is GONE.
+    sender_count: Arc<AtomicUsize>,
+}
+
+impl<T, E> Drop for Sender<T, E> {
+    fn drop(&mut self) {
+        // Each cloned sender calls `Drop` so we need to ensure that only the last sender being
+        // dropped actually sets the `GONE` state.
+        // The result is discarded because it could be updated in-between this line and the next.
+        let _ = self.sender_count.fetch_sub(1, SeqCst);
+        // if it was one before, it *should* be zero now, but let's check just to make sure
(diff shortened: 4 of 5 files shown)


alexcrichton:
Thanks for the PR!

Glancing at the implementation though, it wasn't quite apparent to me where the backpressure came into effect? I've also thought in the past that a mpsc queue with "backpressure" may not make much sense unfortunately. Presumably `NotReady` is returned at some point to signal backpressure, but the `Future` contract means that a task needs to get notified when it would otherwise be ready to go. This means that a channel would have to accommodate an unbounded number of senders attempting to send messages, and store handles to all of them to wake them up when its their turn. This then generates the same problem as before, runaway memory in terms of storing the task handles.

In light of that I'd personally expect a mpsc queue to be very similar to `std::sync::mpsc`, just futures-powered. That is, `send` returns immediately (doesn't return a future) and otherwise `Receiver` still implements `Stream`. Does that make sense? I believe we could actually just lift all the `shared` code in the standard library and apply it here to the futures crate almost blindly, so it in theory would be a good fit with the same API.

carllerche:
I was thinking about mpsc w/ backpressure the other day. The issue is that there needs to be a slot in which to store the task handle when the producer blocks. It seems like calling `Sender::clone` could also allocate this slot. In which case, if the channel is at capacity, the sender would enqueue it's waiter node...

The algorithm that the consumer uses to wake up producers will most likely be complicated, but I assume no more so than any existing concurrent algorithm...

compressed:
Thanks for the comments. In my implementation, the number of outstanding sending tasks is bounded by the number of `Sender` clones. Similar to the existing spsc channel implementation, each `Sender` will be consumed when you call `send` and returns a `FutureSender`. When the `FutureSender` is polled it will either send data (and then return the `Sender`), park, or error because the `Receiver` is gone. I also enforce than any given `FutureSender` can have no more than one task (if it has already stored a task, it won't store another one if the `Receiver` is still busy when it is polled again).  The sending tasks are stored in an `MsQueue` (https://github.com/alexcrichton/futures-rs/pull/218/files#diff-d478aefb513a31350d5fdb19589b417bR77).

I don't believe it's possible to use a just one sender task and still permit `Sender` to be cloned. I'll think about it...

Ironically, your comment about using `std::sync::mpsc` was my first implementation, but then I attempted to implement it using backpressure and wound up diving into the all this atomic fun 😄 ! I can move towards that approach though if that seems to be the best way to proceed.

alexcrichton:
Yeah it's true that we could track clones to manage bits and pieces, but I'm not sure that it's really the backpressure that we want? That is, it's still easy to exhaust memory by just cloning a bunch and spawning more tasks with more handles. 

I feel that for now we don't have a strong enough story around the backpressure here that we should stick to the simple mpsc model like in the standard library. We could perhaps add a fancier queue later but it'd likely end up being much different than this. Additionally, this is often "good enough" for use cases anyway.

carllerche:
Something will have to constrain the number of tasks spawned for producers. It would be totally fine to allocate a slot per sender as that would be constant per producer task (which has a constrained number managed by whatever spawns the producer task). The main goal of the back pressure is that when the consumer gets backed up, it is able to signal to the producers...

The only tricky bit is coordinating how to unpark waiting producers, but that is most likely a standard condvar algorithm

alexcrichton:
True yeah, the number of producers is likely more constant than the number of messages in flight. I talked with @aturon briefly yesterday about the "who to wake up" problem and I feel that currently we pessimistically must wake up all blocked senders if a slot becomes available, which may be unfortunate.

This is why I lean towards just punting on this for now and perhaps reconsidering for 0.2. There's a lot of questions with backpressure and it's not clear to me that this would be crucial to solving problems, so in the meantime we can get something that works 90% of the time and get experience with that.

compressed:
Hey sorry, I was away for a few days. I see there is another discussion about back pressure design for mpsc. Did you still want to move forward with a no back pressure version based on the standard library mpsc? I can close this PR then and submit a new one using that approach.

alexcrichton:
@compressed yeah I think we'll still want a version without backpressure as an option, so I'd be fine still pursuing.

carllerche:
I commented in the other issue, I provided an initial implementation of an mpsc channel: https://github.com/carllerche/futures-mpsc/blob/master/src/lib.rs

The channel supports handling parking senders when the buffer is full, but it also supports an unbound variant by setting the buffer size to `usize::MAX`.

It also has the ability to shutdown the receive half without dropping the receiver. This enables the ability for the receiver to perform a clean shutdown ensuring that there are no messages stuck in the channel.

I also have thoughts on how to support `poll_cancel` on the `Sender`, enabling senders to be notified when the receiver shuts down (aka, senders don't have to fail sending a message to be notified of the receiver going away).

alexcrichton:
Ok, I'm going to close this in favor of https://github.com/alexcrichton/futures-rs/pull/238 and https://github.com/carllerche/futures-mpsc/blob/master/src/lib.rs

PR #219, opened 2016-11-01
Use SeqCst instead of Acquire and Release in Lock.

dwrensha:
This fixes some deadlocking behavior that I noticed in https://github.com/dwrensha/zillions (specifically, the stress test in the [debug branch](https://github.com/dwrensha/zillions/tree/debug), built in release mode).

@alexcrichton helped me to determine that the problem appears to have been introduced in https://github.com/alexcrichton/futures-rs/commit/2ece87ddeb372b384c9a29012fa77851123e085d.

What seems to happen is this:
  1. Thread A is waiting on `oneshot1.join(oneshot2)`.
  2. Thread B calls `complete1.complete(())` and begins calling `complete2.complete(())`.
  3. The first notification reaches thread A, and both oneshots get polled.
  4. During `oneshot2.poll()` [while `self.rx_task()` is locked](https://github.com/alexcrichton/futures-rs/blob/43e290f02782321637947119342e45520a426e24/src/oneshot.rs#L243-L246), `complete2` gets dropped, and [does not `unpark()` anything](https://github.com/alexcrichton/futures-rs/blob/43e290f02782321637947119342e45520a426e24/src/oneshot.rs#L198-L202), because it cannot acquire the lock.
  5. The next step in `oneshot2.poll()` [is to check `self.inner.complete` again](https://github.com/alexcrichton/futures-rs/blob/43e290f02782321637947119342e45520a426e24/src/oneshot.rs#L257). If everything were sequentially consistent, then it would necessarily be `true`, and the future could return the completed value. However, for some reason `oneshot2` again observes `self.inner.complete` to be `false`, and returns `NotReady`.
  6. At this point, there is no remaining way for `oneshot2` to get notified that it is ready, so it hangs forever.

According to my reading of [cppreference.com](http://en.cppreference.com/w/cpp/atomic/memory_order#Release-Acquire_ordering), the `Release`/`Acquire` ordering on `Lock::locked` does not in fact require thread A to see `true` for `self.inner.complete` in step (5) above. For that, we would need `SeqCst`, as in this patch.

Note that I don't have a particularly high confidence about this explanation, and would appreciate a sanity check from someone with more expertise, like maybe @aturon.

See also:

https://doc.rust-lang.org/std/sync/atomic/enum.Ordering.html
http://llvm.org/docs/LangRef.html#memory-model-for-concurrent-operations

### DIFF
--- a/src/lock.rs
+++ b/src/lock.rs
@@ -8,7 +8,7 @@ extern crate core;
 
 use self::core::cell::UnsafeCell;
 use self::core::ops::{Deref, DerefMut};
-use self::core::sync::atomic::Ordering::{Acquire, Release};
+use self::core::sync::atomic::Ordering::{SeqCst};
 use self::core::sync::atomic::AtomicBool;
 
 /// A "mutex" around a value, similar to `std::sync::Mutex<T>`.
@@ -54,7 +54,7 @@ impl<T> Lock<T> {
     /// If `None` is returned then the lock is already locked, either elsewhere
     /// on this thread or on another thread.
     pub fn try_lock(&self) -> Option<TryLock<T>> {
-        if !self.locked.swap(true, Acquire) {
+        if !self.locked.swap(true, SeqCst) {
             Some(TryLock { __ptr: self })
         } else {
             None
@@ -84,7 +84,7 @@ impl<'a, T> DerefMut for TryLock<'a, T> {
 
 impl<'a, T> Drop for TryLock<'a, T> {
     fn drop(&mut self) {
-        self.__ptr.locked.store(false, Release);
+        self.__ptr.locked.store(false, SeqCst);
     }
 }
 


dwrensha:
On further reflection, I think that it might be preferable to fix the implementation of `OneShot` instead of altering `Lock`. We should weigh the performance cost of `SeqCst` vs the potential trickiness (assuming my above explanation is correct) of using `Release` / `Acquire`.

carllerche:
In my experience dealing w/ atomics, it's generally very difficult to correctly implement algorithms with multiple atomics & acq / rel unless there is a very clear hierarchy between them. Aka, the algorithm would be correct w/ a single atomic and the second atomic is an optimization (I don't know if that makes sense).

Looking through this, it doesn't seem obvious that this can be safely implemented using acq / rel w/o some more significant changes.

I vote to just switch to SeqCst for now and an alternate implementation can be experimented with later.

alexcrichton:
Yeah we may be able to be a bit more surgical about this in the future, but for now switching back to `SeqCst` seems good for now. We can focus more energy on these orderings if benchmarks show this as the real hot spot.

alexcrichton:
Thanks for the investigation @dwrensha!

dwrensha:
Ah, I was somehow under the impression that `lock::Lock` was a publicly-exposed type. Now that I see it's a private type, I'm less worried about making this change.

👍

Diggsey:
@dwrensha @carllerche  I'm pretty sure this is not correct - SeqCst is only required when you have more than two threads in play.

It looks to me like the issue is with `self.locked.swap(true, Acquire)` - the swap operation performs both a read *and* a write, and so the memory ordering must be `AcqRel`, not just `Acquire`.

SeqCst comes into play when you have a 3rd thread "observing" a synchronised interaction between two other threads: without SeqCst, the 3rd thread may observe reads/writes to disparate locations happening in a different (but still valid) order to the first two threads.

dwrensha:
> It looks to me like the issue is with self.locked.swap(true, Acquire) - the swap operation performs both a read and a write, and so the memory ordering must be AcqRel, not just Acquire.

I tried this just now and observed the same deadlock behavior.

I've fixed the repro link in my original comment; it was pointing to a branch that was no longer the relevant branch. The commit where I observe the deadlock behavior is: https://github.com/dwrensha/zillions/commit/6399c8818200df941fdcd73ba318211ecba538be

```
cargo build --release
cd impls/gjio
cargo build --reelase
cd ../..
./target/release/stresstest ./impls/gjio/target/release/server
```

... and it eventually gets stuck.

If I do `cargo update -p futures`, it works fine.

Diggsey:
@dwrensha hmm, you're right - still, I'm sure that nothing stronger than Acquire/Release semantics *should* be necessary here.

I simplified what's going on here to this sequence of events:
```
	thread 1	|	thread 2
------------------------+-----------------------
A.store(1, Release)	|	B.store(1, SeqCst)
x := B.load(SeqCst)	|	y := A.load(Acquire)
```

The question is then "can `x == 0 && y == 0` ?"

Apparently, yes (this is Dekker's algorithm): if the store to `B` happens after the load from `B`, then access to `A` is entirely un-synchronised. To fix the problem, SeqCst can be used, *or* the stores can be given Acquire semantics to ensure that the loads cannot be reordered above them. This can be achieve by replacing them with swaps:
```
	thread 1	|	thread 2
------------------------+-----------------------
A.swap(1, AcqRel)	|	B.swap(1, AcqRel)
x := B.load(Acquire)	|	y := A.load(Acquire)
```

This way, whichever order the swaps execute in, it still counts as a synchronisation point.

To test this, I wrote this smaller program which reliably reproduces the issue much more quickly:
```rust
use std::sync::Arc;
use std::sync::mpsc::channel;
use std::thread;

#[derive(Default)]
struct Inner {
    a: AtomicUsize,
    b: AtomicUsize,
}

fn main() {
    let (tx0t0, rx0t0) = channel::<Arc<Inner>>();
    let (tx0t1, rx0t1) = channel::<Arc<Inner>>();
    let (tx1t0, rx1t0) = channel();
    let (tx1t1, rx1t1) = channel();
    
    let t0 = thread::spawn(move || {
        while let Ok(x) = rx0t0.recv() {
            x.a.store(1, Ordering::Release);
            tx1t0.send(x.b.load(Ordering::SeqCst)).unwrap();
        }
    });
    let t1 = thread::spawn(move || {
        while let Ok(y) = rx0t1.recv() {
            y.b.store(1, Ordering::SeqCst);
            tx1t1.send(y.a.load(Ordering::Acquire)).unwrap();
        }
    });
    
    for i in 0..10000 {
        if i % 100 == 0 {
            println!("{}", i);
        }
        let x = Arc::new(Inner::default());
        let y = x.clone();
        tx0t0.send(x).unwrap();
        tx0t1.send(y).unwrap();
        
        let r0 = rx1t0.recv().unwrap();
        let r1 = rx1t1.recv().unwrap();
        assert!(r0 != 0 || r1 != 0);
    }
    
    drop(tx0t0);
    drop(tx0t1);
    t0.join().unwrap();
    t1.join().unwrap();
    println!("Done");
}
```

Unrelatedly, this change would also be a good idea, to avoid running arbitrary code while the lock is held:
```diff
@@ -241,7 +241,11 @@ impl<T> Future for Oneshot<T> {
         } else {
             let task = task::park();
             match self.inner.rx_task.try_lock() {
-                Some(mut slot) => *slot = Some(task),
+                Some(mut slot) => {
+                    let old_task = ::std::mem::replace(&mut *slot, Some(task));
+                    drop(slot);
+                    drop(old_task);
+                },
                 None => done = true,
             }
         }
```

PR #226, opened 2016-11-03
Add Sink trait and re-organize the crate

aturon:
This PR makes several backwards-compatible changes to the futures crate.

### Introduce a new `Sink` trait, which is essentially the reverse of `Stream`

This trait is modeled based on our experiences/needs in `tokio-core` and above, but is also used directly in futures as an interface for sending on channels. The basic idea is that you can "send" values to a sink, which is a multi-step process. The first step is initiating the send, which may be "blocked" on the sink having enough space. The second step is driving the send to completion, which for buffered sinks may involve actually flushing out the data involved. In our async setting, all of this work is driven by the usual polling/task wakeup architecture. The details are in the `Sink` trait.

Of particular note: there is a `send_all` combinator that gives you a future that will send the contents of a stream into a sink. This is expected to be the most common way to ultimately consume a sink, since it allows batching large amounts of items to send without forcing a flush in between.

With some changes we have planned for `tokio-core`, you'll be able to write a simple RPC using code like the following:

```rust
fn simple_rpc(s: TcpStream) -> impl Future<Item = ()> {
    // the split method will produce a pair of a Stream and a Sink;
    // these work by encoding/decoding according to the item types,
    // which here are inferred from `my_service`
    let (input, output) = s.split();
    let responses = input.and_then(my_service);
    output.send_all(responses)
}
```

If you want to get more fancy, you can turn this into a simple pipeliner:

```rust
fn simple_rpc(s: TcpStream) -> impl Future<Item = ()> {
    let (input, output) = s.split();
    output.with(my_service).buffer(MAX_REQUESTS).send_all(input)
}
```

### Introduce `sync` module, migrating existing synchronizers there

The futures library contains two futures-aware synchronizers today: `oneshot` and `channel`. Ultimately, we want to grow this set with additional channel types, future-aware mutexes, and other such tools.

This PR introduces a `sync` module, modeled after the one in `std`, that will contain futures-aware synchronizers. To be "futures-aware" means that the synchronizer doesn't block the OS thread, but instead uses the poll/task system to schedule/deschedule tasks as needed.

The PR also integrates channels with the sink API.

Although the APIs are moved, deprecated re-exports are kept around for the time being; we can remove them along with other breaking changes when we want to head toward 0.2

### Tidy up organization

The PR also:

- Moves most futures-related items into a `future` submodule, while retaining deprecated re-exports. That brings futures into line with streams and sinks.

- Splits `task` into `task` and `executor`; this greater focus should guide people to the right things when using tasks.


src/sink/buffer.rs
```diff
@@ -0,0 +1,65 @@
...
+use {Poll, Async};
+use sink::{Sink, StartSend, AsyncSink};
+
+/// Sink for the `Sink::buffer` combinator, which buffers up to some fixed
+/// number of values when the underlying sink is unable to accept them.
+#[must_use = "sinks do nothing unless polled"]
+pub struct Buffer<S: Sink> {
+    sink: S,
+    buf: VecDeque<S::SinkItem>,
+    cap: usize,
```
alexcrichton: Could this be accessed through `self.buf.capacity()`?

src/sink/buffer.rs
```diff
@@ -0,0 +1,65 @@
...
+        need_close: false,
+    }
+}
+
+impl<S: Sink> Sink for Buffer<S> {
+    type SinkItem = S::SinkItem;
+    type SinkError = S::SinkError;
+
+    fn start_send(&mut self, item: Self::SinkItem) -> StartSend<Self::SinkItem, Self::SinkError> {
+        try!(self.poll_complete());
```
alexcrichton: This seems like it may defeat the purpose, right? This means that you're always trying to write out items, even if you're just starting to push. If the final sink at the bottom is always ready for writing then you'll do syscalls-per-item instead of batching them up into one, right?

Perhaps this could happen only in the case that `self.buf.len() > cap`?

src/sink/mod.rs
```diff
@@ -0,0 +1,298 @@
...
+    ///
+    /// This method returns `AsyncSink::Ready` if the sink was able to start
+    /// sending `item`. In that case, you *must* ensure that you call
+    /// `poll_complete` to process the sent item to completion. Note, however,
+    /// that several calls to `start_send` can be made prior to calling
+    /// `poll_complete`, which will work on completing all pending items.
+    ///
+    /// The method returns `AsyncSink::NotReady` if the sink was unable to begin
+    /// sending, usually due to being full. In that case, the current task is
+    /// automatically scheduled to be polled when the sink may be ready to
```
alexcrichton: s/polled/notified/

src/sink/mod.rs
```diff
@@ -0,0 +1,298 @@
...
+    /// This adapter produces a new sink that passes each value through the
+    /// given function `f` before sending it to `self`.
+    ///
+    /// To process each value, `f` produces a *future*, which is then polled to
+    /// completion before passing its result down to the underlying sink. If the
+    /// future produces an error, that error is returned by the new sink.
+    ///
+    /// Note that this function consumes the given sink, returning a wrapped
+    /// version, much like `Iterator::map`.
+    fn with<U, F, Fut>(self, f: F) -> With<Self, U, F, Fut>
```
alexcrichton: `U` shouldn't need to show up on `With`, right?

src/sink/mod.rs
```diff
@@ -0,0 +1,298 @@
...
+    /// Adds a fixed-size buffer to the current sink.
+    ///
+    /// The resulting sink will buffer up to `amt` items when the underlying
+    /// sink is unwilling to accept additional items. Calling `poll_complete` on
+    /// the buffered sink will attempt to both empty the buffer and complete
+    /// processing on the underlying sink.
+    ///
+    /// Note that this function consumes the given sink, returning a wrapped
+    /// version, much like `Iterator::map`.
+    fn buffer(self, amt: usize) -> Buffer<Self>
```
alexcrichton: I think we'll want to remove `.buffered` on stream eventually to reduce confusion here, but that can be an 0.2 thing.

src/sink/mod.rs
```diff
@@ -0,0 +1,298 @@
...
+    /// Convenience function for turning this sink into a trait object.
+    ///
+    /// This simply avoids the need to write `Box::new` and can often help with
+    /// type inference as well by always returning a trait object. Note that
+    /// this method requires the `Send` bound and returns a `BoxSink`, which
+    /// also encodes this constraint. If you'd like to create a `Box<Sink>`
+    /// without the `Send` bound, then the `Box::new` function can be used
+    /// instead.
+    #[cfg(feature = "use_std")]
+    fn boxed(self) -> BoxSink<Self::SinkItem, Self::SinkError>
```
alexcrichton: This'll conflict with `Stream::boxed`? We may want to elide this for now in the hopes that boxed sinks are rare enough that we may want to avoid the convenience

src/sink/mod.rs
```diff
@@ -0,0 +1,298 @@
...
+    {
+        ::std::boxed::Box::new(self)
+    }
+}
+
+if_std! {
+    // TODO: consider expanding this via e.g. FromIterator
+    impl<T> Sink for ::std::vec::Vec<T> {
+        type SinkItem = T;
+        type SinkError = (); // Change this to ! once it stabilizes
```
alexcrichton: :\

If we really do this I think we should invest in a `enum Void {}` in the meantime

src/sink/send.rs
```diff
@@ -0,0 +1,50 @@
...
+    }
+}
+
+impl<S: Sink> Future for Send<S> {
+    type Item = S;
+    type Error = S::SinkError;
+
+    fn poll(&mut self) -> Poll<S, S::SinkError> {
+        if let Some(item) = self.item.take() {
+            try!(self.sink_mut().poll_complete());
```
alexcrichton: Like with the buffering above, perhaps this could get delayed until after after the start send returned not ready?

src/sink/send_all.rs
```diff
@@ -0,0 +1,79 @@
...
+impl<T, U> Future for SendAll<T, U>
+    where T: Sink,
+          U: Stream<Item = T::SinkItem>,
+          T::SinkError: From<U::Error>,
+{
+    type Item = T;
+    type Error = T::SinkError;
+
+    fn poll(&mut self) -> Poll<T, T::SinkError> {
+        try!(self.sink_mut().poll_complete());
```
alexcrichton: Same concerns as above (delaying until after we see not ready)

src/sink/with.rs
```diff
@@ -0,0 +1,117 @@
...
+impl<S, U, F, Fut> With<S, U, F, Fut>
+    where S: Sink,
+          F: FnMut(U) -> Fut,
+          Fut: IntoFuture<Item = S::SinkItem, Error = S::SinkError>,
+{
+    fn poll(&mut self) -> Poll<(), S::SinkError> {
+        loop {
+            match mem::replace(&mut self.state, State::Empty) {
+                State::Empty => break,
+                State::Process(mut fut) => {
```
alexcrichton: If we want to reduce moves this could perhaps be:

```rust
let item = match self.state {
    State::Empty => break,
    State::Process(ref mut f) => try_ready!(f.poll()),
    State::Buffered(_) => /* do the replace here */,
};
// do the send here
```

alexcrichton:
In talking offline it was also decided that for now we should remove `start_close_sink` and we can re-add it (if necessary) in 0.2, but it seems unlikely as we can perhaps encode those use cases differently.

src/sink/mod.rs
```diff
@@ -0,0 +1,298 @@
...
+    fn with<U, F, Fut>(self, f: F) -> With<Self, U, F, Fut>
+        where F: FnMut(U) -> Fut,
+              Fut: IntoFuture<Item = Self::SinkItem, Error = Self::SinkError>,
+              Self: Sized
+    {
+        with::new(self, f)
+    }
+
+    /*
+    fn with_map<U, F>(self, f: F) -> WithMap<Self, U, F>
```
alexbool: Why is this code commented out?

carllerche:
Re: @alexcrichton 

> I wonder if we want to future-proof ourselves and perhaps call the current spsc channel futures::sync::spsc::sync_channel? We could theoretically add a channel (like the standard library) which is nonblocking on the sender side (e.g. doesn't implement Sink) and it could be useful in some contexts perhaps.

I wouldn't bother with it yet. I think the organization of channels should happen when we are ready to focus on them. I know that I am most likely going to push to have the linked channel take a max capacity and rename the unbounded one -> `mpsc::unbounded()` (or `unbounded_channel()` :) IMO given the danger of unboundedness in an async system, being able to get an unbounded channel **should** be explicit.

carllerche:
The PR looks good. I commented inline when appropriate. I do have a couple of higher level questions. The answer may be to punt until 0.2, but I want to at least get them on the record.

1) There is no way to detect that the remote "hung up". For example, a sink backed by a socket shutdown (cleanly). A consumer dropped interest, etc... This is an important event to be able to detect before having to send a message. For example, if a socket closed, the producer may want to eagerly open a new connection to a different host before it receives a message to send. This is currently a problem w/ Tokio `Service` https://github.com/tokio-rs/tokio-service/issues/4

2) While I understand the reasoning behind returning "not ready" with `AsyncSink::NotReady(T)` vs `SinkError<T>`, I want to point out that this strategy *probably* won't work for Tokio `Service`. In the case of Tokio `Service`, the request being "pushed back" needs to be able to happen as part of the response future. Given this, the current best way that I can think of looks something like this: https://gist.github.com/carllerche/d8724bd98d5f80919dfdf73cb60760c3

It would be unfortunate to have two pretty different ways of handling "push back", but maybe this is unavoidable?

src/sink/buffer.rs
aturon: I didn't want to rely on the underlying capacity, because you don't get exactly what you asked for, and we don't want the buffer to actually be bigger than what you requested. I'll add a comment.

src/sink/mod.rs
aturon: It has to show up, because it's an input (non-associated) type. (One thing that makes this different from `and_then`).

src/sink/mod.rs
aturon: Yes, most likely. I'm not 100% sure on our total buffering strategy yet.

aturon:
@carllerche Yep, I agree with this concern, but explicitly decided to punt on that functionality until 0.2, where we're likely taking an end-to-end look at backpressure and Service integration.

src/sink/buffer.rs
aturon: Let's talk on IRC about this -- we've had a bunch of discussions on how this should work with I/O, but seem to still not be on the same page.

src/sink/mod.rs
aturon: It's definitely useful to have `Vec` be a sink; we could/should consider extending this to anything that implements `Extend`.

I'm not sure I agree it's worth introducing a `Void` type here; is `()` really so bad? Is the concern just about forward-compat with going to `!`?

src/sink/mod.rs
aturon: These are just placeholders for additional combinators, which I wanted to sketch without fully implementing just yet.

src/sink/buffer.rs
aturon: OK, we've resolved things, although this instance of `poll_complete` stays put. (I still don't think this defeats the purpose or fails to batch, FWIW.)

Issue #173, opened 2016-09-22
How can I wait for a future in multiple threads


jamespharaoh:
I can't seem to make this work, perhaps I'm being dumb! In any case, I think this should be in the documentation as an example.

In my case, I have background processes loading data for a given ID. Obviously if a piece of data is already being loaded then I want to join the existing waiters rather than starting a new background process.

I've implemented this using a `Map <Id, Vec <Complete>>>`, then the first loader triggers the completion of subsequent loaders when it has completed. This is a lot of boilerplate.

I've tried all sorts of things to get this to work but somehow I can't get anything else to compile. Waiting for a future consumes it, so I can only do that once. I have tried to replace the future with a new one and wait on that, like I might do in JavaScript, but this also doesn't work.

If anyone can show me an example then that would be great, if not then I'll probably create an abstraction around my current method and submit this as a pull request for the library.

alexcrichton:
Strictly speaking the issue as titled is not possible. To poll a future it requires `&mut self` which means *synchronized*. In other words, only one thread can call `Future::poll` at a time.

What you may want to do is to create your own custom implementation of `Future` for this. You'd have to keep track of all tasks waiting for a value to get produced, and then wake them all up once the value is ready. You'd then also have the logic for the future being immediately ready if the value is already there.

Does that make sense?

jamespharaoh:
Yes, as I explained I've done something like this in my code already and it is working.

The question, I guess, is if you would be interested in this being submitted as a pull request for inclusion in the futures library itself, or if I should just go ahead and implement it separately.

alexcrichton:
Perhaps! Mind if I take a peek at what you've got going on?

jamespharaoh:
Well it's kind of mixed up in the other code, but it's in this file:

https://github.com/wellbehavedsoftware/wbs-backup/blob/6428167c0bfe42144b8d0d42f6e94b597daa33d7/rzbackup/src/zbackup/repo.rs

Here's where I keep a map of "additional" clients waiting for a future to complete, presumably the map key would stay the same but the value would be a struct mapping the desired behaviour:

https://github.com/wellbehavedsoftware/wbs-backup/blob/6428167c0bfe42144b8d0d42f6e94b597daa33d7/rzbackup/src/zbackup/repo.rs#L74

Then in this function, I either create a direct future to produce a result, or I add a new oneshot to the list above. The first one will fire off all the others on completion:

[...]

PR #227, opened 2016-11-03
add shared() to Future trait

liran-ringel:
The shared() method provides a mean to convert any future into a clonable future.
It enables a future to be polled by multiple threads.
Example:
`
>let shared_future = original_future.shared();
>let cloned_future = shared_future.clone();
>let cloned_future2 = cloned_future.clone();
`
#173


src/shared.rs
```diff
@@ -0,0 +1,211 @@
...
+    fn new(error: E) -> Self {
+        SharedError { error: Arc::new(error) }
+    }
+}
+
+struct SyncedInner<F>
+    where F: Future
+{
+    original_future: F,
+    result: Option<Result<Async<SharedItem<F::Item>>, SharedError<F::Error>>>,
```
alexcrichton: I'd (perhaps naively) expect this to be `Option<Result<Item, Error>>`, how come the `Async` is here?

src/shared.rs
```diff
@@ -0,0 +1,211 @@
...
+    tasks_receiver: Mutex<Receiver<Task>>,
```
alexcrichton: It looks like the blocking functionality of this `Mutex` is never used, so perhaps this could be `Lock` (in this repo)

src/shared.rs
```diff
@@ -0,0 +1,211 @@
...
+                            polled_result = inner.result.clone();
+                        }
+                        Ok(Async::NotReady) => {} // Will be handled later
+                    }
+                }
+            }
+            Err(_) => {} // Will be handled later
+        }
+
+        if polled_result.is_some() {
```
alexcrichton: This could be `if let Some(...) = polled_result`, removing the `unreachable!` below I believe.

src/shared.rs
```diff
@@ -0,0 +1,211 @@
...
+        let _ = self.tasks_sender.send(t);
+        if self.inner.tasks_unpark_started.load(Ordering::Relaxed) {
+            // If the tasks unpark has started, synced_inner can be locked for read,
+            // and its result has value (not None).
+            // The result must be read here because it is possible that the task,
+            // t (see variable above), had not been unparked.
+            match self.inner.synced_inner.try_read() {
+                Ok(inner_guard) => {
+                    let ref inner = *inner_guard;
+                    if let Some(ref result) = inner.result {
```
alexcrichton: This should be a `unwrap` essentially, right?

src/shared.rs
```diff
@@ -0,0 +1,211 @@
...
+{
+    original_future: F,
+    result: Option<Result<Async<SharedItem<F::Item>>, SharedError<F::Error>>>,
+    tasks_receiver: Mutex<Receiver<Task>>,
+}
+
+struct Inner<F>
+    where F: Future
+{
+    synced_inner: RwLock<SyncedInner<F>>,
```
alexcrichton: This implementation is making some assumptions about `RwLock` and the various return values of `try_read` and `try_write`. That is, it assumes that if `Err` is returned that it's guaranteed that another task has the lock, as opposed to, for example, also being in contention for the same lock.

Put another way, I could imagine an implementation where two concurrent calls to `try_read` would have one fail and one succeed. 

I think a situation like that could cause this implementation to deadlock. Perhaps we could vendor a copy with known semantics around this form of contention to ensure we're getting what we want?

src/shared.rs
liran-ringel: It matches the type that will be returned by poll().
If it won't have `Async`, a conversion will have to be made.

src/shared.rs
liran-ringel: done

src/shared.rs
liran-ringel: hmm I haven't considered that.
What alternative do we have?
Maybe we should use normal mutex and atomic option instead:
http://aturon.github.io/crossbeam-doc/crossbeam/sync/struct.AtomicOption.html
I guess it won't have the issue you talk about.

src/shared.rs
liran-ringel: done

src/shared.rs
alexcrichton: Yeah a conversion is fine, cuts down on the storage space here anyway

liran-ringel: done

alexcrichton:
Ah sorry I've missed some of the recent updates!

It looks like the amount of `unsafe` code is on the uprise here. I wonder if we should perhaps just stick to a `Mutex`-based approach to start out with, and then if it becomes a bottleneck we can move towards a lock-free approach?

liran-ringel:
@alexcrichton  I'd like that too, but we have the `RwLock` problem you mentioned.
I've searched for an implementation that has the required guarantees (no race for try_read()), and did not find one.
If we write an `RwLock` for that purpose we will eventually write unsafe code.

Beyond the slightly better performance, current implementation solves the `RwLock` issues. I think we should go with it, and just be careful.

If you agree, I will prepare the branch for code review (remove WIP).
Otherwise, tell me if you are familiar with a `RwLock` implementation that will be fine for our purpose.

alexcrichton:
Oh I was thinking of basically starting the entire `poll` method with just a blanket:

```rust
let mut state = self.inner.lock().unwrap();
```

That way the entire function call is synchronized with respect to any other concurrent callers and we don't have to worry about platform-specific guarantees and such.

liran-ringel:
@alexcrichton Ah that will be much simpler than anything else, but I thought we should never block in `poll` method.

alexcrichton:
Indeed yeah, but this kind of synchronization should have a bounded wait time. That is, it's possible for a thread to be blocked for a small amount of time, but it would never wait on something like I/O for an unbounded amount of time.

In other words a targeted use of a mutex or an rwlock is fine so long as the duration for which the lock is held is as small as possible.

liran-ringel:
@alexcrichton I've replaced the unsafe blocks with synced RwLock methods (that are safe).
It's not really necessary, and adds some of overhead, but as you said, if it becomes a bottleneck we can change it.

src/future/shared.rs
```diff
@@ -0,0 +1,222 @@
...
+        // The result was not ready.
+        // Try lock the original future.
+        match self.inner.synced_inner.try_lock() {
+            Some(mut synced_inner) => {
+                let ref mut synced_inner = *synced_inner;
+                let ref mut original_future = synced_inner.original_future;
+                let mut result = self.inner.result.write().unwrap();
+                // Other thread could already poll the result, so we check if result has a value
+                if result.is_none() {
+                    match original_future.poll() {
```
alexcrichton: One part I'd be particularly worried about here is the call to `poll()` inside the lock, as `poll()` is something that could invoke an unbounded amount of code (although it should still return quickly).

When I was thinking of using a lock I think we may be able to simplify this as well, perhaps with [something like this](https://gist.github.com/alexcrichton/bff55d6723991b1b47bd63eecdc7b251#file-main-rs-L38-L72) where we just have a lock around the state and a queue of waiters.

liran-ringel: Is the problem with synced_inner lock or with the result lock?

alexcrichton: Oh it's just that a write lock is held here while the `poll` is done, preventing readers while the polling is happening.

liran-ringel: No thread will try to lock for read when poll is called because tasks_unpark_started is false. 
If it's true and a thread locks for write, it won't call poll() because result is not None.

alexcrichton: Ah yes, that's indeed true! Even in that case though I think we may still want to maintain a `Vec<Task>` rather than a queue of tasks through a sender. Each time `poll` is called it should update the blocked task if there was a previous one, but this way it just always enqueues a new task, perhaps holding onto old tasks for too long.

Additionally, I believe there's still a race where once `unpark_tasks` starts it may miss tasks that haven't been sent yet, causing those tasks to get leaked as they'll just sit around in the channel until the future is destroyed.

liran-ringel: done

liran-ringel:
@alexcrichton I've also removed `shared()` method from `Future` trait because it required std (used `Task`).
Instead, I've created `IntoShared` that implements `shared()` to every `Future`, so the result is the same.

alexcrichton:
Oh it's fine to just tag the method with `#[cfg(feature = "use_std")]` like other methods that require std, an extra trait shouldn't be necessary for this.

src/future/shared.rs
```diff
@@ -0,0 +1,229 @@
...
+            Ok(item) => Ok(Async::Ready(item)),
+            Err(error) => Err(error),
+        }
+    }
+
+    /// Clones the result from self.inner.state.
+    /// Assumes state is `State::Done`.
+    fn read_result(&self) -> Result<Async<SharedItem<F::Item>>, SharedError<F::Error>> {
+        match *self.inner.state.read().unwrap() {
+            State::Done(ref result) => Self::result_to_polled_result(result.clone()),
```
alexcrichton: This could just be:

```rust
result.clone().map(Async::Ready)
```

right?

src/future/shared.rs
```diff
@@ -0,0 +1,229 @@
...
+        // Try lock the original future.
+        match self.inner.original_future.try_lock() {
+            Some(mut original_future) => {
+                // Other thread could already poll the result, so we check if result_ready.
+                if self.inner.result_ready.load(Ordering::Relaxed) {
+                    return self.read_result();
+                }
+
+                match original_future.poll() {
+                    Ok(Async::Ready(item)) => {
```
alexcrichton: As soon as `Ready` happens could this also destroy the original future? (e.g. set it to `None`)

src/future/shared.rs
```diff
@@ -0,0 +1,229 @@
...
+    /// and returns the result.
+    fn store_result(&self,
+                    result: Result<SharedItem<F::Item>, SharedError<F::Error>>)
+                    -> Result<Async<SharedItem<F::Item>>, SharedError<F::Error>> {
+        let ref mut state = *self.inner.state.write().unwrap();
+
+        match mem::replace(state, State::Done(result.clone())) {
+            State::Waiting(waiters) => {
+                self.inner.result_ready.store(true, Ordering::Relaxed);
+                for task in waiters {
```
alexcrichton: Could this happen outside the write lock?

liran-ringel: No, because otherwise another thread could call read_result() that will panic if the state is State::Waiting

src/future/shared.rs
liran-ringel: you're right :+1:

src/future/shared.rs
alexcrichton: It's generally good hygiene to run destructors associated with futures ASAP as it ensure that any remaining resources held internally are all deallocated and/or pending futures that we're not interested in are canceled.

src/future/shared.rs
alexcrichton: Er well this is sort of the point, for `unpark` to happen outside the lock. The `unpark` can run arbitrary code, which can take awhile, and we shouldn't unnecessarily block other threads.

liran-ringel: Ok now the mutex guard is dropped before `unpark` is called.

alexcrichton:
Ok @liran-ringel this looks great to me, thanks!

@aturon @carllerche mind doing a gut-check on naming here? To recap the `shared` method here takes a `Future<T, E>` into what is essentially a `Future<Arc<T>, Arc<E>>`. The `Shared` type returned can be cloned to create many handles to the same future.

aturon:
I'm :+1: on the name and on providing this functionality.

PR #238, opened 2016-11-09
feat(mpsc): add mpsc channel without back pressure

compressed:
As discussed in #218, adding a mpsc channel using `std::sync::mpsc`.

One thing I wasn't sure about was which atomic ordering to use.

Let me know what you think! Thanks.


src/sync/mpsc.rs
```diff
@@ -0,0 +1,134 @@
...
+}
+
+impl<T> Drop for Sender<T> {
+    fn drop(&mut self) {
+        // Each cloned sender calls `Drop` so we need to ensure that only the last sender being
+        // dropped will unpark the `Receiver`.
+        // The result is discarded because it could be updated between this line and the next.
+        let _ = self.inner.sender_count.fetch_sub(1, SeqCst);
+        // if it was one before, it *should* be zero now, but let's check just to make sure
+        if self.inner.sender_count.compare_exchange(0, 0, SeqCst, SeqCst).is_ok() {
```
alexcrichton: I think you can just check here if the previous return value was 1 (from `fetch_sub`) and that'll indicate whomever is last.

src/sync/mpsc.rs
```diff
@@ -0,0 +1,134 @@
...
+}
+
+impl<T> Sender<T> {
+    /// Send a message on the channel
+    pub fn send(&self, t: T) -> Result<(), mpsc::SendError<T>> {
+        match self.tx.send(t) {
+            Ok(_) => {
+                // if an rx_task exists, try to unpark
+                // otherwise someone else is unparking the Receiver
+                if let Some(mut opt_task) = self.inner.rx_task.try_lock() {
```
alexcrichton: I believe this can also interfere with the `try_lock` in the `Receiver`, in which case that'll result in a lost wakeup (e.g. deadlock)

src/sync/mpsc.rs
```diff
@@ -0,0 +1,134 @@
...
+                    if let Some(task) = opt_task.take() {
+                        task.unpark();
```
alexcrichton: Ah I tend to try to be careful to do operations like this *outside* of the lock, so could this be accompanied with a `drop(opt_task)` before?

src/sync/mpsc.rs
```diff
@@ -0,0 +1,134 @@
...
+
+impl<T> Stream for Receiver<T> {
+    type Item = T;
+    type Error = ();
+    fn poll(&mut self) -> Poll<Option<Self::Item>, Self::Error> {
+        match self.rx.try_recv() {
+            Ok(t) => Ok(Some(t).into()),
+            Err(e) => {
+                match e {
+                    mpsc::TryRecvError::Disconnected => Ok(Async::Ready(None)),
```
alexcrichton: This may look a bit nicer as a nested match (e.g. match `Err(TryRecvError::...)`) above.

src/sync/mpsc.rs
alexcrichton: (I think this race applies with the `Sender::drop` below as well)

src/sync/mpsc.rs
```diff
@@ -0,0 +1,134 @@
...
+                }
+            }
+        }
+    }
+}
+
+/// Creates a new multiple-producer, single-consumer channel. `Sender` can be cloned.
+///
+/// This channel does not implement back pressure.
+pub fn channel<T>() -> (Sender<T>, Receiver<T>) {
```
alexcrichton: We're still debating naming a bit, but I think this is most likely to be named `unbounded`, although I'm not sure precisely on how that would affect the names of sender/receiver ...

src/sync/mpsc.rs
```diff
@@ -0,0 +1,134 @@
...
+pub struct Receiver<T> {
+    /// Receiver
+    rx: mpsc::Receiver<T>,
+    /// Inner
+    inner: Arc<Inner>,
+}
+
+impl<T> Sender<T> {
+    /// Send a message on the channel
+    pub fn send(&self, t: T) -> Result<(), mpsc::SendError<T>> {
```
alexcrichton: Could this take `&mut self` for consistency with `Sink` and friends? Essentially it's what `std::sync::mpsc::Sender` *should* have taken ...

src/sync/mpsc.rs
```diff
@@ -0,0 +1,134 @@
...
+use std::sync::atomic::AtomicUsize;
+use stream::Stream;
+use task::{self, Task};
+use std::sync::{Arc, mpsc};
+use lock::Lock;
+
+/// `Sender` of the channel. `Sender` can be cloned.
+pub struct Sender<T> {
+    /// Sender
+    tx: mpsc::Sender<T>,
```
alexcrichton: Eventually we'll probably want to replace this with our own queue (as most of the machinery in std we're not using as we're only using `try_recv` here, but we can always optimize that later.

src/sync/mpsc.rs
compressed: Yes, definitely. Not sure why I wrote it this way to begin with :)

compressed:
Thanks for the comments. I've tried to address them, but there still seems to be a race condition of sorts causing a deadlock. From what I can tell, the `try_recv` function ends up returning either `Empty` when there is still data to be processed or `Empty` when all the `Sender`s have been dropped. As a result, the `Receiver` just parks, and we're stuck...

Maybe we need to use move to using that queue you mentioned for Sender here?

src/sync/mpsc.rs
compressed: Thanks! I've taken care of this.

alexcrichton:
Ah yeah I think the deadlock is the [race I mentioned here](https://github.com/alexcrichton/futures-rs/pull/238#discussion_r87324095).

compressed:
Ok, I think I've fixed all the race conditions.

- To address the issue when the last `Sender` drops, but doesn't have a task to unpark the `Receiver`, I've added an additional task specifically for that condition that is populated by the `Receiver` the first time poll is called.
- To address the issue where the `Receiver` has been unparked with data pending, but we get a `mpsc::TryRecvError::Empty`, I added a `num_msgs_pending` to determine if there is in fact anything pending. If there is, then we try to loop to get that message instead of parking the `Receiver`. I believe this race condition was caused internally by std's `mpsc_queue.pop()` which may give us a `PopResult::Inconsistent` result in some cases.

I'm not sure if this solution is something like you had in mind. Another option could be to just swap out `std::sync::mpsc` entirely and use something like crossbeam's `MsQueue` for storing the data.

src/sync/mpsc.rs
```diff
@@ -0,0 +1,174 @@
...
+
+impl<T> Sender<T> {
+    /// Send a message on the channel
+    pub fn send(&mut self, t: T) -> Result<(), mpsc::SendError<T>> {
+        match self.tx.send(t) {
+            Ok(_) => {
+                let _ = self.inner.num_msgs_pending.fetch_add(1, SeqCst);
+                // if an rx_task exists, try to unpark
+                // otherwise someone else is unparking the Receiver
+                if let Some(mut opt_task) = self.inner.rx_task.try_lock() {
```
compressed: Good catch, thanks! I'll work on fixing this.

compressed:
So I added a bit of synchronization here now by making `is_first_rx_poll` into an `AtomicBool` and putting it in `Inner`. Now the receiver sets `rx_task` and `rx_task_tx_drop` only the first time it `poll`s. After that point, the sender is responsible for storing the `rx_task` when it sends something so there's no race with the receiver anymore. The sender can only try to unpark the receiver after the receiver has set the initial `rx_task`.

I also removed some of the unnecessary `compare_exchange` calls and replaced them with `load` instead.

src/sync/mpsc.rs
```diff
@@ -0,0 +1,175 @@
...
+}
+
+impl<T> Stream for Receiver<T> {
+    type Item = T;
+    type Error = ();
+    fn poll(&mut self) -> Poll<Option<Self::Item>, Self::Error> {
+        // if this is the receiver's first poll call, store two receiver tasks:
+        // rx_task_tx_drop - used only once to wake up receiver when last Sender drops
+        // rx_task - used by senders to wake up the receiver after a message has been sent
+        if self.inner.is_first_rx_poll.load(SeqCst) {
```
alexcrichton: Unfortunately the contract of `Stream::poll` is that it wakes up the *last* task that called `poll`. In that sense we'll need to continuously updated the blocked task as a receiver blocks over time. I think that may unfortunately complicate this wrt the sender trying to unblock a blocked receiver :(

alexcrichton:
Ok, in the meantime I've merged https://github.com/alexcrichton/futures-rs/pull/245 which includes an unbounded implementation as well, but I'd guess that an implementation like this may be faster, so I'd at least still be interested in pursuing this! 

When rebasing, feel free to just delete the unbounded wrappers and add new types like this.

src/sync/mpsc.rs
compressed: Ah I did not know that was required. I guess then I will need to add some synchronization between the Sender and Receiver for the `rx_task` then to satisfy that.

compressed:
Okay - yes I agree this should be faster, not as much overhead. I'll give it another go.

compressed:
So I'm going to close this for now. After adding some synchronization (either via Mutex or a combination of atomics), I wasn't able to improve the performance of the current implementation significantly.

alexcrichton:
Ok, thanks regardless though @compressed! I suspect it's still possible to eek out performance somewhere (even if it's just compile times), so we can perhaps address at a later date.

compressed:
Sure, thanks for your reviews @alexcrichton! Agreed.

PR #240, opened 2016-11-10
Add an unbounded spsc channel

alexcrichton:
This commit vendors the spsc queue implementation in the standard library into
this repo and then builds an unbounded spsc channel on top. Notable a spsc
channel has no backpressure so `start_send` and `poll_complete` always return
ready.

This implementation continues to be lock-free by taking the same strategy as the
bounded channel by having two slots for blocked tasks that are alternated
between to ensure neither the producer nor consumer ever has to wait for the
other.

### DIFF
--- a/src/sync/spsc/bounded.rs
+++ b/src/sync/spsc/bounded.rs
@@ -1,5 +1,3 @@
-//! A single-producer, single-consumer, futures-aware channel
-
 use std::any::Any;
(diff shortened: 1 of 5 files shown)


alexcrichton:
I'm particularly curious about what we think naming should be here, `UnboundedSendError` is pretty wordy!

PR #241, opened 2016-11-10
Add a select combinator for streams

### LINKED ISSUE #239: Version of Stream::merge() without Both?
I have an application where I'd like to receive events from several `Stream`s, but I only want to receive one event at a time.  `Stream::merge()` is *nearly* right, but instead of receiving `Both(a,b)` if both sub-streams are ready, I would prefer to just receive one of the two, and leave the other to the next `poll`.

So my proposal would be to add a new method like `merge` (for lack of a better name, `mesh`), which returns something just like `stream::Merge` (say `Mesh`), which returned a `MeshedItem` which lacks the `Both` variant.

One complication is that you need to have a policy for which one you return; the simplest would be to favour the first (but risk never getting to the second if it's busy); otherwise a round robin policy might be better (but less ideal if you mesh several together in pairs).

In my application I want to expose a blocking "get next event" API to the client, where the events would be things like keypresses, timers, and signals; my current idea is to mesh the event sources together as above and block on `core.run(meshed.into_future())`.  Happy to hear it's a dumb idea and there's a better way to do it.  :-)

alexcrichton:
Similar to merge but requires the items/errors to be the same and then only
yields one at a time.

Closes #239

### DIFF
--- a/src/stream/select.rs
+++ b/src/stream/select.rs
@@ -1,0 +1,64 @@
+use {Poll, Async};
+use stream::{Stream, Fuse};
+
+/// An adapter for merging the output of two streams.
+///
+/// The merged stream produces items from either of the underlying streams as
+/// they become available, and the streams are polled in a round-robin fashion.
+/// Errors, however, are not merged: you get at most one error at a time.
+#[must_use = "streams do nothing unless polled"]
+pub struct Select<S1, S2> {
+    stream1: Fuse<S1>,
+    stream2: Fuse<S2>,
+    flag: bool,
+}
+
+pub fn new<S1, S2>(stream1: S1, stream2: S2) -> Select<S1, S2>
+    where S1: Stream,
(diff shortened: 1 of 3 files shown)


src/stream/select.rs
```diff
@@ -0,0 +1,57 @@
...
+
+    fn poll(&mut self) -> Poll<Option<S1::Item>, S1::Error> {
+        let (a, b) = if self.flag {
+            (&mut self.stream2 as &mut Stream<Item=_, Error=_>,
+             &mut self.stream1 as &mut Stream<Item=_, Error=_>)
+        } else {
+            (&mut self.stream1 as &mut Stream<Item=_, Error=_>,
+             &mut self.stream2 as &mut Stream<Item=_, Error=_>)
+        };
+        self.flag = !self.flag;
```
alexcrichton: True, I'll add a clause later to handle that.

Issue #202, opened 2016-10-10
Stackoverfow on deeply nested join/map


weiznich:
The following code will trigger a stack overflow.

``` rust
extern crate futures;

use futures::{Future, done};

const FUTURE_COUNT: usize = 10_000;

fn main() {
    let f = (0..FUTURE_COUNT)
        .map(|_| done::<i32, i32>(Result::Ok(42)))
        .fold(done(Result::Ok(0)).boxed(),
              |acc, f| acc.join(f).map(|(a, b)| a + b).boxed());

    let r = ::futures::task::spawn(f).wait_future();
    assert_eq!(r, Ok(42 * FUTURE_COUNT as i32));
    println!("Hello, world!");
}

```

[gdb backtrace](https://gist.github.com/weiznich/1e17dcce6eacdb5c145be11e404ee61f)

alexcrichton:
Thanks for the report! This is indeed the case where massively recursive futures like this may stack overflow. This is similar to massively recursive functions in Rust, however, which may also stack overflow. Unfortunately this is likely to be a "WONTFIX" issue in the sense that the design of the `Future` trait precludes any mitigation being done here.

Note that while developing this library this was an issue that we were quite aware of. Initially we tried mitigations through trait methods and such but unfortunately it never panned out, it ended up always imposing a cost on downstream consumers.

Another important point we found, though, is that almost all examples which lead to this sort of recursion can typically be written in such a way to avoid it. For example here instead of 10K `Join` futures it'd be far more efficient to, for example, use the `collect` future (and this would also avoid the stack overflow problem).

Does that make sense? Did you hit this particular case n the wild though where it was tough to work around?

alexcrichton:
Closing due to my previous comment

PR #242, opened 2016-11-11
Fold for list of futures

weiznich:
* implement an ordered and an unordered variant

Basically what I wanted to do in #202.

### DIFF
--- a/src/future/fold.rs
+++ b/src/future/fold.rs
@@ -1,0 +1,145 @@
+use std::mem;
+use std::prelude::v1::*;
+
+use {Future, IntoFuture, Poll, Async};
+
+/// Future for the `fold_unordered` combinator, folds a list of
+/// futures into one Future
+///
+/// This is created by this `fold_unordered` function.
+#[must_use = "futures do nothing unless polled"]
+pub struct FoldUnordered<T, F, R>
+    where F: Fn(R, T::Item) -> R,
+          T: Future
+{
+    fold: F,
+    initial: Option<R>,
+    remaining: Vec<Option<T>>,
+}
+
+/// Creates a new future which will fold a list of futures into one future
+///
+/// The function tries to execute each future in parallel, so the order
+/// of items passed to the fold function may be different as the order of
+/// futures in the initial list
+pub fn fold_unordered<I, R, F>(i: I,
+                               initial: R,
+                               fold: F)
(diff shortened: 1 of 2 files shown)


alexcrichton:
Thanks! I wonder if this could perhaps be a more ergonomic conversion from a list of futures to a stream? That way you could leverage the standard `fold` and `buffer_unordered` combinators

weiznich:
You are thinking about something like this?

```rust
    let data = (0..MAX).rev().collect::<Vec<_>>();
    ::futures::stream::iter(data.clone().into_iter().map(Ok))
          .map(|d| pool.spawn_fn(move || work(d as u32)))
          .buffer_unordered(MAX)
          .fold(0, |acc, a| Ok(fold(acc, a)))
          .wait()
          .unwrap()
```

I'm not sure if this really serves the problem I'm trying to solve. `buffer_unorderd` takes a (ordered) Stream of values which could be evaluated as futures, caches some of those futures and then waits till them are finished. In my usecase I've a list of a possible large number of futures which are started nearly at the same time and which should be evaluated in parallel. Each future reads the content of an file. If an future has finished the content  from those file could be merged into the final result using the fold method. (The order of merging does not matter here.) Using the fold_unorderd method from this pull-requst gives me on my benchmarks a performance improvement of ~3-4% to a plain `::futures::collect(future_list).into_iter().fold(init, fold_function)`. 
I will also try to test the stream approach from above in the next few days.

alexcrichton:
Yeah that's along the lines of what I was thinking in that the implementation of this sort of feature would basically compile down to those operations. If that sequence of operations is too slow, however, I'd love to dig into why and see if there's something we can improve!

Another possibility here is to provide an ergonomic conversion from a list of futures to a stream, basically automating the first steps in the gist:

```rust
    let data = (0..MAX).rev().collect::<Vec<_>>();
    ::futures::stream::iter(data.clone().into_iter().map(Ok))
          .map(|d| pool.spawn_fn(move || work(d as u32)))
          .buffer_unordered(MAX)
```

as just:

```rust
let stream = stream::futures(data);
```

Once you've done that it's just a standard `Stream::fold` away to get the same result as before.

I'd basically just prefer to push as hard as we can on the future/stream combinators (and primitives) before we add similar functionality in a specialized fashion. Having to understand fewer things is always nicer!

weiznich:
Closed in favour of #263

PR #243, opened 2016-11-12
futures-cpupool: enrich `Builder` with thread name prefix

alexbool:
And name threads in the pool accordingly.


alexcrichton:
Nice!

PR #244, opened 2016-11-12
Export BiLockAcquire and BiLockAcquired.

dwrensha:
Currently these types are not publicly exported, but they leak through the `BiLock::lock()` method. (I suppose that means there's a bug in rustc?)


alexcrichton:
Oops! indeed!

Unfortunately not a bug per se in rustc, but would be great to have a fix...

PR #245, opened 2016-11-15
Add a MPSC channel with backpressure handling

### LINKED ISSUE #229: Proposal: Bounded MPSC channel with back pressure handling
# Problem

"Multi producer, single consumer" (MPSC) channels are a great way for communicating across task boundaries. However, providing an implementation of a bounded MPSC channel that is future aware has some complications.

When the channel has not reached capacity, the producer handles are always "ready" to send a value. However, at some point, the channel will become full, at which point the producers will no longer be able to send the value. To be "future aware", when the channel is out of capacity, any request to send a value should be rejected and the producer task should be parked. From there, the easiest implementation would unpark all blocked tasks when the channel has additional capacity.

The issue with the easiest implementation is that it is subject to the [thundering herd problem](https://en.wikipedia.org/wiki/Thundering_herd_problem). Imagine that there are 1000 producer tasks. When the consumer receives the next value, the channel now has a single available slot. The consumer must unpark all 1000 waiting producers. This is because it is not guaranteed that a producer will actually send a value. However, it is *likely* that the producer will send a value, which means that 1000 tasks will race to send a value and only one will "win" and the 999 other tasks will become blocked again.

# Solution

**Note** I have not put much thought into optimizing the proposed implementation. I have only focused on a strategy that is possible. For example, I use Mutex, but really a spin lock would most likely be better or maybe there is a way to implement this without locks.

A MPSC channel is created with a capacity of `cap`, however it actually has a capacity of `cap + num-senders`. The structs look something like this:

```rust
struct Channel<T> {
	len: AtomicUsize,
	cap: usize,
	message_queue: MpscQueue<T>,
	blocked_queue: MpscQueue<Arc<Mutex<SenderState>>>,
}

struct SenderState {
	is_blocked: bool,
	task: Task,
}

pub struct Sender<T> {
	sender_state: Arc<Mutex<SenderState>>,
	channel: Arc<Channel<T>>,
}

pub struct Receiver<T> {
	channel: Arc<Channel<T>>,
}
```

The insight is that by having the true capacity be `cap + num-senders`, then every sender has one available slot once the channel capacity is reached.

So, the algorithms look something like this:

### Send

```
- Lock sender_state.
- Read is_blocked, if true, reject send.
- Increment len.
- If len > cap
	- park task
	- is_blocked = true
	- queue sender_state into blocked_queue
- Queue message on message_queue
```

### Receive

```
- Receive message from message_queue, if none return
- Decrement len
- Receive from blocked_queue, if none return
- Lock received sender_state
- is_blocked = false
- unpark task
```

This also has the advantage of providing fair scheduling to producers since the wait queue is first in, first out.

carllerche:
At the core, the channel uses an atomic FIFO queue for message passing.  This
queue is used as the primary coordination primitive. In order to enforce
capacity limits and handle back pressure, a secondary FIFO queue is used to
send parked task handles.

The general idea is that the channel is created with a `buffer` size of `n`.
The channel capacity is `n + num-senders`. Each sender gets one "guaranteed"
slot to hold a message. This allows `Sender` to know for a fact that a send
will succeed *before* starting to do the actual work of sending the value.
Since most of this work is lock-free, once the work starts, it is impossible
to safely revert.

If the sender is unable to process a send operation, then the the curren task
is parked and the handle is sent on the parked task queue.

Note that the implementation guarantees that the channel capacity will never
exceed the configured limit, however there is no *strict* guarantee that the
receiver will wake up a parked task *immediately* when a slot becomes
available. However, it will almost always unpark a task when a slot becomes
available and it is *guaranteed* that a sender will be unparked when the
message that caused the sender to become parked is read out of the channel.

The steps for sending a message are roughly:

* Increment the channel message count
* If the channel is at capacity, push the task handle onto the wait queue
* Push the message onto the message queue.

The steps for receiving a message are roughly:

* Pop a message from the message queue
* Pop a task handle from the wait queue
* Decrement the channel message count.

It's important for the order of operations on lock-free structures to happen
in reverse order between the sender and receiver. This makes the message queue
the primary coordination structure and establishes the necessary
happens-before semantics required for the acquire / release semantics used by
the queue structure.

Fixes #229

### DIFF
--- a/src/sync/mpsc/mod.rs
+++ b/src/sync/mpsc/mod.rs
@@ -255,13 +255,13 @@ impl<T> Sender<T> {
     fn do_send(&mut self, msg: Option<T>, can_park: bool) -> Result<(), SendError<T>> {
         // First, increment the number of messages contained by the channel.
         // This operation will also atomically determine if the sender task
-        // should be parked and if the receiver task should be unparked.
+        // should be parked.
         //
         // None is returned in the case that the channel has been closed by the
         // receiver. This happens when `Receiver::close` is called or the
         // receiver is dropped.
-        let (park_self, unpark_recv) = match self.inc_num_messages(msg.is_none()) {
-            Some((park_self, unpark_recv)) => (park_self, unpark_recv),
+        let park_self = match self.inc_num_messages(msg.is_none()) {
+            Some(park_self) => park_self,
             None => {
                 // The receiver has closed the channel. Only abort if actually
                 // sending a message. It is important that the stream
@@ -293,33 +293,16 @@ impl<T> Sender<T> {
         // Push the message onto the message queue
         self.inner.message_queue.push(msg);
 
-        // If sending the message caused the number of messages contained by
-        // the channel to go from 0 -> 1, then the sender is responsible to
-        // unpark the receiver (if it is currently parked).
-        if unpark_recv {
-            // Do this step first so that the lock is dropped when
-            // `unpark` is called
-            let task = {
-                let mut recv_task = self.inner.recv_task.lock().unwrap();
-
-                // Setting this flag enables the receiving end to detect that
-                // an unpark event happened in order to avoid unecessarily
-                // parking.
-                recv_task.unparked = true;
-                recv_task.task.take()
-            };
-
-            if let Some(task) = task {
-                task.unpark();
-            }
-        }
+        // Signal to the receiver that a message has been enqueued. If the
+        // receiver is parked, this will unpark the task.
+        self.signal();
 
         Ok(())
     }
 
     // Increment the number of queued messages. Returns if the sender should
     // block.
-    fn inc_num_messages(&self, close: bool) -> Option<(bool, bool)> {
+    fn inc_num_messages(&self, close: bool) -> Option<bool> {
         let mut curr = self.inner.state.load(Ordering::SeqCst);
 
         loop {
@@ -354,13 +337,39 @@ impl<T> Sender<T> {
                     None => false,
                 };
 
-                // Only unpark the receive half if transitioning from 0 -> 1.
-                let unpark_recv = state.num_messages == 1;
-
-                return Some((park_self, unpark_recv));
+                return Some(park_self);
             }
 
             curr = actual;
+        }
+    }
+
+    // Signal to the receiver task that a message has been enqueued
+    fn signal(&mut self) {
+        // TODO
+        // This logic can probably be improved by guarding the lock with an
+        // atomic.
+        //
+        // Do this step first so that the lock is dropped when
+        // `unpark` is called
+        let task = {
+            let mut recv_task = self.inner.recv_task.lock().unwrap();
+
+            // If the receiver has already been unparked, then there is nothing
+            // more to do
+            if recv_task.unparked {
+                return;
+            }
+
+            // Setting this flag enables the receiving end to detect that
+            // an unpark event happened in order to avoid unecessarily
+            // parking.
+            recv_task.unparked = true;
+            recv_task.task.take()
+        };
+
+        if let Some(task) = task {
+            task.unpark();
         }
     }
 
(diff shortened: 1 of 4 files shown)


src/sync/mpsc/mod.rs
```diff
@@ -0,0 +1,693 @@
...
+
+            if actual == curr {
+                return;
+            }
+
+            curr = actual;
+        }
+    }
+
+    fn next_message(&self) -> Async<Option<T>> {
```
alexcrichton: For this function to be safe (the `message_queue.pop` below) could this be tagged `&mut self`?

src/sync/mpsc/mod.rs
```diff
@@ -0,0 +1,693 @@
...
+                        }
+                        TryPark::Closed => {
+                            // The channel is closed, there will be no further
+                            // messages.
+                            return Ok(Async::Ready(None));
+                        }
+                        TryPark::NotEmpty => {
+                            // A message was sent while trying to park. Yield
+                            // the thread and try the poll operation again.
+                            thread::yield_now();
```
alexcrichton: Hm I think I may have missed this before. The `yield_now` calls above for working with `Inconsistent` I'm already slightly uncomfortable with. I feel ok with those though in the sense that there's literally a sequence of two instructions where when preempted in the middle will cause `Inconsistent`.

Here, though, there's perhaps an arbitrarily large amount of time between the increment of the number of messages and when the data's actually sent, causing a receiver to spin here for awhile. Notably in this window a call to `task::park` is made which could involve allocating memory, involving syscalls, etc, etc.

Is there perhaps an alternative to resolve this race?

src/sync/mpsc/mod.rs
```diff
@@ -0,0 +1,693 @@
...
+ * ===== impl Receiver =====
+ *
+ */
+
+impl<T> Receiver<T> {
+    /// Closes the receiving half
+    ///
+    /// This prevents any further messages from being sent on the channel while
+    /// still enabling the receiver to drain messages that are buffered.
+    pub fn close(&mut self) {
```
alexcrichton: I think that we'll want to be sure to add this to other primitives as well to ensure a uniform API surface area.

src/sync/mpsc/mod.rs
carllerche: What are your suggestions for avoiding the yield?

src/sync/mpsc/mod.rs
```diff
@@ -0,0 +1,736 @@
...
+        self.0.poll_complete()
+    }
+}
+
+impl<T> Clone for Sender<T> {
+    fn clone(&self) -> Sender<T> {
+        // Since this atomic op isn't actually guarding any memory and we don't
+        // care about any orderings besides the ordering on the single atomic
+        // variable, a relaxed ordering is acceptable.
+        let mut curr = self.inner.num_senders.load(Ordering::Relaxed);
```
alexcrichton: This whole loop and function can just be a `fetch_add` followed by an abort, right?

src/sync/mpsc/mod.rs
```diff
@@ -0,0 +1,736 @@
...
+                        TryPark::Closed => {
+                            // The channel is closed, there will be no further
+                            // messages.
+                            return Ok(Async::Ready(None));
+                        }
+                        TryPark::NotEmpty => {
+                            // A message has been sent while attempting to
+                            // park. Loop again, the next iteration is
+                            // guaranteed to get the message.
+                            continue;
```
alexcrichton: The `yield_now` was removed here, but can't this still spin? After seeing `num_messages > 1` we're not guaranteed the message is here, right?

(I may be missing something though ... )

carllerche: Yes, it can only loop a single time. It can only reach this point if it previously didn't pop a message, tried to park, and during the park a sender completed the send process (see the unparked flag in `try_park`.

src/sync/mpsc/mod.rs
carllerche: Yeah, but w/ a panic vs. an abort (still safe). Seems better to panic vs. killing the process when possible IMO.

PR #247, opened 2016-11-16
Adds `Stream::forward` combinator as the reverse of `Sink::send_all`

tailhook:
This allows examples like this:

```rust
let (stream, sink) = 
sink.send_all(
  stream.map(|msg| { 
    ..
  })
)
```

To be written as more natural chain:

```rust
let (stream, sink) = ...
stream
  .map(|msg| {
  ...
  })
  .sink_all(sink)
```

The implementation currently uses `sink::SendAll` so these things may be not obvious:

1. As result of the operation (future) `Sink` is returned. This may be ugly because we have applied a combinator on the stream. But returting Stream is not useful. Probably `()` should be returned?

2. Error returned is `Sink::SinkError` rather than `Stream::Error` (i.e. stream error is converted into sink error not vice versa)

I'll fix it if this pull request is going to be accepted. The downside of fixing these, is that `send_all` and `sink_all` operations would be subtly different, so people may mess up with these.

Also is `sink_all` name fine?

### DIFF
--- a/src/stream/forward.rs
+++ b/src/stream/forward.rs
@@ -1,0 +1,90 @@
+use {Poll, Async, Future, AsyncSink};
+use stream::{Stream, Fuse};
+use sink::Sink;
+
+/// Future for the `Stream::forward` combinator, which sends a stream of values
+/// to a sink and then waits until the sink has fully flushed those values.
+#[must_use = "futures do nothing unless polled"]
+pub struct Forward<T: Stream, U> {
+    sink: Option<U>,
+    stream: Option<Fuse<T>>,
+    buffered: Option<T::Item>,
+}
+
+
+pub fn new<T, U>(stream: T, sink: U) -> Forward<T, U>
+    where U: Sink<SinkItem=T::Item>,
+          T: Stream,
+          T::Error: From<U::SinkError>,
+{
+    Forward {
+        sink: Some(sink),
+        stream: Some(stream.fuse()),
+        buffered: None,
+    }
+}
+
+impl<T, U> Forward<T, U>
+    where U: Sink<SinkItem=T::Item>,
+          T: Stream,
+          T::Error: From<U::SinkError>,
+{
+    fn sink_mut(&mut self) -> &mut U {
+        self.sink.as_mut().take()
+            .expect("Attempted to poll Forward after completion")
+    }
+
+    fn stream_mut(&mut self) -> &mut Fuse<T> {
+        self.stream.as_mut().take()
+            .expect("Attempted to poll Forward after completion")
+    }
+
+    fn take_result(&mut self) -> (T, U) {
+        let sink = self.sink.take()
+            .expect("Attempted to poll Forward after completion");
+        let fuse = self.stream.take()
+            .expect("Attempted to poll Forward after completion");
+        return (fuse.into_inner(), sink)
+    }
+
+    fn try_start_send(&mut self, item: T::Item) -> Poll<(), U::SinkError> {
+        debug_assert!(self.buffered.is_none());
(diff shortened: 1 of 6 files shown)


alexcrichton:
I think in terms of API forward-compatibility we'd want to return a new type from this method, but it's fine if it's just wraps `send_all` for now (which is what I'd expect).

I wonder if perhaps we could call this `proxy`? That is, proxy all messages from a stream to a different sink.

alexcrichton:
Another option, `copy`, to perhaps mirror `std::io::copy` somewhat

tailhook:
In my vision `proxy` is tighly coupled with something that represents original object. I mean proxy of a stream is a stream of some kind. Similarly about `copy`: there isn't anything we are copying. In `io::copy` we actually copying bytes, but when we fetch something from stream we don't copy it, we move it to the sink. (on the other hand, I'm not native english, so I often misunderstand original meaning of technical terms :) )

Maybe `forward` is a good name? By the way maybe we should use `Stream::send_all` and `Sink::receive_all` ? Because current `sink.send_all(stream)` is a little bit awkward.

Regardless of the naming. I didn't think about forward compability, so I'll make a new type. But should it be exact the same as `SendAll` or maybe fix return and error value of the future as described above?

alexcrichton:
Ah ok the avoidance of `proxy` and `copy` makes sense to me. I like `forward` though! I was also hesitant about `Sink::send_all` but it made sense to me when I thought about it in terms of:

```rust
tx.send_all(some_list);
```

That is, if you have a `Sender`, then you'd expect to be able to call `send` on it. Extending that to `send_all` seems somewhat reasonable as well.

About the types here. Hm I wonder if we should change both this method and `send_all` to return both the stream and the sink? Rather than always destroying one but returning the other.

Otherwise yeah changing the error here to the stream's error makes sense to me.

tailhook:
Okay, renamed to `forward`, and returns tuple. Should I also include change to send_all in this pull request?

Other note: it looks like we can't use `SendAll` here because of mirrored error conversion (or at least I don't see any simple way of that).

src/stream/forward.rs
```diff
@@ -0,0 +1,90 @@
...
+          T::Error: From<U::SinkError>,
+{
+    type Item = (T, U);
+    type Error = T::Error;
+
+    fn poll(&mut self) -> Poll<(T, U), T::Error> {
+        // If we've got an item buffered already, we need to write it to the
+        // sink before we can do anything else
+        if let Some(item) = self.buffered.take() {
+            try_ready!(self.try_start_send(item))
```
alexcrichton: I think that if this fails with `NotReady` we'll need to guarantee at least one call to `poll_complete`, right?

tailhook: Well, this is a good question. It's not clear for me that `start_send` never does any I/O. Because otherwise every failure of `start_send` is basically a `start_send -> poll_complete -> start_send`.

From what I can see:
1. `tokio_core::Framed` calls poll_complete under pressure.
2. `sink::Send` doesn't call poll_complete on `AsyncSink::NotReady`
3. `SendAll` has same behavior (because I've copied from there)
4. `sink::Buffer` always tries to flush on start_send (which is fine), and does `poll_complete` on `NotReady` [*]

[*] So only a buffer expects this semantics. And yes if we rely on assumption of `start_send` doing I/O under pressure, we have a bug in Buffer (which ironically is fixed by removing `poll_complete` in `Buffer::start_send`)

alexcrichton: Oh nevermind, I'm misremembering. Sink docs:

> The method returns AsyncSink::NotReady if the sink was unable to begin sending, usually due to being full. The sink must have attempted to complete processing any outstanding requests (equivalent to poll_complete) before yielding this result.

So tl;dr; the answer to my question is "no"

Issue #248, opened 2016-11-16
Misleading description of sink error


tailhook:
[Documentation states](http://alexcrichton.com/futures-rs/futures/sink/trait.Sink.html#errors):

> If the sink encounters an error other than being temporarily full, it uses the Err variant to signal that error. In most cases, such errors mean that the sink will permanently be unable to receive items.

I think "In most cases" does mean in all cases here, right? If it is not, there is no way to find out if is it allowed to call `poll_complete` again or not. Because of this:

> # Panics
> [..snip..]
> A previous call to start_send or poll_complete yielded a permanent error.

I.e. there is no way to find out which error is permanent and which isn't.

alexcrichton:
I believe the intended semantics here were:

* If you're working with `S: Sink`, then you have to abide by "error == fatal"
* If you're working with a *specific* sink, you're not contractually obliged to translate "error == fatal"

In other words, sinks may or may not be safe to use after an error (like iterators after they're exhausted).

PR #251, opened 2016-11-18
Implement `Future::flatten_sink` similar to `Future::flatten_stream`

tailhook:
The real use case for this thing is something like:

```
let sink = TcpStream::connect(addr, handle)
  .map(|(sock, _)| {
    let (_, sink) = sock.framed(Codec);
    sink
  })
  .flatten_sink();
```

### DIFF
--- a/src/future/flatten_sink.rs
+++ b/src/future/flatten_sink.rs
@@ -1,0 +1,91 @@
+use std::mem;
+
+use {Async, AsyncSink, StartSend, Poll, Future, Sink};
+
+
+enum State<F: Future>
+    where F::Item: Sink
+{
+    Future(F),
+    Sink(F::Item),
+    Error(<F::Item as Sink>::SinkError),
+    Void,
+}
+
+/// Future for the `into_sink` combinator, flattening a future-of-a-sink to
+/// get just the result of the final sink instead
+///
+/// This is created by the `Future::into_sink` method.
+pub struct FlattenSink<F: Future>
+    where F::Item: Sink
+{
+    state: State<F>,
+}
+
+pub fn new<F>(future: F) -> FlattenSink<F>
+    where F: Future,
+          F::Item: Sink
+{
+    FlattenSink {
+        state: State::Future(future),
+    }
+}
+
+impl<F: Future> Sink for FlattenSink<F>
+    where F::Item: Sink,
+          <F::Item as Sink>::SinkError: From<F::Error>,
+{
+    type SinkItem = <F::Item as Sink>::SinkItem;
+    type SinkError = <F::Item as Sink>::SinkError;
+    fn start_send(&mut self, item: Self::SinkItem)
+        -> StartSend<Self::SinkItem, Self::SinkError>
+    {
+        let (res, state) = match mem::replace(&mut self.state, State::Void) {
+            State::Future(mut fut) => {
+                match fut.poll() {
+                    Ok(Async::Ready(mut sink)) => {
+                        (sink.start_send(item)?, State::Sink(sink))
+                    }
+                    Ok(Async::NotReady) => {
+                        (AsyncSink::NotReady(item), State::Future(fut))
+                    }
+                    Err(e) => {
+                        (AsyncSink::NotReady(item), State::Error(e.into()))
+                    }
+                }
+            }
+            State::Sink(mut sink) => {
+                (sink.start_send(item)?, State::Sink(sink))
+            }
+            State::Void => {
+                panic!("Called FlattenSink::start_send after error");
+            }
+            State::Error(e) => (AsyncSink::NotReady(item), State::Error(e)),
+        };
+        self.state = state;
+        return Ok(res);
+    }
+    fn poll_complete(&mut self) -> Poll<(), Self::SinkError> {
+        let (res, state) = match mem::replace(&mut self.state, State::Void) {
+            State::Future(mut fut) => {
+                match fut.poll()? {
+                    Async::Ready(sink) => {
+                        (Async::Ready(()), State::Sink(sink))
+                    }
+                    Async::NotReady => {
+                        (Async::NotReady, State::Future(fut))
+                    }
+                }
+            }
+            State::Sink(mut sink) => {
+                (sink.poll_complete()?, State::Sink(sink))
+            }
+            State::Void => {
+                panic!("Called FlattenSink::start_send after error");
+            }
+            State::Error(e) => return Err(e),
+        };
+        self.state = state;
+        return Ok(res);
+    }
+}
--- a/src/future/mod.rs
+++ b/src/future/mod.rs
@@ -45,6 +45,7 @@ mod map_err;
 mod or_else;
 mod select;
 mod then;
+mod flatten_sink;
 
 // impl details
 mod chain;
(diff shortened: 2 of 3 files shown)


src/future/flatten_sink.rs
```diff
@@ -0,0 +1,91 @@
...
+    type SinkItem = <F::Item as Sink>::SinkItem;
+    type SinkError = <F::Item as Sink>::SinkError;
+    fn start_send(&mut self, item: Self::SinkItem)
+        -> StartSend<Self::SinkItem, Self::SinkError>
+    {
+        let (res, state) = match mem::replace(&mut self.state, State::Void) {
+            State::Future(mut fut) => {
+                match fut.poll() {
+                    Ok(Async::Ready(mut sink)) => {
+                        (sink.start_send(item)?, State::Sink(sink))
```
tailhook: Well, my assumption was that when working with `T: Sink` we always treat `SinkError` as fatal error (as discussed in #248). So whether we store sink back in our state machine or not, we must not call `start_send` or `poll_complete` again.

Do you think that it's better to call `start_send` and `poll_complete` anyway? I.e. rely on the underlying sink to panic when `start_send` or `poll_complete` is used incorrectly.

tailhook: Another thought: even if original Sink supports resume after error in `start_send` and `poll_complete` we still have a case when initial `Future` returned error, which we can never resume from. I.e. allowing to resume from error in sink here requires from user to know which errors are resumable *and* to ensure that this kind of errors can't be returned from `Future`.

I believe this is very narrow use case and it's too easy to misuse. So it's better to panic on this for now and enable this behavior later if there is a clear use case (I think removing some panics usually is backward compatible?).

alexcrichton: Oh right yeah we shouldn't ourselves call the underlying sink again as part of normal usage, but I figure that we'd delegate to the underlying sink to "act improperly", if at all. In that sense this future would be robust in the face of use-after-error if the underlying sink was as well.

I do think though that if the future returns the error then this implementation should just be fused to forever return `NotReady`

src/future/flatten_sink.rs
```diff
@@ -0,0 +1,91 @@
...
+                    }
+                }
+            }
+            State::Sink(mut sink) => {
+                (sink.start_send(item)?, State::Sink(sink))
+            }
+            State::Void => {
+                panic!("Called FlattenSink::start_send after error");
+            }
+            State::Error(e) => (AsyncSink::NotReady(item), State::Error(e)),
```
alexcrichton: Hm doesn't this mean that if the future returns an error it's never returned upwards?

alexcrichton: (er well at least it seems like here if the future returned an error we'd want to return that upwards to the caller of `start_send` rather than delay until a `poll_complete`

tailhook: In this case, we will loose a request. 

For my use case, where future is a `TcpStream` connect it leads to the following consequences:

1. Let's imagine we have a connection pool connected to the two servers
2. Second, server is unresponsive for some time
3. Pool reconnects to the second server continuously, until it becomes available
4. When pool has new request it forwards it to sinks in round-robin fashion, trying `start_send` of each connection 

So in my implementation on (4) it will receive request back and try another connection, while continuing to do `poll_complete` on every connection, so will receive the error on next call.

In your proposed implementation, pool loses one request on each reconnect. Or I need to call `poll_complete` before sending every request (so can't batch them). And latter case contains a race condition: nothing guarantees that `start_send` will be consistent a millisecond after `poll_complete` returned `NotReady`.

Does it make sense?

src/future/flatten_sink.rs
tailhook: Well, I'll fix the code if we agree on that, but your previous message was confusing :). So I need to make things more clear:

> I do think though that if the future returns the error then this implementation should just be fused to forever return NotReady

Why should it be fused?

1. Most futures panic when poll'ed after error
2. Sink is documented that can panic when polled after error

I think it's more clear to panic, because there is no real use case of the fused sink. And calling poll_complete/start_send again means there is a bug in code that silenced an error.

Even if we do return `NotReady` on the error of the future user still has to differentiate the error generated by Future and error generated by Sink because there is no sense to continue using sink when Future returned error (it just *hangs* from the viewpoint of the user).

And I think this part:
> I figure that we'd delegate to the underlying sink to "act improperly", if at all

.. only works if we return `NotReady` after errors of the Future, right?

src/future/flatten_sink.rs
alexcrichton: Yeah that use case makes sense, but relies on `poll_complete` eventually getting called. We've [already got code](https://github.com/alexcrichton/futures-rs/blob/81273d39d09746729a390e0d3ba39cfc097d840f/src/stream/forward.rs#L69-L73), though, which only calls `start_send` instead of `poll_complete`. Additionally the [current documentation on `start_send`](https://github.com/alexcrichton/futures-rs/blob/81273d39d09746729a390e0d3ba39cfc097d840f/src/sink/mod.rs#L114-L119) may indicate that this should propagate an error?

Now this does mean that the method may not be appropriate for the use case you're thinking of. In general though I think that makes sense, though? If you've got a pooled connection presumably you'd want some cloneable request to send to the pool, because even if the pool *didn't* hit an error in the future, but it hit an error in the sink, you'd want to retry the request, right?

src/future/flatten_sink.rs
tailhook: > I figure that if there's interest in differentiating the future error from the sink error this combinator wouldn't be used, right? Instead it'd be handled manually by having the error handling for the future be slightly different?

Yes, good point. But this only proves what I'm saying. Let me elaborate on it:

1. Let's say we have a `Sink` that returns `io::Error`. And we know that if it returns `io::ErrorKind::Other` it is resumable.
2. Then we use `flatten_stream` with arbitrary future that also returns `io::Error`, errors are compatible, so we're happy to use it.
3. Then on some very rare occasion Future returns `Other` too and we know that we could resume when `Other` error is returned. So we restart and get panic.

This kind of situation is what I mean by this point:

>  user still has to differentiate the error generated by Future and error generated by Sink

So the only safe way to make `flatten_stream` resumable on error is to return `enum Error { FutureError(A), SinkError(B) }`, which is too narrow use case to include here I think.

alexcrichton:
cc @aturon and @carllerche there's some interesting discussion here of guarantees on `Sink` and errors and such.

I think this also relates to https://github.com/alexcrichton/futures-rs/issues/232 and perhaps https://github.com/alexcrichton/futures-rs/issues/237 as well.

src/future/flatten_sink.rs
alexcrichton: I think to support cases like this you're [right below](https://github.com/alexcrichton/futures-rs/pull/251#discussion_r89178646) in that we [may need to change `Sink`](https://github.com/alexcrichton/futures-rs/pull/251#issuecomment-262351631)

tailhook:
Well, it looks like I don't need this anymore for my original use case. Given there are unresolved questions and nobody is interested in this, should it be just closed for now? At least until somebody finds a good use case for the functionality.

alexcrichton:
Ok sounds good to me. I'd still be fine adding this functionality, though, I've already used the stream version before and it was quite convenient!

PR #252, opened 2016-11-19
Implement `s.split()` for `Stream + Sink`

tailhook:
Basically, this replaces `tokio_core::io::Framed::split`, it's unclear why framed needs specific implementation instead of generic one like this.

Also, I think we should remove `Stream` and `Sink` for `BiLock<T>` in favor of this idea. I don't think using Stream on both instances of the same BiLock is generally useful.

### DIFF
--- a/src/stream/split.rs
+++ b/src/stream/split.rs
@@ -1,0 +1,48 @@
+use {StartSend, Sink, Stream, Poll, Async, AsyncSink};
+use sync::BiLock;
+
+/// A `Stream` part of the split pair
+pub struct SplitStream<S>(BiLock<S>);
+
+impl<S: Stream> Stream for SplitStream<S> {
+    type Item = S::Item;
+    type Error = S::Error;
+
+    fn poll(&mut self) -> Poll<Option<S::Item>, S::Error> {
+        match self.0.poll_lock() {
+            Async::Ready(mut inner) => inner.poll(),
+            Async::NotReady => Ok(Async::NotReady),
+        }
+    }
+}
+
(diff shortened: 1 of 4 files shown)


src/stream/split.rs
```diff
@@ -0,0 +1,39 @@
...
+    fn poll_complete(&mut self) -> Poll<(), S::SinkError> {
+        self.0.poll_complete()
+    }
+}
+
+pub fn split<S: Stream + Sink>(s: S) -> (SplitStream<S>, SplitSink<S>) {
+        let (a, b) = BiLock::new(s);
+        let read = SplitStream(a);
+        let write = SplitSink(b);
+        (read, write)
```
alexcrichton: indentation here may be a bit off?

alexcrichton:
Yeah let's go a head and remove the impls directly on `BiLock` itself, this makes more sense to me 👍

alexcrichton:
Also we tend to prefer `let (tx, rx) = foo()` for pair-returning methods, so perhaps the order of the return value could be swapped?

tailhook:
I have to admit that we use `(tx, rx)` for things returning channel. This use case is quite opposite. However, it may happen that `split()` is useful for channel-like things too (not very sure though), and it's still easier to remember if order is same. So, I've changed method to return `(sink, stream)`.

Also fixed indent, rebased and removed Stream/Sink for BiLock.

tailhook:
Okay, last amendment fixes `--no-default-features`

alexcrichton:
Thanks!

PR #259, opened 2016-11-23
Add `FutureFrom` trait for future-based conversion

aturon:
Analogous to the `From` trait in `std`, but works asynchronously.

### DIFF
--- a/src/future/mod.rs
+++ b/src/future/mod.rs
@@ -782,3 +782,18 @@ impl<T, E> IntoFuture for result::Result<T, E> {
         result(self)
     }
 }
+
+/// Asynchronous conversion from a type `T`.
+///
+/// This trait is analogous to `std::convert::From`, adapted to asynchronous
+/// computation.
+pub trait FutureFrom<T>: Sized {
+    /// The future for the conversion.
+    type Future: Future<Item=Self, Error=Self::Error>;
+
+    /// Possible errors during conversion.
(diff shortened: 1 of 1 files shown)


alexcrichton:
I wonder if we should perhaps draw a parallel with `TryFrom`? In the sense that this is fallible it's closer to that rather than `From`.

Maybe `FutureTryFrom`?

aturon:
@alexcrichton Given that `Result`-ness is build into the `Future` type, adding `Try` here feels redundant to me -- it's already implied.

Another open question: what blanket impls, if any, should we have? One lifting `TryFrom` would make the most sense, but one lifting `From` might be more useful...

alexcrichton:
Ah that makes sense to me!

Perhaps we could add a blanket impl for `Result` or anything that implements `IntoFuture`? 

Hmm.... Can we have an `IntoFuture` coupling like libstd does with `From` and `Into`?

aturon:
@alexcrichton 

> Hmm.... Can we have an `IntoFuture` coupling like libstd does with `From` and `Into`?

In `std`, we have a blanket from `From` to `Into`, which unfortunately won't work in futures-rs due to the existing blanket impls (until we can use specialization).

I think for the moment let's avoid any blanket impls, and let this bake.

alexcrichton:
Ok, sounds good to me

Issue #262, opened 2016-11-27
reached the recursion limit during monomorphization (selection ambiguity)


weiznich:
Compiling some futures related code fails with:
`error: reached the recursion limit during monomorphization (selection ambiguity)`
```rust
extern crate futures; // 0.1.6
extern crate futures_cpupool; // 0.1.2

use futures::Future;
use futures_cpupool::CpuPool;

const MAX: usize = 10;

fn work(n: u32) -> Result<u32, ()> {
    ::std::thread::sleep_ms(n * 2);
    Ok(n)
}

fn fold_b(a: u64, b: u32) -> u64 {
    ::std::thread::sleep_ms(b);
    a + b as u64
}

fn main() {
    let pool = CpuPool::new_num_cpus();
    let data = (0..MAX).rev().collect::<Vec<_>>();
    let a = {
        let workers = data.clone()
            .into_iter()
            .map(|d| pool.spawn_fn(move || work(d as u32)));
        ::futures::collect(workers).wait().unwrap().into_iter().fold(0, fold_b)
    };
}
```

`
rustc --version
rustc 1.15.0-nightly (d9bdc636d 2016-11-24)
`

weiznich:
It seems that this was introduced between version 0.1.3 and version 0.1.4.

alexcrichton:
I've opened https://github.com/rust-lang/rust/issues/38033 to track this

dwrensha:
This particular instance of the error appears to have been eliminated, so I think it makes sense to close this issue.

I've found another instance of this error message and reported it as https://github.com/alexcrichton/futures-rs/issues/367.

alexcrichton:
Oops forgot to close this with @dwrensha's last comment.

PR #263, opened 2016-11-27
Add a ergonomic method to convert a list of futures into a stream

weiznich:
* as suggested by @alexcrichton in #242

### DIFF
--- a/src/stream/futures_unordered.rs
+++ b/src/stream/futures_unordered.rs
@@ -1,0 +1,98 @@
+use future::{Future, IntoFuture};
+use stream::Stream;
+use poll::Poll;
+use Async;
+use stack::{Stack, Drain};
+use std::sync::Arc;
+use task::{self, UnparkEvent};
+
+use std::prelude::v1::*;
+
+/// An adaptor for a stream of futures to execute the futures concurrently, if
+/// possible, delivering results as they become available.
+///
+/// This adaptor will return their results in the order that they complete.
+/// This is created by the `futures` method.
+///
+#[must_use = "streams do nothing unless polled"]
+pub struct FuturesUnordered<F>
+    where F: Future
+{
+    futures: Vec<Option<F>>,
+    stack: Arc<Stack<usize>>,
+    pending: Option<Drain<usize>>,
+    active: usize,
+}
+
+/// Converts a `Vec` of `IntoFuture` elements into a `Stream` over the underlying
+/// futures. The created stream will yield into values produced by those futures.
+/// Similar to `buffer_unordered` the order of the produced items will not be equal
+/// with the order of elements in the input vector.
+///
+pub fn futures_unordered<I>(futures: I) -> FuturesUnordered<<I::Item as IntoFuture>::Future>
+    where I: IntoIterator,
+          I::Item: IntoFuture
+{
+    let futures = futures.into_iter().map(IntoFuture::into_future).map(Some).collect::<Vec<_>>();
+    let stack = Arc::new(Stack::new());
+    for i in 0..futures.len() {
+        stack.push(i);
+    }
+    FuturesUnordered {
+        active: futures.len(),
+        futures: futures,
+        pending: None,
+        stack: stack,
+    }
+}
+
+impl<F> FuturesUnordered<F>
+    where F: Future
+{
+    fn poll_pending(&mut self, mut drain: Drain<usize>) -> Option<Poll<Option<F::Item>, F::Error>> {
+        while let Some(id) = drain.next() {
+            let event = UnparkEvent::new(self.stack.clone(), id);
+            let ret = match task::with_unpark_event(event, || {
+                self.futures[id]
+                    .as_mut()
+                    .unwrap()
+                    .poll()
+            }) {
+                Ok(Async::NotReady) => {
+                    continue;
+                }
+                Ok(Async::Ready(val)) => Ok(Async::Ready(Some(val))),
+                Err(e) => Err(e),
+            };
+            self.pending = Some(drain);
+            self.active -= 1;
+            self.futures[id] = None;
+            return Some(ret);
+        }
+        None
+    }
+}
+
+impl<F> Stream for FuturesUnordered<F>
+    where F: Future
+{
+    type Item = F::Item;
+    type Error = F::Error;
+
+    fn poll(&mut self) -> Poll<Option<Self::Item>, Self::Error> {
+        if self.active == 0 {
+            return Ok(Async::Ready(None));
+        }
+        let drain = self.pending
+            .take()
+            .unwrap_or_else(|| self.stack.drain());
+        if let Some(ret) = self.poll_pending(drain){
+            return ret;
+        }
+        let drain = self.stack.drain();
+        if let Some(ret) = self.poll_pending(drain){
+            return ret;
+        }
+        Ok(Async::NotReady)
+    }
+}
--- a/src/stream/mod.rs
+++ b/src/stream/mod.rs
@@ -75,6 +75,7 @@ if_std! {
     mod wait;
     mod channel;
     mod split;
+    mod futures_unordered;
     pub use self::buffered::Buffered;
     pub use self::buffer_unordered::BufferUnordered;
     pub use self::catch_unwind::CatchUnwind;
@@ -82,6 +83,7 @@ if_std! {
     pub use self::collect::Collect;
     pub use self::wait::Wait;
     pub use self::split::{SplitStream, SplitSink};
+    pub use self::futures_unordered::{futures_unordered, FuturesUnordered};
 
     #[doc(hidden)]
     #[cfg(feature = "with-deprecated")]
--- a/tests/futures_unordered.rs
+++ b/tests/futures_unordered.rs
@@ -1,0 +1,44 @@
+#[macro_use]
+extern crate futures;
+
+use futures::sync::oneshot;
+use futures::stream::futures_unordered;
+use futures::Future;
+
+mod support;
+
+#[test]
+fn works_1() {
+    let (a_tx, a_rx) = oneshot::channel::<u32>();
+    let (b_tx, b_rx) = oneshot::channel::<u32>();
+    let (c_tx, c_rx) = oneshot::channel::<u32>();
+
+    let stream = futures_unordered(vec![a_rx, b_rx, c_rx]);
+
+    let mut spawn = futures::executor::spawn(stream);
+    b_tx.complete(99);
+    assert_eq!(Some(Ok(99)), spawn.wait_stream());
+
+    a_tx.complete(33);
+    c_tx.complete(33);
+    assert_eq!(Some(Ok(33)), spawn.wait_stream());
+    assert_eq!(Some(Ok(33)), spawn.wait_stream());
+    assert_eq!(None, spawn.wait_stream());
+}
+
+
+#[test]
+fn works_2() {
+    let (a_tx, a_rx) = oneshot::channel::<u32>();
+    let (b_tx, b_rx) = oneshot::channel::<u32>();
+    let (c_tx, c_rx) = oneshot::channel::<u32>();
+
+    let stream = futures_unordered(vec![a_rx.boxed(), b_rx.join(c_rx).map(|(a, b)| a + b).boxed()]);
+
+    let mut spawn = futures::executor::spawn(stream);
+    a_tx.complete(33);
+    b_tx.complete(33);
+    assert!(spawn.poll_stream(support::unpark_noop()).unwrap().is_ready());
+    c_tx.complete(33);
+    assert!(spawn.poll_stream(support::unpark_noop()).unwrap().is_ready());
+}


weiznich:
While trying to get some benchmark numbers I've hit #262 several times.

src/stream/futures.rs
```diff
@@ -0,0 +1,71 @@
...
+    futures: Vec<Option<F>>,
+    active: usize,
+}
+
+/// Converts a `Vec` of `IntoFuture` elements into a `Stream` over the underlying
+/// futures. The created stream will yield into values produced by those futures.
+/// Similar to `buffer_unordered` the order of the produced items will not be equal
+/// with the order of elements in the input vector.
+///
+pub fn futures<F>(futures: Vec<F>) -> Futures<F::Future>
```
Stebalien: Is there any reason to not just take an `IntoIterator`?

src/stream/futures.rs
```diff
@@ -0,0 +1,71 @@
...
+    where F: Future
+{
+    type Item = F::Item;
+    type Error = F::Error;
+
+    fn poll(&mut self) -> Poll<Option<Self::Item>, Self::Error> {
+        if self.active == 0 {
+            return Ok(Async::Ready(None));
+        }
+        for f in self.futures.iter_mut() {
```
Stebalien: Nit: `for f in &mut self.futures {`

src/stream/futures.rs
```diff
@@ -0,0 +1,71 @@
...
+            let ret;
+            match *f {
```
Stebalien: Nit: `let ret = match *f {`

src/stream/futures.rs
```diff
@@ -0,0 +1,71 @@
...
+                None => continue,
+                Some(ref mut future) => {
+                    match future.poll() {
+                        Ok(Async::NotReady) => {
+                            continue;
+                        }
+                        Ok(Async::Ready(e)) => {
+                            ret = e;
```
Stebalien: Just yield `e` (`=> e,`). However, `e` usually refers to errors so I'd use `v` or `val`.

danburkert:
I didn't see this before opening #265.  I'm biased, but I think the solution outlined there is a little more elegant since it doesn't require allocating in the implementation (although it's arguably less flexible in the inputs).  Whatever the implementation, I think the name should have a hint as to what order the items are returned in the stream, since it's equally valid to expect iteration vs. completion order.

weiznich:
@danburkert 
>I'm biased, but I think the solution outlined there is a little more elegant since it doesn't require allocating in the implementation (although it's arguably less flexible in the inputs).


I've adopted some ideas from #265, so this should now only allocate if `IntoFuture::into_future` change the item type. Hopefully the compiler is able to optimize the allocation caused by the flexible input type away in case of a vector over futures. (I didn't check this.)

> Whatever the implementation, I think the name should have a hint as to what order the items are returned in the stream, since it's equally valid to expect iteration vs. completion order.

I simply used the name @alexcrichton used in his pseudo-code in #242.

src/stream/futures.rs
```diff
@@ -0,0 +1,67 @@
...
+        let mut ready = None;
+        for (id, f) in self.futures.iter_mut().enumerate() {
+            match f.poll() {
+                Ok(Async::NotReady) => {}
+                Ok(Async::Ready(val)) => {
+                    ready = Some((id, val));
+                    break;
+                }
+                Err(e) => {
+                    return Err(e);
```
danburkert: This will cause an error-producing future to be polled repeatedly.  You may want to add a test case with a failing future.

src/stream/futures.rs
```diff
@@ -0,0 +1,67 @@
...
+/// Similar to `buffer_unordered` the order of the produced items will not be equal
+/// with the order of elements in the input vector.
+///
+pub fn futures<I, It, F>(futures: I) -> Futures<F::Future>
+    where I: IntoIterator<Item = F, IntoIter = It>,
+          It: Iterator<Item = F> + ExactSizeIterator,
+          F: IntoFuture
+{
+    Futures {
+        futures: futures.into_iter().map(IntoFuture::into_future).collect(),
```
danburkert: Will the compiler really optimize away this allocation?  Seems simpler to me to just have the caller pass in a `Vec`, it's a much more straightforward signature as well.

src/stream/futures.rs
```diff
@@ -0,0 +1,67 @@
...
+}
+
+/// Converts a `Vec` of `IntoFuture` elements into a `Stream` over the underlying
+/// futures. The created stream will yield into values produced by those futures.
+/// Similar to `buffer_unordered` the order of the produced items will not be equal
+/// with the order of elements in the input vector.
+///
+pub fn futures<I, It, F>(futures: I) -> Futures<F::Future>
+    where I: IntoIterator<Item = F, IntoIter = It>,
+          It: Iterator<Item = F> + ExactSizeIterator,
```
danburkert: Is `ExactSizeIterator` necessary?

src/stream/futures.rs
alexcrichton: I don't believe anything will optimize this allocation away, but I'd also think that it shouldn't matter too much in practice. @danburkert have you seen this show up in profiles though?

I'd actually be more worried about the scalability of this combinator for polling all the futures all the time, but we have methods of fixing that.

src/stream/futures.rs
danburkert: More generally, I think this signature can be simplified to a single type parameter by doing what `select_all` does.

src/stream/futures.rs
danburkert: > @danburkert have you seen this show up in profiles though?

Nope, I don't personally have any futures-based applications that have gotten to the point of perf. profiling.  I was mainly concerned because eliding the copy requires a change to the public interface.

> I'd actually be more worried about the scalability of this combinator for polling all the futures all the time, but we have methods of fixing that.

That's a great point.  Are you thinking something along the lines of putting the futures into a Slab and using `with_unpark_event`?  If so that is a strong reason to keep the iterator based API.

alexcrichton: Ah yeah eventually we'll likely use `with_unpark_event` here, similar to what `buffer_unordered` does right now. We may not literally use a `Slab` as this is a fixed number of futures, but same underlying piece.

src/stream/futures.rs
weiznich: What is the problem with a stream returning the same error when polled again? I think if an error occurred the whole stream should treated as failed. Or did I misunderstand the concept and an error only signals that one item has failed? (In this case the item should be removed similar to the Async::Ready case)

danburkert: > What is the problem with a stream returning the same error when polled again?

Polling an unfused future after it has completed is an error, from `Future::poll`:

> Once a future has completed (returned Ready or Err from poll), then any future calls to poll may panic, block forever, or otherwise cause wrong behavior. The Future trait itself provides no guarantees about the behavior of poll after a future has completed.

> I think if an error occurred the whole stream should treated as failed.

I disagree, its useful to let the application decide this instead of the implementation.

src/stream/futures.rs
```diff
@@ -0,0 +1,80 @@
...
+{
+    type Item = F::Item;
+    type Error = F::Error;
+
+    fn poll(&mut self) -> Poll<Option<Self::Item>, Self::Error> {
+        if self.futures.is_empty() {
+            return Ok(Async::Ready(None));
+        }
+        let mut ready = None;
+        for (id, f) in self.futures.iter_mut().enumerate() {
```
danburkert: This should only be iterating the indices in `stack`.

src/stream/futures.rs
```diff
@@ -0,0 +1,80 @@
...
+                    ready = Some((id, val));
+                    break;
+                }
+                Err(e) => {
+                    return Err(e);
+                }
+            }
+        }
+        if let Some((id, val)) = ready {
+            self.futures.swap_remove(id);
```
danburkert: `swap_remove` will change the index of the last future in the `Vec`, so its likely it will never be polled.  This is why I suggested using a slab, since either you need a collection where the index won't change others on removal, or the swapped future needs to get repolled with an updated unpark event.

danburkert: It occurred to me that swap_remove is no longer needed, since the future should not be polled again.  This also means the non-allocating version which just takes a Vec could work with unpark events.

alexcrichton: Yeah let's just make the vector a vector of `Option` so the indices stay the same

src/stream/futures.rs
alexcrichton: Yeah right now streams can return multiple errors while futures can't, so once we hit an error we should erase the future and go to the next future on the next iteration.

src/stream/futures.rs
```diff
@@ -0,0 +1,83 @@
...
+/// possible, delivering results as they become available.
+///
+/// This adaptor will return their results in the order that they complete.
+/// This is created by the `futures` method.
+///
+#[must_use = "streams do nothing unless polled"]
+pub struct Futures<F>
+    where F: Future
+{
+    futures: Vec<F>,
```
alexcrichton: I think we'll still want this to be a vector of options, as soon as a future is done we should destroy the future to deallocate all resources associated with it.

src/stream/futures.rs
```diff
@@ -0,0 +1,83 @@
...
+
+    fn poll(&mut self) -> Poll<Option<Self::Item>, Self::Error> {
+        if self.active == 0 {
+            return Ok(Async::Ready(None));
+        }
+        let mut drain = if self.pending.is_some() {
+            ::std::mem::replace(&mut self.pending, None).unwrap()
+        } else {
+            self.stack.drain()
+        };
```
alexcrichton: This could be:

```rust
let mut drain = self.pending.take().unwrap_or_else(|| self.stack.drain());
```

src/stream/futures.rs
```diff
@@ -0,0 +1,83 @@
...
+        while let Some(id) = drain.next() {
+            let event = UnparkEvent::new(self.stack.clone(), id);
+            let ref mut f = self.futures[id];
+            match task::with_unpark_event(event, || f.poll()) {
+                Ok(Async::NotReady) => {}
```
alexcrichton: Perhaps this branch could be `continue` and the `active`/`pending` adjustments of the next two branches could be shared?

src/stream/futures.rs
```diff
@@ -0,0 +1,80 @@
...
+                }
+            }else{
+                unreachable!()
+            };
+            self.active -= 1;
+            self.pending = Some(drain);
+            self.futures[id] = None;
+            return ret;
+        }
+        Ok(Async::NotReady)
```
danburkert: should this return `Ok(Async::Ready(None))` if `self.active == 0`?  It's possible the stream will never get woken up again, but I'm not sure if that matters.

alexcrichton: I think the beginning of this function should take care of that, right?

danburkert: yep, you're right.  I mistakenly thought there was a path that would both remove from the vec and fall down to here, but that's not the case.

alexcrichton:
Ah sorry for letting this fall by the wayside a bit, but looking good to me! Some last nits:

* I think that the loop will want to be executed twice. First we want to exhaust `self.pending` but after that we want to exhaust `self.stack`. Right now only at most one of those lists are looked at but we'll need to be sure to peek at both.
* I was a little unsure about the name `futures` but in retrospect that seems ok. Perhaps we can call this `futures_unordered` though to leave room for an implementation of `futures` later that preserves the order of the original iterator?

weiznich:
> I think that the loop will want to be executed twice. First we want to exhaust self.pending but after that we want to exhaust self.stack. Right now only at most one of those lists are looked at but we'll need to be sure to peek at both.

The test case results in a correct execution order. Also the stream finished if all elements have finished. So I'm sot sure why we want to iterate over the stack again.

alexcrichton:
You'll need to iterate over both because the existing pending iterator may not contain all events that need to be processed on this iteration. If all of the pending events are not ready but an event in the stack is ready, then it needs to be looked at. This test case shows the failure:

```rust
#[macro_use]
extern crate futures;

use futures::Future;
use futures::sync::oneshot;
use futures::stream::futures_unordered;

mod support;

#[test]
fn works() {
    let (a_tx, a_rx) = oneshot::channel::<u32>();
    let (b_tx, b_rx) = oneshot::channel::<u32>();
    let (c_tx, c_rx) = oneshot::channel::<u32>();

    let stream = futures_unordered(vec![
        a_rx.boxed(),
        b_rx.join(c_rx).map(|(a, b)| a + b).boxed(),
    ]);

    let mut spawn = futures::executor::spawn(stream);
    a_tx.complete(33);
    b_tx.complete(33);
    assert!(spawn.poll_stream(support::unpark_noop()).unwrap().is_ready());
    c_tx.complete(33);
    assert!(spawn.poll_stream(support::unpark_noop()).unwrap().is_ready());
}
```

weiznich:
@alexcrichton Thanks for the explanation. I've added those test case and  fixed the code.

src/stream/futures_unordered.rs
```diff
@@ -0,0 +1,98 @@
...
+            .unwrap_or_else(|| self.stack.drain());
+        while let Some(id) = drain.next() {
+            if let Some(ret) = self.poll_item(id) {
+                self.pending = Some(drain);
+                return ret;
+            }
+        }
+        for id in self.stack.drain() {
+            if let Some(ret) = self.poll_item(id) {
+                return ret;
```
alexcrichton: Here we'll want to save off all remaining items in the `self.stack.drain()` iterator created above I believe.

src/stream/futures_unordered.rs
```diff
@@ -0,0 +1,103 @@
...
+/// This adaptor will return their results in the order that they complete.
+/// This is created by the `futures` method.
+///
+#[must_use = "streams do nothing unless polled"]
+pub struct FuturesUnordered<F>
+    where F: Future
+{
+    futures: Vec<Option<F>>,
+    stack: Arc<Stack<usize>>,
+    pending: [Option<Drain<usize>>; 2],
```
alexcrichton: I think there only needs to be one of these, right? Either both of these will be `None` or one will be.

src/stream/futures_unordered.rs
```diff
@@ -0,0 +1,103 @@
...
+        }
+        let mut drain = self.pending[1]
+            .take()
+            .unwrap_or_else(|| self.stack.drain());
+        while let Some(id) = drain.next() {
+            if let Some(ret) = self.poll_item(id) {
+                self.pending[1] = Some(drain);
+                return ret;
+            }
+        }
```
alexcrichton: I might recommend writing this as:

```rust
if let Some(s) = self.pending.take() {
    if let Some(ret) = self.poll_pending(s) {
        return ret
    }
}
if let Some(ret) = self.poll_pending(self.stack.drain()) {
    return ret
}

return Ok(Async::NotReady)

alexcrichton:
Awesome, thanks so much again @weiznich!

jimmycuadra:
Is there an equivalent to this where the items are yielded in the order of the original collection? e.g. `futures_ordered`?

alexcrichton:
@jimmycuadra just added in https://github.com/alexcrichton/futures-rs/pull/504!

PR #269, opened 2016-12-01
Revert "Simplify Sender::clone"

### LINKED ISSUE #267: Recent commit to mpsc clone fn introduces potential state hazards
I believe this commit should be reverted: https://github.com/alexcrichton/futures-rs/commit/dd3d102ebb9147f88542792a9ad4ae0b9aa4309b

The original implementation, while more complex, was intended to maintain correctness. The issues with the changes are two fold:

1) It enables overflow of max senders, which in turn could cause more senders than intended to exist.
2) It corrupts the state such that once senders are dropped, enabling the creation of additional senders, `clone` will still fail.

carllerche:
This reverts commit dd3d102ebb9147f88542792a9ad4ae0b9aa4309b.

The commit introduced correctness issues. While the case is unlikely to happen
under normal circumstances, there are ways of hitting the issues and the
previous code, while more verbose, handles the edge cases.

Fixes #267

### DIFF
--- a/src/sync/mpsc/mod.rs
+++ b/src/sync/mpsc/mod.rs
@@ -71,7 +71,7 @@ use std::any::Any;
 use std::error::Error;
 use std::fmt;
 use std::sync::atomic::AtomicUsize;
-use std::sync::atomic::Ordering::{SeqCst, Relaxed};
+use std::sync::atomic::Ordering::SeqCst;
 use std::sync::{Arc, Mutex};
 use std::thread;
 use std::usize;
@@ -510,7 +510,7 @@ impl<T> Clone for Sender<T> {
         // Since this atomic op isn't actually guarding any memory and we don't
         // care about any orderings besides the ordering on the single atomic
         // variable, a relaxed ordering is acceptable.
-        let mut curr = self.inner.num_senders.load(Relaxed);
+        let mut curr = self.inner.num_senders.load(SeqCst);
 
         loop {
             // If the maximum number of senders has been reached, then fail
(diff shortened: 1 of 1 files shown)


alexcrichton:
Can you leave these as `SeqCst`? We've been bitten too much before and I doubt these are perf critical at all.

carllerche:
The max sender atomic is used only as a counter and not to guard any memory or ordering. In other words, the memory access before and after the atomic op do not depend on on any mutations done in another thread and are not coordinated by the max sender atomic. This is the canonical use case for relaxed ordering. 

In other words, clone implies duplicating the state of the current handle, and all the necessary state is already visible to the current thread by virtue of having the handle. Any subsequent operations on the new handle (except for another clone which follows the previously described logic) goes through the state atomic and maintains consistency that way.

carllerche:
That being said, if you are concerned about the ordering of max senders (clone drop) with respect to sending values over the channel, the send operation (with seqcst, but had it been acq / rel it would have been equivalent) ensures correct ordering. 

A dec is ordered after the send and the send of None is ordered after the final dec. 

However, while I am fairly confident that it is correct, I'm not 100% sure (never am with atomics) and would love to discover any correctness flaws. I also care less about relaxed vs seqcst and was just applying a revert. If you want to switch to seqcst I'm fine with that (I still would appreciate having any flaws pointed out if you see them)

alexcrichton:
I mean I'm just in a default mode of "SeqCst unless proven otherwise". We haven't proven we need relaxed here, so in my mind there's no reason to use relaxed.

carllerche:
Updated

Issue #270, opened 2016-12-01
Add If/Else combinator


ghost:
I believe it could be useful to add an if/then/else combinator to return a single type implementing future from an if/then/else, even if the branches don't return the same type (as long as they return futures ultimately resolving to the same type)

Example:

``` rust
(if a {
    IfElse::If(futures::lazy(|| x))
} else {
    IfElse::Else(futures::finished(y))
}).and_then(|z| ...)
```

alexcrichton:
Agreed! I've thought that we should have some useful enums (Perhaps like 2-4 variants) for a situation like this.

Want to send a PR?

alexcrichton:
I believe this is now done with `Either`, so closing.

PR #271, opened 2016-12-01
Combinator to merge two futures yielding the same types

pijul:
Supposed to solve #270.

I chose the name `Branch` to be consistent with future similar types with more variants (if ever useful), like `Branch3`, `Branch4`, etc, instead of Haskell's `Either`.

### DIFF
--- a/src/future/branch.rs
+++ b/src/future/branch.rs
@@ -1,0 +1,23 @@
+use {Future, Poll};
+/// Combines two different futures yielding the same item and error
+/// types into a single type.
+pub enum Branch2<A, B> {
+    /// First branch of the type
+    A(A),
+    /// Second branch of the type
+    B(B),
+}
+
+impl<A, B, Item, Error> Future for Branch2<A, B>
+    where A: Future<Item = Item, Error = Error>,
+          B: Future<Item = Item, Error = Error>
+{
+    type Item = Item;
+    type Error = Error;
+    fn poll(&mut self) -> Poll<Item, Error> {
+        match *self {
+            Branch2::A(ref mut a) => a.poll(),
+            Branch2::B(ref mut b) => b.poll(),
+        }
+    }
+}
--- a/src/future/either.rs
+++ b/src/future/either.rs
@@ -1,14 +1,14 @@
 use {Future, Poll};
 /// Combines two different futures yielding the same item and error
 /// types into a single type.
-pub enum Branch2<A, B> {
+pub enum Either<A, B> {
     /// First branch of the type
     A(A),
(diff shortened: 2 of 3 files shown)


alexcrichton:
Thanks for the PR! Could this be a new module, `src/future/branch.rs`, with reexports in `src/future/mod.rs`? (Like the other combinators).

I also wonder if we should call this `Branch2` to have the naming be extensible into more variants later?

alexcrichton:
cc @tailhook you tend to have a lot of insights on naming these combinators and such, does this look ok?

Stebalien:
FYI, @bluss has already created an [either](https://crates.io/crates/either) crate for use cases like this so you might consider simply implementing `Future` on `Either`. Multiple branches can be achieved through recursion, albeit not quite as efficiently at the moment (better layout optimizations may fix that).

alexcrichton:
I'd prefer to keep the dependencies slim for now if possible, but that's certainly an option to keep open (and can be done in parallel if need be)

carllerche:
Personally I prefer the either name. I don't really think it has a hard implication of two variants.

pijul:
I initially called it `IfElse`.

I don't really mind, but I've tried not to use Ether because it's not very readable in Haskell, where it means both `Result` and "Branch combinator".

spinda:
@pijul Do you plan on updating this so it can be merged in?

pijul:
Update in which way?

alexcrichton:
Ah yeah I'd still like to discuss naming a bit here before landing (there's quite a few possibilities). @aturon do you have thoughts on the particular names used?

tailhook:
Well, the naming part is unclear here (this is actually why I haven't submitted a similar PR a while ago :) ).

Personally, I like current version:
```
if something {
  Branch2::A(something)
} else {
  Branch2::B(something_else)
}
``` 
I.e. it reads like `branch 1 from 2` (but backwards).

The `IfElse` and `Either` could be okay but I think more than two branches are at least as useful as join5 that we already have.

aturon:
My preference is for `Either`, and then later `Either3`, `Either4` etc if needed. Mostly because of the strong precedent and straightforward reading.

Another question is what to call the variants. I've also seen these called `Left` and `Right` or `First` and `Second`, but just using single letters as this PR does seems like an improvement to me!

alexcrichton:
Ok, @pijul want to rename this to `Either` but leave the names `A` and `B`?

pijul:
Sure!

alexcrichton:
Thanks!

PR #272, opened 2016-12-03
remove the `'static` bound on `Unpark`

nikomatsakis:
Because `Arc<Unpark>` defaults to `Arc<Unpark+'static>`, this bound was
useless. Moreover it hinders Rayon integration, since it means
we can't build a `Arc<Unpark+'foo>` that we transmute to a `Arc<Unpark>`
(since we manage the lifetimes).

### DIFF
--- a/src/task_impl/mod.rs
+++ b/src/task_impl/mod.rs
@@ -407,7 +407,7 @@ impl<T> Spawn<T> {
 /// `Spawn::poll_stream` functions. It's transitively used as part of the
 /// `Task::unpark` method to internally deliver notifications of readiness of a
 /// future to move forward.
-pub trait Unpark: Send + Sync + 'static {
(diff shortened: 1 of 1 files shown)


alexcrichton:
I think this is strictly speaking a breaking change because `T: Unpark` would previously imply `'static` and afterwards it wouldn't? I can't imagine anyone's doing that though so it's probably fine to land now?

nikomatsakis:
@alexcrichton yeah, I was wondering about that. I agree it is strictly speaking a breaking change. I also agree it is unlikely to affect actual users.

nikomatsakis:
(The library itself only uses `Unpark` as an object, from what I could tell.)

alexcrichton:
Yeah let's just do it

PR #275, opened 2016-12-04
Send as many items as possible before poll_complete

alexcrichton:
This changes SendAll::poll and Forward::poll to both attempt to send as many
items as possible before calling `poll_complete`. The `Sink` trait already
guarantees that if `start_send` returns `NotReady` that it performed the
equivalent of a `poll_complete`,  and otherwise we only need to call
`poll_complete` when there aren't any more items to pull out from the stream.

This change should allow for sinks which have some level of buffering to fill up
the buffer more with "ready items" if available, while also ensuring that if we
ever return `NotReady` from `poll` the sink has had the equivalent of
`poll_complete` called on it.

### DIFF
--- a/src/sink/send_all.rs
+++ b/src/sink/send_all.rs
@@ -71,17 +71,16 @@ impl<T, U> Future for SendAll<T, U>
         }
 
         loop {
-            try!(self.sink_mut().poll_complete());
-
-            if let Some(item) = try_ready!(self.stream_mut().poll()) {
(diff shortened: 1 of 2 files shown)


aturon:
LGTM.

aturon:
@alexcrichton r=me, are we seeing intermittent deadlocks in general on travis?

alexcrichton:
Bah yes I saw that somewhere else and meant to investigate, I will do so now.

PR #279, opened 2016-12-07
Add tail_fn, a combinator for tail-recursive loops

spinda:
I started writing some skeleton code for an initial proposal and ended up with a complete implementation. Hope it's okay if I submit this directly as a PR instead of asking in an issue first.

`tail_fn` takes an initial state and a loop function which turns that state into a future. The future returns a `Tail<T, S>` enum with either a final result value (`Tail::Done(T)`) or a new input state value to use to reinvoke the loop function (`Tail::Loop(S)`) and generate a new future. The loop function should keep returning `Tail::Loop` until it's time to break out by returning `Tail::Done`.

Some real code I want to write that's motivating this, from a work-in-progress D-Bus client:

```rust
use futures::future::{self, Future, Tail};
use libc;

use super::client::{Authenticator};
use super::types::{AuthError, ClientCommand, ServerCommand, ServerGuid};

pub fn auth_external(auth: Authenticator)
                     -> impl Future<Item = (Authenticator, ServerGuid), Error = AuthError> {
    let uid_str = unsafe { libc::getuid().to_string() };

    // Start by sending an AUTH EXTERNAL command to the server.
    future::tail_fn((auth,
                     ClientCommand::Auth {
                         mechanism: b"EXTERNAL"[..].into(),
                         initial_response: Some(uid_str.into_bytes().into()),
                     }),
                    |(auth, cmd)| {
        auth.send(cmd)
            .and_then(|auth| auth.receive())
            .and_then(|(auth, response)| {
                match response {
                    // If the server replies with OK, complete successfully.
                    ServerCommand::Ok { server_guid } => Ok(Tail::Done((auth, server_guid))),

                    // If the server replies with REJECTED, fail authentication.
                    ServerCommand::Rejected { supported_mechanisms } => {
                        Err(AuthError::Rejected { supported_mechanisms: supported_mechanisms })
                    }

                    // If the server replies with ERROR, it didn't understand our last command.
                    // Abort authentication by sending CANCEL and wait for the server to
                    // reply with REJECTED.
                    ServerCommand::Error => Ok(Tail::Loop((auth, ClientCommand::Cancel))),

                    // If we receive something unexpected, send back ERROR and pretend
                    // we never saw it, as per the D-Bus spec.
                    _ => Ok(Tail::Loop((auth, ClientCommand::Error(None)))),
                }
            })
    })
}
```

### DIFF
--- a/src/future/tail_fn.rs
+++ b/src/future/tail_fn.rs
@@ -1,0 +1,101 @@
+//! Definition of the `TailFn` combinator, implementing tail-recursive loops.
+
+use {Async, Future, IntoFuture, Poll};
+
+/// The status of a `tail_fn` loop.
+pub enum Tail<T, S> {
+    /// Indicates that the loop has completed with output `T`.
+    Done(T),
+
+    /// Indicates that the loop function should be called again with input
+    /// state `S`.
+    Loop(S),
+}
+
+/// A future implementing a tail-recursive loop.
+///
+/// Created by the `tail_fn` function.
+pub struct TailFn<A, F> {
+    future: A,
+    func: F,
+}
+
+/// Creates a new future implementing a tail-recursive loop.
+///
+/// The loop function is immediately called with `initial_state` and should
+/// return a value that can be converted to a future. On successful completion,
+/// this future should output a `Tail<T, S>` to indicate the status of the
+/// loop.
+///
+/// `Tail::Done(T)` halts the loop and completes the future with output `T`.
+///
+/// `Tail::Loop(S)` reinvokes the loop function with state `S`. The returned
+/// future will be subsequently polled for a new `Tail<T, S>` value.
+///
+/// # Examples
+///
+/// ```
+/// use futures::future::{ok, tail_fn, Future, Ok, Tail};
+/// use std::io::Error;
+///
+/// struct Client {
+///     ping_count: u8,
+/// }
+///
+/// impl Client {
+///     fn new() -> Self {
+///         Client { ping_count: 0 }
+///     }
+///
+///     fn send_ping(self) -> Ok<Self, Error> {
+///         ok(Client { ping_count: self.ping_count + 1 })
+///     }
+///
+///     fn receive_pong(self) -> Ok<(Self, bool), Error> {
+///         let done = self.ping_count >= 5;
+///         ok((self, done))
+///     }
+/// }
+///
+/// let ping_til_done = tail_fn(Client::new(), |client| {
+///     client.send_ping()
+///         .and_then(|client| client.receive_pong())
+///         .and_then(|(client, done)| {
+///             if done {
+///                 Ok(Tail::Done(client))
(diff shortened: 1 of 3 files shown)


spinda:
Some alternative naming schemes:

- `tail` / `Tail` and `Loop::Done`, `Loop::Tail`
- `tail` / `Tail` and `Loop::Done`, `Loop::Tail`
- `tail_loop` / `TailLoop` and `Tail::Done`, `Tail::Loop`
- `tail_loop` / `TailLoop` and `Loop::Done`, `Loop::Tail`
- `loop_tail` / `LoopTail` and `Tail::Done`, `Tail::Loop`
- `loop_tail` / `LoopTail` and `Loop::Done`, `Loop::Tail`

carllerche:
Looks good, quick question. Is there prior art that you are drawing the design from? I've been meaning to look more at clojure which I believe has a similar problem. I've seen some fns like trampoline and I wonder how that relates / applies to this case. Do you have any thoughts?

src/future/tail_fn.rs
```diff
@@ -0,0 +1,101 @@
...
+{
+    type Item = T;
+    type Error = A::Error;
+
+    fn poll(&mut self) -> Poll<Self::Item, Self::Error> {
+        let status = try_ready!(self.future.poll());
+        match status {
+            Tail::Done(x) => Ok(Async::Ready(x)),
+            Tail::Loop(s) => {
+                self.future = (self.func)(s).into_future();
```
alexcrichton: I think this `match` will want to be executed in a loop because once this future is created and `NotReady` is returned we're not actually scheduled to receive a wakeup I believe.

alexcrichton:
@spinda I wonder in terms of naming if we could call this `loop_fn` and call the enum `Loop::Done` and `Loop::Continue`?

alexcrichton:
Er maybe even `Loop::Break` instead of `Loop::Done` to look similar to `break`/`continue`

spinda:
I like `loop_fn` / `Loop::{Break, Done}`. Will make those changes as well as add the missing `loop` to the implementation.

src/future/tail_fn.rs
spinda: D'oh, thanks!

46bit: Would looping here cause each `LoopFn` to block the thread until complete? I've been wondering if it would be better to have this in `tokio-core` where it could take a `&Handle` and schedule itself to be polled?

spinda: Only if the future continuously returns `Ok(Async::Ready(Loop::Continue(...)))`, meaning it's only executing blocking computations anyway. The first `Ok(Async::NotReady)` returned would exit the loop and cede control.

spinda:
Sorry for the delay. I've made those changes and brought this patch up to date again.

alexcrichton:
Thanks!

PR #280, opened 2016-12-07
Tiny refactoring of poll::Async

0x7CFE:
This is a small refactoring that may reduce the code complexity by a tiny bit.

Typically, expressions are easier to understand when negations are reduced to a bare minimum. This PR moves the actual logic to `Async::is_ready()`, whereas  `is_not_ready()` is now a simple negation of `is_ready()`. The same goes for `AsyncSink`.

Also, match arms in `Async::map()` method are reversed, so that the most important arm goes first.

I believe this is more natural order of the `Async`'s methods and their logic. Hope this helps :)

### DIFF
--- a/src/poll.rs
+++ b/src/poll.rs
@@ -40,17 +40,17 @@ impl<T> Async<T> {
         }
     }
 
-    /// Returns whether this is `Async::NotReady`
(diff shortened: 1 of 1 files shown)


alexcrichton:
Thanks!

PR #284, opened 2016-12-10
Adding the from_err() combinator

pijul:
Just like `map_err`, but if my code is big enough and I have already defined a `From` instance for some common errors (which I would do for `try!`), this new combinator lets the compiler infer error types.

### DIFF
--- a/src/future/from_err.rs
+++ b/src/future/from_err.rs
@@ -1,0 +1,32 @@
+use {Future, Poll, Async};
+use std::marker::PhantomData;
+/// Future for the `map_err` combinator, changing the error type of a future.
+///
+/// This is created by this `Future::map_err` method.
+#[must_use = "futures do nothing unless polled"]
+pub struct FromErr<A, E> where A: Future {
+    future: A,
+    f: PhantomData<E>
+}
+
+pub fn new<A, E>(future: A) -> FromErr<A, E>
+    where A: Future
+{
+    FromErr {
+        future: future,
+        f: PhantomData
+    }
+}
+
+impl<A:Future, E:From<A::Error>> Future for FromErr<A, E> {
+    type Item = A::Item;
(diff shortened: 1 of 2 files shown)


alexcrichton:
Thanks for the PR! I think the build failures can be alleviated with an import from `core` instead of `std`, but otherwise looks good to me.

@carllerche @aturon any thoughts? Naming-wise I'm not entirely sure but seems fine as a stand-in regardless.

aturon:
Ooh, nice! It'd be even better if we could build in these conversions to the other combinators, but I suspect that would create inference problems -- maybe worth trying at some point?

In any case, I'm happy to land this for now, and I think the name is reasonable.

alexcrichton:
Alright let's merge.

dwrensha:
Yay! This is going to save me a lot of calls of the form `.map_err(|e| e.into())`.

I'm not sold on the name yet. To me, it seems to get the direction backwards. As an analogy: I often call `v.into()`, but `v.from()` doesn't make sense. So I feel like this combinator would make more sense as `.into_err()`. Other possibilities that I would be happier with are`.inject_err()` and `lift()`.

aturon:
I think `.into_err()` is more confusing, since we *do* have conversion methods following this pattern, but they involve converting the entire object into some other type.

`lift_err` seems like a reasonable alternative.

Meanwhile, I'm going to open up an issue for 0.2 to explore whether we can bake in error conversion more broadly, or whether that causes inference problems.

alexcrichton:
Hm yeah in retrospect I like the name `lift_err` as well, @dwrensha do you want to send a PR?

@aturon thanks for opening an issue, I also intended to do so after this!

joepie91:
What happened to this combinator? I can't find it anywhere in the documentation as either `from_err` of `lift_err`.

PR #286, opened 2016-12-11
Forget about the iterator type after constructing a `JoinAll`.

### LINKED ISSUE #285: join_all() is too picky about lifetimes
Consider the following function:

```rust
fn foo<'a>(bufs: Vec<&'a [u8]>) -> Box<::futures::Future<Item=Vec<usize>, Error=()> + 'static> {
    let iter = bufs.into_iter().map(|b| ::futures::future::ok::<usize, ()>(b.len()));
    Box::new(::futures::future::join_all(iter))
}
```

I expect this to typecheck successfully, but I get an error:

```
error[E0495]: cannot infer an appropriate lifetime due to conflicting requirements
  --> src/main.rs:35:21
   |
35 |     let iter = bufs.into_iter().map(|b| ::futures::future::ok::<usize, ()>(b.len()));
   |                     ^^^^^^^^^
   |
   = note: first, the lifetime cannot outlive the lifetime 'a as defined on unknown free region bounded by scope CodeExtent(212/CallSiteScope { fn_id: NodeId(122), body_id: NodeId(276) })...
note: ...so that types are compatible (expected std::vec::Vec<&[u8]>, found std::vec::Vec<&'a [u8]>)
  --> src/main.rs:35:21
   |
35 |     let iter = bufs.into_iter().map(|b| ::futures::future::ok::<usize, ()>(b.len()));
   |                     ^^^^^^^^^
   = note: but, the lifetime must be valid for the static lifetime...
note: ...so that the type `futures::JoinAll<std::iter::Map<std::vec::IntoIter<&[u8]>, [closure@src/main.rs:35:37: 35:84]>>` will meet its required lifetime bounds
  --> src/main.rs:36:5
   |
36 |     Box::new(::futures::future::join_all(iter))
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

```

The problem seems to be that `JoinAll<I>` is parameterized by the type of the `IntoIterator` that is passed into `join_all()`. In the above case, that type is not `'static`, even though its `Item` is `'static`.

dwrensha:
Fixes #285.

### DIFF
--- a/src/future/join_all.rs
+++ b/src/future/join_all.rs
@@ -17,11 +17,10 @@ enum ElemState<T> where T: Future {
 ///
 /// This future is created with the `join_all` method.
 #[must_use = "futures do nothing unless polled"]
-pub struct JoinAll<I>
-    where I: IntoIterator,
-          I::Item: IntoFuture,
+pub struct JoinAll<Item>
+    where Item: IntoFuture,
 {
-    elems: Vec<ElemState<<I::Item as IntoFuture>::Future>>,
+    elems: Vec<ElemState<<Item as IntoFuture>::Future>>,
 }
 
 /// Creates a future which represents a collection of the results of the futures
@@ -60,7 +59,7 @@ pub struct JoinAll<I>
 ///     x
 /// });
 /// ```
-pub fn join_all<I>(i: I) -> JoinAll<I>
+pub fn join_all<I>(i: I) -> JoinAll<I::Item>
     where I: IntoIterator,
           I::Item: IntoFuture,
 {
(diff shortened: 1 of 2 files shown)


dwrensha:
The current type parameter situation is left over from before #255, when `JoinAll` did in fact keep the iterator around as one of its fields.

alexcrichton:
This seems reasonable to me, and it seems like we probably won't be storing the iterator in the `JoinAll` future any time soon as well, so seems fine moving forward. Could you add a test for this as well?

Unfortunately this is a breaking change so this can't land until 0.2, but I don't mind keeping this around until that time.

dwrensha:
Test added.

alexcrichton:
Awesome, thanks! I've tagged this to ensure it makes it into 0.2 and we don't forget about it.

stuhood:
@alexcrichton : It looks like `0.2` might be ramping up. Would it be a good time to tackle this in order to help with #662?

alexcrichton:
Certainly! I'd recommend coordinating with @cramertj though!

cramertj:
@stuhood Can you send a PR to 0.2 in the next hour or so? If not, I'd recommend waiting until later this afternoon once I've had a chance to reorg 0.2 into multiple crates, otherwise the rebase will be a pain.

stuhood:
@cramertj : Should be able to, as this PR is already green.

cramertj:
@stuhood Great! ty :)

stuhood:
@cramertj, @alexcrichton : Opened #732.

alexcrichton:
Landed in https://github.com/rust-lang-nursery/futures-rs/pull/732

PR #287, opened 2016-12-11
Fix crates.io shield link

casey:
The URL wasn't set. Fixed it to point at https://crates.io/crates/futures


alexcrichton:
Thanks!

PR #288, opened 2016-12-11
Unify `future::Ok`, `future::Err`, and `feature::FutureResult`

sgrif:
For users who are writing code that conditionally uses these
combinators, it is more convenient for them to return the same type,
without boxing or forcing a more verbose invocation. For example:

```
let x = match optional_result {
    Some(result) => future::result(result),
    None => future::err("Expected a value"),
}
```

Prior to this patch, the futures would need to be boxed, or the second
branch would need to be changed to `future::result(Err("Expected a
value"))`. Now this code compiles as expected.


alexcrichton:
This seems like a reasonable feature to me, and the example here seems well motivated. A downside of this is that the in-memory representation is now inflated by the discriminant of `Result`, but that seems like a totally negligible "downside".

Looks like there's some CI failures though?

sgrif:
CI is green

alexcrichton:
Thanks!

PR #294, opened 2016-12-16
Make `after_start` and `before_stop` take an index

Hoverbear:
Since these functions are primarily intended for bookkeeping and
monitoring it seems like they should have some knowledge of which thread
they are.

Means you can do something like:

```rust
let pool = PoolBuilder::new()
        .pool_size(4)
        .after_start(|num| println!("THREAD {} START", num))
        .before_stop(|num| println!("THREAD {} STOP", num))
        .create();
```

This is a **breaking change**.

### DIFF
--- a/futures-cpupool/src/lib.rs
+++ b/futures-cpupool/src/lib.rs
@@ -82,8 +82,8 @@ pub struct CpuPool {
 pub struct Builder {
     pool_size: usize,
     name_prefix: Option<String>,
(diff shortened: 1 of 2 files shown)


alexcrichton:
Looks good to me! Unfortunately this is a breaking change so I'm not going to merge just yet. I'll categorize it for 0.2, though, and we'll be sure to merge it then.

alexcrichton:
Ok merged into 0.2!

Hoverbear:
🎉  Yay!

PR #295, opened 2016-12-16
Add CpuFuture::forget() function.

### LINKED ISSUE #293: fire and forget style futures_cpupool::CpuPool::spawn() api?
Currently API users have to keep the returned CpuFuture to keep the computation alive, but can we also have API that doesn't require this management?

kkimdev:
CpuFuture::forget() consumes the future but does not cancel the computation.

Closes #293

### DIFF
--- a/futures-cpupool/src/lib.rs
+++ b/futures-cpupool/src/lib.rs
@@ -44,7 +44,7 @@ extern crate num_cpus;
 
 use std::panic::{self, AssertUnwindSafe};
 use std::sync::Arc;
-use std::sync::atomic::{AtomicUsize, Ordering};
+use std::sync::atomic::{AtomicBool, AtomicUsize, Ordering};
 use std::thread;
 
 use crossbeam::sync::MsQueue;
@@ -89,6 +89,7 @@ pub struct Builder {
 struct MySender<F, T> {
     fut: F,
     tx: Option<Sender<T>>,
+    keep_running_flag: Arc<AtomicBool>,
 }
 
 fn _assert() {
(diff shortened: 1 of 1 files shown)


futures-cpupool/src/lib.rs
```diff
@@ -258,8 +269,10 @@ impl<F: Future> Future for MySender<F, Result<F::Item, F::Error>> {
 
     fn poll(&mut self) -> Poll<(), ()> {
         if let Ok(Async::Ready(_)) = self.tx.as_mut().unwrap().poll_cancel() {
-            // Cancelled, bail out
-            return Ok(().into())
+            if !self.keep_running_flag.load(Ordering::SeqCst) {
```
alexcrichton: I think it's ok to switch the ordering of these two conditions, check `keep_running_flag` and then check `poll_cancel`. That way we can avoid messing with that piece if the flag is set.

alexcrichton:
Looks great to me, thanks!

futures-cpupool/src/lib.rs
kkimdev: Thank!, my concern was the following case:
1. `MySender` checks `keep_running_flag`
2. `CpuFuture::forget()` called that sets `keep_running_flag = true` and drops `CpuFuture`
3. `MySender` checks `self.tx.as_mut().unwrap().poll_cancel()` and cancels the computation.

It's unlikely, but I guess it's still possible so..

kkimdev: Done!

alexcrichton:
Thanks!

PR #305, opened 2016-12-25
Fix a race in Shared where it blocks forever

### LINKED ISSUE #304: future::Shared can hang if a clone gets dropped
Consider the following program:

```rust 
    use futures::Future;
    use futures::sync::oneshot;

    let mut core = ::tokio_core::reactor::Core::new().unwrap();
    let handle = core.handle();

    let (tx, rx) = oneshot::channel::<u32>();
    let f1 = rx.shared();
    let f2 = f1.clone();

    let (tx2, rx2) = oneshot::channel::<u32>();

    handle.spawn(
        f1.map_err(|_| ()).map(|x| *x).select(rx2.map_err(|_| ()))
            .and_then(|x| { println!("first task got {}", x.0); Ok(()) }).map_err(|_| ()));

    let (tx3, rx3) = oneshot::channel::<u32>();

    handle.spawn(f2.map(|x| tx3.complete(*x)).map_err(|_| ()));

    tx2.complete(11); // Causes `f1` to get cancelled.
    core.turn(Some(::std::time::Duration::from_millis(1)));

    tx.complete(42); // Should cause `f2` and then `rx3` to get resolved.
    let result = core.run(rx3).unwrap();
    assert_eq!(result, 42);
```

I expect this to terminate successfully. Instead, it hangs at the `core.run()`. The problem, as I understand it, is that once `f1` is cancelled, `tx.complete(42)` has no way of waking up the proper task. One way to solve this might be to add some `unpark()` logic in the destructor of `Shared<F>`.

alexcrichton:
A `Shared` represents multiple handles to an original future, which can have
many tasks waiting on it. The original future, however, can only have at most
one task waiting on it. If a `Shared` handle is dropped which belongs to the
task that's blocking on the inner future, then once the inner future is done
we're not guaranteed to ever be able to wake up again.

To solve this case for now this commit adds a destructor to the `Shared` future
which wakes up all waiting tasks on destruction. This has the downside of a
"thundering stampede" problem but we can't be certain which task should be woken
up, so we just wake up all of them.

Closes #304

### DIFF
--- a/tests/shared.rs
+++ b/tests/shared.rs
@@ -1,9 +1,9 @@
 extern crate futures;
 
 use std::thread;
+
 use futures::sync::oneshot;
 use futures::Future;
-
 
 fn send_shared_oneshot_and_wait_on_multiple_threads(threads_number: u32) {
     let (tx, rx) = oneshot::channel::<u32>();
@@ -37,4 +37,32 @@ fn two_threads() {
 #[test]
 fn many_threads() {
     send_shared_oneshot_and_wait_on_multiple_threads(1000);
+}
+
+#[test]
+fn drop_on_one_task_ok() {
+    let (tx, rx) = oneshot::channel::<u32>();
+    let f1 = rx.shared();
+    let f2 = f1.clone();
+
+    let (tx2, rx2) = oneshot::channel::<u32>();
+
+    let t1 = thread::spawn(|| {
+        let f = f1.map_err(|_| ()).map(|x| *x).select(rx2.map_err(|_| ()));
+        drop(f.wait());
+    });
+
+    let (tx3, rx3) = oneshot::channel::<u32>();
+
+    let t2 = thread::spawn(|| {
+        drop(f2.map(|x| tx3.complete(*x)).map_err(|_| ()).wait());
+    });
+
+    tx2.complete(11); // cancel `f1`
+    t1.join().unwrap();
+
+    tx.complete(42); // Should cause `f2` and then `rx3` to get resolved.
+    let result = rx3.wait().unwrap();
+    assert_eq!(result, 42);
+    t2.join().unwrap();
 }
--- a/src/future/shared.rs
+++ b/src/future/shared.rs
@@ -99,7 +99,8 @@ impl<F> Future for Shared<F>
         //    - If the future is not ready, or if the lock failed:
         // 3. Lock the state for write.
         // 4. If the state is `State::Done`, return the result. Otherwise:
-        // 5. Create a task, push it to the waiters vector, and return `Ok(Async::NotReady)`.
+        // 5. Create a task, push it to the waiters vector, and return
+        //    `Ok(Async::NotReady)`.
 
         if !self.inner.result_ready.load(SeqCst) {
             match self.inner.original_future.try_lock() {
@@ -132,7 +133,9 @@ impl<F> Future for Shared<F>
 
                         match mem::replace(&mut *state, State::Done(result)) {
                             State::Waiting(waiters) => waiters,
-                            State::Done(_) => panic!("store_result() was called twice"),
+                            State::Done(_) => {
+                                panic!("store_result() was called twice")
+                            }
                         }
                     };
                     for task in waiters {
@@ -164,6 +167,34 @@ impl<F> Clone for Shared<F>
 {
     fn clone(&self) -> Self {
         Shared { inner: self.inner.clone() }
+    }
+}
+
+impl<F: Future> Drop for Shared<F> {
+    fn drop(&mut self) {
+        // A `Shared` represents a bunch of handles to one original future
+        // running on perhaps a bunch of different tasks.  That one future,
+        // however, is only guaranteed to have at most one task blocked on it.
+        //
+        // If our `Shared` handle is the one which has the task blocked on the
+        // original future, then us being dropped means that we won't ever be
+        // around to wake it up again, but all the other `Shared` handles may
+        // still be interested in the value of the original future!
+        //
+        // To handle this case we implement a destructor which will unpark all
+        // other waiting tasks whenever we're dropped. This should go through
+        // and wake up any interested handles, and at least one of them should
+        // make its way to blocking on the original future itself.
+        if self.inner.result_ready.load(SeqCst) {
+            return
+        }
+        let waiters = match *self.inner.state.write().unwrap() {
+            State::Waiting(ref mut waiters) => mem::replace(waiters, Vec::new()),
+            State::Done(_) => return,
+        };
+        for waiter in waiters {
+            waiter.unpark();
+        }
     }
 }
 


tests/shared.rs
```diff
@@ -38,3 +38,31 @@ fn two_threads() {
 fn many_threads() {
     send_shared_oneshot_and_wait_on_multiple_threads(1000);
 }
+
+#[test]
+fn drop_on_one_task_ok() {
```
dwrensha: This test case passes for me even *before* applying the rest of this commit, so I doubt that it is providing the coverage that we want.

dwrensha:
I suppose this could be optimized by giving each clone a unique ID and keeping track of which one was responsible for the most recent call to the inner `.poll()`.

dwrensha:
Hm. This destructor helps, but I'm still a bit worried about the fact that everything can grind to a halt if one of the clones leaks and therefore never gets polled again and never gets dropped.

I wonder if it would be better it modify the design of `shared()` so that it can be predictably driven by a single task. For example, maybe the initial `Shared` could drive the execution and would be the only clone that is allowed to poll the underlying future. We would probably want to avoid `Clone` is this case, to emphasize that there is a semantic difference between the copies.

tests/shared.rs
alexcrichton: I think it's somewhat nondeterministic as we can't be sure that the threads are blocking at a particular point. I had to run this a few times to get it to deadlock beforehand.

alexcrichton:
@dwrensha yeah I do think that a unique ID would help here in not waking up too many, and it's true yeah that leaks will cause everything to grind to a halt. I'm not *too* worried about leaks, though, and it may mean that we should just provide a `oneshot`-like future were the `Receiver` implements `Clone`. That way we don't have to talk about spawns in futures itself just yet (which doesn't understand spawning like tokio-core does).

dwrensha:
To be more concrete, my proposal is that we split `Shared<F>` into something like `SharedRoot<F>` and `SharedBranch<F>`, where:

  1.  `Future::shared()` returns a `SharedRoot<F>`.
  2.  Both `SharedRoot<F>` and `SharedBranch<F>` implement `Future`.
  3. `SharedRoot<F>::poll()` polls the underlying future.
  4. `SharedRoot<F>::add_branch()` creates a new `SharedBranch<F>` from a `SharedRoot<F>`.
  5. Neither `SharedRoot<F>` nor `SharedBranch<F>` implements `Clone`.
  6. `SharedBranch<F>::poll()` does *not* poll the underlying future. Instead, it checks whether a value is currently ready and if not parks the current task.
  7. Dropping a `SharedRoot<F>` cancels the underlying future.
  8. Dropping a `SharedBranch<F>` does not cancel the underlying future.

This would allow one to pass a `SharedBranch<F>` to untrusted code without worrying about leaks, and would avoid the need for fancy bookkeeping.

tests/shared.rs
dwrensha: Ah. Indeed, if I run it a few times I do see the deadlock.

liran-ringel:
@dwrensha a similar idea was presented here:
https://github.com/alexcrichton/futures-rs/compare/master...liran-ringel:feature/multishot

The disadvantage is that only the root can generate `SharedBranch<F>`, while current `Shared` can be cloned from cloned `Shared`s.

dwrensha:
> The disadvantage is that only the root can generate `SharedBranch<F>`

No, a `SharedBranch` would be able to generate other `SharedBranch`s. I imagine both types having the same representation as the current `Shared<F>`, that is:

```rust
pub struct SharedRoot<F> where F: Future
{
    inner: Arc<Inner<F>>,
}

pub struct SharedBranch<F> where F: Future
{
    inner: Arc<Inner<F>>,
}
```

So it should should be easy to have `impl <F> Clone for SharedBranch<F> { ... }`.

(The first version of my previous comment proposed that neither new type should implement `Clone`; I've since updated the comment.)

@liran-ringel the multishot proposal is interesting. One difference from my present proposal is that `<Multishot as Future>::Item = ()` and `<Multishot as Future>::Error = ()`, while `<SharedRoot<F> as Future>::Item = SharedItem<F::Item>` and `<SharedRoot<F> as Future>::Error = SharedError<F::Error>`.

liran-ringel:
@dwrensha if `Shared` drop could be implemented more efficiently, O(1), will `SharedRoot` and `SharedBranch` have a value?

alexcrichton:
@dwrensha an interesting idea! That'd definitely make destruction more deterministic in all cases, especially in terms of time complexity.

I think, though, that with this PR the implementation is correct in the face of early drops (cancellations). The downside of this PR, however, is that it's probably not up to par with the perf of the implementation that you're proposing. 

You mentioned that leaks can cause this to go wrong, but don't leaks make everything go wrong in a futures system? Even with a root/branch distinction if you leak a root the whole system grinds to a halt?

liran-ringel:
@alexcrichton @dwrensha I propose that instead of `SharedRoot` and `SharedBranch` we do that:
https://github.com/alexcrichton/futures-rs/compare/master...liran-ringel:bugfix/shared-blocks-if-poller-is-dropped?expand=1

The `drop` implementation is O(1).
Do you still see any advantage to the `SharedRoot` and `SharedBranch` concept? I think it adds some complexity.
Also, the `SharedBranch`es are canceled if the `SharedRoot` is dropped (which does not happen in current `Shared`).

dwrensha:
Unparking only one task makes me nervous. What if that task does not poll the shared future this time around?

liran-ringel:
@dwrensha If the task doesn't poll the shared future this time it means one of the following:
1. `result_ready` is true
2. Other `Shared` is making progress on the future - in that case he becomes the "poller".

dwrensha:
Consider the following future:

```rust
enum Which { Left, Right }

struct ModedFuture<F> where F: Future {
    mode: Mode,
    left: F;
    right: F;
}

impl <F> ModedFuture<F> where F: Future {
    pub fn new(left: F, right: F, mode: Mode) -> ModedFuture<F> {
        ModedFuture { left: left, right: right, mode: mode }
    }

    pub fn set_mode(&mut self, mode: Mode) {
        self.mode = mode;
    }
}

impl <F> Future for ModedFuture<F> where F: Future {
    type Item = F::Item;
    type Error = F::Error;
        match self.mode {
            Mode::Left => self.left.poll(),
            Mod::Right => self.right.poll(),
        }
    }
}
```

Now suppose that I wrap a `Shared` as part of a `ModedFuture`, allow it to get polled a few times, and then change the mode of the `ModedFuture`. If this is the single task that gets unparked when a clone of the `Shared` gets dropped, the `Shared` could get stuck; there might be other clones that are waiting for it to complete, but they will never get unparked because the designated poller (i.e. the task with the `ModedFuture`) has decided not to poll the `Shared`.

So if we want to have `Shared::drop()` only unpark a single task, then I think that we need to declare that `ModedFuture` is somehow an invalid impl of `Future`. I suppose we could say that a future *must* poll all futures that it owns/wraps. I worry that this is a rather fuzzy property that would be difficult to formalize and document. Moreover, although I don't currently have any need for something like `ModedFuture`, it seems kind of sad to rule to it out, as it might be useful.

liran-ringel:
@dwrensha Good point, it will indeed block until the other future is finished, and `ModedFuture` will be destroyed.
But that situation could happen even if no `Shared` is dropped:
One `Shared` (called f1) polled the original future, and another `Shared` (f2) didn't. If f1 is in something like `ModedFuture`, f2 will hang until `ModedFuture` is dropped.
:(

I think we cannot say that a future must poll all futures that it owns/wraps, can we @alexcrichton?

So I guess we need to replace it with something like `Multishot`/`SharedRoot`&`SharedBranch`?

dwrensha:
(Note: I've updated my above implementation `ModedFuture` to unpark itself when its mode has changed.)

> One `Shared` (called f1) polled the original future, and another `Shared` (f2) didn't. If f1 is in something like `ModedFuture`, f2 will hang until `ModedFuture` is dropped.

I'm not sure that I follow. Why would `f2` not poll the original future? Does `f2` poll the `Shared` at all? If not, then presumably it doesn't actually care about the result of the `Shared`.

liran-ringel:
@dwrensha `f2` won't poll the original future if `f1` locked it here:
https://github.com/alexcrichton/futures-rs/blob/6468b266c674d71d4209d985312522b57f7a0776/src/future/shared.rs#L106

dwrensha:
In that case `f2` parks its task and will get re-polled when the lock is released, right? So therefore `f2` will poll the original future soon. As far as I understand, something like `ModedFuture` can't cause the lock to be held forever.

Hm. This suggests that it might be worth thinking about whether something like `futures::sync::Bilock` could have the potential to get stuck in a locked state if it were used with `ModedFuture`. If so, then maybe we are forced to accept that "futures must poll wrapped futures".

liran-ringel:
@dwrensha `f2` won't get re-polled when the lock is released. `f2` will get re-polled only when the original future will be polled as ready. Look here:
https://github.com/alexcrichton/futures-rs/blob/master/src/future/shared.rs#L117

If `f1` polled the original future and `f2` didn't (because it failed to lock the original future), but `f1` won't poll the original future again when it's ready (e.g. if he's in `ModedFuture`), `f2` will hang.


I wonder if we could use `futures::executor::spawn`, to create something like `SharedRoot` (not exposed as `pub`, only for implementation), that will unpark all `Shared`s tasks. That `SharedRoot`, if I understand it correctly, will poll the original future as fast as the `Executor` runs it, and won't have the discussed issues.

alexcrichton:
In general there's no guarantee that if you unpark a task that the task will go back and poll the relevant future. In that sense I may have been incorrect when I mentioned that we could optimize this in the future to not unpark all waiters, we may be forced to always unpark all waiters.

This point is solved by @dwrensha's proposed split, though, as it's clear who's responsible for making progress on the original future.

liran-ringel:
I wonder about the Item and Error types of `SharedBranch`.
For `SharedRoot`, it's as @dwrensha described:
`<SharedRoot<F> as Future>::Item = SharedItem<F::Item>` and `<SharedRoot<F> as Future>::Error = SharedError<F::Error>`.

But what will it be for `SharedBranch`?
`<SharedBranch<F> as Future>::Item = Result<SharedItem<F::Item>, SharedError<F::Error>>` and `<SharedBranch<F> as Future>::Error = Canceled`, or:
`<SharedBranch<F> as Future>::Item = SharedItem<F::Item>` and `SharedBranch<F> as Future>::Error = ErrorEnum<F::Error>`, where `ErrorEnum` can be either `SharedError<F::Error>` or `Canceled`.

alexcrichton:
Yes it's true that there's some type trickery with a root/branch approach as well. Going down that road is somewhat orthogonal to fixing the problem at hand, though. I'm going to go ahead and merge to we can get the fix out, but we can continue discussing alternate strategies for 0.2!

PR #314, opened 2016-12-29
Use single thread in `one_thread` test

Byron:
Previously two threads were used, which makes the test
equivalent to the `two_threads` test.


alexcrichton:
Ok!

PR #321, opened 2017-01-01
Generalize `Stream::for_each` using `IntoFuture`

aturon:
Makes the `for_each` combinator significantly more useful. This is
technically a breaking change due to adding a type parameter to
`ForEach` and changing the method signature. However, both are very
unlikely to affect code in practice.

### DIFF
--- a/src/stream/for_each.rs
+++ b/src/stream/for_each.rs
@@ -1,4 +1,4 @@
-use {Async, Future, Poll};
+use {Async, Future, IntoFuture, Poll};
 use stream::Stream;
 
 /// A stream combinator which executes a unit closure over each item on a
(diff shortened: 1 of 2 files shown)


aturon:
@alexcrichton do you feel comfortable landing this in the 0.1 series? It will help on the tutorial front.

alexcrichton:
Looks good to me, I did an initial run of `cargo-crusader` and it didn't turn up any real breakage. Due to the closure I doubt anyone is doing much naming a `ForEach` type specifically, and I think inference still works out, so I think we can get away with this.

Some other PRs though have shown that we may not want to put `F: IntoFuture` in type parameter lists on structs but rather `F::Future` directly. It has implications on variance I think mainly.

Would that work here to do that?

alexcrichton:
Oh I should also mention that I did some initial investigation and found no breakage here, so in that respect we're good to go here.

aturon:
I don't think we can avoid `IntoFuture` here, since we need to do the conversion fresh every time the closure is invoked.

PR #340, opened 2017-01-11
Resolve a race in the implementation of Shared

### LINKED ISSUE #339: Race in Shared
There's a race when:

* Task A task polls the future, the future isn't ready, the task unlocks the future
* Task B polls the future, it still isn't ready
* Task B's `Shared` handle is dropped, waking up no other tasks
* The future is completed, no tasks are waiting on it
* Task A finishes blocks on the `Shared` handle

This means that task A is wedged forever when the future is completed. Not entirely sure how to solve this yet.

cc @dwrensha @liranringel

alexcrichton:
Let's just go to a plain old `Mutex` instead of trying to be fancy with an
opportunistic lock. It weeds out more bugs and we can deal with perf
improvements later if a correct implementation comes along.

Closes #339

### DIFF
--- a/src/future/shared.rs
+++ b/src/future/shared.rs
@@ -15,41 +15,22 @@
 
 use std::mem;
 use std::vec::Vec;
-use std::sync::{Arc, RwLock};
-use std::sync::atomic::AtomicBool;
-use std::sync::atomic::Ordering::SeqCst;
+use std::sync::{Arc, Mutex};
 use std::ops::Deref;
 
 use {Future, Poll, Async};
 use task::{self, Task};
-use lock::Lock;
 
 /// A future that is cloneable and can be polled in multiple threads.
 /// Use Future::shared() method to convert any future into a `Shared` future.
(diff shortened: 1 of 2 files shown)


dwrensha:
Is there a possible deadlock if a `poll()` causes a clone of the `Shared` on the same thread to be dropped? The `poll()` is already locking the `inner`, so the `drop()` will block forever, I think. So I believe `original_future` needs to stay in its own field, and not be folded into `State`.

dwrensha:
(There is a [similar potential pitfall](https://github.com/dwrensha/capnp-rpc-rust/blob/cc37fafef3a985d0b6db7a7efcd797f8b35cd795/src/forked_promise.rs#L71-L75) in my `ForkedPromise` implementation.)

alexcrichton:
@dwrensha good point! I think we can solve that with a `try_lock` in the destructor ignoring any failed attempt, and I've pushed that up with a test as well to ensure we don't regress that.

PR #346, opened 2017-01-12
Update documentation for flatten

Idolf:


alexcrichton:
Thanks! Perhaps this could be changed to say "roughly equivalent"? I like the brevity of just using `then`, and I feel like being 100% correct is why the combinator exists and we perhaps don't need to spell it out in the docs.

Idolf:
Perhaps I am misunderstanding something, but `self.then(|x| x)` does not typecheck in my head, nor can I get a variation of it to typecheck in my test program.

Perhaps you meant "roughly equivalent to `self.and_then(|x| x)`? I actually considered writing something along those lines initially.

alexcrichton:
Oh hm yes I think you're right! I do think this is more equivalent to `and_then(|x| x)`

Idolf:
PR updated!

alexcrichton:
Thanks!

PR #347, opened 2017-01-12
Export SharedItem and SharedError

Idolf:
The documentation for `Shared` was lacking, since its `Item` and `Error` were
not exported.


alexcrichton:
Oops, thanks!

PR #349, opened 2017-01-12
Replaced `list` with `iter`, since that is the function arguments' name

Idolf:
- Replaced `list` with `iter`, since that is the function arguments' name
- Put a function name in backticks.


alexcrichton:
Thanks!

PR #351, opened 2017-01-12
Restrict the Spawn::execute function

Idolf:
Since rust-lang/rust#36155 has been fixed, we can avoid an extra box
allocation when calling Spawn::execute with an already boxed future.

The downside of this is that it is compatibility breaking. I am now sure
if it is worth the breakage to get the performance improvement in a few
cases.

### DIFF
--- a/futures-cpupool/src/lib.rs
+++ b/futures-cpupool/src/lib.rs
@@ -184,7 +184,7 @@ impl CpuPool {
             tx: Some(tx),
             keep_running_flag: keep_running_flag.clone(),
         };
-        executor::spawn(sender).execute(self.inner.clone());
+        executor::spawn(sender.boxed()).execute(self.inner.clone());
         CpuFuture { inner: rx , keep_running_flag: keep_running_flag.clone() }
     }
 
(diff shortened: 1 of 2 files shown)


Idolf:
The only references to this in the tokio-rs organization are on  [the website][website] and in [tokio-hyper][tokio-hyper].

[website]: https://github.com/tokio-rs/website/blame/35c5634ae76016a97929532be596455fea3e57d9/content/docs/going-deeper/tasks.md#L219

[tokio-hyper]: https://github.com/tokio-rs/tokio-hyper/blob/5d1947a201a84d661296d5865a4292daaed970d3/src/lib.rs#L189

alexcrichton:
Thanks for the PR! I think I'll hold off on this until the 0.2 release (as it's a breaking API change), but otherwise looks great to me!

carllerche:
@alexcrichton I'm pretty sure `Spawn::execute` is deprecated (though it isn't marked as such). Is this correct?

alexcrichton:
I think it's ok to wait and see what happens with `Spawn` and close this for now.

PR #352, opened 2017-01-13
Introduce yield_after combinator for improving fairness

drewhk:
This PR is an attempt to create a `yield_after(u64)` combinator that limits the number of polls on a stream before it resubmits the task. The main purpose of this combinator to improve fairness if there are multiple competing streams for the same execution resources (especially in event loops) and some of the streams can run for a long time without returning `Async::NotReady`.

Currently this is PR is far from ready, because two things are unclear for me:
 - the purpose of Sink adapters (like in the most similar combinator, `take.rs`)
 - how to test this properly, since `task::park` will panic if there is no `Task` available because the stream has not been submitted anywhere, and I want to unit test the `poll()` method somehow. Any pointers?

Do you think this combinator is worthy for consideration? Does it actually work as I hope or are there hidden pitfalls of this approach?

### DIFF
--- a/src/stream/mod.rs
+++ b/src/stream/mod.rs
@@ -48,6 +48,7 @@ mod then;
 mod unfold;
 mod zip;
 mod forward;
+mod yield_after;
 pub use self::and_then::AndThen;
 pub use self::chain::Chain;
 pub use self::empty::{Empty, empty};
@@ -71,6 +72,7 @@ pub use self::take::Take;
 pub use self::take_while::TakeWhile;
 pub use self::then::Then;
 pub use self::unfold::{Unfold, unfold};
+pub use self::yield_after::YieldAfter;
 pub use self::zip::Zip;
 pub use self::forward::Forward;
 use sink::{Sink};
@@ -664,6 +666,33 @@ pub trait Stream {
         where Self: Sized
     {
         take::new(self, amt)
+    }
+
+    /// Creates a new stream that will resubmit its `Task` after a certain number
+    /// of successfully polled elements. The purpose of this combinator is to
+    /// ensure fairness between competing `Stream` instances on the same
+    /// executor, especially on event loops. Without yielding a long-running
+    /// stream (one that can be polled successfully for a large number of elements),
+    /// it can cause other, unrelated streams to starve for execution resources.
+    ///
+    /// Using this combinator the stream will produce only up to `yield_after`
+    /// elements before it returns `Async::NotReady`, then it immediately
+    /// unparks its `Task` so the executor can continue the stream later.
+    ///
+    /// If the original stream suspends itself then the yield counter is
+    /// reset, i.e. this limit only takes effect if the original stream
+    /// does not suspend itself after the specified elements have been polled.
+    /// For example if `yield_after` is set to 100, but the original stream
+    /// always returns `Async::NotReady` after 10 elements then this
+    /// combinator will not intervene as its counter is reset to 100 every
+    /// time the original stream signals it is not ready.
+    ///
+    /// Please note that this combinator can only ensure fairness if the
+    /// underlying executor is fair.
+    fn yield_after(self, yield_after: u64) -> YieldAfter<Self>
+        where Self: Sized
+    {
+        yield_after::new(self, yield_after)
     }
 
     /// Creates a new stream which skips `amt` items of the underlying stream.
--- a/src/stream/yield_after.rs
+++ b/src/stream/yield_after.rs
@@ -1,0 +1,58 @@
+use {Async, Poll};
+use stream::Stream;
+use task;
+
+/// A stream combinator that resubmits the task after a certain number of elements
+/// have been successfully polled, even if the stream would be ready for further polls.
+///
+/// This structure is produced by the `Stream::yield_after` method.
+#[must_use = "streams do nothing unless polled"]
+pub struct YieldAfter<S> {
+    stream: S,
+    yield_after: u64,
+    remaining: u64,
+}
+
+pub fn new<S>(s: S, yield_after: u64) -> YieldAfter<S>
+    where S: Stream,
+{
+    YieldAfter {
+        stream: s,
+        yield_after: yield_after,
+        remaining: yield_after
+    }
+
+}
+
+//TODO: What about the forwarding Sink? What should be done here?
+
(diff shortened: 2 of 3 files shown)


alexcrichton:
Seems like a neat combinator, thanks! I wonder if we want to perhaps consider adding a similar one to futures as well? I'm also curious if we could perhaps think of a more evocative name than `yield_after`, not that I'm exactly brimming with ideas in that department...

Stebalien:
This could be useful to quickly fix a bug but, IMO, fairness is the event loop's job. For example, `select` and friends currently poll in a round-robin fashion. This seems like a hack to get around unfair scheduling.

However, it might be useful to implement some form of yield combinator for Futures (filed under #353 to avoid getting off topic).

drewhk:
> Seems like a neat combinator, thanks! I wonder if we want to perhaps consider adding a similar one to futures as well?

Hm, I was thinking this makes no sense as a future can only return one `Ready` in a poll-cycle but I see now in #353# that it can still make sense to add preemption points to a larger chain. I can attempt to add that.

> This could be useful to quickly fix a bug but, IMO, fairness is the event loop's job. For example, select and friends currently poll in a round-robin fashion. This seems like a hack to get around unfair scheduling.

Unfortunately this is not true. This is not a hack to get around unfair scheduling but it is a workaround to get around limitations of the current architecture. The basic issue is that many (most?) streams are submitted to an executor as a `Future` where the `Future` itself drives the stream to completion. That leaves no preemption point for the executor to intervene. As an example, look into the loop in the `fold` implementation. The executor has no way to preempt that loop. This combinator would allow preemption to happen. Of course *ideally* the executor should be able to preempt without an explicit combinator but I don't see how that is possible without *all* (or most of) the combinators cooperating somehow (like reading a counter from the associated `Task` and returning `NotReady` when the counter is zero, and decrementing the counter on every `Ready`).

drewhk:
One option can be that `Task` gains another method `preempt() -> Boolean` which is up to the executor to handle or ignore. This means that all operators would need likely a macro, called as the first thing in `poll()` and which expands to something like `if task.yield() { return NotReady }`. Feels a bit brittle, but at least a start for brainstorming.

Stebalien:
> The basic issue is that many (most?) streams are submitted to an executor as a Future where the Future itself drives the stream to completion.

Ah, you mean like a `my_stream().for_each(...)`? I see what you mean. In that case, I'm worried that forgetting to insert a yield will be a foot gun. I've raised #354 to discuss the design space a bit.

drewhk:
> Ah, you mean like a my_stream().for_each(...)?

That is one example yes. Please note that this is not limited to futures backed by streams. The key is to have a "divergent" section, i.e. one that can make unbounded progress (or just long progress) without any event on another boundary (in this case the outermost poll). For example if you have a short stream of userIds, which you then flatten into a longer stream of friends of each of those users, then you filter down those to a few elements then you can have lots of internal "chit-chat" between the expanding and filtering phases without hitting the outermost poll. This might be a contrived example, but if there is any chance in the future that this library will support feedback cycles in streams than this will be no longer a weird edge case but a problem that needs to be tackled. But yes, we are talking about the same thing :)

As for this ticket, I am still not sure about the role of the `Sink` thingies and what should I supposed to to there. Also, another question is how to raise an error on an invalid argument, like passing zero to this combinator. Should it panic?

alexcrichton:
With all the various discussion on yielding I'm now tempted to hold off on this for now until we get a chance to have a broader discussion about what we're going to do in that domain. We can perhaps be conservative for 0.1 and consider deeper changes for 0.2, but I just want to make sure we're not too hasty on the lower layers.

carllerche:
Given that task wakeup is unpredictable, I'm not sure if the `u64` argument provides much additional value. Secondly, I am not sure if this makes sense as a combinator vs. a helper fn that provides does `park().unpark()` or which yields a `Future<Item = ()>` representing the yield.

Aka, I think that the yield concept is advanced enough to not be ideal as a fn on the trait.

drewhk:
> Given that task wakeup is unpredictable, I'm not sure if the u64 argument provides much additional value. 

All I can say that I have implemented and used this pattern (yielding after a count) on the JVM for different problems. Starving threadpools is very real. It is a significant percentage of our commercial support cases.   These counters+wakeup patterns many times improved latency last percentiles. I don't know if the combinator is the best approach, but you should take any chance you have to protect your threadpools if it is not a big cost.

carllerche:
@drewhk I never said anything against yielding :) Only that, in the case of futures, including a yield counter doesn't seem like it makes sense. Like, what would the reason be to have yield(5) vs. just a yield (and conditionally yield again). The issue is that there is no guarantee as to how long the task goes to sleep for. Spurious wake ups are to be expected, which makes something that implies precision seem confusing to me.

drewhk:
>  Only that, in the case of futures, including a yield counter doesn't seem like it makes sense.

I agree with that, but this PR is for the stream combinator, the Future combinator is another PR (https://github.com/alexcrichton/futures-rs/pull/359/files) done by @Stebalien, which does not take an u64 argument.

carllerche:
@drewhk by futures I meant this library (naming overload)

drewhk:
Well, so having the counter in the combinator is certainly not optimal, it is just that there is currently no other way given the current API. I would personally prefer a method, maybe called `should_yield() -> Boolean` that enables an operator to interface with the executor and then depending on the result either return `NotReady` or continue. It could be even neatly wrapped in a macro and then used either as the first statement in `poll()` or added to loop edges. The executor is clearly a better place to have these decisions as it can handle the whole Task together and make more educated decisions.

> Spurious wake ups are to be expected, which makes something that implies precision seem confusing to me.

Spurious wakeups do not matter much as the operator here is implemented in terms of _emitted elements_, i.e. if the task suspends for any other reason, the counter is reset. In other words this counter only takes effect if there was a long period of actual work done without signalling `NotReady`. It establishes an upper bound of elements processed without `NotReady` but does not intervene if there were any parks. Of course a timer can be more "accurate", but more costly as well. In practice these counters served me well in the past and worked just as well as timers.

drewhk:
I propose moving the discussion to the discussion ticket https://github.com/alexcrichton/futures-rs/issues/354. I will close this PR for now as there is no clear consensus on how to proceed.

PR #355, opened 2017-01-14
Optional futures

Ralith:
This provides another convenience tool for branching futures. Algebraically redundant to `Either`, so not necessarily needed, but it might make some code clearer.

### DIFF
--- a/src/future/mod.rs
+++ b/src/future/mod.rs
@@ -12,6 +12,7 @@ mod poll_fn;
 #[path = "result.rs"]
 mod result_;
 mod loop_fn;
+mod option;
 pub use self::empty::{empty, Empty};
 pub use self::lazy::{lazy, Lazy};
 pub use self::poll_fn::{poll_fn, PollFn};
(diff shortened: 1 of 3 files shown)


alexcrichton:
Seems like an interesting idea to me! I think I'd prefer to avoid the `Future::some` method (we don't have methods like `Future::ok`, etc.

cc @carllerche @aturon, thoughts about this?

Ralith:
> we don't have methods like `Future::ok`

Is this a good analogy? There's no `impl<F: Future, E> Future for Result<F, E>` either, afaik. I'm not strongly attached to the method, but it's consistent with the proposed additions in #356.

aturon:
I agree with @alexcrichton that it's a bit confusing to have this as a method, rather than a function, since it's drawing a strong analogy with the `Option::Some` constructor.

alexcrichton:
@Ralith want to trim this back to just the impl? We can always add the method later if need be and otherwise that's just the more conservative route.

Ralith:
Works for me; done.

alexcrichton:
Thanks!

PR #356, opened 2017-01-15
Combinators for wrapping futures in Either.

withoutboats:
If you are returning different futures in different branches of a
conditional expression, you need to wrap them in a construct which
unifies their types.

Currently, the easiest way to do this is to called the `.boxed()`
method on both of them, but a lower overhead solution is to wrap them
in an `Either`. This commit enables that by defining a `left` and
`right` method on `Future` which wrap the future in different variants
of the `Either`.

### DIFF
--- a/src/future/mod.rs
+++ b/src/future/mod.rs
@@ -834,6 +834,60 @@ pub trait Future {
     {
         Shared::new(self)
     }
+
+    /// Wrap this future in an `Either` future, making it the left-hand variant
+    /// of that `Either`.
+    ///
+    /// This can be used in combination with the `right` method to write `if`
+    /// statements that evaluate to different futures in different branches.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use futures::future::*;
+    ///
+    /// let x = 6;
+    /// let future = if x < 10 {
+    ///     ok::<_, bool>(x).left()
+    /// } else {
+    ///     empty().right()
+    /// };
+    ///
+    /// assert_eq!(x, future.wait().unwrap());
+    /// ```
+    fn left<B>(self) -> Either<Self, B>
+        where B: Future<Item = Self::Item, Error = Self::Error>,
+              Self: Sized
+    {
+        Either::A(self)
+    }
+
+    /// Wrap this future in an `Either` future, making it the right-hand variant
+    /// of that `Either`.
+    ///
+    /// This can be used in combination with the `left` method to write `if`
+    /// statements that evaluate to different futures in different branches.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use futures::future::*;
+    ///
+    /// let x = 6;
+    /// let future = if x < 10 {
+    ///     ok::<_, bool>(x).left()
+    /// } else {
+    ///     empty().right()
+    /// };
+    ///
+    /// assert_eq!(x, future.wait().unwrap());
+    /// ```
+    fn right<A>(self) -> Either<A, Self>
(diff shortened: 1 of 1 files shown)


withoutboats:
Probably there is some design work this could go through in terms of the discussion around making `Either`-like futures with more than 1 variant and such.

alexcrichton:
Seems nifty to me! We probably can't have much of an extensibility story here for types like `Either3` and beyond, and otherwise my only worry would be that left/right terminology, while it makes sense, doesn't match the `Either` type itself.

Overall I'm inclined to merge, but curious what others think!

withoutboats:
> my only worry would be that left/right terminology, while it makes sense, doesn't match the Either type itself.

`Either`'s variants could be renamed `Left` and `Right` at 0.2. `A` and `B` aren't very evocative names.

Ralith:
Renaming the variants sounds nice to me too. It's more readable (imo) and consistent with Haskell's precedent, and having methods on `Future` named `a`/`b` is unappealing.

alexcrichton:
Yes `A` and `B` were meant to be extensible, not necessarily evocative. If we add `Either3` it'd break the naming convention with `Either` (2)

alexcrichton:
@aturon do yo have thoughts on this perhaps? It does seem related to https://github.com/alexcrichton/futures-rs/pull/355, but I'd lean towards merging this one but not the `some` method there. Not entirely sure why...

tailhook:
I think `fut.left()` isn't very useful name (`Either::Left` is more clear comparing to `Future::left`, but not extensible).

What about `.either_a()` or `.as_either_a()` or `.into_either_a()` (`.into_a()`)? I think it might be even possible to make it work both for Either and Either3 with some trait magic.

tailhook:
Maybe `.wrap_a()` or `.wrap_as_a()` (since the title of this issue mentions "wrapping"), if you find `into_` or `as_` meaning something else. But I think `into_` in `Into::into` or `IntoFuture::into_future` means same thing. We need to chose or A/B, though.

alexcrichton:
Oh I kinda like `.either_a()` and that way we could have `.either3_a()` if we really needed to eventually

alexcrichton:
@withoutboats thoughts on `.either_a()` and such?

withoutboats:
@alexcrichton I think `left`/`right` are enough more evocative that they are easier to use (here is the 'left' branch, here is the 'right' branch), but ultimately I'll let you decide.

KeenS:
What the status of this PR? If you mind extensibility of the APIs, how about #522 ?

alexcrichton:
@KeenS this just stalled out on naming, and didn't make progress.

P-E-Meunier:
Related to this, I'm using only two extra combinators of this form (Either3 and Either4) on nest.pijul.com, and for larger branches I write explicit enums. Since writing Futures implementations for big enums is not particularly interesting, I wrote the futures-derive crate, which you can use by doing:

```
#[derive(Future)]
enum Either3<A, B, C> { A(A), B(B), C(C) }
```

This way, if you don't like the default naming of `Either` (I do like it), you can easily replace it in one readable line.

carllerche:
Why not just name it `either` and let the positions of the arguments determine `A` / `Left` or `B` / `Right`?

```rust
fn either<B>(self, b: B) -> Either<Self, B> { ... }
```

cramertj:
@withoutboats FYI now that we've moved to using the `either` crate, the variants are called `Left` and `Right`, so this PR seems like an obvious addition. I'm going to rebase this work and open a new PR.

PR #357, opened 2017-01-15
Unsynchronized bounded mpsc channel

Ralith:
I've attempted to implement a variant of `sync::mpsc` specialized for communication between tasks on the same thread. Atomicity overhead is avoided and message storage is pre-allocated, so sending messages through the channel should be fast. Sending a message will only produce a dynamic allocation when the number of blocked sender tasks increases substantially over the previous high watermark, which should only happen a small number of times per lifetime in typical use-cases.

This is my first attempt at inter-task synchronization code, so please review carefully. I'm not sure whether the unit tests adequately test inter-task behavior, though I was able to successfully test some simple multitasking network code I have using this channel.

To support the introduction of this and future additional channel types, I moved `SendError` out of `sync::mpsc` and into `sink` for easy reuse, `pub use`ing the old location with a deprecation attribute to maintain backwards compatibility.


src/unsync/mpsc.rs
```diff
@@ -0,0 +1,127 @@
...
+impl<T> Stream for Receiver<T> {
+    type Item = T;
+    type Error = ();            // TODO: Replace with uninhabited type, as this can never fail
+
+    fn poll(&mut self) -> Poll<Option<Self::Item>, Self::Error> {
+        if let Async::Ready(msg) = self.shared.as_ref().unwrap().borrow_mut().recv() {
+            return Ok(Async::Ready(Some(msg)));
+        }
+
+        match Rc::try_unwrap(self.shared.take().unwrap()) {
```
alexcrichton: The ownership movement here seems sort of unfortunate, but I think we've got some other strategies:

* Use `get_mut` instead of this to test "unique ownership"
* My preferred strategy, store a weak pointer in the `Receiver`, and a failed upgrade means the channel is terminated.

One case here I think that isn't handled is where once all senders are dropped it should unpark the receiver, and that could be handled naturally with a `Drop` implementation if the receiver stores a weak pointer.

tests/unsync.rs
```diff
@@ -0,0 +1,72 @@
+#![cfg(feature = "use_std")]
```
alexcrichton: Ah we only have tests that actually run with std so it's ok to remove this.

src/unsync/mpsc.rs
```diff
@@ -0,0 +1,127 @@
...
+use {Async, AsyncSink, Poll, StartSend, Sink, Stream};
+
+/// Creates a bounded in-memory channel with pre-allocated storage.
+///
+/// This method creates concrete implementations of the `Stream` and `Sink`
+/// traits which can be used to communicate a stream of values between tasks
+/// with backpressure. The channel capacity is exactly `buffer`.
+pub fn channel<T>(buffer: usize) -> (Sender<T>, Receiver<T>) {
+    let shared = Rc::new(RefCell::new(Shared {
+        buffer: VecDeque::with_capacity(buffer),
```
alexcrichton: Perhaps the capacity could be stored on the side and this could just allocate a `VecDeque`? We may not actually hit the capacity ever and this way we'd save memory for large `buffer` sizes passed in (perhaps)

src/unsync/mpsc.rs
```diff
@@ -0,0 +1,127 @@
...
+                Async::NotReady
+            }
+            Some(msg) => {
+                if let Some(task) = self.blocked_senders.pop_front() {
+                    task.unpark();
+                }
+                Async::Ready(msg)
+            }
+        }
+    }
```
alexcrichton: These methods are small enough that it should be fine to inline them below.

src/sink/send_error.rs
```diff
@@ -0,0 +1,35 @@
+use std::fmt;
+use std::error::Error;
+use std::any::Any;
+
+/// Error type for sending, used when the receiving end of a channel is
+/// dropped
+pub struct SendError<T>(pub T);
```
alexcrichton: Despite this being the same across the two modules I think I'd prefer to keep separate `SendError` types for now if possible. That way we can customize them (if need be) and then perhaps later in 0.2 or such if it's required we can unify them.

In the meantime though I don't think we get *too* much benefit from unifying the types, other than a little deduplication in this crate.

src/unsync/mpsc.rs
alexcrichton: Oh and could a test be added for dropping all senders waking up the receiver as well?

tests/unsync.rs
Ralith: Are you sure? I actually added this to resolve a CI failure. `cargo test --no-default-features` fails without it.

alexcrichton: Oh maybe we did start testing that at some point. In that case carry on!

Ralith:
I've implemented all of your suggestions that made sense to me, and fixed a potential deadlock when the receiver is dropped while any senders are blocked. I can't imagine why `SendError` would need to be different between any channel implementations, but I suppose it doesn't cost much to be conservative here. There were a few suggestions I didn't implement:

> store a weak pointer in the Receiver

This would cause in-flight messages to be silently lost when all senders are dropped. I instead simplified the ownership story with `Rc::get_mut`, which preserves property of never losing messages so long as a receiver exists.

> once all senders are dropped it should unpark the receiver, and that could be handled naturally with a Drop implementation if the receiver stores a weak pointer.

Because a weak pointer wasn't used, this doesn't fall out very well. I opted instead to unpark the receiver on every `Sender` drop for simplicity, since spurious wakeups should be tolerated; I can change this to manually reference-count senders if that's preferred. If the `rc_counts` feature was stabilized, `Rc::strong_count` would mitigate the need for additional manual reference-counting.

> could a test be added for dropping all senders waking up the receiver

I'm not sure how to write a test that models task (un)parking. Can you point me at an example?

Ralith:
On further consideration, I think it actually makes sense for the *senders* to hold weak pointers. I've made that change, resulting in some slight simplification.

src/sync/mpsc/mod.rs
```diff
@@ -130,9 +130,9 @@ pub struct Receiver<T> {
 /// `unbounded` method.
 pub struct UnboundedReceiver<T>(Receiver<T>);
 
-/// Error type for sending, used when the receiving end of the channel is
+/// Error type for sending, used when the receiving end of a channel is
 /// dropped
-pub struct SendError<T>(T);
+pub struct SendError<T>(pub T);
```
alexcrichton: Can this stay private?

src/unsync/mpsc.rs
```diff
@@ -0,0 +1,198 @@
...
+/// This is created by the `channel` function.
+pub struct Sender<T> {
+    shared: Weak<RefCell<Shared<T>>>,
+}
+
+impl<T> Drop for Sender<T> {
+    fn drop(&mut self) {
+        if let Some(task) = self.shared.upgrade().and_then(|shared| shared.borrow_mut().blocked_recv.take()) {
+            // Wake up receiver to avoid deadlock in case this was the last sender
+            task.unpark();
```
alexcrichton: Perhaps a count could be maintained to ensure this only happens when the last sender is dropped?

src/unsync/mpsc.rs
```diff
@@ -0,0 +1,198 @@
...
+                _ => {
+                    shared.buffer.push_back(msg);
+                    if let Some(task) = shared.blocked_recv.take() {
+                        task.unpark();
+                    }
+                    Ok(AsyncSink::Ready)
+                }
+            }
+        } else {
+            Err(SendError(msg))
```
alexcrichton: I tend to like to format these to optimize for less drift, perhaps like:

```rust
let shared = match self.shared.upgrade() {
    Some(shared) => shared,
    None => return Err(SendError(msg)),
};

// ...
```

src/unsync/mpsc.rs
```diff
@@ -0,0 +1,198 @@
...
+        }
+    }
+}
+
+impl<T> Stream for Receiver<T> {
+    type Item = T;
+    type Error = ();            // TODO: Replace with uninhabited type, as this can never fail
+
+    fn poll(&mut self) -> Poll<Option<Self::Item>, Self::Error> {
+        if let Some(shared) = Rc::get_mut(&mut self.shared) {
```
alexcrichton: I think this succeeding means all senders are gone, right? If so, could you add a comment to that effect?

src/unsync/mpsc.rs
```diff
@@ -0,0 +1,198 @@
...
+///
+/// Identical semantics to `channel`, except with no limit to buffer size.
+pub fn unbounded<T>() -> (UnboundedSender<T>, UnboundedReceiver<T>) {
+    let (send, recv) = channel_(None);
+    (UnboundedSender(send), UnboundedReceiver(recv))
+}
+
+/// Error type for sending, used when the receiving end of a channel is
+/// dropped
+pub struct SendError<T>(pub T);
```
alexcrichton: Could this field stay private with an `into_inner` method to access it?

Ralith:
Applied the second batch of comments. Any tips on testing task dynamics?

alexcrichton:
Oh right sorry I missed that. In this test suite the easiest thing historically has been to spawn threads and call `.wait()`, but given how this is not a synchronized version that may not be easy to do so.

To write a test with multiple tasks you may have to cook up a local executor which uses rc handles and such to emulate a mini event loop.

Ralith:
Pushed an attempt at that. I'm slightly suspicious because at no point did I actually implement `Executor`, but it seems to work. I'm not sure why `Unpark` implementations need to be `Sync`--surely they're only ever seen by one thread at a time in idiomatic code?

src/lib.rs
```diff
@@ -206,7 +206,9 @@ if_std! {
 
     pub mod task;
     pub mod executor;
+    pub mod local_executor;
```
alexcrichton: Oh sorry I more meant that this could be used in tests, we should avoid exposing it in the crate for now

src/local_executor.rs
```diff
@@ -0,0 +1,70 @@
...
+    pub fn spawn<F>(&mut self, f: F)
+        where F: Future<Item=(), Error=()> + 'static
+    {
+        self.live.insert(self.next_id, executor::spawn(Box::new(f)));
+        self.unpark_send.send(self.next_id).unwrap();
+        self.next_id += 1;
+    }
+
+    /// Run the loop until all futures previously passed to `spawn` complete.
+    pub fn run(&mut self) {
```
alexcrichton: This file looks pretty god! Should definitely work for running tests. If you feel the need to you can also add `F` as a type parameter to this `run` method which basically runs everything until that future completes.

Ralith:
Moved local_executor into tests/support and tweaked its API to be a bit closer to tokio-core's as suggested. I've also got most of a `oneshot` implementation written and will open a separate PR for it once it has tests and this one's been merged.

Ralith:
I feel like something like `local_executor` might be useful to have publicly available for e.g. handling large numbers of `sync` streams, but I suppose such a thing would be better off in another crate, seeing as it only needs public APIs anyway.

alexcrichton:
Awesome, thanks @Ralith! 

I'll look to merging soon

PR #358, opened 2017-01-15
Future merge combinator

Ralith:
This functions like `Select`, except that it does not require its component futures to be the same type, and allows the incomplete future to be recovered as its original type. Similar in spirit to `Stream::merge`. If the `merge` name sticks, their interfaces should be brought closer together, but the proposed interface is more flexible than the existing `Stream::merge` form so I'm hesitant to just copy the latter.

I find this much more natural than having to manually map heterogeneous futures into an enum prior to selecting them. Tests to be added if other people agree and this is deemed desirable.

Note that `Either` here is used purely as a convenient meaningless sum type, since for some reason rust doesn't otherwise have one of those, afaik.

### DIFF
--- a/src/future/hselect.rs
+++ b/src/future/hselect.rs
@@ -1,0 +1,36 @@
+use {Future, Poll, Async};
+use future::Either;
+
+/// Future for the `hselect` combinator, waiting for one of two differently-typed
+/// futures to complete.
+///
+/// This is created by the `Future::hselect` method.
+#[must_use = "futures do nothing unless polled"]
+pub struct HSelect<A, B> {
+    inner: Option<(A, B)>,
+}
+
+pub fn new<A, B>(a: A, b: B) -> HSelect<A, B> {
+    HSelect { inner: Some((a, b)) }
+}
+
+impl<A, B> Future for HSelect<A, B> where A: Future, B: Future {
+    type Item = Either<(A::Item, B), (B::Item, A)>;
+    type Error = Either<(A::Error, B), (B::Error, A)>;
+
+    fn poll(&mut self) -> Poll<Self::Item, Self::Error> {
+        let (mut a, mut b) = self.inner.take().expect("cannot poll HSelect twice");
+        match a.poll() {
+            Err(e) => Err(Either::A((e, b))),
+            Ok(Async::Ready(x)) => Ok(Async::Ready((Either::A((x, b))))),
+            Ok(Async::NotReady) => match b.poll() {
+                Err(e) => Err(Either::B((e, a))),
+                Ok(Async::Ready(x)) => Ok(Async::Ready((Either::B((x, a))))),
+                Ok(Async::NotReady) => {
+                    self.inner = Some((a, b));
+                    Ok(Async::NotReady)
+                }
+            }
+        }
+    }
+}
--- a/src/future/mod.rs
+++ b/src/future/mod.rs
@@ -49,6 +49,7 @@ mod map_err;
 mod from_err;
 mod or_else;
 mod select;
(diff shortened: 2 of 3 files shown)


alexcrichton:
Thanks for the PR! I'm a little unsure on the name here as well. It's not *quite* the same thing as `Stream::merge` as you're not necessarily merging the output, you're still just waiting for the first of two futures. 

I'm not thrilled about the usage of the `Either` type which was intended purely for left/right future branches. The more we use it the more we'll be expected to build up an either `Either` library, which seems unfortunate.

Ralith:
Other name ideas: `hselect`, `first`, `or`, replace `select` with this outright.

We could use an external `Either` library, or have a dedicated enum for this future. I'm not a huge fan of the latter approach, since it leads to duplication of convenience methods and doesn't IMO help readability.

alexcrichton:
It's true yes that there's a number of external libraries that we could use, but I've been hoping to keep the dependencies of this crate pretty lean.

To step back a bit could you elaborate on the rationale for this combinator? When I've had to do something like this locally I've been able to do:

```
a.map(Ok).select(b.map(Err))
```

Basically just quickly wrap up the two futures in a `Result` and then they've both got the same type to be selected over.

Ralith:
The biggest part of my motivation is that every time I've needed to use `select` myself, it's been on heterogeneous futures, so having to do a dance like that every time was getting annoying. However, there is one plausibly-useful thing this can do that `select` as it stands cannot: when it resolves, you get back *exactly* the original future you put in for the case that didn't complete first, instead of a black box.

This is this useful for cases where you need to do something with it that the `Future` interface alone doesn't support, and provides greater opportunity for static program correctness, because the type that the incomplete future will resolve to is known at compiletime: you don't have to manually dynamically assert that it's the case that wasn't completed. In that regard, the types of this proposal are more specific in a way that the existing `select` operator cannot emulate.

alexcrichton:
@Ralith oh right yeah that's a very good point. I know we've discussed recursive situations in the past where getting back the exact future is useful. I'm not sure that this version would be more ergonomic, though, because you'd have to import `Either` and still do a bunch of matching, right?

Regardless I think getting back the exact same future is alone enough motivation for such a combinator. I can live with growing a mini `Either` library in the meantime until we think of something better.

The only real sticking point for me is the name of this combinator. We could perhaps consider `select2` and then if we add more `Either` variants we could add `select3`, etc, maybe?

Ralith:
That naming seems reasonable to me, although it seems strange for `select` and `select2` to both exist (as different things, no less). It feels a bit weird to have three different variations on the theme of summing futures in the library as well, but perhaps the diversity of use cases justifies it.

alexcrichton:
It's true yeah that we're growing a lot of names for similar operations...

@aturon, @carllerche, thoughts?

aturon:
I wonder if it's feasible to deprecate the homogeneous version in favor of this one? We could offer some methods to get something equivalent from the output, e.g.:

```rust
impl<T, A, B> Either<(T, A), (T, B)> {
    fn split(self) -> (T, Either<A, B>) { ... }
}
```

carllerche:
I'm with @aturon on preferring this version of `select` over the original.

alexcrichton:
@aturon that sounds pretty reasonable, yeah

In terms of immediate-term naming that may make me think that `select2` is the best option and we can rename to `select` at the 0.2 mark

alexcrichton:
Ok I've merged this (sorry for the delay) with the name `select2`. I've opted to not deprecate `select` just yet to get some experience with it, but we can deprecate after the next release.

Issue #354, opened 2017-01-14
Handle long-running, always-ready futures fairly


Stebalien:
In #352, @drewhk introduced a yield combinator for streams as a way to to ensure fairness. As I understand it, the problem is that `my_stream().for_each(...)` will block until `my_stream()` returns `NotReady`. If `my_stream()` is always ready, it will block forever and never let anything else run.

**Ideas:**

1. @drewhk's proposal was to introduce a `yield_after(items)` combinator for streams. One would write `my_stream().yield_after(n_items).for_each(...)` and the `yield_after` combinator would force the stream to return `NotReady` after returning `n_items`. This way, the future returned by `my_stream().yield_after(n_items).for_each(...)` would return a `NotReady` after processing `n_items` and the scheduler would have the opportunity to schedule something else.

2. An alternative solution is to have `for_each` (and friends) return `NotReady` after every item.  Basically, as a rule, futures would always do as little work as possible and then let the scheduler decide whether or not you should do more work. This could be a performance issue due to extra bookkeeping (I'm mostly worried about potentially slow reads/writes of atomic flags) and might make it harder for the compiler to optimize loops.

3. Provide some sort of `task::should_yield()` function and check it before yielding as in solution (2). This could reduce some of the overhead.

4. Extend `Async` to include an `Async::Yield` state (in addition to either 2 or 3). This could be used to avoid the overhead of park/unpark but is a big breaking change and adds significant complexity.

5. Instead of 4,  implement `task::yield()` (#177) as a replacement for `park().unpark()`. Given that this guarantees that `unpark()` will be called from the task's thread, we could probably avoid the overhead of atomics described in (2).

6. Introduce a yield combinator for futures as described in #353. One would then write `my_stream().for_each(|item| { ...; futures::finished(()).then_yield() })` to force a yield after every item. IMO, for this particular use case, 1 is a better option (less cumbersome).

---

Thoughts? Other ideas?

drewhk:
> @drewhk's proposal was to introduce a yield_after(items) combinator for streams.

I am not seriously proposing, I was just thinking about the problem, realized that I cannot solve it at the executor level, then on a sleepness night I realized that a combinator can do that, so I just sketched it up and submitted here :)

>  This could be a performance issue due to extra bookkeeping (I'm mostly worried about potentially slow reads/writes of atomic flags) 

[...]

PR #359, opened 2017-01-16
[WIP] Add a then_yield combinator for futures.

### LINKED ISSUE #353: Yield combinator for futures
To avoid blocking for too long, it could be useful to have a yield combinator for futures. Specifically, one that returns `NotReady` once after the wrapped future completes and then returns the result on the next poll. For example, in:

```rust
let my_future = some_call()
    .yield()
    .and_then(do_second);
```

Even if `some_call` and `do_second` are actually synchronous, the `yield` call will force `my_future` to return `NotReady` after completing `some_call` and before running `do_second`.

Stebalien:
This is useful when one wants to yield to the scheduler between expensive,
synchronous computations.

Fixes #353

TODO:

* [ ] Better documentation? Should I be using a stream in the example? Really,
  the `for_each` stream combinator should probably be fixed separately as
  described in #354.
* [ ] Better name? I'd like to use `yield` but that's reserved.

### DIFF
--- a/src/future/mod.rs
+++ b/src/future/mod.rs
@@ -50,6 +50,7 @@ mod from_err;
 mod or_else;
 mod select;
 mod then;
+mod then_yield;
 mod either;
 
 // impl details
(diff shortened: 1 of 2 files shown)


Stebalien:
This really needs some discussion/design work so I'm closing it for now.

PR #365, opened 2017-01-19
Correct comment in docs

rhys-vdw:
Comment references use of `.boxed()` which was removed from example in 4d4ce22f.


alexcrichton:
Thanks!

PR #373, opened 2017-01-24
Use with_unpark_event() to fix some possible deadlocks in Shared.

### LINKED ISSUE #330: Shared can interact badly with futures that don't always poll their subfutures
(I'm opening this issue to continue the discussion from https://github.com/alexcrichton/futures-rs/pull/305.)

The problem is that `Shared`, as it is currently designed, can interact poorly with certain futures, such as the `ModedFuture` sketched below. We should either formalize some reason why `ModedFuture` is an invalid implementation of `Future`, or we should redesign `Shared` to accommodate this kind of situation.

```rust
extern crate futures;
extern crate tokio_core;

use futures::{Future, Poll};
use futures::sync::oneshot;
use std::rc::Rc;
use std::cell::RefCell;

enum Mode { Left, Right }

struct ModedFutureInner<F> where F: Future {
    mode: Mode,
    left: F,
    right: F,
    task: Option<::futures::task::Task>,
}

struct ModedFuture<F> where F: Future {
    inner: Rc<RefCell<ModedFutureInner<F>>>,
}

struct ModedFutureHandle<F> where F: Future {
    inner: Rc<RefCell<ModedFutureInner<F>>>,
}

impl <F> ModedFuture<F> where F: Future {
    pub fn new(left: F, right: F, mode: Mode) -> (ModedFutureHandle<F>, ModedFuture<F>) {
        let inner = Rc::new(RefCell::new(ModedFutureInner {
            left: left, right: right, mode: mode, task: None,
         }));
        (ModedFutureHandle { inner: inner.clone() }, ModedFuture { inner: inner })
    }
}

impl <F> ModedFutureHandle<F> where F: Future {
    pub fn switch_mode(&mut self, mode: Mode) {
        self.inner.borrow_mut().mode = mode;
        if let Some(t) = self.inner.borrow_mut().task.take() {
            // The other future may have become ready while we were ignoring it.
            t.unpark();
        }
    }
}

impl <F> Future for ModedFuture<F> where F: Future {
    type Item = F::Item;
    type Error = F::Error;
        let ModedFutureInner { ref mut mode, ref mut left, ref mut right, ref mut task } =
            *self.inner.borrow_mut();
        *task = Some(::futures::task::park());
        match *mode {
            Mode::Left => left.poll(),
            Mode::Right => right.poll(),
        }
    }
}

pub fn main() {
    let mut core = ::tokio_core::reactor::Core::new().unwrap();
    let handle = core.handle();

    let (tx, rx) = oneshot::channel::<u32>();
    let f1 = rx.shared();
    let f2 = f1.clone();

    let (mut mfh, mf) = ModedFuture::new(
        Box::new(f1.map_err(|_| ()).map(|v| *v)) as Box<Future<Item=u32, Error=()>>,
        Box::new(::futures::future::empty()) as Box<Future<Item=u32, Error=()>>,
        Mode::Left);

    let (tx3, rx3) = oneshot::channel::<u32>();
    handle.spawn(f2.map(|x| tx3.complete(*x)).map_err(|_| ()));

    core.turn(Some(::std::time::Duration::from_millis(1)));

    handle.spawn(mf.map(|_| ()));

    core.turn(Some(::std::time::Duration::from_millis(1)));

    mfh.switch_mode(Mode::Right);

    tx.complete(11); // It seems like this ought to cause f2 and then rx3 to get resolved.

    // This hangs forever.
    core.run(rx3).unwrap();
}
```

dwrensha:
Fixes #330 and adds two tests that fail on the current master branch.

### DIFF
--- a/src/future/shared.rs
+++ b/src/future/shared.rs
@@ -128,11 +128,15 @@ impl<F> Future for Shared<F>
             None => unreachable!(),
         };
 
+        // Lock `state` before dropping `original_future` so that another thread
+        // cannot observe a situation where `original_future` is None and `state`
+        // is `Waiting`.
+        let mut state = self.inner.state.lock().unwrap();
+
         // We can now drop the original future to free up any resources it holds.
         original_future.take();
         drop(original_future);
 
-        let mut state = self.inner.state.lock().unwrap();
         match mem::replace(&mut *state, State::Done(done_val.clone())) {
             State::Waiting(ref unparker) => unparker.unpark(),
             _ => unreachable!(),
--- a/src/future/shared.rs
+++ b/src/future/shared.rs
@@ -14,7 +14,7 @@
 //! ```
 
 use std::mem;
-use std::sync::{Arc, Mutex, RwLock, TryLockError};
+use std::sync::{Arc, Mutex, TryLockError};
 use std::sync::atomic::{AtomicBool, Ordering};
 use std::ops::Deref;
 use std::collections::HashMap;
@@ -38,7 +38,7 @@ struct Inner<F: Future> {
     /// without holding a lock on `state`.
     original_future: Mutex<Option<F>>,
 
-    state: RwLock<State<F>>,
+    state: Mutex<State<F>>,
 }
 
 enum State<F: Future> {
@@ -57,7 +57,7 @@ impl<F> Shared<F>
                 Inner {
                     next_clone_id: Mutex::new(1),
                     original_future: Mutex::new(Some(future)),
-                    state: RwLock::new(State::Waiting(Arc::new(Unparker::new()))),
+                    state: Mutex::new(State::Waiting(Arc::new(Unparker::new()))),
                 }),
         }
     }
@@ -70,7 +70,7 @@ impl<F> Future for Shared<F>
     type Error = SharedError<F::Error>;
 
     fn poll(&mut self) -> Poll<Self::Item, Self::Error> {
-        let state = self.inner.state.read().unwrap();
+        let state = self.inner.state.lock().unwrap();
         let (mut original_future, event) = match *state {
             State::Waiting(ref unparker) => {
                 match self.inner.original_future.try_lock() {
@@ -123,7 +123,7 @@ impl<F> Future for Shared<F>
         original_future.take();
         drop(original_future);
 
-        let mut state = self.inner.state.write().unwrap();
+        let mut state = self.inner.state.lock().unwrap();
         match mem::replace(&mut *state, State::Done(done_val.clone())) {
             State::Waiting(ref unparker) => unparker.unpark(),
             _ => unreachable!(),
@@ -153,12 +153,13 @@ impl<F> Clone for Shared<F>
 
 impl<F: Future> Drop for Shared<F> {
     fn drop(&mut self) {
-        let state = self.inner.state.read().unwrap();
-        match *state {
-            State::Waiting(ref unparker) => {
-                unparker.remove(self.id);
+        if let Ok(state) = self.inner.state.lock() {
+            match *state {
+                State::Waiting(ref unparker) => {
+                    unparker.remove(self.id);
+                }
+                State::Done(_) => (),
             }
-            State::Done(_) => (),
         }
     }
 }
--- a/src/future/shared.rs
+++ b/src/future/shared.rs
@@ -17,6 +17,7 @@ use std::mem;
 use std::sync::{Arc, Mutex};
 use std::ops::Deref;
 use std::collections::HashMap;
+use std::vec::Vec;
 
 use {Future, Poll, Async};
 use task;
@@ -36,7 +37,7 @@ struct Inner<F: Future> {
 
 enum State<F: Future> {
     Waiting(Arc<Unparker>, F),
-    Polling(Arc<Unparker>),
+    Polling(Arc<Unparker>, Vec<task::Task>),
     Done(Result<Arc<F::Item>, Arc<F::Error>>),
 }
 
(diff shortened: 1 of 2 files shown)


src/future/shared.rs
```diff
@@ -51,85 +70,77 @@ impl<F> Future for Shared<F>
...
+            Some(ref mut future) => {
+                match task::with_unpark_event(event, || future.poll()) {
+                    Ok(Async::NotReady) => return Ok(Async::NotReady),
+                    Ok(Async::Ready(v)) => Ok(Arc::new(v)),
+                    Err(e) => Err(Arc::new(e)),
+                }
             }
-            State::Done(Ok(ref e)) => Ok(SharedItem { item: e.clone() }.into()),
-            State::Done(Err(ref e)) => Err(SharedError { error: e.clone() }.into()),
+            None => unreachable!(),
```
alexcrichton: I think this panic can get triggered, but we can just ignore the case, right?

I'm thinking of a case where:

* One task wins the race to do the poll (turns off `original_future_needs_poll`)
* The future is unparked, setting `original_future_needs_poll` to true
* Another task wins the race to poll (again turning off `original_future_needs_poll`)
* The first task finishes the future, setting it to none
* The second task then panics when it sees the none

Or maybe I missed something?

alexcrichton: If this is true then I think that the reading of `state` may need to be updated as well

dwrensha: Hm. Yeah, I agree that the panic can be triggered. Worse, if `original_future()` unparks the current task and then causes a `Shared` clone to recursively call `poll()` on the same thread, then that thread might attempt to recursively grab the lock for `original_future`.

alexcrichton: Hm good point yeah. Maybe this could get by with a `try_lock` and if it's missed it just falls through to storing a parked task?

dwrensha: No, that would leave `original_future_needs_poll` in the wrong state.

I think it works to probe the `original_future` mutex before doing the swap on `original_future_needs_poll`, to check whether we are in the recursive case.

dwrensha: I've pushed a commit that should fix this issue and includes a test that covers it.

dwrensha: Hm. On second thought, the new branch that I added does not only apply to the recursive case. If a `poll()` on a separate thread fails to grab the `original_future` mutex, it could now get into a busy wait loop of repeatedly calling `task::park().park()`.

dwrensha: I've updated the code comments in that branch. It does seem unfortunate that the `task::park().unpark()` might cause waiting threads to consume a lot of cpu, but I haven't been able to think of a better solution. Fortunately, the problematic case seems like it would be rare: `original_future.poll()` would need to both call `task::park().unpark()` *and* take a lot of time before returning.

dwrensha:
I've updated `Unparker` so that it wraps the `original_future_needs_poll` flag and the `tasks` hashmap in a single mutex. This makes it possible to avoid calling `task::park().unpark()` in the common cases. Calling `task::park().unpark()` is still necessary in the case where `original_future.poll()` wakes up its own task.

I *think* this addresses the concerns brought up so far, but... it's subtle. @alexcrichton, can you find any other potential issues?

src/future/shared.rs
alexcrichton: I think with the recent changes this case of panic can still be hit, but double check me:

* A task gets all the way to `drop(original_future)` below, unlocking the lock. The future is completed and dropped
* Another task comes in an polls, sees the `Waiting` state, acquires the original future lock, sees that a poll is needed (some stray unpark set the flag perhaps)
* The second task comes down to this branch

Does that sound right?

alexcrichton:
@dwrensha could you elaborate on the motivation for multiple locks here? I'd imagine that keeping just one mutex would simplify this perhaps, no?

alexcrichton:
(I'm not *too* worried about the poll-during-unpark case btw, just wanted to weed out other issues before scrutinizing that too much)

dwrensha:
> could you elaborate on the motivation for multiple locks here?

Good question! The idea was to avoid locking `state` while `original_future.poll()` is being called. However, I'm not sure why it didn't occur to me that we don't actually need to hold *any* lock during `original_future.poll()`. I've pushed a commit that pushes `original_future` back into `State`, avoiding the need for a separate `original_future` mutex.

src/future/shared.rs
dwrensha: Yes, I agree. I think https://github.com/alexcrichton/futures-rs/pull/373/commits/30d90d7bd497c30a5e6b1a4daf8c0b593a37aab2 fixes this, and then https://github.com/alexcrichton/futures-rs/pull/373/commits/92c89415d614eda6725ab57631df3209cfbe2869 sidesteps the entire issue.

alexcrichton:
Looks good to me! Threading ownership of the future looks like an elegant solution to solve a number of problems.

So turning now to the performance characteristics, I've got two points of hesitation:

1. The thundering stampede problem for *any* unpark. Unfortunately I think this is unavoidable (e.g. the whole impetus for this PR) at this point, and you can always optimize this by using something like a `oneshot` in the middle which is guaranteed to have zero spurious unparks.
2. After the thundering stampede, possible spinning for awhile by all threads.

On the latter point I think it has specifically do do with the `park().unpark()` business. On an unpark all tasks wake up but if they all hit the lock at the same time then one will start polling and N-1 will enter "busy loops". Let's assume that the future won't be ready for some time, so then N-2 will be in a busy loop, etc. (if that makes sense). 

Well hm now after writing that I realize it's only the case when a bunch of unparks are coming in in which case you sorta have to check for it anyway. So maybe it's not that big of a problem!

I wonder, though, perhaps this could be handled by the original poller? Instead of putting threads into a park/unpark loop, perhaps the owner of the future could atomically lock and ensure that polling isn't required? That way the owner would `loop` around polling the future until the "needs poll" flag is unset.

A scheme like that I think would help make a bit more progress (no threads in a "park loop") although it runs the risk of starving the owner of the future by looping for awhile. That being said, that's a pervasive problem, so I don't think we'd need to solve it specifically here.

dwrensha:
> Instead of putting threads into a park/unpark loop, perhaps the owner of the future could atomically lock and ensure that polling isn't required? That way the owner would `loop` around polling the future until the "needs poll" flag is unset.

Do you mean wrap the body of `Shared::poll()` in a `loop {}` that continues until `original_future_needs_poll` is `false`? That sounds wrong to me. If `task::pask().unpark()` gets called during `original_future.poll()`, presumably that means that `original_future` wants to yield control to other tasks running on the thread, but if we add a loop we'll just immediately call `original_future.poll()` again.

I've pushed a commit with a nicer solution: we keep a `Vec` of tasks that will be unconditionally unparked when the state transitions away from `State::Polling`. I had in fact considered this approach previously, but when there were separate locks for `state` and `original_future` it was prone to race conditions. Now that `original_future` is part of `state`, I think it works.

alexcrichton:
Just to write down my thoughts, the race condition we're considering here is:

* The future is actively being polled
* An unpark happens, setting the "needs a poll" flag
* Another task concurrently comes in, sees the flag, and doesn't see the future

So in general unparks during a poll are totally normal and happen all the time, they just schedule the task to go back and re-poll the future again. The race here though is we have to be sure to not lose that poll request because it could result in the future getting stuck.

With that in mind, I think the case to worry about is when the future is being polled, an unpark happens, all other tasks miss the future itself, then the task polling the future unlocks the future but then forgets about its instance of `Shared` (somehow). We need to make sure that the request to keep polling the future is communicated to other blocked tasks (if any).

Having written that out, the `Vec` feels right here. It's a list of otherwise blocked tasks. When the original future leaves the block it wakes them all up. I believe the task working on the future may *not* have received an unpark from the future's current unpark as it may not have been blocked yet maybe?

In any case, the `Vec` sounds good to me, thanks!

Issue #129, opened 2016-09-06
Consider passing the Task that is driving the future to Future::poll and renaming the function


rjuse:
Currently Future::poll seems to be expected to call task::park which then fetches the current task from TLS and panics if there is no task in TLS.

This results in an unintuitive API (it's not clear at first glance that poll()'s expected interface/implementation is related to tasks) and a potential run-time failure that could be checked at compile time.

So my suggestion is to instead pass the task driving the Future explicitly to Future::poll as an additional function argument, either as a Task reference, a closure calling task::park() (if that's enough), or a similar mechanism, instead of storing it in the TLS variable CURRENT_TASK.

Also, "poll" is a confusing name, since it creates the expectation that it is a function that anyone can call to get the value of the future if it has already completed, but it is in fact an "internal" function that drives future execution instead and currently even panics if called outside a task.

Something like "drive", "execute", "run", "run_next_step" or similar would be a better name.

aturon:
FWIW, we actually started with a design that passed an explicit `&mut Task` argument. That ended up posing problems for things like implementing the direct `Read` trait, which also wants access to a task. Mediating this through thread-local storage made it possible to support standard I/O APIs while implicitly accessing the task, and also greatly improved ergonomics.

I do agree that thread-local storage has downsides, in terms of being less explicit and in our case opening you up to runtime errors. However, I think that once you understand the basics of the library, this kind of mistake is very rare and will very reliably product panics (so shouldn't be hard to debug).

Regarding the naming, I can see what you mean but none of the suggested names particularly evoke "tasks" for me in a way that the current name doesn't.

rjuse:
I think implementing the Read trait this way might not be the optimal design (and also might be broken currently if it's called outside of a Future's poll function, although I suppose that's easily fixed by doing nothing if there is no current task).

A possible alternative that seems more sound is to implement, if desired, the Read trait in a way that is simply purely nonblocking and thus does not access the current task.

End users would normally use APIs that return futures (and that thus aren't Read or similar traits), while the internal implementation of those APIs can work with internal functions that take a task argument instead of the Read trait.

aturon:
[...]

PR #374, opened 2017-01-25
[experiment] remove CURRENT_TASK thread-local variable

dwrensha:
`CURRENT_TASK` was introduced in #77, a change that bundled the removal of the `schedule()` method together with the removal of the `task` parameter to `poll()`. There has been some discussion, e.g. in #129, about whether the latter removal has been a net win. My feeling is that is has *not*, and this pull request represents my attempt to see what would be involved in bringing back explicitly-passed tasks.

 I have not yet figured out a way to get task-local storage working with this approach, so at the moment this pull request removes `task_local!()`. I'm interested in hearing about any ideas for how we could continue to support task-local storage, or how we might design something that fills the same need.


dwrensha:
Hm. Another issue is that `Task` depends on `::std::sync::Arc`, and therefore will not be available in `no_std` environments.

Diggsey:
For efficient task-local storage, you might need to have a `LocalTask`/`RemoteTask` pair like in tokio, and only allow access to task-local storage through a `LocalTask`.

This could be implemented as a `RefCell<AnyMap>` on the `LocalTask`. Only the `RemoteTask` would need `Arc` to be implemented.

Diggsey:
On second thoughts, I think the important distinction is not between Local/Remote: it's actually between whether we're *in* the task, or simply referring to the task.

Instead of introducing `LocalTask`/`RemoteTask` types, add a new type, `TaskHandle` through which a task can be unparked. `Task` would be neither `Sync` nor `Clone`, and `with_unpark_event` would invoke a closure with a `&Task` scoped to the lifetime of that closure.

A `&Task` can at any time be downgraded to a `TaskHandle`, which can be freely cloned and stored, while task-local storage can only be accessed through a `&Task`.

carllerche:
The discussion around the concepts introduced in the PR are being discussed in #129.

alexcrichton:
Note that with https://github.com/alexcrichton/futures-rs/pull/436 this crate is now compatible with `no_std` entirely (including the task system) with only a requirement you call the [init function](https://github.com/alexcrichton/futures-rs/blob/b24dea100c01224ca75bb73e9a1e852e284e176a/src/task_impl/core.rs#L130-L137) with thread-local-like semantics. In that sense we still rely on a thread-local *somewhere* but it may impact this somewhat as well.

alexcrichton:
Ok at this point this is a pretty old PR without a lot of recent activity, so I'm going to close this. Discussion can of course continue on https://github.com/alexcrichton/futures-rs/issues/129 though!

PR #375, opened 2017-01-25
update docstring from notify() to unpark()

dwrensha:
When `notify()` was renamed to `unpark()` in https://github.com/alexcrichton/futures-rs/commit/cb863674fc6b8148649484191bc5b7f84d07a726, the docstring did not get updated.


alexcrichton:
Thanks!

PR #383, opened 2017-02-03
Add `peek()` for Shared

stuhood:
In many situations it is useful to be able to get the value of a `Shared` future if it has already completed for some other purpose... particularly since `wait`ing for it might trigger work that would not have been necessary otherwise.

To avoid a schrödinger's cat situation, this change adds a `peek` method for Shared.


dwrensha:
This looks reasonable to me.

alexcrichton:
Thanks!

PR #387, opened 2017-02-08
Implement Sink::map_err

Ralith:
Not sure if this was omitted for some particular reason. It should make it much more convenient to e.g. `Stream::forward` into a `Sink` with a different error type.

### DIFF
--- a/src/sink/map_err.rs
+++ b/src/sink/map_err.rs
@@ -1,0 +1,30 @@
+use sink::Sink;
+
+use {Poll, StartSend};
+
+/// Sink for the `Sink::map_err` combinator.
+#[must_use = "sinks do nothing unless polled"]
+pub struct MapErr<S, F> {
+    sink: S,
+    f: Option<F>,
+}
+
+pub fn new<S, F>(s: S, f: F) -> MapErr<S, F> {
+    MapErr { sink: s, f: Some(f) }
+}
+
+impl<S, F, E> Sink for MapErr<S, F>
+    where S: Sink,
+          F: FnOnce(S::SinkError) -> E,
+{
+    type SinkItem = S::SinkItem;
+    type SinkError = E;
+
+    fn start_send(&mut self, item: Self::SinkItem) -> StartSend<Self::SinkItem, Self::SinkError> {
+        self.sink.start_send(item).map_err(|e| self.f.take().expect("cannot use MapErr after an error")(e))
+    }
+
+    fn poll_complete(&mut self) -> Poll<(), Self::SinkError> {
+        self.sink.poll_complete().map_err(|e| self.f.take().expect("cannot use MapErr after an error")(e))
+    }
+}
--- a/src/sink/map_err.rs
+++ b/src/sink/map_err.rs
@@ -2,18 +2,18 @@ use sink::Sink;
 
 use {Poll, StartSend};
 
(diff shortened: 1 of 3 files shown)


alexcrichton:
Thanks! Since it's common to have `Stream + Sink` implemented, perhaps this could be called `sink_map_err` to avoid conflicts?

Ralith:
There are certainly a few big primitives that naturally implement `Stream + Sink`, but how common is it really to work with them in non-split form, considering that most applications will want to read and write concurrently?

carllerche:
I personally almost never use them in the split form, so I guess that would
be one data point towards it not being uncommon.

carllerche:
That being said, having to prefix all Sink combinators w/ `sink_` does not
sound like a great situation.

alexcrichton:
Historically we've concluded that for objects like transports to tokio proto it's best for them to implement both the `Sink` and `Stream` traits. We've also concluded that for ergonomic combinators with these types we'll continue to forward implementations of these two traits through combinators. For example `stream::Map` has a [forwarding implementation of `Sink`](https://github.com/alexcrichton/futures-rs/blob/c53bad27f61494d4cfaa0086409802ef5c7bfd4b/src/stream/map.rs#L24-L38).

So given that these objects (types with Stream+Sink) exist *and* we intend them to be used with combinators (with forwarding impls) then it means that `map_err` is by default ambiguous if added to `Sink`. As a result I think we need *some* disambiguation here. I agree that `sink_*` isn't the best, but it's at least a convention that can be stuck to

Ralith:
I'm really not a fan of the `Stream+Sink` pattern, to be honest. It seems to me that you nearly always need to end up splitting things, as this is the only way to accomplish concurrent reading and writing. Handling the `Stream` and `Sink` separately would simplify code and reduce the confusion that I've repeatedly seen in newcomers trying to understand how they're supposed to achieve that concurrency. Furthermore, the generic `split` implementation must be very conservative--its `BiLock` should never be necessary for TCP and UDP sockets or for other cases where you've glued entirely separate `Stream` and `Sink` together. Letting the splitting of sockets into a separate `Stream` and `Sink` occur at an earlier stage, where the concrete type is known, therefore both simplifies code and reduces overhead. This is also consistent with (what I hope to be) the movement of 0.2 away from monolithic bidirectional objects, which is all together making the abstractions tokio provides broadly more useful for unidirectional primitives like stdio streams and pipes.

However, I concede that these concerns are out of scope for this PR. I'll rename the method for now and perhaps it can be reconsidered for 0.2.

Ralith:
Other benefits of ditching `Stream+Sink`: allows the removal/precludes the creation of large amounts of forwarding boilerplate, and allows the `split` method to be removed from its awkward placement on the `Stream` trait and relocated to more logical inherently-splittable concrete objects like `TcpStream`.

src/sink/map_err.rs
```diff
@@ -0,0 +1,30 @@
+use sink::Sink;
+
+use {Poll, StartSend};
+
+/// Sink for the `Sink::map_err` combinator.
+#[must_use = "sinks do nothing unless polled"]
+pub struct MapErr<S, F> {
```
alexcrichton: Could this be called `SinkMapErr` for consistency? (also the comment above this may need updating)

Ralith:
Naming tweaked.

PR #388, opened 2017-02-08
remove unneeded macro_use attributes

dwrensha:
The newest beta `rustc` warns on these, so let's remove them.


alexcrichton:
Thanks!

Issue #320, opened 2017-01-01
Lots of public types do not implement `Debug`


Diggsey:

alexcrichton:
I believe this was implemented in #392

ipetkov:
#392 picked off the low hanging fruit, but there are definitely a more than a few public types left which don't implement `Debug`. I remember leaving some of these out due to impossibility of printing something sane (e.g. trait objects) or because I was unfamiliar with the internals and wasn't sure what made sense to stub out a `Debug` impl and what didn't.

That said, to get closure on this, I think its worth putting up a `#![deny(missing_debug_implementations)]` at the crate root and whitelisting any (current) deviants with `#[allow(missing_debug_implementations)]`. That way new additions will be `Debug` from the get go, and other gaps can be addressed when needed.

alexcrichton:
@ipetkov oh right good point. I think I cleaned up everything else in https://github.com/alexcrichton/futures-rs/commit/202c9b8b894ea220d578843f6daa602406a72916

ipetkov:
@alexcrichton looks good to me, thanks 👍

PR #392, opened 2017-02-18
Implement `Debug` on a number of public items

ipetkov:
Implement `Debug` where possible for "simple" cases. The remaining public items that are not `Debug` transitively depend on `Task` implementing debug, which isn't possible due to various trait objects. Still, we could eventually implement a stub `Debug` for `Task` to allow upper layers to become `Debug` as well.

Partially addresses #320


alexcrichton:
Thanks!

Issue #390, opened 2017-02-17
Consider adding `Stream::collect_bytes`


alexcrichton:
Some I/O applications are starting to represent streaming bodies as a `Stream` where each item is a chunk of bytes. For convenience, though, it might be nice to have a combinator such as:

```rust
trait Stream {
    fn collect_bytes(self) -> impl Future<Item = Vec<u8>, Error = Self::Error>
        where Self::Item: AsRef<[u8]>,
    { /* ... */ }
}
```

luser:
The explicit motivator for this was [`hyper::Body`](https://hyper.rs/hyper/master/hyper/struct.Body.html), which is a `Stream` with `Item = [hyper::Chunk](https://hyper.rs/hyper/master/hyper/struct.Chunk.html)` where `hyper::Chunk` is `Deref<[u8]>` and `AsRef<[u8]>`. I can imagine that working with `Stream`s whose items are effectively byte slices is going to be pretty common for users of tokio, and often it's necessary to process the entire stream contents once available.

carllerche:
I wonder if this should go on `Stream` itself, or perhaps there should be a separate (sub?) trait that has additional features specifically for streams of byte chunks.

For example, there could be additional fns providing the total size of the payload (if known), and there could be more than helper function, `collect_bytes` could allocate a buffer internally, `collect_bytes_into` (naming tbd) could take an existing buffer, etc...

Perhaps this functionality should live outside of the `futures` crate.

seanmonstar:
It being tied directly to "bytes" sounds odd in the futures crate. It could be generic, just played with this in a playpen, just a minimal version that isn't yet async:

```rust
trait Stream {
    // ...

    // name as 'concat' isn't important part
    fn concat<T, I>(&mut self, fill: T) -> SomeReturnTypeThatImplsStream
    where Self::Item: IntoIterator<Item=I>,
          T: Extend<I>,

    {
        // ...
    } 
}
```

This works to fill up a `Vec<u8>` or other container when receiving a stream of `Vec`s (or stream of `IntoIterator<Item=u8>`, or a `Vec<Entry>` when receiving a few entries at a time. Or a HashMap, etc.

stuhood:
It would be a true test of specialization if a completely generic form of `flatten` worked here.

seanmonstar:
[...]

PR #394, opened 2017-02-20
add Stream::concat

seanmonstar:
A possible solution to #390. Another name I considered was `extend`, but maybe that could be confused with some how extending the actual stream.

### DIFF
--- a/src/stream/concat.rs
+++ b/src/stream/concat.rs
@@ -1,0 +1,81 @@
+use core::mem;
+
+use {Poll, Async};
+use future::Future;
+use stream::Stream;
+
+/// A stream combinator to concatenate the results of a stream into the first
+/// yielded item.
+///
+/// This structure is produced by the `Stream::concat` method.
+#[derive(Debug)]
+#[must_use = "streams do nothing unless polled"]
+pub struct Concat<S>
+    where S: Stream,
+{
+    stream: S,
+    extend: Inner<S::Item>,
+}
+
+pub fn new<S>(s: S) -> Concat<S>
+    where S: Stream,
+          S::Item: Extend<<S::Item as IntoIterator>::Item> + IntoIterator,
+{
+    Concat {
+        stream: s,
+        extend: Inner::First,
+    }
+}
+
+impl<S> Future for Concat<S>
+    where S: Stream,
+          S::Item: Extend<<S::Item as IntoIterator>::Item> + IntoIterator,
+
+{
+    type Item = S::Item;
+    type Error = S::Error;
+
+    fn poll(&mut self) -> Poll<Self::Item, Self::Error> {
+        loop {
+            match self.stream.poll() {
+                Ok(Async::Ready(Some(i))) => {
+                    match self.extend {
+                        Inner::First => {
+                            self.extend = Inner::Extending(i);
+                        },
+                        Inner::Extending(ref mut e) => {
+                            e.extend(i);
+                        },
+                        Inner::Done => unreachable!(),
+                    }
+                },
+                Ok(Async::Ready(None)) => return Ok(Async::Ready(expect(self.extend.take()))),
+                Ok(Async::NotReady) => return Ok(Async::NotReady),
+                Err(e) => {
+                    self.extend.take();
+                    return Err(e)
+                }
+            }
+        }
+    }
+}
+
+#[derive(Debug)]
+enum Inner<E> {
+    First,
+    Extending(E),
+    Done,
+}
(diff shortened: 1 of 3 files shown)


carllerche:
Given that the container is provided, I would suggest that `extend` is not the correct fn name.

It seems that this could be better functionality provided as a separate function? Something like:

```rust
future::extend(my_buffer, my_stream) -> impl Future<Item = MyBuffer>`
```

alexcrichton:
Yeah I'm not totally sure on the naming here, I typically expect "concatenate" to not take a buffer as an argument and maybe even take a separator, but it does neither here! I sorta like the name 'extend' actually as you're extending a provided buffer.

@carllerche I'd prefer a trait method for this as it's a bit more ergonomic to call (one less import), but do you think it fits more with the other free-function combinators that we have?

carllerche:
I generally prefer mirroring std as much as possible. In this case, the goal is to provide an equivalent to the Extend trait for futures. That would mean creating a new trait: FutureExtend or something like that. 

I also prefer to not use a few traits (Future, Stream) as essentially a flat namespace for all functions. It hurts discoverability. 

I do think that before commiting to an API it would be worth exploring how specialization could potentially handle the case of the Stream of bytes knowing the final lengh.

alexcrichton:
@carllerche I'm not sure I follow the point about `FutureExtend`? With this use case at least it seems like there's not much of a use case for appending an element to a collection taking a long time (e.g. needing to itself return a future). I'd imagine that `Extend` (or some similar trait) ends up being fine in this case.

I agree that a flat namespace isn't great, but we also don't want to leave useful functionality for everyone to pull in an extra crate, so we need to make sure we've at least got the essentials in this crate.

seanmonstar:
As for the name `concat`, it came to mind because of my JavaScript background, where `Array.concat` takes any number of arrays and concatenates them all into a single array: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/concat

I can see `stream.extend(vec)` making sense, but I can also see someone reading it like `stream.extend(other_stream)` (if they hadn't checked the docs).

As for the failing tests, I noticed that `catch_unwind` was failing for me locally on nightly also. I haven't looked to see if this was a bug in nightly, or if futures needs to be updated.

carllerche:
After re-reading, I think that I am (mostly) happy w/ the PR. Maybe, the one thing I would suggest is since this is `concat`, to not pass in an argument and instead, extend onto the first value returned by the stream.

I *believe* that the `Concat` type could, in the future, be specialized to be smart and `reserve` the amount of space necessary if the stream can provide that info. For example a HTTP body stream impl could specialize the `Future::poll` implementation and reserve the amount of space specified by content-length.

So, given all that, w/ the change I suggested above, I'm 👍 

@alexcrichton Re my comment about `Extend`, it isn't actually applicable to this specific case because this is a "flat extend" (concat). (I was misunderstanding).

However, to elaborate. [`Extend`](https://doc.rust-lang.org/std/iter/trait.Extend.html) takes an `Iterator`. Given that `Stream` is the async version of `Iterator`, `FutureExtend` would take a stream, and extend self w/ the received values. It needs to be "async" since values are received from the stream "async".

seanmonstar:
I want to make the changes needed to get this done, but I admit I'm not sure what exactly they are. Sounds like Alex prefers `Extend`, Carl prefers `Concat`.

Additionally, there's a suggestion to not provide a target to fill up, but to make use of the first item in the stream. This could allow something neat, like if it is a `Stream` of `Bytes`, the `Bytes` could maybe just expand its length instead of copying the other `Bytes`. However, providing the target directly allows someone to `reserve` beforehand the size they expect (which is usually pretty easy to know in something like HTTP, by checking the `Content-Length` header first).

alexcrichton:
Oh I don't feel too strongly about `Extend`, `Concat` I could live with. @carllerche how do you feel about taking in an argument to do optional reservation if necessary?

carllerche:
If `concat` uses the first item, reservation could be done w/ `chain`:

```rust
let body = BytesMut::with_capacity(content_length)
   .chain(body_stream)
   .concat();
```

Reservation could also be done in the future w/ specialization.

Given those two options, I would opt for the more ergonomic API: not taking an argument.

seanmonstar:
Updated to make use of the first item from the stream, so it looks like `let body = stream.concat()`.

carllerche:
Seems good to me.

alexcrichton:
👍

PR #404, opened 2017-03-06
Add a first-class `Sink::close` method

alexcrichton:
This commit adds a new method to the `Sink` trait, `close`. The intention of
this method is to model closing a sink with an indication that no more messages
will be pushed into it. This indication can then be acted upon to do blocking
and/or other asynchronous work necessary to perform a clean shutdown.

For many protocols this is then followed by some subsequent work to shut down
the sink. For example TLS has a specific shutdown handshake which is applicable
for any TLS-backed `Sink` types.

The upcoming [tokio-io] crate recently [added] a `shutdown` method to the
`AsyncWrite` trait which this method is intended to connect up with. A call to
`Sink::close` will translate to an `AsyncWrite::shutdown` call.

Ideally this method would be a required method, but in the interest of not
breaking API compatibility just yet the method is introduced as a default method
on the `Sink` trait which calls `poll_complete`. It is strongly recommended,
however, that all implementations of `Sink` will implement this method.

More documentation about the method itself can be found in the commit with the
associated documentation.

### DIFF
--- a/src/sink/mod.rs
+++ b/src/sink/mod.rs
@@ -42,6 +42,10 @@ if_std! {
         fn poll_complete(&mut self) -> Poll<(), Self::SinkError> {
             Ok(::Async::Ready(()))
         }
+
+        fn close(&mut self) -> Poll<(), Self::SinkError> {
+            Ok(::Async::Ready(()))
+        }
     }
 
     /// A type alias for `Box<Stream + Send>`
@@ -59,6 +63,10 @@ if_std! {
 
         fn poll_complete(&mut self) -> Poll<(), Self::SinkError> {
             (**self).poll_complete()
+        }
+
+        fn close(&mut self) -> Poll<(), Self::SinkError> {
+            (**self).close()
         }
     }
 }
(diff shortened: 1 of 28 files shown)


src/sink/mod.rs
```diff
@@ -182,6 +190,80 @@ pub trait Sink {
...
+    /// to implement it whenever you implement `Sink` locally. It is especially
+    /// crucial to be sure to close inner sinks, if applicable.
+    #[cfg(feature = "with-deprecated")]
+    fn close(&mut self) -> Poll<(), Self::SinkError> {
+        self.poll_complete()
+    }
+
+    /// dox (you should see the above, not this)
+    #[cfg(not(feature = "with-deprecated"))]
+    fn close(&mut self) -> Poll<(), Self::SinkError>;
```
carllerche: Maybe elaborate a bit more about the contract around calling `start_send` / `poll_complete` after `close`. There is a note about panicking, but it seems like it would be good to elaborate a bit.

PR #408, opened 2017-03-07
remove (unused) log dependency

seanmonstar:
There was a warning as I built this crate that the `#[macro_use]` on the `log` crate was not used. It seems the log macros are no longer used at all, so futures can now have 0 dependencies.


alexcrichton:
👍

Issue #409, opened 2017-03-07
Consider adding a fn to `Sink` that can be used to check if the sink is ready to accept a value without actually sending one.


carllerche:
I propose `Sink::poll_ready` (in part so this issue shows up in searches).

I remember discussing this at the December 16 work week, but I don't entirely remember all the details. I didn't find another issue tracking this, so I am opening one (please close if this is a dup). This function can help in cases where it is useful to know if a `start_send` will succeed before creating the value. Imagine that storing the value is has a cost compared to generating & sending it immediately with no buffered storage.

Relates to #256.

alexcrichton:
I'd be ok adding this, but I think it's important to default to `Ready(())` to be ergonomic

carllerche:
Ok, so you have to help me remember the conclusion we reached in Hawaii.. maybe @aturon has notes!

So, it's important for poll_ready() in this case to not return any false positives (otherwise it defeats the purpose of avoiding storage).

I also agree that ergonomics are important, so maybe the conclusion was that this shouldn't be a fn on Sink?

Maybe it was a separate trade w/ some sort of combinator that provided poll_ready?

carllerche:
Basically, you can always make any `Sink` be able to provide poll_ready by wrapping it with a 1 slot buffer, but you want to avoid that storage in the case that the Sink knows the answer, like mpsc::Sender.

aturon:
Here's what I have in my notes:

- Add combinator for `poll_ready`, basically follow the fuse pattern, use the code in proto
  - This guarantees that if `poll_ready` succeeds then it will accept a message

carllerche:
Thanks... now hopefully that jogs someones memory :P

alexcrichton:
Hm so "like the fuse pattern" may imply:

* A separate `PollReady` trait
* A `Sink::poll_ready` *combinator* which returns a `SinkPollReady<T>`
* Implement a `SinkPollReady::poll_ready` function
* Specialize `SinkPollReady` to do nothing for types that implement `PollReady` and have a one-slot storage for all other types

(modulo all the naming conflicts I guess)

carllerche:
So, after talking a bunch about this in IRC, the conclusion is that this is tricky.

* `Sink` cannot have a `poll_ready` function because it doesn't make sense for all `Sink` implementations.
[...]

PR #414, opened 2017-03-14
Add `BufferOne` sink combinator

carllerche:
The decorator can buffer a single value without requiring any internal allocations. The type also provides `poll_ready`. I left out a `Sink::buffer_one` for this one in favor of `BufferOne::new`.

Relates to #409


alexcrichton:
I wonder if we could perhaps push on the `buffered` type first before this?  I think the semantics of `poll_ready` are implementable over there, right? We may also be able to special case w/ a "small vec" sort of

PR #417, opened 2017-03-16
Implement `Mutex<T>` syncronization primitive.

ticki:
`Mutex<T>` acts kind of like a classical mutex, except that it doesn't block: `lock()` produces a future, which completes, when the lock can be grabbed.

It very simple. It consists of the data and then an atomic boolean. As there is no spin logic, there is no need for things like `futex` calls.

### DIFF
--- a/src/sync/mod.rs
+++ b/src/sync/mod.rs
@@ -13,5 +13,7 @@
 pub mod oneshot;
 pub mod mpsc;
 mod bilock;
+mod mutex;
 
 pub use self::bilock::{BiLock, BiLockGuard, BiLockAcquire, BiLockAcquired};
+pub use self::mutex::{Mutex, MutexGuard, MutexFuture};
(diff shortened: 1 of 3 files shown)


ticki:
Welp. A timeout in some travis tests....

carllerche:
Thanks for the PR. Looking at the code, I don't think that the `poll` implementation satisfies the required guarantees of `Poll::poll`. Specifically, if `poll` returns `NotReady`, then the task **must** be notified when the state changes such that `poll` could succeed.

ticki:
@carllerche

> Thanks for the PR. Looking at the code, I don't think that the poll implementation satisfies the required guarantees of Poll::poll. Specifically, if poll returns NotReady, then the task must be notified when the state changes such that poll could succeed.

Could you elaborate on this. I am not sure I follow.

alexcrichton:
I'd recommend reading [the documentation of the `poll` function](https://docs.rs/futures/0.1.11/futures/future/trait.Future.html#tymethod.poll), the implementation here is missing a crucial aspect of that documentation, namely:

> When a future is not ready yet, the Async::NotReady value will be returned. In this situation the future will also register interest of the current task in the value being produced.

ticki:
rtfm, ticki.

PR #420, opened 2017-03-20
Proposal for a try_result macro.

partim:
This PR proposes a `try_result!()` macro that returns early on `Async::NotReady` and otherwise converts the inner expression into a `Result<T, E>`.

I relatively often have encountered cases where processing continues both on success and error but needs to halt if a poll is not ready. The macro would improve the ergonomics of these cases.

### DIFF
--- a/src/poll.rs
+++ b/src/poll.rs
@@ -11,6 +11,18 @@ macro_rules! try_ready {
     })
 }
 
+/// A macro for extracting a `Result<T, E>` from a `Poll<T, E>`.
+///
+/// The macro returns early if an expression is `NotReady`.
+#[macro_export]
+macro_rules! try_result {
+    ($e:expr) => (match $e {
+        Ok($crate::Async::Ready(t)) => Ok(t),
+        Ok($crate::Async::NotReady) => return Ok($crate::Async::NotReady),
+        Err(e) => Err(e),
+    })
+}
+
 /// Return type of the `Future::poll` method, indicates whether a future's value
 /// is ready or not.
 ///


alexcrichton:
Thanks for the PR! I'm somewhat wary to add another macro as I feel that we may have already blown our "one macro budget" on the `try_ready!` macro, unfortunately though. Do you find yourself commonly needing this? I forget how much it came up in this lib itself...

In general this is where I'd love for something like this to work:

```rust
let a: Result<T, E> = poll?;
let a: T = poll?;
```

Although I don't think that `?` allows us to do that :(

burdges:
Just curious : Was it ever decided what `?` should ideally mean for futures?

alexcrichton:
I believe the status quo is https://github.com/rust-lang/rfcs/pull/1859

partim:
As a data point, it looks like it came up about five times in my DNS resolver. I think it’ll be fairly often encountered when building combined futures manually with structs or enums. Admittedly, this is probably not the most common way of using futures. It also isn’t all that much of an issue, in many cases just an extra line in a match. It just bugs me esthetically.

carllerche:
One option would be to create a futures-macros crate or something w/ all of
these macros as a "nursery" to see how much the various macros end up being
helpful in practice :)

scottmcm:
I'd encourage people to post in [that RFC](https://github.com/rust-lang/rfcs/pull/1859) with futures-related thoughts.  It got bogged down in FCP, and new thoughts on what `Poll`s would want might just help unstick it.

For example, I made [a post](https://github.com/rust-lang/rfcs/pull/1859#issuecomment-286393549) that claimed `?` on poll should be `try_ready!` as an implication of the `Ok(Async::Ready(()))` problem, and you should definitely correct me if I'm wrong :smile:

Ralith:
As an additional data point, I find myself wanting this macro fairly often as well.

alexcrichton:
Ok there hasn't been much activity on this in awhile, so I'm going to close this. The good news though is that importable macros are nearing completion on their implementation and may hopefully stabilize soon!

PR #422, opened 2017-03-21
Made SendError Clone where possible.

46bit:
Hi. It's fairly crucial for [sirpent-team/comms](https://github.com/sirpent-team/comms) that Sink errors are Clone. When testing I've noticed that `mpsc::SendError` isn't Clone when it can be, so this adds that derivation.


alexcrichton:
Thanks!

PR #423, opened 2017-03-21
Stream::from_err and Sink::from_err combinators

46bit:
Hi! This is a very useful combinator on `Future` and so I've extended it to `Stream` and `Sink`.


alexcrichton:
Thanks!

PR #425, opened 2017-03-22
Use AtomicPtr instead of AtomicUsize in Stack

MaloJaffre:


alexcrichton:
Thanks for the PR! I'd imagine the motivation for a PR like this though is to reduce casts, but it looks like this increases the number of casts?

MaloJaffre:
The real motivation is to use an AtomicPtr because an AtomicUsize doesn't appears as a casted pointer, and doesn't indicates what it references (a `Node<T>`).

alexcrichton:
Sure yeah but I feel like this PR isn't actually leveraging the benefit of using `AtomicPtr`:

* the `_marker` field isn't removed
* Lots of `as` casts still exist

Can that be handled?

MaloJaffre:
Sure, it's now done.

src/stack.rs
```diff
@@ -32,18 +31,16 @@ unsafe impl<T: Sync> Sync for Drain<T> {}
...
+            head: AtomicPtr::default(),
         }
     }
 
     pub fn push(&self, data: T) {
-        let mut node = Box::new(Node { data: data, next: 0 as *mut _ });
+        let mut node = Box::new(Node { data: data, next: ptr::null_mut() });
         let mut head = self.head.load(SeqCst);
         loop {
             node.next = head as *mut _;
```
alexcrichton: I believe this `as` can be removed, right?

MaloJaffre: Oops, fixed.

alexcrichton:
Thanks!

PR #426, opened 2017-03-23
Added a `Sink` implementation to `stream::FromErr`.

46bit:
This support is very useful from my experience using the combinator with `Sink + Stream`.

It [simplifies](https://github.com/sirpent-team/comms/blob/master/src/client/mod.rs#L47-L49) a type from `
Client<I, Unsplit<SinkFromErr<SplitSink<C>, ErrorString>, FromErr<SplitStream<C>, ErrorString>>>` to `Client<I, FromErr<C>>`. Without this patch that'll be the general case of using it with, e.g., `TcpStream` - not pleasant at all.

I'm not sure if you want this in the library, so feel free to decline if you'd rather.

### DIFF
--- a/futures-cpupool/Cargo.toml
+++ b/futures-cpupool/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "futures-cpupool"
-version = "0.1.4"
+version = "0.1.5"
 authors = ["Alex Crichton <alex@alexcrichton.com>"]
 license = "MIT/Apache-2.0"
 repository = "https://github.com/alexcrichton/futures-rs"
--- a/src/sink/from_err.rs
+++ b/src/sink/from_err.rs
@@ -40,3 +40,12 @@ impl<S, E> Sink for SinkFromErr<S, E>
         self.sink.close().map_err(|e| e.into())
     }
 }
+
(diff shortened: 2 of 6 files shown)


alexcrichton:
Thanks for the PR! A number of other stream combinators also forward `Sink` implementations but they do so transparently. Perhaps this could transparently forward the `Sink` implementation (like other stream combinators) and then this could be achieved with `stream.from_err().sink_from_err()` perhaps?

46bit:
> forward Sink implementations

Aha, that's the terminology I was looking for.

I agree that would be a better implementation; I'll get it done :-)

Cheers,
Michael

alexcrichton:
I'm curious, did you agree that we should split this into `sink_from_err`? Or do you feel like it's best to just use this one combinator?

46bit:
Hey, sorry, this has been on my back burner. I definitely think the two combinators is the best way forward - larger type signatures, but much more flexible usage. I'll update this before go to sleep.

46bit:
I've updated this so that `stream::FromErr` and `sink::SinkFromErr preserve `Sink + Stream`, and added basic tests. There's no test for preserving `Sink + Stream` as I wasn't sure where in the testsuite it belongs.

alexcrichton:
Thanks!

PR #428, opened 2017-03-26
Use `debug_assert` for invariants.

leodasvacas:
Particularly in `Stream` combinators.


NeoLegends:
I'm not sure this is the right approach. Lowering asserts to debug-only reduces the guarantees / invariants for release code.

alexcrichton:
Did you find that this came up in profiles in release code?

leodasvacas:
@alexcrichton No, this is just speculation, it seems difficult to make this show up in a benchmark.

alexcrichton:
Ok, in that case I'm going to close as I'd prefer to not move these to debug asserts, but if they show up in profiles please let me know!

PR #431, opened 2017-03-27
Add newlines to docs (for correct rendering)

jaystrictor:
Without the newlines, the code appears inline and the language tag `rust` is rendered in the docs.


alexcrichton:
Oh dear it looks like this uncovered a bug as well! Mind fixing it while you're at it?

alexcrichton:
Thanks!

Issue #341, opened 2017-01-11
Take `&Arc<T>` instead of `Arc<T>` where possible


alexcrichton:
Most fast paths don't end up cloning arcs at all, so we shouldn't impose an extra clone by taking a reference in all sorts of methods.

This is a breaking change, so tagging for 0.2

alexcrichton:
This is now effectively done for non-deprecated functions

PR #433, opened 2017-03-28
Take `&Arc` instead of `Arc` in `poll_future` and `poll_stream`

leodasvacas:
Part of issue #341 (fixes it?). As can be seen in the wait methods this prevents one `Arc` clone for every polling after the first one. Did not benchmark but seems like a nice performance optimization. However this makes the syntax around polling complicated because one does not simply cast `&Arc<T>` to `&Arc<Trait>`.

Thanks @stephaneyfx for helping me with the trait object magic in IRC.

### DIFF
--- a/src/task_impl/mod.rs
+++ b/src/task_impl/mod.rs
@@ -333,8 +333,8 @@ impl<F: Future> Spawn<F> {
     /// scheduled to receive a notification when poll can be called again.
     /// Otherwise if `Ready` or `Err` is returned, the `Spawn` task can be
     /// safely destroyed.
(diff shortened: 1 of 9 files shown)


alexcrichton:
Looks great! This is a breaking change though so we'll hold off until 0.2

alexcrichton:
I think it's actually looking like we're going to deprecate these functions in https://github.com/alexcrichton/futures-rs/pull/436, so I'm going to close this.

alexcrichton:
(they'll just be removed for 0.2 anyway)

PR #434, opened 2017-03-29
executor::Spawn: add an optional park timeout for blocking methods

abonander:
This is useful for futures which want to be polled even without being able to move forward; otherwise, it would be necessary to spawn an extra thread to send notifications to the future. For my use-case, I want to animate a "Working..."  message on stdout while waiting for the completion of a future on a background thread, and polling on regular intervals seems to be the easiest way to go about this.

Since `thread::park()` can wake spuriously anyways, this doesn't seem to break any guarantees of the current API.

### DIFF
--- a/src/task_impl/mod.rs
+++ b/src/task_impl/mod.rs
@@ -6,6 +6,7 @@ use std::mem;
 use std::sync::Arc;
 use std::sync::atomic::{Ordering, AtomicBool, AtomicUsize, ATOMIC_USIZE_INIT};
 use std::thread;
+use std::time::Duration;
 
 use {Poll, Future, Async, Stream, Sink, StartSend, AsyncSink};
 use future::BoxFuture;
@@ -283,6 +284,7 @@ pub struct Spawn<T> {
     obj: T,
     id: usize,
     data: data::LocalMap,
+    park_timeout: Option<Duration>,
 }
 
 /// Spawns a new future, returning the fused future and task.
@@ -299,6 +301,7 @@ pub fn spawn<T>(obj: T) -> Spawn<T> {
         obj: obj,
         id: fresh_task_id(),
         data: data::local_map(),
+        park_timeout: None,
     }
 }
 
@@ -316,6 +319,15 @@ impl<T> Spawn<T> {
     /// Consume the Spawn, returning its inner object
     pub fn into_inner(self) -> T {
         self.obj
+    }
+
+    /// Set a timeout to use when parking the current thread in the `wait_*()` methods.
+    ///
+    /// Can be either `Duration` or `Option<Duration>`. Setting the timeout to `None`
+    /// means no timeout will be used.
+    pub fn park_timeout<D: Into<Option<Duration>>>(&mut self, timeout: D) -> &mut Self  {
+        self.park_timeout = timeout.into();
+        self
     }
 }
 
@@ -342,9 +354,10 @@ impl<F: Future> Spawn<F> {
     ///
     /// This function will call `poll_future` in a loop, waiting for the future
     /// to complete. When a future cannot make progress it will use
-    /// `thread::park` to block the current thread.
+    /// `thread::park`, or `thread::park_timeout()` if a timeout was set
+    /// with `park_timeout()`, to block the current thread.
     pub fn wait_future(&mut self) -> Result<F::Item, F::Error> {
-        let unpark = Arc::new(ThreadUnpark::new(thread::current()));
+        let unpark = Arc::new(ThreadUnpark::new(thread::current(), self.park_timeout));
         loop {
             match try!(self.poll_future(unpark.clone())) {
                 Async::NotReady => unpark.park(),
@@ -380,6 +393,7 @@ impl<F: Future> Spawn<F> {
                 id: self.id,
                 data: self.data,
                 obj: self.obj.boxed(),
+                park_timeout: self.park_timeout,
             },
             inner: Arc::new(Inner {
                 exec: exec,
@@ -399,7 +413,7 @@ impl<S: Stream> Spawn<S> {
     /// Like `wait_future`, except only waits for the next element to arrive on
     /// the underlying stream.
     pub fn wait_stream(&mut self) -> Option<Result<S::Item, S::Error>> {
-        let unpark = Arc::new(ThreadUnpark::new(thread::current()));
+        let unpark = Arc::new(ThreadUnpark::new(thread::current(), self.park_timeout));
         loop {
             match self.poll_stream(unpark.clone()) {
                 Ok(Async::NotReady) => unpark.park(),
@@ -439,7 +453,7 @@ impl<S: Sink> Spawn<S> {
     /// be blocked until it's able to send the value.
     pub fn wait_send(&mut self, mut value: S::SinkItem)
                      -> Result<(), S::SinkError> {
-        let unpark = Arc::new(ThreadUnpark::new(thread::current()));
+        let unpark = Arc::new(ThreadUnpark::new(thread::current(), self.park_timeout));
         let unpark2 = unpark.clone() as Arc<Unpark>;
         loop {
             value = match try!(self.start_send(value, &unpark2)) {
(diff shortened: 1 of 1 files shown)


alexcrichton:
Thanks for the PR! This wasn't quite how I envisioned timeout support but this seems like, but it seems nifty! 

I was thinking that when a timeout elapsed it'd be returned to the user instead of simply re-polling the future so the timeout could be acted on, but maybe that could be a suite of extra methods? I kinda like the ability to set the timeout globally for a `Spawn` and then other operations just fall out without extra apis.

I'm curious, do you think there's much of a use case for acting externally in the event of a timeout?

carllerche:
I would prefer keeping the timeout out of the struct. `Spawn` is a fundamental construct that and **must** be used by all schedulers. It would be better to keep it as small as possible IMO and provide an additional abstraction for convenience.

If there is a way to execute a future w/o using `Spawn`, let me know, because I think it is the only way to correctly setup the necessary thread locals.

abonander:
@carllerche there's not any additional cost to code that doesn't use timeouts so I don't see the issue of maintaining it in `Spawn`, unless you're really concerned about the additional size an `Option<Duration>` adds. 

Re: handling externally, I don't know of any use cases specifically but re-polling after the timeout is simple enough so it makes more sense. However, this should be in separate methods which can have return types to signify the timeout; the `Poll` result type works as the `NotReady` value could signify that the operation timed out. If we're adding separate methods then storing the timeout in `Spawn` becomes less obligatory.

carllerche:
I'm definitely concerned about the `Option<Duration>` storage space. A cache line is 64 bytes and this adds an extra 16, aka takes up 25% of the cache line for something that will be rarely used.

abonander:
So are we thinking a more traditional approach, i.e. adding methods that take a timeout as a parameter? I'm thinking:

* `wait_future_timeout`
* `wait_stream_timeout`
* `wait_send_timeout`
* `wait_flush_timeout`

Or something more like a wrapper struct that adds the timeout? This can be done in client code but it's a bit unwieldy with having to provide your own semi-redundant `Unpark` impl.

alexcrichton:
I think in the long run these are just going to be conveniences for what you can build externally, so I'm not too too worried about ergonomics here (they're already pretty bad with `Spawn`). 

In that sense I think basically adding `_timeout` methods is fine.

abonander:
@alexcrichton One concern with having the timeout methods meant to be called in a loop from the outside is that the `Arc<Unpark>` that they use will be getting repeatedly allocated and deallocated. There's not really a good way around this with the current design; exposing it to the user so they can hang onto it is little more gain than having them create this abstraction themselves.

I'm playing around with the idea of having a `WithTimeout` adapter, though, which gives us a little more flexibility. We can keep the `Arc<Unpark>` and potentially the timeout in the adapter while not making `Spawn` itself any bigger. It could also be generic over `BorrowMut<Spawn<_>>` so it can take ownership or mutably borrow.

There is an issue with it being sent across threads as it'll retain the wrong thread token to unpark, though this could be remedied by replacing the thread token if necessary--we can check equality by comparing the raw thread handles. Alternately, we could just opt-out of `Send` for the `WithTimeout` adapter, though it's not strictly necessary for thread safety.

carllerche:
Why not take `&Arc<Unpark>` instead?

abonander:
Most of the point of this was to avoid the user having to create their own `Unpark` impl which is redundant to the private one in the module. Otherwise there's no point over the user creating their own abstraction.

alexcrichton:
So one thing I'm not sure about is how this'll interact with various executors. Does the desire for timeouts only arise from blocking on futures? Or also during the execution of a future in a typical setting?

If it's only related to blocking on futures then we may wish to adapt `Stream::wait` and `Sink::wait` return values to have a configurable timeout. We could then add something like a `Future::waiting` method which returns an adapter which has a configurable timeout. (or something like that).

Does that make sense?

abonander:
> Does the desire for timeouts only arise from blocking on futures? Or also during the execution of a future in a typical setting?

For my use-case, I want the future to be re-polled even if it can't move forward. While futures *should* support being re-polled even if they can't move forward (they have to when working with `Spawn::wait_*()` since parked threads can wake spuriously), I don't know your opinion on whether or not they should be allowed to *assume* they'd be re-polled on a regular basis, since that does imply some coupling between the future and the executor. I mentioned my use-case in the OP--for that, coupling is fine because it's all in the same file anyway.

> If it's only related to [...]. Does that make sense?

It does, and that would work for my use-case, but since these methods already delegate to `Spawn`, what's the point in having a high-level abstraction without exposing the underlying mechanism? Both would be preferable, I think.

carllerche:
@abonander I agree that there should be a way to poll a future off task w/o passing an unpark handle, but I don't think `Spawn` would be the place to do that. I haven't put much thought in trying to figure it out though.

abonander:
That was my point in suggesting an adapter/wrapper for `Spawn` that holds the `Unpark` and can take timeouts. 

Alternately, though this perhaps isn't that good of an idea, we can expose the `ThreadUnpark` type that `Spawn` uses, maybe hiding the `thread::current()` in the constructor and also providing a constructor that directly returns an `Arc` just for a little more convenience. Then the solution to the timeout issue is as simple as adding a `park_timeout()` method. 

A potential issue, like with the previously suggested wrapper type, is that the user could move it to a different thread and then end up with a deadlock because the unparks are being sent to the wrong thread handle. We can't opt-out of `Send` here because it's required by `Unpark`, though. I don't think it's that big of an issue, a documentation warning should suffice, or maybe an assert in the parking methods that check that the stored thread handle is the current thread.

alexcrichton:
Ah yeah so I would imagine that if a future wants to be polled on a regular interval it would internally use a future to schedule such a notification, not relying on the executor (in the general abstract sense). This makes sense from an implementation perspective but not necessarily a compositional perspective I think?

abonander:
That works, and it's pretty much already been done in a third-party crate, but the issue is that it requires an extra thread to send the notification, or an OS message loop to use a timer primitive.

So unless you want to introduce an interface for a future to ask to be polled again on a timeout--which may not be a bad idea but may be complicated to implement and not supported by all event loops so it would need some kind of fallback--then supporting it in `Spawn` is probably the better way anyway, even if it's at the wrong level conceptually. 

What do you think of just exposing `ThreadUnpark` to provide a basic `Unpark` impl to client code, and adding a `park_timeout()` method? It would save some boilerplate for users that want to do their own thing with `Spawn`.

abonander:
I've now realized that `tokio-core` basically has everything I want already, but I'd consider it pretty heavyweight for my use case since I'm not doing anything with evented I/O.

alexcrichton:
@abonander oh right yeah I wouldn't want this to *require* an extra thread, but I'd imagine that the support could be implemented in the same fashion that tokio-core does today (just without I/O). Put another way I'd imagine that most executors would provide the ability to have some form of timeout, it's just that the source of the timeout may vary across exeuctors. 

We could expose `ThreadUnpark` yeah, but I fear that it wouldn't solve much of the problem here. Once you reach down into the `futures::executor` module you're already pretty much in the weeds and I'd imagine that not having to reimplement `ThreadUnpark` would be the least of your worries :(

abonander:
So are you thinking something more like lifting the timeout support from tokio-core itself? That could be an option.

alexcrichton:
Yeah so I've recently wanted to be able to do that in the `futures` crate itself. Basically support the `Timeout` and `Interval` types in `futuers` instead of `tokio-core`. That implies, however, that all executors are able to support timeouts, which is not required of executors today. As a result I think it'd be a breaking change, so I haven't thought through it too much.

In general though, yes. I'd expect this problem to be solved with types like `Timeout` and `Interval`, regardless of whether they come from tokio-core or another executor.

abonander:
Okay, closing this then if you'd prefer to go in that sort-of direction for this use case.

alexcrichton:
Ok. If that ends up being too onerous to implement or doesn't work well locally, I don't mind revisiting this!

Issue #432, opened 2017-03-27
Consider switching Arc<Unpark> to an opaque type


carllerche:
This issue is mostly a placeholder for a discussion leading up to 0.2. 

Currently, Arc<Unpark> is used to pass around the unpark handle. This API has a hard requirement on Arc which means that an allocation is required even if it isn't strictly necessary.

I propose switching that to an opaque struct, something like UnparkHandle which would allow the internal representation to change in the future. 

There could be a conversion from Arc<Unpark> as I think we would need to maintain that indefinetly. 

Relates to #430

alexcrichton:
I'd be surprised if we could do this, but I'd be willing to try!

alexcrichton:
Done in #436!

Issue #360, opened 2017-01-16
RFC: Customizable `wait()` behavior


nikomatsakis:
The existing `wait()` method *always* blocks the current thread. However, as the docs themselves state, this is very often not the thing you want to do:

> Note: This method is not appropriate to call on event loops or similar I/O situations because it will prevent the event loop from making progress (this blocks the thread). This method should only be called when it's guaranteed that the blocking work associated with this future will be completed by another thread.

In Rayon, I made the `wait()` method (when invoked on a `RayonFuture`, the only kind I have control of) simply `panic()` if it is used on a Rayon thread. I then added a second method, `rayon_wait()`, that will interoperate properly with Rayon's scheduler.

It seems to me it would make sense for the futures library to offer some kind of thread-local interface to customize what `wait()` does. The default would no doubt be the current behavior, but executors like Rayon could -- for their threads, or within their scope -- override this by setting an appropriate thread-local callback.

The main counterargument I see is that this might encourage people to write code that uses `wait()`, which I would consider largely an anti-pattern. The better thing to do is to create (and schedule) a derived future. In this case, I have to wonder if including `wait()` in the Futures interface *at all* makes sense, though.

alexcrichton:
Makes sense to me! You could even imagine that this feature could be leveraged to implement coroutines-as-futures where `.wait()` just does the context switch away if it's not ready yet.

alexcrichton:
This relates to some of the discussion @carllerche @aturon and I had about https://github.com/tokio-rs/tokio-core/issues/79 as well. I think this would also be quite useful for tokio-core clients. What I'm thinking is:

* At the beginning of your thread you create `Core` and customize the `wait` behavior to use this core.
* Otherwise in your program you just call `wait`. 

The calls to `wait` would run the event loop on the current thread, waiting for the passed in future to resolve.

nikomatsakis:
@alexcrichton 

> Otherwise in your program you just call wait.

By "otherwise", do you mean "elsewhere"? i.e., once you have created a `Core` to customize the wait behavior, you can just call `wait`? If so, it seems reasonable, though I worry about footguns if users forget to create a `Core`. But I'm lacking context.

alexcrichton:
Ah yes yeah my command of the English language isn't always there... 

[...]

Issue #312, opened 2016-12-28
Consider renaming task park / unpark


carllerche:
Placeholder issue for an 0.2 discussion.

Originally, the names were picked for symmetry with thread park / unpark. However, the behavior is sufficiently difficult that I believe the naming is leading to unnecessary confusion (this is also based off of observing questions in Gitter).

leodasvacas:
To unpark a thread means immediately wake it, to unpark a task it usually means to schedule it to be polled. I'd suggest park -> suspend/delay/pause and unpark -> schedule. Since future is a time metaphor it seems helpful to use other time metaphors.

carllerche:
> To unpark a thread means immediately wake it

That's not strictly true. unparking a thread has no guarantees in terms of **when** the thread gets scheduled to execute. Just like w/ futures, the OS scheduler is free to schedule the thread as appropriate.

carllerche:
I propose either:

* `task::current()` & `Task::notify()`
* Leave it as is.

Now is the time to make the change (#436). It should either happen or we leave the naming as is.

ipetkov:
+1 on `current`/`notify`. It always bothered me that `park` implied side effects when it had none

alexcrichton:
Done! (#436)

PR #436, opened 2017-03-31
Task system overhaul

carllerche:
## Updated description

The primary change made in this PR is to restructure memory management and
notifications throughout the "task system" in the `futures` crate. It is
intended that this will have very little impact, if any, on consumers of
futures. Implementations of runtimes of futures (e.g. crates like tokio-core)
are the target for this series of changes, enabling a suite of possible
optimizations that were not previously feasible. One major use case that is 
now enabled is usage of the `task` and `executor` modules in the `no_std` ecosystem.
This means that bare-metal applications of futures should be able to use the same
task system that the std-based futures ecosystem uses.

One of the largest changes being made to support this is an update to the memory management of
objects behind `Task` handles. Previously it was required that `Arc<Unpark>`
instances were passed into the various `Spawn::poll_*` functions, but new
functions `Spawn::poll_*_notify` were added which operate with a `NotifyHandle`
instead. A `NotifyHandle` is conceptually very similar to an `Arc<Unpark>`
instance, but it works through an `unsafe trait UnsafeNotify` to manage memory
instead of requiring `Arc` is used. You can still use `Arc` safely, however, if
you'd like.

In addition to supporting more forms of memory management, the `poll_*_notify`
functions also take a new `id` parameter. This parameter is intended to be an
opaque bag-of-bits to the `futures` crate itself but runtimes can use this to
identify the future being notified. This is intended to enable situations where
the same instance of a `NotifyHandle` can be used for all futures executed by
using the `id` field to distinguish *which* future is ready when it gets
notified.

## API Additions

* A `FuturesUnordered::push` method was added and the `FuturesUnordered` type
  itself was completely rewritten to efficiently track a large number of
  futures.
* A `Task::will_notify_current` method was added with a slightly different
  implementation than `Task::is_current` but with stronger guarantees and
  documentation wording about its purpose.

## Compatibility Notes

As with all 0.1.x releases this PR is intended to be 100% backwards compatible.
All code that previously compiled should continue to do so with these changes.
As with other changes, though, there are also some updates to be aware of:

* The `task::park` function has been renamed to `task::current`.
* The `Task::unpark` function has been renamed to `Task::notify`, and in general
  terminology around "unpark" has shifted to terminology around "notify"
* The `Unpark` trait has been deprecated in favor of the `Notify` trait
  mentioned above.
* The `UnparkEvent` structure has been deprecated. It currently should perform
  the same as it used to, but it's planned that in a future 0.1.x release the
  performance will regress for crates that have not transitioned away. The
  primary primitive to replace this is the addition of a `push` function on the
  `FuturesUnordered` type. If this does not help implement your use case though,
  please let us know!
* The `Task::is_current` method is now deprecated, and you likely want to use
  `Task::will_notify_current` instead, but let us know if this doesn't suffice!

## Original description

**This PR is a work in progress**

I'm submitting this PR early to hopefully try to illustrate my thoughts and get some early feedback.

# Checklist

* [x] Land #442 
* [x] Update tokio-core
* [x] Run sscache tests using new task system
* [x] Switch `Arc<Unpark>` to `UnparkHandle` #432
* [x] Decide on #312 (leaning towards yes)
* [x] Allow executors to customize wait behavior #360 (deferring this until later)
* [x] Fix `Task::is_current`
* [x] Remove `Notify::is_current`, I don't think this is needed anymore.
* [x] Consider `GetNotifyHandle` https://github.com/alexcrichton/futures-rs/issues/129
* [x] Should `ref_inc` -> `ref_dec` be moved to `UnsafeNotify` (@alexcrichton says no).
* [x] Consider getting rid of `poll_*_notify` on `Stream` and `Sink`. Also, maybe name it `poll_notify` if it is only for `Future`.
* [x] Merge https://github.com/carllerche/futures-rs/pull/4
* [x] `u64` vs. `usize`

# Overview

The previous implementation of the task system required a number of allocations per task instance. Each task required a dedicated `Arc<Unpark>` handle which means that executors require at least two allocations per task.

Things get worse when using `with_unpark_event` as nested calls to `with_unpark_event` result in Vec allocation and cloning during each call to `task::park`.

This commit provides an overhaul to the task system to work around these problems. The `Unpark` trait is changed so that only one instance is required per executor. In order to identify which task is being unparked, `Unpark::unpark` takes an `unpark_id: u64` argument.

`with_unpark_event` is removed in favor of `UnparkContext` which satisfies a similar end goal, but requires only a single allocation per lifetime of the `UnparkContext`.

# The new `Unpark` trait

In general, tasks are driven to completion by executors and executors are able to handle large number of tasks. As such, the `Unpark` trait has been tweaked to require a single allocation per executor instead of one per task. The idea is that the executor creates one `Arc<Unpark>` handle and uses the `unpark_id: u64` to identify which task is unparked.

In the case of tokio-core, each task is stored in a slab and the the `unpark_id` is the slab index. Now, given that an Arc is no longer used, it can be that a slab slot is released and repurposed for a different task while there are still outstanding Task handles referencing the now released task.

There are two potential ways to deal with this.

a) Not care. Futures need to be able to handle spurious wake ups already. Spurious wake ups can be reduced by splitting the `u64` into 28 bits for the slab offset and use the rest of the u64 as a slot usage counter.

b) Use `Unpark::ref_inc` and `Unpark::ref_dec` to allow the executor implementation to handle its own reference counting.

Option b) would allow an executor implementation to store a pointer as the `unpark_id` and the `ref_inc` and `ref_dec` allow for atomic reference counting. This could be used in cases where using a slab is not an option.

# `UnparkContext`

This is quite similar in spirit to `with_unpark_event` except it requires significantly less allocations when used in a nested situation.

It does have some different behavior. Given the following:

```rust
// Currently in task A
let my_context = UnparkContext::new(my_unpark);

my_context.with(0, || {
    let task = task::park();

    thread::spawn(move || {
        thread::sleep(a_few_secs);
        task.unpark();
    });
});

my_executor.spawn(move || {
    // Currently in task B

    my_context.with(0, || {
        // some work here
    });
});
```

Task B will be the root task that is notified.


carllerche:
I should elaborate a bit.

This change should also enable to implement a tokio-core server that can run without allocations. Given that `UnparkContext` is now pretty lightweight, it seems reasonable for a server task to store a slab itself for the futures handling each accepted connection. In this case, the only allocations would be for the task handling the `TcpListener` and a pre-allocated slab for futures handling client sockets. The task then uses `UnparkContext` to either accept from the listener or poll a socket connection task.

carllerche:
If this ends up being a direction that we want to pursue, I think that it would be safe to keep this in a `task2` module. I'm fairly confident that the existing task API (`task` module) could use `task2` under the hood.

carllerche:
This PR is related to #432. If this strategy is taken, I don't think #432 is relevant anymore.

alexcrichton:
This looks pretty slick, I really like it how it's making `with_unpark_event` essentially free (which is not the case today). 

Your code snippet of how behavior changes was intially troubling to me, but in practice I don't think it will matter. Each `UnparkContext` will typically be associated with a particular future, which means that if you call `with` you're typically in a `Future::poll`. In that case you always want to override any previously stored task.

So to make sure I get this straight in my head:

* All top-level executors assign ids to futures.
* Top-level executors create **one** `Arc<Unpark>` for the entire duration of that executor.
* Code currently using `with_unpark_event` will now instead use `UnparkContext`
* `UnparkContext` requires on allocation at creation, and no more afterwards
* Code uses `UnparkContext::with` to receive a custom notification with a custom id whenever something is unparked, allowing it to arrange for it to learn later about precisely which ids were unparked.

Does that sound like a correct interpretation? I can't really see many downsides to this!

Some other random thoughts:

* This is essentially creating a linked list of `Arc`, right? I'd want to think careful about cycles here to see if they're possible, and if necessary we may wish to use weak pointers in some locations.
* Can you explain in more detail how this would work in tokio-core? Currently we create a `SetReadiness` for each future, but would we perhaps no longer do that and just have a linked list of "ready futures"?

carllerche:
Re: cycles, I don't think they are possible. `UnparkContext::with` takes `&mut self` and a closure, which prevents nesting `UnparkContext::with` calls w/ the same `self`. Only currently running tasks can be linked together... so it should be fine.

Re: how to hook this into tokio-core, you could most likely get rid of the `SetReadiness` in favor of a linked list of readiness. The slab in -core that tracks tasks could have have a `next_readiness: Option<usize>` for each entry..

alexcrichton:
Ok yeah, that both makes sense to me.

Before merging I'd want to see tokio-core working (ideally something like sccache compiling llvm) just to be sure though.

carllerche:
Seems fine to me, there is still work to be done though!

carllerche:
Ok, I have a tweaked proposal.

* Not introduce `task_impl2` (as requested by @alexcrichton).
* Deprecate the `Unpark` trait.
* Introduce the `Notify` trait:

```rust
pub trait Notify: Send + Sync {
    /// Indicates that an associated future and/or task are ready to make
    /// progress.
    ///
    /// Typically this means that the receiver of the notification should
    /// arrange for the future to get poll'd in a prompt fashion.
    fn notify(&self, id: u64);

    /// Called when an instance of `NotifyHandle` has been cloned.
    fn ref_inc(&self);

    /// Called when an instance of `NotifyHandle` has been dropped.
    fn ref_dec(&self);

    /// A new `Task` handle referencing `id` has been created.
    fn id_ref_inc(&self, id: u64);

    /// A `Task` handle referencing `id` has been dropped.
    fn id_ref_dec(&self, id: u64);
}
```

* Instead of `Arc<Notify>` (or `Arc<Unpark>`), introduce `NotifyHandle` as a wrapper struct:

```rust
struct NotifyHandle {
    ptr: *const NotifyHandle,
}

impl NotifyHandle {
    unsafe pub fn from_ptr(ptr: *const NotifyHandle) -> NotifyHandle {
        unimplemented!();
    }

    pub fn noop() -> NotifyHandle {
        unimplemented!();
    }

    pub fn notify_fn<F: Fn() + Sync + Send>(f: F) -> NotifyHandle {
        unimplemented!();
    }
}
```

* `Rename `task::park()` -> `task::current()` and `Task::unpark()` -> `Task::notify()` (#312 after discussing w/ a number of users, I think this is much clearer naming).

On one hand, `Notify` is a much more complex trait to implement... and very unsafe. However, the `Notify` trait would be able to support executors that run without `std`. `NotifyHandle` would provide a couple of additional constructors (`noop`, `notify_fn`) to assist in constructing "simple" `Notify` implementations.

Another option would be to provide `trait NotifySimple` which is safe and has fewer methods to implement. I prefer sticking with one trait `Notify` as increasing the number of traits increases the complexity of both decisions that users have to make + the implementation.

The one unknown here is how to add `Notify` support to `Spawn`. The easiest option would be to deprecate all existing `poll_*` fns and introduce new ones with a `_notify` syntax.

IMO, `poll_future()` (and co.) should be repurposed in 0.2 to be:

```rust
fn poll_future(&mut self) -> Poll<F::Item, F::Error> {
    self.poll_future_notify(&NotifyHandle::noop())
}
```

carllerche:
So, it seems that `NotifyContext` has a relatively big flaw when it comes to nesting them... this will take some thinking.

carllerche:
One thought:

Perhaps `Notify::ref_inc` and `Notify::ref_dec` should be moved to `UnsafeNotify`. These functions are probably going to be implemented using unsafe code (like transmuting a ptr to u64). I'm not sure having a default empty impl is great.

carllerche:
Regarding #360 (customizable wait), do you (@alexcrichton) have any thoughts on how you would implement this? It seems like one way would be to add a `wait` fn on `Notify`. However, I think that you originally suggested making it a fn passed to `poll_future`. I *think* this would be the better strategy as the wait fn could be stored on `BorrowedTask` and doing so would not block this PR.

If that's what you were thinking, then we can check off the item in the list.

carllerche:
It looks like there is also a bunch of other cleanup to do.

Is the `executor` vs. `task` module split still wanted (I don't have a huge opinion). The reason I ask is that it looks like the PR as it stands moves some fns around and I don't know if that is deliberate or accidental. For example `executor::spawn` -> `task::spawn`.

alexcrichton:
Ok I got around to sketching out what I was thinking for #360: https://github.com/alexcrichton/futures-rs/commit/390f34bcd9617b658facae7b501d5d63b6e0aef2

I'm not super happy about `with_executor` as I was hoping we could avoid a closure, but other solutions I thought of always had hidden panics you had to be aware of. I'm relatively confident this'll cover tokio-core's use case and I *think* it'll cover rayon. My thinking is that you'd have something like:

```rust
extern crate tokio_core;

fn main() {
    tokio_core::run(|handle| {
        let future = ...;
        let result = future.wait();
        println!("{:?}", result);
    });
}
```

carllerche:
@alexcrichton Thanks. I looked over really quick. As far as I can tell, this doesn't touch anything that would need to be bundled up w/ this PR. So, we can safely punt this work until later?

alexcrichton:
Agreed with that yeah, I'll check off the box.

alexcrichton:
Ok I'm successfully compiling LLVM with sccache *only* updating this crate. In other words I think our backcompat story here is solid, looks like nothing broke (yay!).

Gonna update tokio-core itself to use the new task system and make sure that works too now.

alexcrichton:
Ok sccache with tokio-core updated also works. I'll do some quick benchmarking now to make sure we didn't regress.

alexcrichton:
No regression on the tokio-minihttp throughput at least

alexcrichton:
Added a checkbox for https://github.com/alexcrichton/futures-rs/issues/129

alexcrichton:
Ok I've merged in #129 with `T: Clone + Into<NotifyHandle>`

carllerche:
@alexcrichton I was thinking ditching `mio::SetReadiness` completely at some point to take advantage of this PR.

carllerche:
Basically, the futures-rs task infrastructure would enable more efficient integration w/ tokio-core's reactor than mio can provide (as it has less to build on).

carllerche:
The tokio-core PR looks good as an initial "get off the deprecated APIs" pass 👍

alexcrichton:
Ok I've pushed up a commit now as well to implement `is_current`, deprecate it, and at the same time add a new `will_unpark_current` method.

carllerche:
What's left for this?

alexcrichton:
I think to merge we need an up-to-date summary of what this PR does, how it changes things, and why the changes are all justified. Basically just a correct PR description and/or correct commit message.

alexcrichton:
Other than that though I believe I'm comfortable landing everything here. I can try to get around to the docs next week if it's not done already.

alexcrichton:
Ok I started writing up documentation for this but ended up finding what I believe is some [lingering unsafety and memory leaks](https://github.com/carllerche/futures-rs/pull/4).

leodasvacas:
If `drop_raw` takes `&self` now (I think it was `&mut self` at some point) then should we change `NotifyHandle::new` to take a `*const UnsafeNotify` rather than `*mut UnsafeNotify`?

leodasvacas:
Why is `id` an `u64` rather than `usize`? Given that a lot of this is for `no_std` compatibility, isn't `usize` better for compatibility and performance on non-64bit architectures?

alexcrichton:
@leodasvacas the intention is to take `*mut Self` in `drop_raw` but we're not currently allowed to do that for object safety. The `&self` pointer has "less guarantees" in terms of uniqueness and such that the compiler knows about, so we're currently using `&self` instead of `&mut self`.

@carllerche can speak more though to the choice of `u64` instead of `usize`

carllerche:
I opted for `u64` instead of `usize` specifically because the extra bits available w/ 64 are very useful when implementing things like a treiber stack (extra bits are used to guard against ABA).

Picking u64 definitely gears towards optimizing for 64 bit platforms...

I'm not sure how common it is to have 32 bit platforms and care about the extra perf hit.

THAT said, it occurs to me that hardcoding to u64 would be a hazard if there ever are 128bit CPUs... though I don't know how realistic that is :) (especially since ptrs these days are limited to 48bit).

If there is strong preference to switching to `usize` and a good argument for it, I would be fine to do so.

carllerche:
@alexcrichton said he wanted to keep the various `_notify` functions on `Spawn` for consistency reasons. They have a low risk of naming conflict and could be easily deprecated later if needed.

alexcrichton:
Ok here's my updated summary of the PR: @carllerche mind checking?

---

## Major Changes

The primary change made in this PR is to restructure memory management and
notifications throughout the "task system" in the `futures` crate. It is
intended that this will have very little impact, if any, on consumers of
futures. Implementations of runtimes of futures (e.g. crates like tokio-core)
are the target for this series of changes, enabling a suite of possible
optimizations that were not previously feasible.

One of the largest changes being made is an update to the memory management of
objects behind `Task` handles. Previously it was required that `Arc<Unpark>`
instances were passed into the various `Spawn::poll_*` functions, but new
functions `Spawn::poll_*_notify` were added which operate with a `NotifyHandle`
instead. A `NotifyHandle` is conceptually very similar to an `Arc<Unpark>`
instance, but it works through an `unsafe trait UnsafeNotify` to manage memory
instead of requiring `Arc` is used. You can still use `Arc` safely, however, if
you'd like.

In addition to supporting more forms of memory management, the `poll_*_notify`
functions also take a new `id` parameter. This parameter is intended to be an
opaque bag-of-bits to the `futures` crate itself but runtimes can use this to
identify the future being notified. This is intended to enable situations where
the same instance of a `NotifyHandle` can be used for all futures executed by
using the `id` field to distinguish *which* future is ready when it gets
notified.

## API Additions

* A `FuturesUnordered::push` method was added and the `FuturesUnordered` type
  itself was completely rewritten to efficiently track a large number of
  futures.
* A `Task::will_notify_current` method was added with a slightly different
  implementation than `Task::is_current` but with stronger guarantees and
  documentation wording about its purpose.

## Compatibility Notes

As with all 0.1.x releases this PR is intended to be 100% backwards compatible.
All code that previously compiled should continue to do so with these changes.
As with other changes, though, there are also some updates to be aware of:

* The `task::park` function has been renamed to `task::current`.
* The `Task::unpark` function has been renamed to `Task::notify`, and in general
  terminology around "unpark" has shifted to terminology around "notify"
* The `Unpark` trait has been deprecated in favor of the `Notify` trait
  mentioned above.
* The `UnparkEvent` structure has been deprecated. It currently should perform
  the same as it used to, but it's planned that in a future 0.1.x release the
  performance will regress for crates that have not transitioned away. The
  primary primitive to replace this is the addition of a `push` function on the
  `FuturesUnordered` type. If this does not help implement your use case though,
  please let us know!
* The `Task::is_current` method is now deprecated, and you likely want to use
  `Task::will_notify_current` instead, but let us know if this doesn't suffice!

carllerche:
@alexcrichton seems good to me.

alexcrichton:
Alright after some discussion with @aturon I reached the conclusion that I'd personally at least be much more comfortable landing this if we could achieve one of the original goals (`no_std` usage or a better unpark event). Along those lines I've just [pushed up a commit](https://github.com/alexcrichton/futures-rs/pull/436/commits/4d3d46bb428d37235ed0e1c2b6ca0c3f8fc237a9) which enables usage of the entire non-deprecated surface area of the task system in `no_std` contexts. 

The crux here is that no_std contexts must explicitly initialize the task system with a getter/setter for thread-local-storage-like information. This is lazily provided when running in the context of libstd but when disabling the default features of this crate it's required for usage.

carllerche:
As far as I know, everything that is needed for this PR has happened...

alexcrichton:
Alright, let's do it!

PR #441, opened 2017-04-04
Add Future::map_result function

bennofs:
I have needed this function multiple times in my own code. Right now, the
functionality can be emulated by:

```
f.and_then(|v| future::future_result({
   // code
}))
```

but this does not play nicely with `try!/?`. Using `map_result`, you can write:

```
f.map_result(|v| {
  let x = try!(...);
  let y = try!(...);
  Ok(x + y)
})
```

which is much nicer than the alternatives currently possible.

### DIFF
--- a/src/future/map_result.rs
+++ b/src/future/map_result.rs
@@ -1,0 +1,38 @@
+use {Future, Poll, Async};
+
+/// Future for the `map_result` combinator, changing the type of a future.
+///
+/// This is created by the `Future::map_result` method.
+#[derive(Debug)]
+#[must_use = "futures do nothing unless polled"]
+pub struct MapResult<A, F> where A: Future {
+    future: A,
+    f: Option<F>,
+}
+
+pub fn new<A, F>(future: A, f: F) -> MapResult<A, F>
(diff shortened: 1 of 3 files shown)


alexcrichton:
Thanks for the PR! In theory though the `.then` combinator works for this purpose as well, although it doesn't provide as much guidance to the compiler about type inference. Do you think it's still worth having the dedicated method though?

bennofs:
@alexcrichton Oh, because of the `IntoFuture` impl for `Result<T, E>`? That was less than obvious to me. If we decide against this, then we should at least add a note that `then` may be used this way.

One place where a separate `.map_result` is useful is if you want to destructure the argument, like:

```
fut.and_result(|(x,y,z)| { ... })
```

with then, you'd need to write:

```
fut.then(|r| {
    let (x,y,z) = try!(r);
    ...
})
```

bennofs:
Oh wait, I just noticed that one can use `and_then` with the `IntoFuture` variant as well. And it's already there in the docs. I should've read them, thank you for mentioning this! I think this can be closed then.

PR #442, opened 2017-04-04
Do not use unpark events as part Shared

carllerche:
Shared used unpark events internally. However, the way it used it was not part of the original intention for the feature. The correct way to manage unpark logic for a future is to use `Spawn`. This commit
re-implements `Shared` to use `Spawn` instead.

This work blocks #436 as unpark events are being deprecated in favor of `UnparkContext`. `UnparkContext` requires `&mut self` access, which does not fit well with the `Shared` requirements.

The original logic of `Shared` was a bit hard for me to follow, so I tried to restructure it to use clearer state transitions.

cc @dwrensha @alexcrichton 

Please squash when merging.

### DIFF
--- a/src/future/shared.rs
+++ b/src/future/shared.rs
@@ -63,24 +63,28 @@ const REPOLL: usize = 2;
 const COMPLETE: usize = 3;
 const POISONED: usize = 4;
 
+pub fn new<F: Future>(future: F) -> Shared<F> {
+    Shared {
+        inner: Arc::new(Inner {
+            next_clone_id: AtomicUsize::new(1),
+            unparker: Arc::new(Unparker {
+                state: AtomicUsize::new(IDLE),
+                waiters: Mutex::new(HashMap::new()),
+            }),
+            future: UnsafeCell::new(executor::spawn(future)),
+            result: UnsafeCell::new(None),
+        }),
+        waiter: 0,
+    }
+}
+
 impl<F> Shared<F> where F: Future {
     // TODO: make this private
     #[deprecated(since = "0.1.12", note = "use `Future::shared` instead")]
     #[cfg(feature = "with-deprecated")]
     #[doc(hidden)]
     pub fn new(future: F) -> Self {
-        Shared {
-            inner: Arc::new(Inner {
-                next_clone_id: AtomicUsize::new(1),
-                unparker: Arc::new(Unparker {
-                    state: AtomicUsize::new(IDLE),
-                    waiters: Mutex::new(HashMap::new()),
-                }),
-                future: UnsafeCell::new(executor::spawn(future)),
-                result: UnsafeCell::new(None),
-            }),
-            waiter: 0,
-        }
+        new(future)
     }
 
     /// If any clone of this `Shared` has completed execution, returns its result immediately
--- a/src/future/shared.rs
+++ b/src/future/shared.rs
@@ -47,7 +47,7 @@ impl<F> fmt::Debug for Shared<F>
 
 struct Inner<F: Future> {
     next_clone_id: AtomicUsize,
-    future: UnsafeCell<Spawn<F>>,
+    future: UnsafeCell<Option<Spawn<F>>>,
     result: UnsafeCell<Option<Result<SharedItem<F::Item>, SharedError<F::Error>>>>,
     unparker: Arc<Unparker>,
 }
@@ -71,7 +71,7 @@ pub fn new<F: Future>(future: F) -> Shared<F> {
                 state: AtomicUsize::new(IDLE),
                 waiters: Mutex::new(HashMap::new()),
             }),
-            future: UnsafeCell::new(executor::spawn(future)),
+            future: UnsafeCell::new(Some(executor::spawn(future))),
             result: UnsafeCell::new(None),
         }),
         waiter: 0,
@@ -114,6 +114,7 @@ impl<F> Shared<F> where F: Future {
     }
 
     fn complete(&self) {
+        unsafe { *self.inner.future.get() = None };
         self.inner.unparker.state.store(COMPLETE, SeqCst);
         self.inner.unparker.unpark();
     }
@@ -164,7 +165,7 @@ impl<F> Future for Shared<F>
             let unpark: Arc<Unpark> = self.inner.unparker.clone();
 
             // Poll the future
-            match unsafe { (*self.inner.future.get()).poll_future(unpark) } {
+            match unsafe { (*self.inner.future.get()).as_mut().unwrap().poll_future(unpark) } {
                 Ok(Async::NotReady) => {
                     // Not ready, try to release the handle
                     match self.inner.unparker.state.compare_and_swap(POLLING, IDLE, SeqCst) {
--- a/src/future/shared.rs
+++ b/src/future/shared.rs
@@ -13,22 +13,23 @@
 //! assert_eq!(6, *shared2.wait().unwrap());
 //! ```
 
-use std::fmt;
-use std::mem;
-use std::sync::{Arc, Mutex};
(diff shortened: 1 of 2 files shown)


src/future/shared.rs
```diff
@@ -38,64 +39,76 @@ impl<F> fmt::Debug for Shared<F>
 {
     fn fmt(&self, fmt: &mut fmt::Formatter) -> fmt::Result {
         fmt.debug_struct("Shared")
-            .field("id", &self.id)
```
tikue: Any reason you don't want `waiter` in the debug string?

src/future/shared.rs
```diff
@@ -38,64 +39,76 @@ impl<F> fmt::Debug for Shared<F>
...
-{
-    fn fmt(&self, fmt: &mut fmt::Formatter) -> fmt::Result {
-        fmt.debug_struct("Inner")
-            .field("next_clone_id", &self.next_clone_id)
-            .field("state", &self.state)
-            .finish()
-    }
+struct Unparker {
+    state: AtomicUsize,
+    waiters: Mutex<HashMap<usize, Task>>,
```
tikue: I know the performance probably won't matter for this use case, but `Mutex<HashMap<...>>` never sits well with me.

[**tangent**] If there were a concurrent hashmap in libstd, that'd be used here, but since there's not, and since futures-rs is such a fundamental crate, I assume a 3rd party crate won't be considered. I think it's worth noting because usually we think of crates.io as zero cost compared to a batteries-included std.

As a thought experiment, under what conditions could a crates.io concurrent map be considered? Would it need a certain amount of popularity/provenness? A 1.0 release? Some commitment to maintenance?

src/future/shared.rs
```diff
@@ -38,64 +39,76 @@ impl<F> fmt::Debug for Shared<F>
...
+const REPOLL: usize = 2;
+const COMPLETE: usize = 3;
+const POISONED: usize = 4;
 
-impl<F> Shared<F>
-    where F: Future
-{
+impl<F> Shared<F> where F: Future {
     /// Creates a new `Shared` from another future.
     pub fn new(future: F) -> Self {
```
alexcrichton: Could you add a TODO and a `#[doc(hidden)]` here while you're at it? This wasn't ever supposed to be public...

src/future/shared.rs
```diff
@@ -105,188 +118,161 @@ impl<F> Future for Shared<F>
...
+                    self.complete();
+                    return unsafe { self.clone_result().map(Async::Ready) };
+                }
+                Err(e) => {
+                    unsafe {
+                        (*self.inner.result.get()) = Some(Err(SharedError { error: Arc::new(e) }));
+                    }
+
+                    self.complete();
+                    return unsafe { self.clone_result().map(Async::Ready) };
```
alexcrichton: Perhaps the `NotReady` branch could `continue` and then these two branches could fall through to share the same "complete" path?

src/future/shared.rs
carllerche: The previous implementation used `Mutex<HashMap<...>>`... I was trying to do as little work as possible to get `Shared` off of w/ unpark events.

src/future/shared.rs
carllerche: What is the TODO? I can just deprecate it...

carllerche:
r?

stuhood:
Just to chime in quickly here: initial reports from incorporating this into our repo is that it might fix the issue I was reporting on https://github.com/alexcrichton/futures-rs/issues/330 ... so that could be good?

carllerche:
Implemented `Drop` for `Shared` at @dwrensha's request

carllerche:
Ah right I messed up deprecating `new`

src/future/shared.rs
```diff
@@ -105,188 +122,165 @@ impl<F> Future for Shared<F>
...
+            match unsafe { (*self.inner.future.get()).poll_future(unpark) } {
+                Ok(Async::NotReady) => {
+                    // Not ready, try to release the handle
+                    match self.inner.unparker.state.compare_and_swap(POLLING, IDLE, SeqCst) {
+                        POLLING => {
+                            // Success
+                            return Ok(Async::NotReady);
+                        }
+                        REPOLL => {
+                            // Gotta poll again!
```
dwrensha: Interesting! It seems plausible to me that this "loop on REPOLL" business fixes the bug that @stuhood has been encountering in the current implementation.

src/future/shared.rs
```diff
@@ -105,188 +126,165 @@ impl<F> Future for Shared<F>
...
-            State::Waiting(ref unparker, ref _original_future) => {
-                let mut unparker_inner = unparker.inner.lock().unwrap();
-                if unparker_inner.original_future_needs_poll {
-                    unparker_inner.original_future_needs_poll = false;
-                    unparker.clone()
-                } else {
-                    unparker_inner.insert(self.id, task::park());
-                    return Ok(Async::NotReady)
-                }
+        self.set_waiter();
```
dwrensha: Why is `set_waiter()` called here rather than just in the `POLLING | REPOLL` branch? Would it be incorrect to put it only in that branch? I think answer to these questions deserves at least a comment in the code.

dwrensha: Ah, I see that we need it in the `IDLE` branch too. But we don't need it in the `COMPLETE` case, right? Is this call here just to deduplicate code, or is it important for the ordering of things?

src/future/shared.rs
```diff
@@ -38,64 +39,84 @@ impl<F> fmt::Debug for Shared<F>
...
+            .field("waiter", &self.waiter)
             .finish()
     }
 }
 
 struct Inner<F: Future> {
-    next_clone_id: Mutex<u64>,
-    state: Mutex<State<F>>,
+    next_clone_id: AtomicUsize,
+    future: UnsafeCell<Spawn<F>>,
```
dwrensha: Ideally, we should drop the wrapped future as soon as it completes, to free any resources that it holds. The representation here does not allow that to happen. Maybe it should be `UnsafeCell<Option<Spawn<F>>>`? Or, even better, maybe we could pack the `future` and `result` fields into a single enum field. Would that work?

stuhood:
Only one backwards compatibility comment from [integrating this](https://github.com/pantsbuild/pants/pull/4429): the addition of `Clone` for `SharedError`/`SharedItem` meant that calls which used to use `DeRef` to access `Clone` on the target now require an explicit deref, ie:
```
let si: SharedItem<_> = ...;
// Used to work:
... = si.clone();
// Now requires:
... = (*si).clone();
```

We only had two callsites doing this, so not a big deal, but.

carllerche:
@stuhood Thanks for pointing this out. I will remove the `Clone` impl.

src/future/shared.rs
carllerche: It is not needed in the complete branch, but avoiding it adds somewhat significant code. I am favoring simpler code here in this PR and improvements can be done in later PRs.

carllerche:
Alright, I applied the proposed changes. The PR does not represent the best possible implementation but it is just trying to get off of unpark events to unblock the task revamp PR.

src/future/shared.rs
```diff
@@ -38,64 +39,85 @@ impl<F> fmt::Debug for Shared<F>
...
+        waiter: 0,
+    }
 }
 
-impl<F> Shared<F>
-    where F: Future
-{
-    /// Creates a new `Shared` from another future.
+impl<F> Shared<F> where F: Future {
+    // TODO: make this private
```
alexcrichton: 👍

alexcrichton:
Glad to hear @stuhood, thanks for confirming!

Thanks @dwrensha for taking a look too!

PR #453, opened 2017-04-19
better call-out of `Stream::concat` panic behaviour

srijs:
When I put in the doco for this earlier, I didn't realise there was already a convention in the docs for calling out if a function panics under certain circumstances.


alexcrichton:
Thanks!

PR #454, opened 2017-04-21
Implement Into for MergedItem

DenisKolodin:
I think we should have a capability to convert `MergedItem` into a tuple. I've implemented `Into` for this struct. What do you think about it?

I choose between:
1. To implement `Into<(Option<I1>, Option<I2>)>`
2. Add a method like `Option::take`: `fn take(&self) -> (Option<I1>, Option<I2>)`
_But for the second `MergedItem` doesn't have an empty state._

### DIFF
--- a/src/stream/merge.rs
+++ b/src/stream/merge.rs
@@ -36,6 +36,17 @@ pub enum MergedItem<I1, I2> {
     Both(I1, I2),
 }
 
+impl<I1, I2> Into<(Option<I1>, Option<I2>)> for MergedItem<I1, I2> {
+    fn into(self) -> (Option<I1>, Option<I2>) {
+        use self::MergedItem::*;
+        match self {
+            First(first) => (Some(first), None),
+            Second(second) => (None, Some(second)),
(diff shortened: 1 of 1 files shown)


carllerche:
(unrelated to the PR) Does using `Stream::select` work for you instead of `Stream::merge`? I believe that deprecating `Stream::merge` is being considered.

DenisKolodin:
@carllerche Using `Stream::select` is suitable if I wrap/convert items to union type. Will this be the preffered approach?

alexcrichton:
Thanks for the PR! 

I agree with @carllerche that leveraging `Stream::select` is likely the best option here, we're pretty likely to remove `Stream::merge` when we release 0.2 b/c `select` seems nicer to use. We could consider providing a combinator like `select` though which is heterogeneous and basically applies the wrapper for you, though.

In any case though I don't mind adding this method in the meantime, although I'd personally prefer a method on `MergedItem` mainly for discoverability.

DenisKolodin:
You've convinced me to avoid `Select::merge`. I close it.

PR #455, opened 2017-04-21
Add a generic `Executor` trait 

### LINKED ISSUE #313: Consider providing a Spawn trait
Given the number of different types of executors for driving futures, it seems worth providing a trait representing the ability to spawn a task with a future.

One option:

```rust
/// Value that can spawn a future
///
/// On spawn, the executor takes ownership of the future and becomes responsible
/// to call `Future::poll()` whenever a readiness notification is raised.
pub trait Spawn<T: Future<Item = (), Error = ()>> {

    /// Spawns a future to run on this `Spawn`.
    ///
    /// This function will return immediately, and schedule the future `f` to
    /// run on `self`. The details of scheduling and execution are left to the
    /// implementations of `Spawn`.
    fn spawn_detached(&self, f: T);
}
```

[Implementation](https://github.com/carllerche/futures-spawn)

The reason the function is `spawn_detached` is that there also are variants built on top of this primitive that return a future representing the completion of the spawned future:https://github.com/carllerche/futures-spawn/blob/inference/src/lib.rs#L21-L73

`SpawnHelper` exists to make rustc happy in some cases. There could be better ways of implementing this entire abstraction so it is up for discussion.

I also updated CPU pool to work with this `Spawn` trait: https://github.com/carllerche/futures-threadpool

Relates to: tokio-rs/tokio-core#57, tokio-rs/tokio-core#77

alexcrichton:
This commit adds a new trait to the `future` module, `Executor`. This trait has
only one method:

    trait Execute<F: Future<Item = (), Error = ()>> {
        fn execute(&self, f: F) -> Result<(), ExecuteError<F>>;
    }

The purpose of this trait is to unify the various executors found throughout the
ecosystem. Crates which require the ability to spawn futures will now have the
option ot operate generically over all `Executor` instances instead of a
particular executor.

A `Future::background` method was also added to ergonomically pass a future to
an executor and run it in the background. The two `oneshot` modules (sync and
unsync) also grew `spawn` and `spawn_fn` functions which will spawn a future
onto an executor, returning a handle to the resulting future. This can be used
to spawn work onto a specific executor and still get notified when the work
itself is completed.

Finally, an implementation of the `Executor` trait was added to `CpuPool`. Due
to the differences in unwinding/panic semantics, though, the `CpuPool::spawn`
method which previously existed was not deprecated.

Closes #313

### DIFF
--- a/src/future/mod.rs
+++ b/src/future/mod.rs
@@ -3,6 +3,7 @@
 //! This module contains the `Future` trait and a number of adaptors for this
 //! trait. See the crate docs, and the docs for `Future`, for full detail.
 
+use core::fmt;
 use core::result;
 
 // Primitive futures
@@ -956,4 +957,90 @@ pub trait FutureFrom<T>: Sized {
 
     /// Consume the given value, beginning the conversion.
     fn future_from(T) -> Self::Future;
+}
+
+/// A trait for types which can spawn fresh futures.
+///
+/// This trait is typically implemented for "executors", or those types which
+/// can execute futures to completion. Futures passed to `Spawn::spawn`
+/// typically get turned into a *task* and are then driven to completion.
+///
+/// On spawn, the executor takes ownership of the future and becomes responsible
+/// to call `Future::poll()` whenever a readiness notification is raised.
+pub trait Executor<F: Future<Item = (), Error = ()>> {
+    /// Spawns a future to run on this `Executor`, typically in the
+    /// "background".
+    ///
+    /// This function will return immediately, and schedule the future `future`
+    /// to run on `self`. The details of scheduling and execution are left to
+    /// the implementations of `Executor`, but this is typically a primary point
+    /// for injecting concurrency in a futures-based system. Futures spawned
+    /// through this `execute` function tend to run concurrently while they're
+    /// waiting on events.
+    ///
+    /// # Errors
+    ///
+    /// Implementors of this trait are allowed to reject accepting this future
+    /// as well. This can happen for various reason such as:
+    ///
+    /// * The executor is shut down
+    /// * The executor has run out of capacity to execute futures
+    ///
+    /// The decision is left to the caller how to work with this form of error.
+    /// The error returned transfers ownership of the future back to the caller.
+    fn execute(&self, future: F) -> Result<(), ExecuteError<F>>;
+}
+
+/// Errors returned from the `Spawn::spawn` function.
+pub struct ExecuteError<F> {
+    future: F,
+    kind: ExecuteErrorKind,
+}
+
+/// Kinds of errors that can be returned from the `Execute::spawn` function.
+///
+/// Executors which may not always be able to accept a future may return one of
+/// these errors, indicating why it was unable to spawn a future.
+#[derive(Debug, Copy, Clone, PartialEq)]
+pub enum ExecuteErrorKind {
+    /// This executor has shut down and will no longer accept new futures to
+    /// spawn.
+    Shutdown,
+
+    /// This executor has no more capacity to run more futures. Other futures
+    /// need to finish before this executor can accept another.
+    NoCapacity,
+
+    #[doc(hidden)]
+    __Nonexhaustive,
+}
+
+impl<F> ExecuteError<F> {
+    /// Create a new `ExecuteError`
+    pub fn new(kind: ExecuteErrorKind, future: F) -> ExecuteError<F> {
+        ExecuteError {
+            future: future,
+            kind: kind,
+        }
+    }
+
+    /// Returns the associated reason for the error
+    pub fn kind(&self) -> ExecuteErrorKind {
+        self.kind
+    }
+
+    /// Consumes self and returns the original future that was spawned.
+    pub fn into_future(self) -> F {
+        self.future
+    }
+}
+
+impl<F> fmt::Debug for ExecuteError<F> {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        match self.kind {
+            ExecuteErrorKind::Shutdown => "executor has shut down".fmt(f),
+            ExecuteErrorKind::NoCapacity => "executor has no more capacity".fmt(f),
+            ExecuteErrorKind::__Nonexhaustive => panic!(),
+        }
+    }
 }
--- a/futures-cpupool/src/lib.rs
+++ b/futures-cpupool/src/lib.rs
@@ -46,9 +46,9 @@ use std::sync::mpsc;
 use std::thread;
 
 use futures::{IntoFuture, Future, Poll, Async};
-use futures::future::lazy;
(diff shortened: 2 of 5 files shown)


alexcrichton:
I was pretty happy with everything *until* I ended up hitting `CpuPool` and the differences in panicking semantics. Overall though I'm pretty happy with how this turned out. I'd still like to play around with this in tokio-core though before landing.

r? @carllerche

src/future/mod.rs
```diff
@@ -957,3 +971,20 @@ pub trait FutureFrom<T>: Sized {
...
+///
+/// On spawn, the executor takes ownership of the future and becomes responsible
+/// to call `Future::poll()` whenever a readiness notification is raised.
+pub trait Spawn<F: Future<Item = (), Error = ()>> {
+    /// Spawns a future to run on this `Spawn`.
+    ///
+    /// This function will return immediately, and schedule the future `future`
+    /// to run on `self`. The details of scheduling and execution are left to
+    /// the implementations of `Spawn`.
+    fn spawn(&self, future: F);
```
carllerche: This function should return `Result<(), Error<F>>` to allow the executor to return the future if it fails to spawn it.

The docs should say that returning the future on error is "best effort" and returning `Ok` is not a guarantee that the future will be executed.

alexcrichton: Oops forgot that! I've updated with the error

carllerche:
Originally experimentation: https://github.com/carllerche/futures-spawn, though the latest is on a branch https://github.com/carllerche/futures-spawn/tree/restructure.

Once this PR is merged, I need to remember to deprecate that crate (leaving this comment to remind me!)

bluetech:
I saw this PR, and wanted to see if I understand the goal correctly (I hope you don't mind).

Consider a Hyper tokio server for example. Currently, hyper depends on tokio-core and tokio-proto -> tokio-core. If I understand this trait correctly it can help with removing this dependency -- instead of taking `Core`/`Handle`, it will take this trait. Then it should be possible to run a Hyper server with e.g.

- A tokio-core/mio event loop
- A cached thread pool (`CpuPool`)
- Any other `Spawn` implementation

by just changing a few lines in the application code (the library is agnostic).

To test this assumption and check if it's viable, I tried to do the following. First in Hyper:

- Removed all `client/` code (irrelevant for this discussion)
- Removed the helper `Server` trait, leaving only `Http` to use directly (`Server` uses tokio-core so I'm just removing it from consideration for now).
- Replaced all uses of `Handle` with an appropriate `Spawn` bound

and then in `tokio-proto`:

- Removed `tcp_client`/`tcp_server` - not used by hyper so ignoring that for now.
- Removed all of mentions of `old_io` (backward compatibility support for `tokio_core::io`, replaced by `tokio_io`).
- Replaced all uses of `Handle` with an appropriate `Spawn` bound

In the last bullet it got messy and I ran out of time to play with it, so I didn't reach a conclusion. But I think that it can be done - am I correct?

If it would be possible to statically parametrize a Hyper server (as an example) over a `Spawn` trait, and it would all work flawlessly when they are switched, I think it would really show the strength of the abstractions here - it would be quite a feat :)

alexcrichton:
@bluetech you are indeed spot on! That's the intended use case, decoupling crates like Hyper from tokio-core specifically and allowing them to interoprate more gracefully with other crates like futures-glib. The `Spawn` trait isn't sufficient for *everything* tokio-proto and hyper use, but everything minus the conveniences are sufficient to use `Spawn` (or at least that's the intention).

leodasvacas:
Name conflict with the `Spawn` struct seems bad. `Executor` seems like the right name here, we should steal that name from the current `Executor` trait and rename that (to `Dispatch` perhaps?). It is an unfortunate name anyway since when the docs say executor they mean something that spawns futures rather than something that actually implements the current `Executor` trait.

Going maybe out of scope from this PR, the current `Executor` starts looking like an odd duck since it is an executor but doesn't implement `Spawn`. Maybe we should make the current `Executor` into a struct that implements `Spawn` trait, and deprecate the `execute` method from the `Spawn` struct. It dosen't fit well anyways since `task.execute(executor)` reads backwards.

The renames would go like:
`Executor` trait -> `Dispatch` trait
`Spawn` trait -> `Executor` trait

Refactoring `Dispatch`:
Create a `Dispatcher` struct that wraps `Arc<Dispatch>`.
`Dispatcher` implements new `Executor` trait (the trait proposed here as `Spawn`).
Deprecate `Spawn::execute` from the `Spawn` struct.

alexcrichton:
@leodasvacas that sounds pretty reasonable to me! I definitely think the current `Executor` trait is basically deprecated and should likely be replaced by this.

@carllerche what do you think about the naming here?

carllerche:
Renaming seems fine. I would rename the trait fn to match the trait name.

`Executor` is deprecated, so reusing the name seems fine.

Other names:

* `Schedule`
* `Run`.

I don't have a huge preference when it comes to the actual name. I do think avoiding the `Spawn` conflict seems good.

alexcrichton:
Ok I've updated to `Executor`. r? @carllerche?

leodasvacas:
Nice! The fate of "OldExecutor" will be decided later? `background` is ergonomic and I'm not against it, but there are some points to be discussed:
- It's offering both a panicking and non-panicking version of the same API, `std` generally avoids this with a preference towards non-panicking. 
- Naming: I think we should just call it `execute_on` for consistency and neutrality (the executor may block for all we know).

src/future/mod.rs
```diff
@@ -887,6 +888,25 @@ pub trait Future {
...
+    ///
+    /// # Panics
+    ///
+    /// This function will panic if the instance of `Executor` provided is
+    /// unable to execute the future `f` provided.
+    fn background<E>(self, executor: &E)
+        where E: Executor<Self>,
+              Self: Future<Item = (), Error = ()> + Sized,
+    {
+        executor.execute(self).expect("failed to execute future in background")
```
carllerche: I agree that we should avoid panicking. An executor being unable to execute a task is a very plausible situation to get into.

The two options are for `background` to return a `Result` OR for it to silently drop the error.

I'm leaning towards just dropping the error because:

a) `background` is a convenience API
b) There is no guarantee that an executor returns an error on execution failure. It's completely acceptable for the executor to silently drop the task as well.

alexcrichton:
Ok I've updated to just delete `background` for now.

@leodasvacas I'd like to deprecate the old `Executor` trait as part of this. I meant to investigate [your suggestion](https://github.com/alexcrichton/futures-rs/pull/455#issuecomment-302889159) a bit more but I'll go do that now.

alexcrichton:
Ok I've commented that we're likely to deprecate `executor::Executor` and `Spawn::execute` but I haven't done so just yet. I don't think we have tons of users of that trait so I don't want to agonize too much over names just yet. in that case I'm going to merge!

PR #460, opened 2017-04-25
Update the doc test for Future::from_err

munckymagik:
# What?

* Updated the types to try to explicitly show that a type change occurs
* Renamed the variables to add further clarity
* Changed the type of the `T` parameter in the call to `err` to be the unit tuple. The intention being to prevent it distracting the reader from the value of the `E` parameter.

# Question ...

What do you think about using the unit tuple for type parameters that are unused in an example? If you like that, then I might raise another PR to apply it in other suitable places.

### DIFF
--- a/src/future/mod.rs
+++ b/src/future/mod.rs
@@ -388,8 +388,8 @@ pub trait Future {
     /// ```
     /// use futures::future::*;
     ///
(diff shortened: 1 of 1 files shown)


alexcrichton:
sounds good to me! Using unit also seems ok to me, thanks!

PR #464, opened 2017-05-03
[Task rewrite] Refactor Notify

leodasvacas:
**This PR is against task-rewrite not master**

This illustrates some proposals for the new Unpark/Notify. They are:

- Move `'static` bound to `Notify`, a small win for everyone.
- Change the `UnsafeNotify` impl for `Arc` to leverage `Arc::{from|into}_raw` which just landed on stable, making it safer by avoiding transmutes (the `transmute::<Arc<T>, *mut ArcWrapped<T>>` was rather fishy). This requires a rustc bump to 1.17. Also actually impl `UnsafeNotify` for `Arc` in preparation for the next step (this impl ends up changed to `OuterUnsafeNotify` in the last commit).
- Take the `UnsafeNotify` by reference rather than by value, this saves an `Arc` clone when polling. This means that `UnparkHandle` would be taken by value. But I skipped this step and went straight to removing `UnparkHandle` entirely since I don't see what role it would play, instead we take `&UnsafeNotify` directly, which makes for a clean API.
- Split `OuterUnsafeNotiy` from `UnsafeNotify` for flexibility. Need better names though.

Pros: Nicer API, safer implementation. Con: Bump to rustc 1.17.

### DIFF
--- a/src/executor.rs
+++ b/src/executor.rs
@@ -10,4 +10,4 @@
 #[allow(deprecated)]
 pub use task_impl::{Spawn, spawn, Unpark, Notify, Executor, Run};
 
-pub use task_impl::{UnsafeNotify, NotifyHandle};
+pub use task_impl::UnsafeNotify;
--- a/src/future/shared.rs
+++ b/src/future/shared.rs
@@ -15,7 +15,7 @@
 
 use {Future, Poll, Async};
 use task::{self, Task, Spawn};
-use executor::{Notify, NotifyHandle};
+use executor::Notify;
 
 use std::{fmt, mem, ops};
 use std::cell::UnsafeCell;
@@ -163,9 +163,8 @@ impl<F> Future for Shared<F>
 
             // Poll the future
             let res = unsafe {
-                let notify  = NotifyHandle::from(self.inner.notifier.clone());
                 (*self.inner.future.get()).as_mut().unwrap()
-                    .poll_future_notify(&notify, 0)
+                    .poll_future_notify(&self.inner.notifier, 0)
             };
             match res {
                 Ok(Async::NotReady) => {
--- a/src/task_impl/mod.rs
+++ b/src/task_impl/mod.rs
@@ -652,7 +652,7 @@ impl Unpark for RunInner {
 /// Instances of `Notify` must be safe to share across threads, and the methods
 /// be be invoked concurrently. They must also live for the `'static` lifetime,
 /// not containing any stack references.
-pub trait Notify: Send + Sync {
+pub trait Notify: Send + Sync + 'static {
     /// Indicates that an associated future and/or task are ready to make
     /// progress.
     ///
@@ -691,7 +691,7 @@ pub struct NotifyContext<T> {
     inner: Arc<NotifyContextInner<T>>,
 }
 
-impl<T: Notify + 'static> NotifyContext<T> {
+impl<T: Notify> NotifyContext<T> {
     /// Create a new `NotifyContext` backed by the provided instance of
     /// `notify`.
     ///
@@ -1080,7 +1080,7 @@ impl Drop for NotifyHandle {
 
 struct ArcWrapped<T>(PhantomData<T>);
 
-impl<T: Notify + 'static> Notify for ArcWrapped<T> {
+impl<T: Notify> Notify for ArcWrapped<T> {
     fn notify(&self, id: u64) {
         unsafe {
             let me: *const ArcWrapped<T> = self;
@@ -1090,7 +1090,7 @@ impl<T: Notify + 'static> Notify for ArcWrapped<T> {
     }
 }
 
-unsafe impl<T: Notify + 'static> UnsafeNotify for ArcWrapped<T> {
+unsafe impl<T: Notify> UnsafeNotify for ArcWrapped<T> {
     unsafe fn clone_raw(&self) -> NotifyHandle {
         let me: *const ArcWrapped<T> = self;
         let ptr = (*(&me as *const *const ArcWrapped<T> as *const Arc<T>)).clone();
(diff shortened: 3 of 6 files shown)


carllerche:
Sorry, but I'm not following the plot. Can we break it down?

Starting w/ adding `'static` to `Notify`. What is driving this?

alexcrichton:
Note that it's [intentional `Notify` does not have `'static`](https://github.com/alexcrichton/futures-rs/pull/272)

alexcrichton:
It also looks like [you've removed the old unpark system](https://github.com/alexcrichton/futures-rs/pull/464/commits/893221ad64da004ff518cfd5de8485a496524540) which was intentionally left in, I believe it has an unacceptable perf cost for all users of the old API.

alexcrichton:
Also note that it was intentional that `UnsafeNotify for Arc<T>` doesn't exist, as I think you've discovered with `OuterUnsafeNotify` the impl is the wrong layer of indirection.

leodasvacas:
@carllerche I think everywhere that actually used `Notify` required `'static` and the docs say:
"Instances of `Notify` must be safe to share across threads, and the methods be invoked concurrently. They must also live for the `'static` lifetime, not containing any stack references"
So putting it directly in the trait clarifies intent and cleans up signatures a bit. But @alexcrichton makes a fair point that there was a reason for this and we would need to review if rayon still needs the bound left out if we go with the rest of the changes. It seems more conservative to keep the bound in the methods for now.

Second point is that `transmute::<Arc<T>, *mut ArcWrapped<T>>` dosen't look very safe, does std guarantee that `Arc` shall forever be only a pointer and of the same size as a raw pointer?

Third point is taking the notify handles by reference to avoid cloning unless strictly necessary. This is equivalent to taking `&Arc`, taking `&UnparkHandle` is not quite the same since you need to clone the `Arc` into the handle. This leads to `impl From<&Arc> for UnparkHandle`. I went further and removed `UnparkHandle` since I didn't see a use for it, it wasn't really opaque because `UnsafeNotify` anyhow, so it seemed like just noise. I also don't see what type of future proofing it would provide.

@alexcrichton `OuterUnsafeNotify` is a subset of the behaviour of `UnsafeNotify` with the bonus of not requiring `Send + Sync`. I see nothing technically wrong with `UnsafeNotify for Arc<T>` though.

leodasvacas:
@alexcrichton Regarding the removal of the old system, I mostly regret that, only after I removed it I realized how things broke down for `Arc<Unpark>` and the double boxing was a workaround, I figured maybe we didn't care, but if we do then certainly the right thing to do is to leave the compatibility layer.

alexcrichton:
@leodasvacas 

Note that at this point adding `'static` to `Unpark` is a breaking change, so I don't think we can do it in the 0.1.x series regardless.

> does std guarantee that Arc shall forever be only a pointer and of the same size as a raw pointer?

Yes

> Third point is taking the notify handles by reference to avoid cloning unless strictly necessary. 

Apis currently take `&UnparkHandle`? Structs can store `UnparkHandle` if they want to avoid cloning.

> I went further and removed UnparkHandle since I didn't see a use for it, it wasn't really opaque because UnsafeNotify anyhow, so it seemed like just noise

It's a safe abstraction around `*mut UnsafeNotify`, it's *the* point of where the trait object is handled. Adding a trait like `OuterUnsafeNotify` for this seems like overkill to me as it's basically just "get an `UnparkHandle` from this object".

leodasvacas:
@alexcrichton 
I'm convinced adding the `'static` bound is the wrong thing to do, even if rayon doesn't care someone else might.
> Yes

If std makes that guarantee then I guarantee then it's settled, we can revisit this when for some other reason the minimum rustc is bumped.
Regarding the `NotifyHandle` I want to avoid the initial clone to create one, you can see in the tests that this is  achieved. There are other ways to achieve this, I'll make a new PR focused on this issue that hopefully will clarify my thoughts here.

alexcrichton:
When you say adding `'static` is the wrong thing to do, do you actually mean that not having the `'static` bound is the wrong thing to do? (just clarifying)

If we *really* want to avoid one extra clone then I don't think we need many contortions, just one trait perhaps:

```rust
trait GetUnparkHandle {
    fn get(&self) -> UnparkHandle;
}
```

I will agree that technically being generic over `GetUnparkHandle` is faster, but I would also argue that the complexity it adds is not worth it.

carllerche:
@alexcrichton perhaps `AsRef<UnparkHandle>` would work? That said, I've unloaded most of this from cache...

leodasvacas:
@alexcrichton I mean I've changed my mind and we should keep the status quo and not do what this PR does w.r.t the `'static'` bound. I've come to the same conclusion as you and will make a PR using `GetUnparkHandle`, I now see `UnparkHandle` has it's value.

@carllerche I don't think any built in trait quite fits, `ToOwned<UnparkHandle>` is almost it but not quite, `Into<UnparkHandle> + Copy` would work but at that point it's simpler to use our own trait.

alexcrichton:
@carllerche ah yeah I think `AsRef<UnparkHandle>` won't work in the sense of the management of resources here is somewhat unsafe, so we can't implent `AsRef<UnparkHandle>` for `Arc<T>` (I think).

@leodasvacas oh I think that `GetUnparkHandle` could work, we'd just store a trait object in the `BorrowedTask` variable to have a vtable dispatch to acquire the actual handle, which in terms of cost should be quite low in terms of the atomic add and such. 

Note that I'm also fine re-adding the `'static` bound eventually, but I'd just want to clear it with rayon first to make sure it's still workable.

leodasvacas:
I've opened a new PR for `GetUnparkHandle` (with the name `CloneIntoUnparkHandle`). After the task rewrite has settled I'll open issues to track other aspects of this PR we may want to revist:
- Moving the `'static` bound directly to `Notify`
- Changing the implementation to use `Arc::{from|into}_raw`

PR #468, opened 2017-05-07
[Task rewrite] Optimize NotifyHandle by avoiding an Arc clone

leodasvacas:
**This PR is against task-rewrite not master**

Previous discussion in #464. The motivation for this is to avoid the `Arc` clone when creating a `NotifyHandle`. This delays cloning the `Arc` until a task handle is requested through `current()`.

This is done by introducing `CloneIntoNotifyHandle` which is implemented for `Arc`.

For example the signature of `poll_future_notify` goes from:
`fn poll_future_notify(&mut self, notify: &NotifyHandle, id: u64) -> ...`
To:
`fn poll_future_notify(&mut self, notify: &CloneIntoNotifyHandle, id: u64) -> ...`

### DIFF
--- a/src/future/shared.rs
+++ b/src/future/shared.rs
@@ -15,7 +15,7 @@
 
 use {Future, Poll, Async};
 use task::{self, Task, Spawn};
-use executor::{Notify, NotifyHandle};
+use executor::Notify;
 
 use std::{fmt, mem, ops};
 use std::cell::UnsafeCell;
@@ -163,9 +163,8 @@ impl<F> Future for Shared<F>
 
             // Poll the future
             let res = unsafe {
-                let notify  = NotifyHandle::from(self.inner.notifier.clone());
                 (*self.inner.future.get()).as_mut().unwrap()
-                    .poll_future_notify(&notify, 0)
+                    .poll_future_notify(&self.inner.notifier, 0)
             };
             match res {
                 Ok(Async::NotReady) => {
--- a/src/task_impl/mod.rs
+++ b/src/task_impl/mod.rs
@@ -37,7 +37,7 @@ struct BorrowedTask<'a> {
 #[allow(deprecated)]
 enum BorrowedUnpark<'a> {
     Old(&'a Arc<Unpark>),
-    New(&'a NotifyHandle, u64),
+    New(&'a CloneIntoNotifyHandle, u64),
 }
 
 #[derive(Copy, Clone)]
@@ -137,9 +137,10 @@ pub fn current() -> Task {
         let unpark = match borrowed.unpark {
             BorrowedUnpark::Old(old) => TaskUnpark::Old(old.clone()),
             BorrowedUnpark::New(new, id) => {
+                let notify = new.clone_into_handle();
                 // A new handle is being created, increment the ref count
-                new.ref_inc(id);
-                TaskUnpark::New(new.clone(), id)
+                notify.ref_inc(id);
+                TaskUnpark::New(notify, id)
             }
         };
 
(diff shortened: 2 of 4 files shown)


alexcrichton:
Thanks for the PR! This looks good to me from an API-point-of-view, but I personally feel that this level of abstraction is going into the weeds a little. In some sense the task system is already billed as relatively complex, and adding yet-another-trait probably isn't helping it :(

In that sense, if you're able to come up with some convincing benchmarks, that'd help advocate for this quite a bit!

leodasvacas:
@alexcrichton Good point. If you can just store your `NotifyHandle` in a struct, then the single arc clone saved is unlikely to help performance. But for example in `wait_future` we create a new `ThreadUnpark` each time. I've benchmarked `wait_future` before and after this PR. The new implementation is measurably faster.

The tested future pathologicaly does nothing and never blocks, but it's conceivable that a future would be cheap to run and rarely block to the point where this optimization would matter.

Benchmark:

```rust
struct NeverBlocks;
impl Future for NeverBlocks {
    type Item = ();
    type Error = ();
    fn poll(&mut self) -> Poll<(), ()> { Ok(Async::Ready(())) }
}

#[bench]
fn wait_new(b: &mut Bencher) {
    let mut task = executor::spawn(NeverBlocks);
    b.iter(|| { task.wait_future() });
}

#[bench]
fn wait_old(b: &mut Bencher) {
    let mut task = executor::spawn(NeverBlocks);
    b.iter(|| { task.wait_future_old() });
}
```
Results (the ones with the smallest deviation I could get):
```rust
test wait_new ... bench:          60 ns/iter (+/- 6)
test wait_old ... bench:          78 ns/iter (+/- 4)
```

alexcrichton:
@leodasvacas is that benchmarked against master? or against the `task-rewrite` branch? On the task-rewrite branch the clone doesn't happen each iteration of the loop?

leodasvacas:
@alexcrichton against `task-rewrite`, it builds a fresh `NotifyHandle` for each call to `wait_future`, cloning the arc in the process.

alexcrichton:
Interesting! I'm surprised the allocation, `Arc::new`, right before that doesn't dwarf the increment. (note that `thread::park` internally is already doing atomic increments. Could you share the source for the benchmark you had?

leodasvacas:
@alexcrichton This code never calls `thread::park` because the future never blocks. I guess allocators are fast!

I put the source up [here](https://github.com/leodasvacas/futures-benchmark) ready to benchmark with `cargo bench`. It's pulling the futures dependency from [this branch](https://github.com/leodasvacas/futures-rs/blob/task-rewrite-benchmark/src/task_impl/mod.rs#L377) which is the same as the branch in this PR but with the `wait_future_old` for comparision.

carllerche:
If we are concerned about the cost of allocating the `Arc` in each call to `wait`, why don't we just throw the `ThreadUnpark` instance into a thread local variable?

alexcrichton:
@carllerche oh I think this is moreso the cost of cloning an Arc to give it to `poll_*` rather than allocating the Arc itself.

FWIW so far 100% of situations I've called `poll_*` would benefit from this, nothing naturally stores an `UnparkHandle` as-is. I'm leaning towards merging this as it may be worth the complexity, but it'd be great if we could figure out how to hide this away without worrying about the complexity bubbling up too much.

alexcrichton:
Ok @carllerche had the idea of `T: Clone + Into<NotifyHandle>` which I think is great, so I'm gonna pull this in and update it, thanks again for the PR @leodasvacas!

alexcrichton:
I'm gonna manage the branches manually though so closing instead of merging.

leodasvacas:
@alexcrichton Clever workaround object safety! I had considered but discarded that idea because I thought object safety would prevent it from working. Given that it works, I propose we do `T: Copy + Into<NotifyHandle>` and take `T` by value. This generalises the API since references are copy, we would add `impl Into<NotifyHandle> for &Arc`. The motivation is that it would work very nicely with the following impls we may want:
`impl Into<NotifyHandle> for &'static T where T : Notify`
`impl Into<NotifyHandle> for fn(u64)`

PR #471, opened 2017-05-11
Add `Stream::by_ref` method

dmizuk:
This is a convenience method which works just like `Iterator::by_ref` method.

Conceptually, `Future` trait can have a method like this, but I didn't add that because I failed to come up with a use case for that.


alexcrichton:
Thanks!

PR #477, opened 2017-05-16
[stream] add a way to execute futures serially

sid0:
I've found this to be useful when creating a list of futures upfront and then
wanting them all to be executed in the order they were passed in.

In this implementation the next future starts executing only after the previous
future is done. I believe this makes the most sense semantically, and this
avoids nasty ordering issues where the futures could be accessing a shared
resource (DB etc).

### DIFF
--- a/src/stream/futures_ordered.rs
+++ b/src/stream/futures_ordered.rs
@@ -1,0 +1,101 @@
+//! Definition of the FuturesOrdered combinator, executing each future in a sequence serially
+//! and streaming their results.
+
+use std::fmt;
+
+use {Async, Future, IntoFuture, Poll, Stream};
+
+/// A stream which takes a list of futures and executes them serially.
+///
+/// This future is created with the `futures_ordered` method.
+#[must_use = "streams do nothing unless polled"]
+pub struct FuturesOrdered<I>
+    where I: IntoIterator,
+          I::Item: IntoFuture
+{
+    elems: I::IntoIter,
+    current: Option<<I::Item as IntoFuture>::Future>,
+}
+
+impl<I> fmt::Debug for FuturesOrdered<I>
+    where I: IntoIterator,
+          I::Item: IntoFuture,
+          <<I as IntoIterator>::Item as IntoFuture>::Future: fmt::Debug,
+{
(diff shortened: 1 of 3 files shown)


alexcrichton:
Thanks for the PR! Just for a sanity check, this is roughly equivalent to this, right?

```rust
pub fn futures_ordered<I>(iter: I)
    -> impl Stream<Item = I::Item::Item, Error = I::Item::Error>
    where I: IntoIterator,
          I::Item: IntoFuture
{

    stream::iter(iter.map(Ok)).and_then(|future| future)
}
```

sid0:
Yeah, they do the same thing. However without `impl Trait` the function sig becomes pretty unwieldy. How do you feel about this?

This is what compiles, btw:

```
    stream::iter(iter.into_iter().map(ok)).and_then(|future| future)
```

alexcrichton:
Ah ok yeah thanks just wanted to make sure I was understanding this correctly. I generally personally at least prefer to avoid similar-ish combinators and such. This is *mostly* `stream::iter` plus `and_then` (modulo https://github.com/alexcrichton/futures-rs/issues/480) which doesn't quite cross my own threshold for "complex enough it's worth adding". It's definitely true that the type name is unwieldy, but that's the same for just about all more-than-one combinator abstraction :(

@carllerche what do you think?

carllerche:
I don't have a strong opinion either way. I would lean towards not adding a new API unless necessary.

alexcrichton:
Ok for now I think due to this being so similar to the [construction above](https://github.com/alexcrichton/futures-rs/pull/477#issuecomment-302239691) I'm going to close this, but if this becomes a common thing we can certainly reconsider!

sid0:
So, uhh, @khuey landed something that's effectively this in https://github.com/alexcrichton/futures-rs/commit/bc39aa13c94e1f5937dc59aedc6554c413347aa8.

sid0:
(though it looks like that actually executes futures in parallel)

alexcrichton:
@sid0 ah yeah that was the major difference, the capability to run futures concurrently vs hardcoded one at a time.

PR #483, opened 2017-05-20
Fix confusing wording in oneshot::Sender docs

### LINKED ISSUE #482: Minor wording issue in docs for futures::sync::oneshot::Sender
Docs for `oneshot::Sender` in the description of the [`send` method](https://docs.rs/futures/0.1.13/futures/sync/oneshot/struct.Sender.html#method.send) say:

> Completes this oneshot with a successful result.
> 
> This function will consume `self` and indicate to the other end, the `Receiver`, that the error provided is the result of the computation this represents.

The *the error provided* part sounds off and confusing. Shouldn’t it be *the value provided*, as the method is used to complete the future with a successful result? It seems like a copy-paste mistake.

silmeth:
Fixes #482.

### DIFF
--- a/src/sync/oneshot.rs
+++ b/src/sync/oneshot.rs
@@ -120,7 +120,7 @@ impl<T> Sender<T> {
     /// Completes this oneshot with a successful result.
     ///
     /// This function will consume `self` and indicate to the other end, the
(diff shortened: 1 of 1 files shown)


alexcrichton:
Thanks!

PR #485, opened 2017-05-22
SendAll::poll() should not close underlying Sink

eldargab:
Hello. May be I am missing something, but it seems that `sink.send_all(stream)`
should not close itself on completion.

### DIFF
--- a/src/sink/send_all.rs
+++ b/src/sink/send_all.rs
@@ -75,7 +75,7 @@ impl<T, U> Future for SendAll<T, U>
             match try!(self.stream_mut().poll()) {
                 Async::Ready(Some(item)) => try_ready!(self.try_start_send(item)),
                 Async::Ready(None) => {
-                    try_ready!(self.sink_mut().close());
+                    try_ready!(self.sink_mut().poll_complete());
                     return Ok(Async::Ready(self.take_result()))
                 }
                 Async::NotReady => {


alexcrichton:
Thanks for the PR @eldargab! This is related to https://github.com/alexcrichton/futures-rs/issues/462 where unfortunately this is a breaking change that can't yet land, but I've tagged that issue for 0.2-cleanup to ensure we handle that during the transition.

While we're at it though, mind commenting as to your use case for using the sink after `send_all`? (just to have it on record)

eldargab:
Basically, I was just inspecting the code and noticed, that implementation doesn't much neither docs nor my expectations.

As of use case, it's not hard to imaging. First of all it is an alternative to `stream.chain()` and `.chain()` is not always possible. For example, you can have interaction between client and server like this (pseudocode)

```rs
fn do_transaction(rx, tx) {
    let s1 = rx.recv();
    open(s1).send_all(tx);
    let s2 = rx.recv(); // waiting for another stream request, but this time it depends on s1
    open(s2).send_all(tx);
}
```
which translates to something like

```rs
fn do_transaction(rx: Rx, tx: Tx) -> Future {
    rx.recv().and_then(|req1| {
        let s1 = open(req1);
        s1.send_all(tx)
    }).and_then(|(_, tx)| {
        rx.recv().join(Ok(tx))        
    }).and_then(|req2, tx| {
        // req2 depends on contents of s1
        open(req2).send_all(tx)
    })
}
```

alexcrichton:
Ok cool thanks for the use case! We'll be sure to try and cover this in the 0.2 release (not currently planned, but will happen eventually!)

alexcrichton:
Merged in https://github.com/rust-lang-nursery/futures-rs/pull/745

PR #487, opened 2017-05-25
Add an `executor::Wait` trait

### LINKED ISSUE #360: RFC: Customizable `wait()` behavior

alexcrichton:
This commit adds a new trait, `Wait`, to serve as the implementation for
functions like `Future::wait`. The intention for this trait is for executors to
be able to customize the blocking behavior of futures.

Closes #360

### DIFF
--- a/src/task_impl/std/mod.rs
+++ b/src/task_impl/std/mod.rs
@@ -5,8 +5,9 @@ use std::fmt;
 use std::marker::PhantomData;
 use std::mem;
 use std::ptr;
-use std::sync::{Arc, Once, ONCE_INIT};
+use std::rc::Rc;
 use std::sync::atomic::{AtomicBool, Ordering};
+use std::sync::{Arc, Once, ONCE_INIT};
 use std::thread;
 
 use {Future, Stream, Sink, Poll, Async, StartSend, AsyncSink};
@@ -239,14 +240,15 @@ impl<F: Future> Spawn<F> {
     /// to complete. When a future cannot make progress it will use
     /// `thread::park` to block the current thread.
     pub fn wait_future(&mut self) -> Result<F::Item, F::Error> {
-        let unpark = Arc::new(ThreadUnpark::new(thread::current()));
-
-        loop {
-            match try!(self.poll_future_notify(&unpark, 0)) {
-                Async::NotReady => unpark.park(),
-                Async::Ready(e) => return Ok(e),
+        with_current_wait(|wait| {
+            let handle = WaitToHandle(wait);
+            loop {
+                match try!(self.poll_future_notify(&handle, wait.notify_id())) {
+                    Async::NotReady => wait.wait(),
+                    Async::Ready(e) => return Ok(e),
+                }
             }
-        }
+        })
     }
 
     /// A specialized function to request running a future to completion on the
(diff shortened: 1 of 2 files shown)


carllerche:
As a reference, here is the tokio-core snippet from https://github.com/alexcrichton/futures-rs/pull/436#issuecomment-301673150.

```rust
extern crate tokio_core;

fn main() {
    tokio_core::run(|handle| {
        let future = ...;
        let result = future.wait();
        println!("{:?}", result);
    });
}
```

src/task_impl/std/mod.rs
```diff
@@ -651,3 +632,207 @@ impl<T> From<Arc<T>> for NotifyHandle
...
+    /// If this function returns without the above handle being notified, that's
+    /// ok.
+    ///
+    /// # Panics
+    ///
+    /// It is expected that this method will not panic. This method may be
+    /// called recursively due to multiple invocations of `Future::wait` on the
+    /// stack.  In other words its expected for executors to handle
+    /// waits-in-waits correctly.
+    fn wait(&self);
```
carllerche: I think that this function should be permitted to panic if the current executor does not allow for blocking the current thread.

src/task_impl/std/mod.rs
```diff
@@ -651,3 +632,207 @@ impl<T> From<Arc<T>> for NotifyHandle
...
+    /// Note that it is permissible for this function to have spurious wakeups.
+    /// If this function returns without the above handle being notified, that's
+    /// ok.
+    ///
+    /// # Panics
+    ///
+    /// It is expected that this method will not panic. This method may be
+    /// called recursively due to multiple invocations of `Future::wait` on the
+    /// stack.  In other words its expected for executors to handle
+    /// waits-in-waits correctly.
```
tanriol: Does this imply that the inner `wait` invocation may block the outer one until the inner `wait` finishes? This does not sound like "shortly thereafter".

carllerche: @tanriol you are very correct. If nested calls to wait are done, then there are no guarantees as to when `wait` returns. In fact, there could be cases where `wait` never returns (because someone called `wait` on `tcp_listener.incoming().for_each(...)`.

This is definitely a significant hazard and great care should be taken with this feature.

src/task_impl/std/mod.rs
alexcrichton: Ah right yes good point.

alexcrichton:
Ok, I've pushed an updated which I believe addresses comments. I'd like to integrate this into tokio-core first before landing.

carllerche:
I think it would be wise to first agree how tokio-core should use this. I have significant reservations about supporting recursive calls to wait w/ tokio-core as there is no way to guarantee that futures ever complete. This isn't just a hypothetical concern as historically, other "future" platforms built on top of Java's Fork/Join pool did exactly this and my recollection is that it did not end well.

IIRC, this was attempted w/ Akka and other less known libs and has since been removed. Maybe someone who is more familiar w/ these libs could provide further insight.

carllerche:
I believe this this is now a "won't fix" with the direction being taken in #645. Please re-open if I am wrong.

PR #499, opened 2017-06-03
Make trait assertions nicer

rkarp:


3Hren:
It pollutes traits namespace, making compilation slower.

rkarp:
I've just tested it with `cargo clean && time cargo build` commands (on Windows 7 x64) and there is no measurable difference in build times to be seen. Whether it's built with stable or nightly, debug or release mode, it's all within the same roughly 1-2% time window.

alexcrichton:
Thanks @rkarp! Lokks good to me and yeah I don't think this'll slow down compilation at all, so let's merge!

PR #513, opened 2017-06-11
refactor: prevent double boxing in `boxed`

3Hren:
Proof of concept as an alternative of #512.

Do not merge, since it requires unstable feature and some of the code are not wrapped with `cfg's`.


3Hren:
Well, this PR has completed its proof-of-concept job, all further discussion will be in #512.

PR #528, opened 2017-07-04
Add Clone implementation and take() function to SharedItem and SharedError

alteous:
The documentation says these data structures are cloneable but the implementation is missing.

Side note: This future is awesome and has saved me a lot of time, so thanks!

### DIFF
--- a/src/future/shared.rs
+++ b/src/future/shared.rs
@@ -251,7 +251,7 @@ impl<F> fmt::Debug for Inner<F>
 
 /// A wrapped item of the original future that is clonable and implements Deref
 /// for ease of use.
-#[derive(Debug)]
+#[derive(Clone, Debug)]
 pub struct SharedItem<T> {
     item: Arc<T>,
 }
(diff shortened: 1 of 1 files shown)


alteous:
I've added a `take()` method to both data structures so the user can obtain the inner value. I need this functionality for my crate.

alexcrichton:
Thanks for the PR! The intention here though was to not expose the implementation detail of the `Arc`, though. Mind elaborating on why it needs to be extracted?

alteous:
Actually, it may not be necessary after all. The `gltf` crate is using shared futures to lazily load binary data. The idea behind `take` was  to allow the user to keep hold of this loaded data, and I assumed returning an `Arc<[u8]>` would be better for interoperability than a `future::SharedItem<[u8]>`. Hiding the implementation details is probably a better idea, though.

alexcrichton:
Ah ok in that case yeah I'd prefer to keep the sharing details (e.g. `Arc`) private, so I'm going to close, but thanks regardless!

PR #530, opened 2017-07-05
Add Clone implementation for SharedItem and SharedError

alteous:
This addresses https://github.com/alexcrichton/futures-rs/pull/528#issue-240380014.

This change would benefit users of the `gltf` crate as they would be able to make cheap immutable copies of the acquired data.


alexcrichton:
Thanks!

PR #537, opened 2017-07-22
Add with_flat_map sink combinator

exrook:
Add a `with_flat_map` combinator to `Sink`, similar to `flat_map` from `std::Iterator`

### DIFF
--- a/src/sink/mod.rs
+++ b/src/sink/mod.rs
@@ -12,6 +12,7 @@ use {IntoFuture, Poll, StartSend};
 use stream::Stream;
 
 mod with;
+mod with_flat_map;
 // mod with_map;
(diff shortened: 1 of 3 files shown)


alexcrichton:
Thanks! I wonder though, would this perhaps be better to return a stream rather than an iterator?

exrook:
That's probably a better idea, I only used iterator because I wasn't sure how to implement it with Stream. I'll see if I can figure it out sometime this week.

exrook:
I found the time to rewrite this patch to work with Streams rather than iterators. Let me know if you have any other suggestions.

alexcrichton:
Thanks!

PR #539, opened 2017-07-23
ignore IntelliJ project files

tdyas:


alexcrichton:
Thanks for the PR! I personally tend to prefer these go in the global gitignore files, however, so closing.

PR #540, opened 2017-07-23
Add end_after combinator on Streams

tdyas:
Add an end_after combinator on Streams that ends the stream after the element for which the provided predicate returns `true`. The stream yields all of the elements for which the predicate returns false and then yields the element for which the predicate returned true as the stream's final element.

This is useful, for example, to cleanly close a stream representing the responses from a server after the server has emitted its final response.

In some sense, `end_after` is a counterpart to `take_while` in that `take_while` cuts the stream before the element for which the predicate is true, while `end_after` cuts the stream after that
element.

### DIFF
--- a/src/stream/end_after.rs
+++ b/src/stream/end_after.rs
@@ -1,0 +1,95 @@
+use {Async, Poll};
+use stream::Stream;
+
+/// A stream combinator used to end the stream after the predicate returns
+/// true for an element. The stream yields all of the elements for which
+/// the predicate returns false and then yields the element for which the
+/// predicate returned true as the stream's final element.
+#[derive(Debug)]
+#[must_use = "streams do nothing unless polled"]
+pub struct EndAfter<S, P> {
+    stream: S,
+    pred: P,
+    ended: bool,
+}
+
+pub fn new<S, P>(s: S, p: P) -> EndAfter<S, P>
+    where S: Stream,
+          P: FnMut(&S::Item) -> bool,
+{
+    EndAfter {
+        stream: s,
+        pred: p,
+        ended: false,
+    }
+}
+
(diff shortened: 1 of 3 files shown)


tdyas:
It seems like an unrelated test failed for the CI building off of nightly:

```error: internal compiler error: /checkout/src/librustc_trans/context.rs:739: failed to get layout for `&mut futures::future::join::MaybeDone<futures::future::MapErr<futures::stream::Forward<futures::stream::Iter<std::iter::Map<std::vec::IntoIter<i32>, fn(i32) -> std::result::Result<i32, futures::unsync::mpsc::SendError<i32>> {std::result::Result<i32, futures::unsync::mpsc::SendError<i32>>::Ok}>>, futures::unsync::mpsc::UnboundedSender<i32>>, [closure@tests/unsync.rs:84:22: 84:57]>>`: the type `futures::future::join::MaybeDone<futures::future::MapErr<futures::stream::Forward<futures::stream::Iter<std::iter::Map<std::vec::IntoIter<i32>, fn(i32) -> std::result::Result<i32, futures::unsync::mpsc::SendError<i32>> {std::result::Result<i32, futures::unsync::mpsc::SendError<i32>>::Ok}>>, futures::unsync::mpsc::UnboundedSender<i32>>, [closure@tests/unsync.rs:84:22: 84:57]>>` has an unknown layout```

alexcrichton:
I think on iterators in the standard library this is called `take_while`, so perhaps we could mirror that?

tdyas:
There is already a `take_while` combinator on futures/streams. I wanted something that would let the final element go through the stream. Kinda like the difference between inclusive and exclusive ranges. Going to withdraw the PR. Closing.

PR #544, opened 2017-07-28
Add `reunite` for BiLock and Split{Stream,Sink}

spinda:
In both cases, consumes two halves of a split-up value and recovers the original.

```rust
let foo: Foo = Foo();
let (sink, stream) = foo.split();
let foo: Foo = sink.reunite(stream).expect("sink and stream aren't from the same original Stream + Sink");
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,7 +12,7 @@ matrix:
       after_success:
         - travis-cargo doc-upload
     - os: linux
-      rust: 1.10.0
+      rust: 1.17.0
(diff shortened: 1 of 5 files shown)


alexcrichton:
Seems like a slick API to me, thanks! I think the usage of `Arc::ptr_eq` is rasining the minimum Rust version though, mind bumping that in the travis config?

alexcrichton:
Thanks again @spinda!

PR #545, opened 2017-07-31
Implement stream::poll_fn

ishitatsuyuki:
WIP. Using CI because I'm affected by https://github.com/rust-lang/rust/issues/43132.

EDIT: seems passed.


alexcrichton:
Shouldn't this be included in `stream/mod.rs` as well? Can you be sure to test this patch locally and include a unit test as well?

ishitatsuyuki:
Uh, can't right now. I will have to do a rebuild.

I modified mod.rs initially but I accidentally removed it alongside with whitespace change.

I will address the issues including tests tomorrow.

src/stream/poll_fn.rs
```diff
@@ -0,0 +1,49 @@
...
+///
+/// let read_stream = poll_fn(move || -> Poll<Option<String>, std::io::Error> {
+///     if counter == 0 { return Ok(Async::Ready(None)); }
+///     counter -= 1;
+///     Ok(Async::Ready(Some("Hello, World!".to_owned())))
+/// });
+/// ```
+pub fn poll_fn<T, E, F>(f: F) -> PollFn<F>
+where
+    F: FnMut() -> Poll<T, E>,
```
alexcrichton: I think inference may be helped here by having this be `Poll<Option<T>, E>`, right?

ishitatsuyuki: Agreed.

PR #546, opened 2017-08-04
Unsync stream spawn

tmccombs:


src/stream/mod.rs
```diff
@@ -1051,6 +1053,19 @@ pub trait Stream {
...
+    /// Converts a Stream<Item=T,Error=E> into a `Stream<Item=Result<T,E>>`.
+    ///
+    /// The returned `Stream` will never return an `Err` on poll.
+    /// Instead, the item type is a result, matching the result
+    /// of polling the original stream.
+    ///
+    /// This is useful if the stream is attached to a sink,
+    /// and the sink should handle any errors rather than the
+    /// stream itself.
+    fn results<E>(self) -> Results<Self, E> where Self: Sized {
```
alexcrichton: Could this PR hold off on adding this for now? Perhaps this could remain private for just here?

alexcrichton:
Ah unfortunately it looks like `pub(crate)` was stabilized pretty recently so this breaks the 1.10 minimum requirement. I think `pub(crate)` may have been a little *too* recent to use here just yet?

tmccombs:
Well, I think this should work pre-1.18. But, I'm not sure if it is the best solution.

alexcrichton:
👍

PR #548, opened 2017-08-08
Add with/and_with/else_with combinators to Future

spinda:
These facilitate threading extra context through a chain of futures.

Summary:

```rust
fn with<T>(self, context: T) -> impl Future<Item = (Self::Item, T), Error = (Self::Error, T)>;
fn and_with<T>(self, context: T) -> impl Future<Item = (Self::Item, T), Error = Self::Error>;
fn else_with<T>(self, context: T) -> impl Future<Item = Self::Item, Error = (Self::Error, T)>;
```

_____

When implementing client objects using futures, a common pattern is to have a method consume an instance of the client and return a future which holds onto it while doing work, then spits it back out along with the result. Such methods may return the client instance in both success and error cases, or may "eat" the client in case of an error, signifying that the underlying resource may no longer be in a valid state.

I've often ended up writing code along the lines of

```rust
do_thing(&mut resource)
    .and_then(move |value| {
        do_other_thing(&mut resource, value).map(move |value| (value, resource))
    })
    .and_then(|(value, resource)| {
        do_third_thing(&mut resource, value).map(move |value| (value, resource))
    })
```

or even

```rust
do_thing(&mut resource)
    .then(move |result| {
        match result {
            Err(err) => Either::A(err((err, resource))),
            Ok(value) => {
                Either::B(do_other_thing(&mut resource, value).map(move |value| (value, resource)))
            }
        }
    })
    .then(|result| {
        match result {
            Err((err, resource)) => Either::A(err((err, resource))),
            Ok((value, resource)) => {
                Either::B(do_third_thing(&mut resource, value).map(move |value| (value, resource)))
            }
        }
    })
```

`and_with` can turn the first example into

```rust
do_thing(&mut resource).and_with(resource)
    .and_then(|(value, resource)| do_other_thing(&mut resource, value).and_with(resource))
    .and_then(|(value, resource)| do_third_thing(&mut resource, value).and_with(resource))
```

and `with` can turn the second into

```rust
do_thing(&mut resource).with(resource)
    .and_then(|(value, resource)| do_other_thing(&mut resource, value).with(resource))
    .and_then(|(value, resource)| do_third_thing(&mut resource, value).with(resource))
```

`else_with` fulfills the case where you either want to consume the resource object in creating the success value or pass it back to the user in case of error (think `Sink`'s `start_send`).

### DIFF
--- a/src/future/and_with.rs
+++ b/src/future/and_with.rs
@@ -1,0 +1,34 @@
+use {Future, Poll};
+
+/// Future for the `and_with` combinator, passing state along into the result
+/// of a future which completes successfully.
+///
+/// This is created by the `Future::and_with` method.
+#[derive(Debug)]
+#[must_use = "futures do nothing unless polled"]
+pub struct AndWith<A, T> where A: Future {
+    future: A,
+    context: Option<T>,
+}
+
+pub fn new<A, T>(future: A, context: T) -> AndWith<A, T>
+    where A: Future,
+{
+    AndWith {
+        future: future,
+        context: Some(context),
+    }
+}
+
+impl<A, T> Future for AndWith<A, T>
+    where A: Future,
+{
+    type Item = (A::Item, T);
+    type Error = A::Error;
+
+    fn poll(&mut self) -> Poll<(A::Item, T), A::Error> {
+        let item = try_ready!(self.future.poll());
+        let context = self.context.take().expect("AndWith polled after complete");
+        Ok((item, context).into())
+    }
+}
--- a/src/future/else_with.rs
+++ b/src/future/else_with.rs
@@ -1,0 +1,39 @@
+use {Async, Future, Poll};
+
+/// Future for the `else_with` combinator, passing state along into the result
+/// of a future which produces an error.
+///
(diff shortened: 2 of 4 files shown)


alexcrichton:
So for comparison with what I believe these combinators are equivalent to:

```rust
future.and_with(t)
future.map(|x| (t, x))
future.map(move |x| (t, x)) // needed if `t` is Copy

future.else_with(t)
future.map_err(|x| (t, x))
future.map_err(move |x| (t, x)) // needed if `t` is Copy

future.with(t)
future.then(|t| {
    match t {
        Ok(e) => Ok((e, t)),
        Err(e) => Err((e, t)),
    }
})
```

I'm not 100% sold on the `and_with` and `else_with` front as usage of `map` and `map_err` (which I think is equivalent) shouldn't be too much less ergonomic. For `then` though I can see how `with` is a clear winner!

I'm curious, do you think that `with` will be very commonly used? Or do you expect most usage to use `and_with` or `else_with`?

carllerche:
Closing due to inactivity. Please let me know if this is incorrect.

PR #549, opened 2017-08-10
Make oneshot::Sender cloneable

spinda:
This type is a wrapper around Arc anyway, and everything within is concurrency-protected. A unit test is added for good measure.


spinda:
Upon further review I may have been too hasty in submitting this&mdash;it's not so simple as it first appeared. I'll revise and resubmit.

PR #550, opened 2017-08-13
Expose AtomicTask

carllerche:
This PR exposes AtomicTask as it is a generally useful API. It also removes `unsafe` from the API.

### DIFF
--- a/src/task_impl/atomic_task.rs
+++ b/src/task_impl/atomic_task.rs
@@ -7,7 +7,25 @@ use core::cell::UnsafeCell;
 use core::sync::atomic::AtomicUsize;
 use core::sync::atomic::Ordering::{Acquire, Release};
 
-/// A coordinated `Task` handle enabling concurrent operations on a task.
+/// A synchronization primitive for task notification.
+///
+/// `AtomicTask` will coordinate concurrent notifications with the consumer
+/// potentially "updating" the underlying task to notify. This is useful in
+/// scenarios where a computation completes in another thread and wants to
+/// notify the consumer, but the consumer is in the process of being migrated to
+/// a new logical task.
+///
+/// Roughly, consumers should call `park` before checking the result of a
+/// computation and producers should call `notify` after producing the
+/// computation. It is also permitted for `notify` to be called **before**
+/// `park`. This results in a no-op.
+///
+/// A single `AtomicTask` may be reused for any number of calls to `park` or
+/// `noitfy`.
+///
+/// `AtomicTask` does not provide any guaranteed memory ordering, as such the
+/// user should use caution and use other synchronization primitives to guard
+/// the result of the underlying computation.
 pub struct AtomicTask {
     state: AtomicUsize,
     task: UnsafeCell<Option<Task>>,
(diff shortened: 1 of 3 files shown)


carllerche:
CI is failing due to rustc regressions:

* https://github.com/rust-lang/rust/issues/43852
* https://github.com/rust-lang/rust/issues/43853

alexcrichton:
Ah yeah no worries about failures on beta. Could you add a detailed comment about the concurrent behavior of park?

src/task_impl/atomic_task.rs
```diff
@@ -52,30 +70,35 @@ impl AtomicTask {
         }
     }
 
-    /// The caller must ensure mutual exclusion
-    pub unsafe fn park(&self) {
-        if let Some(ref task) = *self.task.get() {
```
alexcrichton: Wasn't this one of the big use cases of this cell? Is the perf impact not measureable?

src/task_impl/atomic_task.rs
```diff
@@ -7,7 +7,25 @@ use core::cell::UnsafeCell;
...
-/// A coordinated `Task` handle enabling concurrent operations on a task.
+/// A synchronization primitive for task notification.
+///
+/// `AtomicTask` will coordinate concurrent notifications with the consumer
+/// potentially "updating" the underlying task to notify. This is useful in
+/// scenarios where a computation completes in another thread and wants to
+/// notify the consumer, but the consumer is in the process of being migrated to
+/// a new logical task.
+///
+/// Roughly, consumers should call `park` before checking the result of a
```
alexcrichton: This isn't quite true, right? Instead consumers shoud call `park` after they determine the value isn't ready, right?

src/task_impl/atomic_task.rs
carllerche: It was never measured. The main point was managing the logic around atomically swapping the consumer task while the producer is notifying. The original API was `unsafe` mostly because it was for a specific use case (`FuturesUnordered`) where we knew there was only a single consumer and it had to be `&self`.

Either way, if it proves to be a perf issue, we can always bring it back as `park_unchecked`.

src/task_impl/atomic_task.rs
carllerche: It is? If consumers call `park` after they determine the value isn't ready, then there is a race condition where:

* Consumer checks value
* Producer produces value
* Producer calls notify (no task registered -> no-op)
* Consumer calls park: nothing happens.

carllerche:
@alexcrichton I added additional notes to `park`, but I'm not sure if you had anything specific in mind.

src/task_impl/atomic_task.rs
alexcrichton: Isn't that what [this block](https://github.com/alexcrichton/futures-rs/blob/10e48d878897efcf0c7e2223661f994b0cf64b87/src/task_impl/atomic_task.rs#L79-L87) is handling?

src/task_impl/atomic_task.rs
alexcrichton: Sounds reasonable to me!

alexcrichton:
>I'm not sure if you had anything specific in mind.

Nah looks great to me, thanks!

src/task_impl/atomic_task.rs
carllerche: That block is handling concurrent calls to `park`. Those states are hit when a concurrent thread is currently in the park path, in which case the op is aborted as we consider the other thread has "won" the race.

alexcrichton: Hm ok reading over the park/unpark implementation again it turns out it's significantly different in this aspect than I was expecting. I initially expected this to behave like `std::thread`'s park/unpark where if you call `unpark` (or `notify` in this case) it "buffers a notification" if there's no blocking task. I think that would mean that you're not *required* to call `park` before you do any work. Do you know if there's a reason to *not* implement that?

carllerche: Yeah, that pattern was my original attempt. However, it doesn't really work in the face of the consumer **changing** its task in the same way that it does w/ `thread::park()`.

As you said, `thread::park()`  / `unpark()` behaves as a buffered notification. If `unpark` is called first, then that notification gets buffered until `park` is called to "consume" the notification.

AtomicTask scenario:

* Consumer calls `park` w/ task A.
* Consumer migrates to task B & checks computation result (not completed).
* Producer completes computation.
* Producer calls `notify`
* Consumer calls park` on task B.

The notification will be sent to task `A` and task `B` has deadlocked.

This primitives is more as a strategy to manage consumers migrating across tasks. This is why the consumer must call `park` before.

One disconnect might be that this PR leaves the original `park` naming. I'm open to naming suggestions.

alexcrichton: Ok yeah that makes sense to me. Perhaps something like:

* Rename the method to `store_current` from `park`
* Explicitly mention in the documentation that "park before you check" may not be expected, and it's different from other "park-like apis" like `std::thread`?

alexcrichton:
👍

PR #555, opened 2017-08-15
Rename `UnboundedSender::send` to `send_now`

### LINKED ISSUE #261: Ergonomics of channel send
Consider this code:

```rust
use futures::*;

fn main() {
    let (mut tx, rx) = futures::sync::mpsc::unbounded();
    tx.send(1);
}
```
You might think this is actually sending the message on the channel, but it's not because `Sink` has been imported via the glob import (and I'd imagine most users are using the glob import) so it's calling `Sink::send` not `UnboundedSender::send` (since `UnboundedSender` impls `Sink`).

We'd have to use UFCS instead:

```rust
fn main() {
    let (mut tx, rx) = futures::sync::mpsc::unbounded();
    UnboundedSender::send(&mut tx, 1).unwrap();
}
```

I guess the compiler warnings will lead the user to fix the code, but figured I would at least document it here.

alexcrichton:
This method conflicts with `Sink::send` which is unlikely to be renamed, so this
commit renames the method on `UnboundedSender`.

Closes #261

### DIFF
--- a/src/sync/mpsc/mod.rs
+++ b/src/sync/mpsc/mod.rs
@@ -518,7 +518,18 @@ impl<T> UnboundedSender<T> {
     /// This is an unbounded sender, so this function differs from `Sink::send`
     /// by ensuring the return type reflects that the channel is always ready to
     /// receive messages.
+    #[deprecated(note = "renamed to `unbounded_send`")]
+    #[doc(hidden)]
     pub fn send(&self, msg: T) -> Result<(), SendError<T>> {
+        self.unbounded_send(msg)
+    }
+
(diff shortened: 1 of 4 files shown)


alexcrichton:
I'm not super in love with the name `send_now`, and would be curious to hear others' suggestions!

carllerche:
Yeah, `send_now` isn't super great. Maybe just `unbounded_send`.. which isn't super great either, but at least matches the `Unbounded` prefix on the types.

alexcrichton:
Sounds good to me!

👍

PR #556, opened 2017-08-15
Deprecate `boxed` methods

### LINKED ISSUE #228: Consider removing `.boxed()` combinators and the `Box` aliases

alexcrichton:
These methods have caused far more confusion than they were ever intended to
cause, and they're use rarely enough in practice "to success" that they're
likely to be removed from any 0.2 release. As a result this commit starts out
the removal process by deprecating them and removing usage internally.

Closes #228

### DIFF
--- a/src/future/mod.rs
+++ b/src/future/mod.rs
@@ -97,6 +97,10 @@ if_std! {
     pub use self::join_all::JoinAll as Collect;
 
     /// A type alias for `Box<Future + Send>`
+    #[doc(hidden)]
+    #[deprecated(note = "removed without replacement, recommended to use a \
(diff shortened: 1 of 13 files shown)


carllerche:
👍

PR #557, opened 2017-08-15
Add an `iter_ok` adaptor for non-Result iterators

### LINKED ISSUE #480: Change stream::iter to not require an iterator of Result
Just work over an iterator of T and don't fail!

alexcrichton:
The `iter` adaptor has been hard to use at least for me historically as it's
pretty rare to have an iterator of results. In addition with the planned change
to have errors terminate a `Stream` it doesn't necessarily make sense to
continue with an iterator of results!

Closes #480


carllerche:
👍

PR #558, opened 2017-08-15
Add `mpsc::Sender::try_send`

### LINKED ISSUE #478: MPSC: Send unless blocks
Using a bounded MPSC channel, would it be possible to send conditionally on the condition that the channel won't block?

Something like: `futures::unsync::mpsc::Sender::send_unless_blocks(&mut self, item: Self::SinkItem) -> Result<(), WouldBlock>`.

Here's my motivation for such an API: I want to send a `Task` handler from a stream that's going to be parked, to an outer receiver that's going to unpark it, but I don't want to send it multiple times in the case the stream is polled multiple times: if the stream won't block, I want to sent the `Task` and return an `Async::NotReady`. If the stream would block, that means that the stream isn't unparked yet, so I want just to return an `Async::NotReady`.

alexcrichton:
This is an API which serves for sending a message along a `mpsc::Sender` without
attempting to park the current task, making it safe to invoke from a helper
thread or a non-futures-aware thread.

Closes #478

### DIFF
--- a/src/sync/mpsc/mod.rs
+++ b/src/sync/mpsc/mod.rs
@@ -484,6 +484,24 @@ impl<T> Sender<T> {
             Async::Ready(())
         }
     }
+
+    /// Attempts to send a message on this `Sender<T>` without blocking.
+    ///
+    /// This function, unlike `start_send`, is safe to call whether or not it's
+    /// being called on a task or not. Note that this function, however, will
+    /// *not* attempt to block the current task if the message cannot be sent.
+    ///
+    /// It is not recommended to call this function from inside of a future,
+    /// only from an external thread where you've otherwise arranged to be
+    /// notified when the channel is no longer full.
+    pub fn try_send(&self, msg: T) -> Result<(), SendError<T>> {
+        match self.inc_num_messages(false) {
+            Some(_park_self) => {}
+            None => return Err(SendError(msg)),
+        };
+        self.queue_push_and_signal(Some(msg));
+        Ok(())
+    }
 }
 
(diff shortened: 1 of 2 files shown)


src/sync/mpsc/mod.rs
```diff
@@ -484,6 +484,24 @@ impl<T> Sender<T> {
...
+    /// This function, unlike `start_send`, is safe to call whether or not it's
+    /// being called on a task or not. Note that this function, however, will
+    /// *not* attempt to block the current task if the message cannot be sent.
+    ///
+    /// It is not recommended to call this function from inside of a future,
+    /// only from an external thread where you've otherwise arranged to be
+    /// notified when the channel is no longer full.
+    pub fn try_send(&self, msg: T) -> Result<(), SendError<T>> {
+        match self.inc_num_messages(false) {
+            Some(_park_self) => {}
```
alexcrichton: I'm not 100% sure it's safe to ignore this value, would be good for a double check from @carllerche!

src/sync/mpsc/mod.rs
```diff
@@ -484,6 +484,24 @@ impl<T> Sender<T> {
             Async::Ready(())
         }
     }
+
+    /// Attempts to send a message on this `Sender<T>` without blocking.
+    ///
+    /// This function, unlike `start_send`, is safe to call whether or not it's
+    /// being called on a task or not. Note that this function, however, will
```
danburkert: extra 'or not' in sentence.

alexcrichton:
I'll try to get back around to this when I have time.

PR #564, opened 2017-08-19
replace try! macro with ?

cssivision:


cssivision:
fix https://github.com/alexcrichton/futures-rs/issues/563

alexcrichton:
Thanks! Can you update the minimum version on CI?

cssivision:
yes, but which version should be?

alexcrichton:
The version of rust that added `?`

cssivision:
yes, the minimum version that add `?`

alexcrichton:
The process you would go through to figure that out is the same that I'd do

cssivision:
ok, i will fix this.

cssivision:
it, 1.13.0, i found it in the [release page](https://github.com/rust-lang/rust/blob/master/RELEASES.md#version-1130-2016-11-10).

PR #565, opened 2017-08-22
Add Future::inspect to peek at the item in a future

### LINKED ISSUE #562: Add a method to Future like map that does not allow for mutation of the resolution
Unsure of a good name but I use `tap` in the example.

```rust
make_request().tap(|| {
  // [...] Executes on successful resolution of the future but does not see or 
  // alter the future
})
```

Recently I've been playing with futures and I've had a need to do _something_ when a future is resolved but that _something_ has little to do with the future being resolved. I could use `map` and forward the value but I'd rather express my intent more clearly.

```rust
// Functionally equivalent to the proposed 'tap'
make_request().map(|response| {
  // [...] Executes on successful resolution of the future but does not alter the future
  response
})
```

---

For more context, see https://github.com/mehcode/salt-rs/blob/master/examples/time.rs#L21-L29

mehcode:
Fixes #562

### DIFF
--- a/src/future/inspect.rs
+++ b/src/future/inspect.rs
@@ -1,0 +1,64 @@
+use {Future, Poll, Async};
+
+/// Do something with the item of a future, passing it on.
+///
+/// This is created by the `Future::inspect` method.
+#[derive(Debug)]
+#[must_use = "futures do nothing unless polled"]
+pub struct Inspect<A, F> where A: Future {
+    future: A,
+    f: Option<F>,
+}
+
+pub fn new<A, F>(future: A, f: F) -> Inspect<A, F>
+    where A: Future,
+{
+    Inspect {
+        future: future,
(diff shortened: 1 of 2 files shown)


alexcrichton:
Thanks! Mind moving the tests to the `tests/` folder? (I think there's a failure in 1.13.0 as well)

Also, while you're at it, mind adding a combinator for `Stream` as well?

mehcode:
Thanks for finishing this up for me. Apologies I didnt get back to it quickly. I got so wrapped up working on [Shio](https://github.com/mehcode/shio-rs) that I forgot I was in the middle of a couple other things, haha.

alexcrichton:
Oh no worries at all! I had a few minutes this morning and was gonna publish anyway, so figured I'd go ahead and help out

Issue #132, opened 2016-09-07
Check if task is parked before parking thread.


Stebalien:
So, I'm worried about a race condition in future's like `Join` when one of the sub-futures calls `thread::park()` internally. Basically, what if we have the following sequence of events:

1. Someone calls `spawn(future).wait_future()` where `future` is a `Join` of two sub-futures.
2. Sub-future 1 parks the task and schedules it to be unparked.
3. Sub-future 1 returns `NotReady`.
4. An event happens and the task is unparked. This will call `thread::unpark` on the current thread.
5. Future 2 communicates with another thread and, in the process, some internal method calls `Thread::park`. This isn't really the "correct" way to communicate in an async environment but I can see some library doing this if it knows that it's thread will be unparked *really* quickly.
6. The `Join` returns `NotReady`.
7. `Spawn`'s `wait_future` method calls `thread::park` and waits indefinitely because step 4 cleared the park flag.

A fix for this is to set a `task_is_parked` flag, set/unset it from within `task::park()`/`Task::unpark()` and check this before calling `thread::park()`. This is still racy but the race is safe: If `task_is_parked` is unset *after* we check it but before we call `thread::park()`, we *know* that no user-specified code could have called `thread::park()` and swallowed our unpark event in the interim. This also protects us from user code that returns `NotReady` without parking.

Is this actually a problem? I haven't tested it (writing concurrent tests like this is not fun) but I can't see how the current code could prevent this.

alexcrichton:
Thanks for the report! This is something thought that I believe is handled by `thread::park` and such. A call to `thread::current` doesn't clear any flags or actually block the current thread, and `thread::park` itself isn't called until a call to `poll` is finished entirely.

In that sense if a call to `Thread::unpark` comes it'll just prevent the later `thread::park` from blocking. That is, in step 4 above the flag I believe isn't cleared.

Stebalien:
The problem is that the internal `thread::park` eats that flag.

* We tell some scheduler to unpark us when ready.
* Our scheduler unparks us.
* Some user-specified code (on our thread) calls `while !some_atomic_condition() { thread::park() }`.
* Eventually some other user-specified code (not on our thread) sets the condition and unparks our thread.
* We call `thread::park()` and wait forever because that park loop above ate our flag.

alexcrichton:
[...]

PR #566, opened 2017-08-26
unset ThreadUnpark.ready when waking from park

oconnor663:
Previously `ThreadUnpark::park()` would exit with the `ready` flag still set to true. That meant that the next call to `park()` would short-circuit parking, even if no calls to `notify()` had happened in between.

I noticed this issue when I was playing with a toy "sleep" future. (Spawn a background thread, sleep on that thread, and then unpark the original task after sleeping.) I noticed that I was sometimes seeing an extra call to `poll` immediately after the sleep started, in particular on the second sleep when two of them were chained together with `then`. I tracked that down to [one of the calls to `ThreadUnpark::park`](https://github.com/alexcrichton/futures-rs/blob/0.1.15/src/task_impl/std/mod.rs#L245), which was returning immediately without blocking. The extra poll this was causing was a no-op, and it shouldn't do any harm to a properly-written future, but it's probably a performance waste in some circumstances.

This change seems like the smallest possible fix, but as part of making it I started wondering: What's the point of the `ready` flag in `ThreadUnpark`? `std::thread::Thread` will already "buffer" an unpark that comes while the thread is running. Why does `ThreadUnpark` need to repeat that logic? Can we get rid of the flag entirely, and make it a stateless wrapper? Advice appreciated.

### DIFF
--- a/src/task_impl/std/mod.rs
+++ b/src/task_impl/std/mod.rs
@@ -475,14 +475,35 @@ impl ThreadUnpark {
     }
 
     fn park(&self) {
-        if !self.ready.swap(false, Ordering::SeqCst) {
-            thread::park();
+        // If `ready` is true, that means we need to immediately re-poll, so we
+        // short-circuit instead of parking.
+        if self.ready.swap(false, Ordering::SeqCst) {
+            return;
         }
+
+        // Block until the next `notify`. Note that if a call to `notify` races
+        // in between our check above and this `park`, that's ok, because
+        // `thread::park` sets an internal flag of its own similar to our
+        // `ready`.
+        thread::park();
+
+        // Now we've awoken from `park`. If it happened the usual way, by
+        // another thread calling `notify` while we slept, then the `ready` flag
+        // has been set back to true. Since the wait loop is going to `poll`
+        // again as soon as we return, and there's no need to `poll` twice, we
+        // clear it.
+        self.ready.store(false, Ordering::SeqCst);
     }
 }
 
 impl Notify for ThreadUnpark {
     fn notify(&self, _unpark_id: usize) {
+        // The target thread might already be in the middle of `poll` when
+        // `notify` is called. In that case it might've just missed the
+        // readiness of some part of its future, and it needs to immediately
+        // call `poll` all over again as soon as the current `poll` is done.
+        // Setting the `ready` flag makes this happen by skipping the next
+        // `park`.
         self.ready.store(true, Ordering::SeqCst);
         self.thread.unpark()
     }
--- a/src/future/mod.rs
+++ b/src/future/mod.rs
@@ -194,9 +194,9 @@ pub trait Future {
     /// When a future is not ready yet, the `Async::NotReady` value will be
     /// returned. In this situation the future will *also* register interest of
     /// the current task in the value being produced. This is done by calling
-    /// `task::park` to retrieve a handle to the current `Task`. When the future
-    /// is then ready to make progress (e.g. it should be `poll`ed again) the
-    /// `unpark` method is called on the `Task`.
+    /// `task::current` to retrieve a handle to the current `Task`. When the
+    /// future is then ready to make progress (e.g. it should be `poll`ed again)
+    /// the `notify` method is called on the `Task`.
     ///
     /// More information about the details of `poll` and the nitty-gritty of
     /// tasks can be [found online at tokio.rs][poll-dox].
@@ -237,8 +237,8 @@ pub trait Future {
     ///
     /// If `NotReady` is returned, then the future will internally register
     /// interest in the value being produced for the current task (through
-    /// `task::park`). In other words, the current task will receive a
-    /// notification (through the `unpark` method) once the value is ready to be
+    /// `task::current`). In other words, the current task will receive a
+    /// notification (through the `notify` method) once the value is ready to be
     /// produced or the future can make progress.
     ///
     /// Note that if `NotReady` is returned it only means that *this* task will


oconnor663:
Here's a playground example that repros the "spurious poll" when it runs against the current version of futures, 0.1.15. https://play.rust-lang.org/?gist=78a5e723f2a52c4954601f8a51a80a0a&version=stable

alexcrichton:
Thanks for this PR!

The boolean here I believe was added in https://github.com/alexcrichton/futures-rs/commit/e52f14c9ece6691b79f55978e0605b51daa1b59b to fix https://github.com/alexcrichton/futures-rs/issues/132, the basic idea being that `unpark` externally should get routed to `ThreadUnpark` instead of accidentally waking up an unrelated `unpark` in other code. The boolean there, though, is just to ensure that if the `notify` was for sure received before that we don't block the thread, but if not we rely on thread park/unpark to do blocking and whatnot.

Thinking on this patch, though, I think I'm a little worried about it. If we spuriously wake up from `park` (which I believe is allowed), then we could, at that moment, get a `notify`. We could then overwrite `ready` and lose it later.

I haven't thought too much about this problem, though, but it definitely seems like it should be fixable!

oconnor663:
Oh wow, that's one of those issues I'll have to read over a few times :)

> If we spuriously wake up from `park` (which I believe is allowed), then we could, at that moment, get a `notify`. We could then overwrite `ready` and lose it later.

I agree that that could happen, but I wonder if the following two scenarios are indistinguishable. The first is this race:

1. Thread 1 parks.
2. Thread 1 unparks spuriously, but doesn't get to line 482 (set ready false) yet.
3. Thread 2 calls `notify` on Thread 1, setting ready true.
4. Thread 1 gets to line 482 and sets ready false, "erasing" the `notify`.

In this scenario, Thread 1 ends up awake, and the ready flag ends up false. Compare this to the "normal unpark" flow (with this PR):

1. Thread 1 parks.
2. Thread 2 calls `notify` on Thread 1 while it's parked.
3. Thread 1 unpacks and sets ready false.

In this scenario, the same result happens: Thread 1 ends up awake, and the ready flag ends up false. If the race above is equivalent to the "normal" notify flow, I think it's not a problem?

Another way to phrase the same thing, my understanding of the requirements for `notify` was something this: "If the task is already unparked, and it has done any useful work at all, guarantee that the next call to `park` doesn't block, so that that useful work gets repeated." If that understanding is right, then maybe we can say that no "real work" is possible between lines 479 and 482, so it's alright for a `notify` in there to get dropped. But now I'm rethinking all of that in the context of the issue you linked (extra parks and unparks happening inside a Future) and I'm not sure if it's still true...

alexcrichton:
Ah so I think what happens is:

* Thread 1 parks
* Thread 1 wakes up spuriously
* Thread 2 sets `ready` to `true`
* Thread 1 sets `ready` to `false`
* Thread 1 goes and calls `park` again due to unrelated code (e.g. some threading abstraction)
* Thread 2 calls `unpark`
* Thread 1 wakes up again, futures isn't ready, goes to block
* Thread 1 see that `ready` is `false`
* Thread 1 calls `park` in `ThreadUnpark`

I think that basically the unconditional store of `false` makes #132 possible to resurface again? I definitely agree, though, that #132 is a pretty obscure bug...

oconnor663:
I think I understand this a little better. So the answer to my question up top ("Can we get rid of the `ready` flag entirely?") is no as per #132: Even though `std::thread::Thread` does "buffer" an unpark, it's always possible for other parking/unparking code (which can crop up anywhere) to "consume" that buffer. That means the wait loop can't rely on it to guarantee that `poll` will run again once the loop is finished.

As far as I can tell, though, the smaller change in this PR is correct. Looking at @alexcrichton's scenario above, I read it as "step 3 is where the other thread calls `notify`" and "steps 5-7 represent the main thread doing a complete call to `poll`". If that's the right way to read it, then I think we're upholding `notify`'s contract (a call to `poll` must come promptly after every call to `notify`), and sleeping the main thread in step 9 is the correct thing to do.

The big assumption I'm making is that, as soon as `ThreadUnpark::park` returns, the wait loop is definitely going to run `poll` again. Is there any case where that wouldn't be true?

I've added some comments and a small bit of refactoring to the PR, but please let me know if they're excessive :)

src/task_impl/std/mod.rs
```diff
@@ -475,14 +475,35 @@ impl ThreadUnpark {
...
 impl Notify for ThreadUnpark {
     fn notify(&self, _unpark_id: usize) {
+        // The target thread might already be in the middle of `poll` when
+        // `notify` is called. In that case it might've just missed the
+        // readiness of some part of its future, and it needs to immediately
+        // call `poll` all over again as soon as the current `poll` is done.
+        // Setting the `ready` flag makes this happen by skipping the next
+        // `park`.
         self.ready.store(true, Ordering::SeqCst);
         self.thread.unpark()
```
oconnor663: Thinking out loud about a side issue that probably doesn't matter: If `notify` is called while the target thread is running, it will set _two_ ready flags, both `self.ready` right here and also the buffered unpark flag set by `std::thread::Thread::unpark`. As we've been discussing re: #132, it's possible that that buffered unpark might get eaten by user code at any point. But if that doesn't happen, then I think the result is that `poll` will get called again _twice_:

1. `notify` comes in while a `poll` is running.
2. The wait loop finishes the current `poll` and calls `ThreadUnpark::park`.
3. The `ready` flag is true, so `park` swaps it and short-circuits.
4. The wait loop calls `poll` again and then `ThreadUnpark::park` again. (This satisfies `notify`'s contract.)
5. The `ready` flag is false, so `park` calls `thread::park`. But the "buffered unpark" created in step 1 makes it return immediately (maybe).
6. The wait loop calls `poll` again and then `ThreadUnpark::park` again. (This is arguably wasted work.)
7. Finally `park` calls `thread::park` and blocks.

I don't think this is a big deal. `std::thread::park` is allowed to spuriously return whenever it feels like it anyway. I suppose if we _really_ cared about avoiding this, we could do something like this:

```rust
fn park(&self) {
    // Clear any buffered unparks on the current thread, with a park that's
    // guaranteed to return immediately.
    self.thread.unpark();
    thread::park();

    // Do all the other stuff we were doing before.
    if self.ready.swap(false, Ordering::SeqCst) {
        return;
    }
    thread::park();
    self.ready.store(false, Ordering::SeqCst);
}
```

But I'm not sure that's actually correct, and I don't know how expensive it is, and I don't know if the extra complexity is worth it anyway.

oconnor663: (Apologies for all the longwindness here.) Thinking more about reordering atomic operations in the _current_ (prior to this PR) version of this code. Is the compiler/CPU allowed to reorder the `self.ready.store` and the `self.thread.unpark` in our implementation of `ThreadUnpark::notify`? If so, I worry the following race is possible:

1. The main thread is in the middle of a `poll`.
2. The other thread calls `ThreadUnpark::notify`.
3. `notify` calls `self.thread.unpark()` _first_ (assuming this reordering is possible).
4. The main thread does some more work, which (a la #132) happens to consume the buffered unpark.
5. The main thread finishes `poll` and calls `ThreadUnpark::park`.
6. The main thread loads `self.ready` and finds that it's false.
7. Finally control resumes in `notify` on the other thread and (again assuming reordering is possible) sets `self.ready`, too late for the load above to see it.
8. The main thread calls `thread::park()` and blocks because of step 4, violating `notify`'s contract.

Is there some larger rule like "no memory operations will ever get reordered around acquiring or releasing any mutex" that protects us here? I'm afraid I don't know enough about how all these primitives work.

alexcrichton: An excellent point about the spurious wakeup! It's true that up to this point it hasn't been considered too too serious because most implementations of `poll` are "fast enough" and otherwise spurious wakeups are expected every now and then.

In general though yeah the usage of `SeqCst` should mean that nothing is reordered here too much, in theory.

carllerche: @oconnor663 A mutex is the strongest form of memory ordering, so there won't be any reordering around it.

alexcrichton:
This has been inactive for awhile so I'm going to close this for now, but let me know if this is in error!

oconnor663:
It looks like the functions this was fixing have since been rewritten anyway. Thanks for the cleanup.

PR #567, opened 2017-08-29
Add Void type and expect_no_error method

srijs:
As discussed in #476.

Went with the name `Void` for now, but as far as I can see we have the following choices:

- Haskell: [`Void`](https://hackage.haskell.org/package/base-4.10.0.0/docs/Data-Void.html) (could be confusing for people coming from C-like languages?)
- Elm: [`Never`](http://package.elm-lang.org/packages/elm-lang/core/5.1.1/Basics#Never) (works nicely for the error case IMO, smth can  "never" fail)
- Scala: [`Nothing`](http://www.scala-lang.org/api/2.9.1/scala/Nothing.html) (subtype of every other type, referring to the bottom of the OO type hierarchy)
- Type Theory: [`Bottom`](https://en.wikipedia.org/wiki/Bottom_type)

I also added a combinator method `expect_no_error` that can be used to convert existing cases of `Future<Error=()>` into `Future<Error=Void>`. It might also make sense to add a function going the other way, e.g. `Future<Error=Void>` into `Future<Error=()>` in order to conform with existing APIs expecting `Error=()`. I'm also happy to remove this from the PR for now and open a separate one once we landed the `Void` type.

Let me know what you think!

### DIFF
--- a/src/future/mod.rs
+++ b/src/future/mod.rs
@@ -115,7 +115,8 @@ if_std! {
     }
 }
 
-use {Poll, stream};
+use {Async, Never, Poll, stream};
+use never::InfallibleResultExt;
 
 /// Trait for types which are a placeholder of a value that may become
 /// available at some later point in time.
@@ -271,6 +272,16 @@ pub trait Future {
     /// error.
     fn poll(&mut self) -> Poll<Self::Item, Self::Error>;
 
+    /// Poll a future that can not fail.
+    ///
+    /// Works similar to `poll`, except that it returns an `Async` value directly
+    /// rather than `Poll`.
+    fn poll_infallible(&mut self) -> Async<Self::Item>
+        where Self: Future<Error=Never> + Sized
+    {
+        self.poll().infallible()
+    }
+
     /// Block the current thread until this future is resolved.
     ///
     /// This method will consume ownership of this future, driving it to
@@ -297,6 +308,17 @@ pub trait Future {
         where Self: Sized
     {
         ::executor::spawn(self).wait_future()
+    }
+
+    /// Wait on a future that can not fail.
+    ///
+    /// Works similar to `wait`, except that it returns the item directly rather
+    /// than a `Result`.
+    #[cfg(feature = "use_std")]
+    fn wait_infallible(self) -> Self::Item
+        where Self: Future<Error=Never> + Sized
+    {
+        self.wait().infallible()
     }
 
     /// Convenience function for turning this future into a trait object which
@@ -410,8 +432,6 @@ pub trait Future {
     {
         assert_future::<Self::Item, E, _>(map_err::new(self, f))
     }
-
-
 
     /// Map this future's error to any error implementing `From` for
     /// this future's `Error`, returning a new future.
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -209,6 +209,8 @@ pub mod sync;
 #[cfg(feature = "use_std")]
 pub mod unsync;
 
+mod never;
+pub use never::Never;
 
 if_std! {
     #[doc(hidden)]
--- a/src/never.rs
+++ b/src/never.rs
@@ -1,0 +1,62 @@
+use core::fmt;
+#[cfg(feature = "std")]
+use std::error::Error;
+
+/// A value that can never happen!
+///
+/// For context:
+///
+/// - The boolean type `bool` has two values: `true` and `false`
+/// - The unit type `()` has one value: `()`
+/// - The empty type `Never` has no values!
+///
+/// You may see it in the wild in `Future<Error = Never>`,
+/// which means that this future will never fail.
+#[derive(Copy, Clone, Debug)]
+pub enum Never {}
+
+impl Never {
+    /// Convert into any other type.
+    ///
+    /// Since it is impossible for `Never` to exist,
+    /// we can exploit this to convert it into any type.
+    ///
+    /// This is also called the [Principle of Explosion]
+    /// (https://en.wikipedia.org/wiki/Principle_of_explosion).
+    #[inline(always)]
+    pub fn never<T>(self) -> T {
+        match self {}
+    }
+}
+
+impl fmt::Display for Never {
+    fn fmt(&self, _: &mut fmt::Formatter) -> fmt::Result {
+        self.never()
+    }
+}
+
+#[cfg(feature = "std")]
+impl Error for Never {
+    fn description(&self) -> &str {
+        self.never()
+    }
+
+    fn cause(&self) -> Option<&Error> {
+        self.never()
+    }
+}
+
+pub trait InfallibleResultExt {
+    type Item;
+
+    fn infallible(self) -> Self::Item;
+}
+
+impl<T> InfallibleResultExt for Result<T, Never> {
+    type Item = T;
+
+    #[inline]
+    fn infallible(self) -> Self::Item {
+        self.unwrap_or_else(|x| x.never())
+    }
+}
--- a/src/stream/mod.rs
+++ b/src/stream/mod.rs
@@ -15,7 +15,7 @@
 //!
 //! [online]: https://tokio.rs/docs/getting-started/streams-and-sinks/
 
-use {IntoFuture, Poll};
+use {IntoFuture, Never, Poll};
 
 mod iter;
 #[allow(deprecated)]
@@ -58,6 +58,8 @@ mod then;
 mod unfold;
 mod zip;
 mod forward;
+#[cfg(feature = "use_std")]
+mod wait_infallible;
 pub use self::and_then::AndThen;
 pub use self::chain::Chain;
 pub use self::concat::{Concat, Concat2};
@@ -88,6 +90,8 @@ pub use self::then::Then;
 pub use self::unfold::{Unfold, unfold};
 pub use self::zip::Zip;
 pub use self::forward::Forward;
+#[cfg(feature = "use_std")]
+pub use self::wait_infallible::WaitInfallible;
 use sink::{Sink};
 
 if_std! {
@@ -239,6 +243,17 @@ pub trait Stream {
         where Self: Sized
     {
         wait::new(self)
+    }
+
+    /// Wait on a stream that can not fail.
+    ///
+    /// Works similar to `wait`, except that it yields the items directly rather
+    /// than a `Result`.
+    #[cfg(feature = "use_std")]
+    fn wait_infallible(self) -> WaitInfallible<Self>
+        where Self: Stream<Error=Never> + Sized
+    {
+        wait_infallible::new(self)
     }
 
     /// Convenience function for turning this stream into a trait object.
--- a/src/stream/wait_infallible.rs
+++ b/src/stream/wait_infallible.rs
@@ -1,0 +1,30 @@
+use {Stream, Never};
+use stream::Wait;
+use never::InfallibleResultExt;
+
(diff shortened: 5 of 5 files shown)


carllerche:
I don't really know if this crate wants to add a `Void` type. If using some sort of `Void` type is something that the lib really wants to do, then it probably should be provided by a 1.x crate entirely dedicated to this type so that other crates can also depend on it.

srijs:
Worth pointing out that this crate already exists as `void`, with nearly 4k daily downloads, and being dependent on by crates like `unreachable`, `nix` and `rotor`.

I'd be happy to change the PR to use this crate. Curious what @alexcrichton thinks.

FWIW I think there are also good arguments for `Void` being part of this crate. The futures/tokio ecosystem is the first place where I have seen a major use-case for `Void` being part of a public API (as documented by the linked futures and tokio issues).

I'll do some more research into how the existing `void` crate is used by it's dependents, and also in which ways other language ecosystems use uninhabitable types.

alexcrichton:
I've been feeling lately that the `futures` ecosystem needs to canonicalize around *some* definition of an empty type. Fundamentally not all futures need `type Error;`, and we're just seeing that pop up from time to time, and I think everyone who doesn't need `type Error;` should be interacting with the `futures` crate in similar fashions.

I also don't really see a huge amount of benefit in sharing a an empty type with other crates in the Rust ecosystem. It seems like the type will never really need to unify as the whole point is to "get out of the world of `Result`" as quickly as possible. In that sense I don't see a need here personally that it's critical to pick up an existing abstraction.

The most valuable thing here, though, I believe is how you interact with `Future<Error = Void>`. Right now this adds `expect_no_error` which is sort of like "`unwrap` for futures" but I'd like to also see combinators along the lines of an infallible `wait` if the error type is `Void`, an infallible `poll`, etc. @srijs would you mind gaming out these areas?

srijs:
Not at all! Couple of questions/thoughts on my mind:

- Some of the cases you mention (infallible `wait` and `poll`) as well as potentially others (e.g. infallible `then`/`and_then`) could also be covered by some sort of `InfallibleResultExt` extension trait. Any thoughts on that?

- Should the set of infallible methods live in the top level futures trait? If yes, does it make sense to agree on a standard naming convention?

- Alternatively, I was reading Alexis RFC on fallible/infallible allocation in collections, and it got me thinking: How about a method like `Future::infallible(self) -> InfallibleFuture<Self> where Self: Future<Error=Void>`, where `InfallibleFuture<F>` can implement the infallible method variants w/o any name clashes, and also implements `IntoFuture<Future=F>`?

Let me know what you think. Also happy to explore multiple avenues.

alexcrichton:
Hm good questions! I'm not really sure what the best answers are though or what this should look like in the long run. The idea of an `InvallibleFuture<F>` seems plausible but I'm not sure what the relative costs and such would be to alternate strategies.

My thinking originally was that the methods and extensions would live on the `Future` trait itself, but it's a good point that maybe they shouldn't! If you wanna play around locally with various solutions are and see what the ergonomics are, maybe that'd help guide us? I could also try to generate some use cases for infallible futures and see if that helps.

srijs:
@alexcrichton just pushed the result of my experimentation to the branch.

I have experimented with three approaches; the `InfallibleFuture` struct, an `InfallibleFutureExt` extension trait, as well as adding the combinators directly to the `Future` trait. I've stuck with the latter, since the first had slightly weird ergonomics and could be confusing (for instance, does an infallible async function now return `InfallibleFuture<impl Future<...>>` or `impl Future<...>`), and the second seemed was making things more cumbersome by requiring an additional import at every point where you want to e.g. chain another future with some infallible future.

The most important combinators for both streams and futures are `expect_infallible` and `infallible_cast_err`. These two combinators are duals of each other, in a sense; the former converts a future with an arbitrary error into an infallible future, while the latter converts an infallible future into a future with an arbitrary error type. 

I also added a few convenience combinators that make it easier to combine infallible futures with fallible ones, w/o requiring the use of `infallible_cast_err` in lots of places.

Additionally, both Future and Stream now have an infallible variant of `wait`, which returns a simple value instead of a result.

Please let me know what you think so far, if I should remove combinators, or add different ones, etc!

srijs:
@alexcrichton Friendly ping, would appreciate some feedback. Let me know if you're busy currently, then I'll back off a little :)

alexcrichton:
Oh oops sorry I thought I commented on this...

In any case, thanks for the investigation here! This is looking pretty nifty to me. My main thought is that this seems like it may be biasing a bit much towards *creating* a `Future` with an error type of `Never`. I'd imagine, for type inference reasons, that we'd instead want to go the other way and get futures to play nicely with other types in the world without `Never`. 

For example this adds an `async_fn` future with an error type of `Never`, but I think the equivalent `poll_fn` should also work, right? That is, if you use `poll_fn` and don't actually constrain the error type, then it's a generic type variable that could then also get bound to `Never`? Although I could imagine in some circumstances this'd require an explicit type annotation to give it *some* value which can certainly be annoying.

Similarly though I'm wary of a combinator like `expect_infallible` as it's essentially "unwrap for futures", but the return value still has an error type in it. For example `Result::unwrap` discards the error and lets you work with a T, but `Future::expect_infallible` forces you to still work with a future and the possibility of an error (it doesn't statically remove the error type).

Some other combinators I'd expect to sort of fall out of the various parts of other combinators, for example:

```rust
a.infallible_cast_err() => a.map_err(|e| e.never()) 
```

I was thinking that `infallible_then` may fall into this category but I think it may not quite be the case...

I guess in general, I wonder if we can possibly trim this down to be a little more conservative? We've always got the option of adding more combinators and types and such in the future, but removing things tends to be pretty hard...

srijs:
Thanks for the feedback @alexcrichton!

I appreciate that the number of new combinators added here is quite high, and am happy to cut back.

> My main thought is that this seems like it may be biasing a bit much towards creating a `Future` with an error type of `Never`. I'd imagine, for type inference reasons, that we'd instead want to go the other way and get futures to play nicely with other types in the world without `Never`.

I agree in so far that when a generic combinator exists that will work for futures of any error type, and thus works for `Never` as well, a specialised variant is probably not necessary (`async_fn` probably falls into this camp). Furthermore if the future is already generic, there might be less need for using `infallible_cast_err` often, which is a good thing.

However I believe that in some cases this results in having to stick a lot of manual type arguments on function calls. But for these cases it's probably okay to not include specialised variants for now and just see what issues crop up in practice and need fixing.

> Some other combinators I'd expect to sort of fall out of the various parts of other combinators

I guess if you took it to the extreme, literally all combinators presented here can be emulated via already existing ones:

- `a.infallible_cast_err()` as you pointed out is `a.map_err(|e| e.never())`
- `async_fn(f)` is `poll_fn::<_, Never, _>(|| Ok(f()))`
- `a.expect_infallible()` is `a.then(|r| ok::<_, Never>(r.expect("...")))`
- `a.infallible_then(f)` is `a.then(|r| f(r.unwrap()))`
- `a.infallible_poll()` is `a.poll().unwrap()`
- `a.infallible_wait()` is `a.wait().unwrap()`

Although all of these combinators provide a somewhat less-noisy way of achieving the same outcome compared to existing combinators, broadly speaking it seems they fall into two camps:

- The type of combinator mentioned earlier that assists type inference in some cases / might require less explicit type annotations in some cases (`infallible_cast_err`, `async_fn` and `expect_infallible`)
- The type of combinator that provides a safe way of unwrapping (`infallible_then`, `infallible_poll` and `infallible_wait`)

Going back to what you said in an earlier comment:

> The most valuable thing here, though, I believe is how you interact with Future<Error = Void>. Right now this adds expect_no_error which is sort of like "unwrap for futures" but I'd like to also see combinators along the lines of an infallible wait if the error type is Void, an infallible poll, etc.

The combinators you reference here fall into the latter camp, so would it be fair to say that this is where you see the bigger value?

If that's the case, I think I agree. Having "safe" variants for result unwrapping helps a lot in writing safe code, and prevents me from hitting false positives when grepping for `unwrap` to identify places in a code-base where an error might lead to panic.

So my proposal would be to remove `infallible_cast_err`, `async_fn` and `expect_infallible` and retain `infallible_then`, `infallible_poll` and `infallible_wait`.

An alternative could be to expose the `InfallibleResultExt` extension trait so e.g. `.infallible_poll()` becomes `.poll().unwrap_infallible()`, however I believe `infallible_poll` to be more ergonomic since it doesn't require importing an additional extension trait.

Let me know what you think!

alexcrichton:
@srijs that sounds great to me! I agree that the most value from this is avoiding the false positive of "I need to audit this `unwrap`". I'm *slightly* worried that it's more ergonomic to unwrap than it is to call `infallible_foo`, but that's something we can perhaps smooth out over time.

carllerche:
I agree w/ @alexcrichton that all the combinators should be punted on. They can always be added later but not removed. It also adds a bunch of noise to the docs and presents decision points.

srijs:
@alexcrichton Updated the PR to only include `poll` and `wait` variants. I've also decided to remove  `infallible_then`, since I realised it's unnecessary if the future is already universally quantified over `Error`.

I do believe once we change traits like `Receiver`/`UnboundedReceiver` to impl `Future`/`Stream` with `type Error = Never`, we might see a stronger need for a combinator like `infallible_cast_err` for ergonomic reasons (perhaps with a better name though).

In terms of ergonomics for the `poll` and `wait` variants, do you have other ideas how we could make them more ergonomic to use? Is it the conciseness of the name, or something else that is bothering you?

@carllerche Not sure if I understand what you are saying. Are you proposing to drop all combinators for now?

srijs:
I also have a slightly different proposal I've been thinking about for a bit. If you feel it's to complex, I'm happy to table it, but I thought I'd just put it out there.

The goal of the proposal would be to safely take advantage of infallibility for things like `Receiver`/`UnboundedReceiver` as well as some of the cases in other crates (primarily tokio), without forcing a breaking change.

I'm thinking along the lines of this (and a similar setup for `Stream`):

```
trait InfallibleFuture {
  type Item;
  fn poll_infallible(&mut self) -> Async<Self::Item>;
}

impl<F: Future<Error=Never>> InfallibleFuture for F {
  type Item = <F as Future>::Item;
  fn poll_infallible(&mut self) -> Async<Self::Item> {
    self.poll().infallible()
  }
}
```

This should then allow us to then implement the infallible trait for existing structs:

```
impl<T> InfallibleStream for Receiver<T> {
  type Item = T;
  fn poll_infallible(&mut self) -> Option<Async<T>> {
    ...
  }
}
```

As soon as we change `Receiver` to impl `Stream<Error=Never>` in a future breaking release, we could simply remove the custom impl for`InfallibleFuture`.

An alternative to that which would be more in-line with the current approach in this branch would be  to introduce new structs such as `InfallibleReceiver` that implements `Stream<Error=Never>`, move the logic from `Receiver` to `InfallibleReceiver`, change `Receiver` to wrap `InfallibleReceiver` to re-use the logic, and then expose a method like `Receiver::infallible` that returns `InfallibleReceiver`.

srijs:
I've gamed out the alternative strategy in #591

srijs:
Friendly ping, @alexcrichton. Apologies if you are occupied at the moment!

alexcrichton:
Hey @srijs sorry for the delay, I've been super busy and haven't had a chance to get back to this yet :(

I haven't forgotten about it though!

alexcrichton:
Ok! We've started working on 0.2 and we've got a `Never` type there. I'm going to close this for now in favor of future work on that branch.

PR #568, opened 2017-08-31
Add a `prelude` module to the crate

alexcrichton:
The intention here is that almost all modules need at least a subset of these
reexports and they shouldn't otherwise cause problems to import extra items.
Similar to the standard library prelude, only not automatically included!


alexcrichton:
I've updated the documentation of the `Future` trait, but if we'd like to pursue this then I'd like to update the `Stream`/`Sink` traits as well in addition to the tests.

PR #569, opened 2017-09-03
Add `AsyncSink::map`

semyon2105:
I noticed `map` method was missing for `AsyncSink` when I was making a transport wrapper.


alexcrichton:
Thanks!

PR #570, opened 2017-09-04
Implements `sync::mpsc::tip` abstraction

tailhook:
This is a stream that only remembers last value and also receives
updates.

The common use case for such thing is when you have a coroutine that
refreshes a value (e.g. polls DNS server for an address), and a coroutine
that uses that value in possibly blocking fashion (e.g. fetches some
data from the address). Using unbounded stream for this case means we
can have a memory leak when a consumer is too slow. And using `channel(1)`
preserves the first value while what we really need is the most recent
value generated by a producer.

The name "tip" is in the sense of end or peak, not in the sense of gratitude :) But suggestions of the better name are appreciated.

What do you think?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,7 +12,7 @@ matrix:
       after_success:
         - travis-cargo doc-upload
     - os: linux
-      rust: 1.13.0
+      rust: 1.15.0
       script: cargo test
 sudo: false
 script:
--- a/src/sync/slot.rs
+++ b/src/sync/slot.rs
@@ -1,0 +1,170 @@
+//! An unbounded channel that only stores last value sent
+
+use std::sync::{Arc, Weak, Mutex};
+
+use task::{self, Task};
+use {Sink, Stream, AsyncSink, Async, Poll, StartSend};
+
+/// Slot is very similar to unbounded channel but only stores last value sent
+///
+/// I.e. if you want to send some value between from producer to a consumer
+/// and if consumer is slow it should skip old values, the slot is
+/// a structure for the task.
+
+/// The transmission end of a channel which is used to send values
+///
+/// If the receiver is not fast enough only the last value is preserved and
+/// other ones are discarded.
+#[derive(Debug)]
+pub struct Sender<T> {
+    inner: Option<Weak<Mutex<Inner<T>>>>,
+}
+
+/// The receiving end of a channel which preserves only the last value
+#[derive(Debug)]
+pub struct Receiver<T> {
+    inner: Arc<Mutex<Inner<T>>>,
+}
+
+/// Error type for sending, used when the receiving end of a channel is
+/// dropped
+#[derive(Clone, PartialEq, Eq, Debug)]
+pub struct SendError<T>(T);
+
+#[derive(Debug)]
+struct Inner<T> {
+    value: Option<T>,
+    task: Option<Task>,
+}
+
+trait AssertKindsSender: Send + Sync + Clone {}
+impl AssertKindsSender for Sender<u32> {}
+
+trait AssertKindsReceiver: Send + Sync {}
+impl AssertKindsReceiver for Receiver<u32> {}
+
+impl<T> Sender<T> {
+    /// Sets the new new value of the stream and notifies the consumer if any
+    pub fn swap(&self, value: T) -> Result<Option<T>, SendError<T>> {
+        let result;
+        // Do this step first so that the lock is dropped when
+        // `unpark` is called
+        let task = {
+            let strong = self.inner.as_ref()
+                .expect("sending to a closed slot");
+            if let Some(ref lock) = strong.upgrade() {
+                let mut inner = lock.lock().unwrap();
+                result = inner.value.take();
+                inner.value = Some(value);
+                inner.task.take()
+            } else {
+                return Err(SendError(value));
+            }
+        };
+        if let Some(task) = task {
+            task.notify();
+        }
+        return Ok(result);
+    }
+}
+
+impl<T> Sink for Sender<T> {
+    type SinkItem = T;
+    type SinkError = SendError<T>;
+    fn start_send(&mut self, item: T) -> StartSend<T, SendError<T>> {
+        self.swap(item)?;
+        Ok(AsyncSink::Ready)
+    }
+    fn poll_complete(&mut self) -> Poll<(), Self::SinkError> {
+        Ok(Async::Ready(()))
+    }
+    fn close(&mut self) -> Poll<(), Self::SinkError> {
+        // Do this step first so that the lock is dropped *and*
+        // weakref is dropped when `unpark` is called
+        let task = {
+            if let Some(weak) = self.inner.take() {
+                if let Some(ref lock) = weak.upgrade() {
+                    drop(weak);
+                    let mut inner = lock.lock().unwrap();
+                    inner.task.take()
+                } else {
+                    None
+                }
+            } else {
+                None
+            }
+        };
+        // notify on any drop of a sender, so eventually receiver wakes up
+        // when there are no senders and closes the stream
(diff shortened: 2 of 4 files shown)


src/sync/mpsc/tip.rs
```diff
@@ -0,0 +1,135 @@
...
+}
+
+impl<T> Stream for TipReceiver<T> {
+    type Item = T;
+    type Error = ();  // actually void
+    fn poll(&mut self) -> Poll<Option<Self::Item>, Self::Error> {
+        let result = {
+            let mut inner = self.inner.lock().unwrap();
+            if !inner.closed {
+                inner.task = Some(task::current());
```
alexcrichton: Ideally this'd only happen if `NotReady` is returned, right?

src/sync/mpsc/tip.rs
```diff
@@ -0,0 +1,135 @@
...
+/// and if consumer is slow it should may skip old values, the tip is
+/// a structure for the task.
+
+/// The transmission end of a channel which is used to send values
+///
+/// If the receiver is not fast enough only the last value is preserved and
+/// other ones are discarded.
+#[derive(Debug)]
+pub struct TipSender<T> {
+    inner: Arc<Mutex<Inner<T>>>,
```
alexcrichton: One trick I've done in the past is for all senders to store a `Weak` and the receiver to store an `Arc`. That way there's no need for an extra `closed` flag and all the races and such should in theory be taken care of.

src/sync/mpsc/tip.rs
```diff
@@ -0,0 +1,135 @@
...
+
+trait AssertKindsSender: Send + Sync + Clone {}
+impl AssertKindsSender for TipSender<u32> {}
+
+trait AssertKindsReceiver: Send + Sync {}
+impl AssertKindsReceiver for TipReceiver<u32> {}
+
+impl<T> TipSender<T> {
+    /// Sets the new new value of the stream and notifies the consumer if any
+    pub fn set(&self, value: T) {
```
alexcrichton: This may want to return some form of `Result` to reacquire ownership of this value if the receiver has gone away

src/sync/mpsc/tip.rs
```diff
@@ -0,0 +1,135 @@
...
+        self.set(item);
+        Ok(AsyncSink::Ready)
+    }
+    fn poll_complete(&mut self) -> Poll<(), Self::SinkError> {
+        Ok(Async::Ready(()))
+    }
+    fn close(&mut self) -> Poll<(), Self::SinkError> {
+        let task = {
+            let mut inner = self.inner.lock().unwrap();
+            inner.closed = true;
```
alexcrichton: I think this may not be quite right? This is only done when all senders have gone away, right?

I'd imagine that doing this in the destructor is fine, `close` is generally intended for shutdown logic like "tell the remote end that we're going away"

src/sync/mpsc/tip.rs
```diff
@@ -0,0 +1,135 @@
+use std::sync::{Arc, Mutex};
+
+use task::{self, Task};
+use {Sink, Stream, AsyncSink, Async, Poll, StartSend};
+
+/// Tip is very similar to unbounded channel but only stores last value sent
+///
+/// I.e. if you want to send some value between from producer to a consumer
+/// and if consumer is slow it should may skip old values, the tip is
```
carllerche: `should may` typo

src/sync/mpsc/tip.rs
carllerche: I'd consider naming this `swap` and always return the previous value.

So the fn would be:

```
pub fn swap(&self, value: T) -> Result<Option<T>, SendError<T>> { ... }
```

hawkw:
> But suggestions of the better name are appreciated.

How about `Single`?

src/sync/mpsc/tip.rs
```diff
@@ -0,0 +1,135 @@
...
+        Ok(Async::Ready(()))
+    }
+}
+
+impl<T> Drop for TipSender<T> {
+    fn drop(&mut self) {
+        // if there are only two strong counts: this one and receiver we
+        // should notify receiver that stream is closed. If these two
+        // references are just two senders the wake up isn't problematic.
+        if Arc::strong_count(&self.inner) == 2 {
```
twmb: Does this race if two senders in two threads drop at the same time? Wouldn't both check this, see 3, and do nothing?

davidblewett:
This is conceptually similar to log compaction in Kafka ( https://kafka.apache.org/documentation/#compaction ), so maybe look there for names?

tailhook:
Okay, next take:
1. Named it `sync::slot::channel`. (Maybe `Head` from @davidblewett's link?)
2. `swap` instead of `set` with both ways of value recovery
3. Eliminated race-condition in `Drop`, but checking in receiver (so waking up receiver on each sender drop, I think it shouldn't be a problem)
4. `close()` now works by dropping Weak ref (we need to drop it in destructor anyway to avoid race conditions)

Overall looks like all suggestions are fixed. Also rebased on top of the new master.

src/stream/mod.rs.orig
```diff
@@ -0,0 +1,840 @@
+//! Asynchronous streams
```
alexcrichton: Stray file?

.travis.yml
```diff
@@ -12,7 +12,7 @@ matrix:
       after_success:
         - travis-cargo doc-upload
     - os: linux
-      rust: 1.13.0
+      rust: 1.15.0
```
alexcrichton: How come this was required?

src/sync/slot.rs
```diff
@@ -0,0 +1,154 @@
...
+        if let Some(task) = task {
+            task.notify();
+        }
+        return Ok(result);
+    }
+}
+
+impl<T> Sink for Sender<T> {
+    type SinkItem = T;
+    type SinkError = SendError<T>;  // actually void
```
alexcrichton: I think the `// actually void` comment here isn't true any more?

src/sync/slot.rs
```diff
@@ -0,0 +1,154 @@
...
+        }
+    }
+}
+
+/// Creates an in-memory Stream which only preserves last value
+///
+/// This method is somewhat similar to `channel(1)` but instead of preserving
+/// first value sent (and erroring on sender side) it replaces value if
+/// consumer is not fast enough and preserves last values sent on any
+/// poll of a stream.
```
alexcrichton: Could this include a doc example showing the overwriting behavior?

.travis.yml
tailhook: It's described in the commit. We need `Arc::weak_count`

tailhook:
Added an example and removed a stray file and outdated commit.

.travis.yml
alexcrichton: Ah sorry, missed that. Could this use `get_mut` to avoid the version bump here?

tailhook: https://github.com/tailhook/futures-rs/commit/fd741a31710d87483bada9390ca69adb85b1d5db

Code looks more complex (we can't check at the same place because of borrow checker, checking before is racy). 

Also, I'm not sure if [cleaning task reference](https://github.com/tailhook/futures-rs/commit/fd741a31710d87483bada9390ca69adb85b1d5db#diff-5d38d21573ff6c00c115f52e230ab8beR127) in on EOS is a good idea.

So is there any reason to not to bump version?

alexcrichton: Er sorry thought I replied to this...

In any case, there was some consternation upgrading to 1.13.0 recently so I was just wondering if it was "easy" to stick to 1.13.0, but looks like no! That's fine, looks good to me.

alexcrichton:
Thanks @tailhook!

PR #575, opened 2017-09-09
Add sync::mpsc::Sender::poll_ready

danburkert:
This is useful a API for situations where it's advantageous to check for
channel capacity before creating or procuring an item to send.

### DIFF
--- a/src/sync/mpsc/mod.rs
+++ b/src/sync/mpsc/mod.rs
@@ -39,7 +39,7 @@
 // Since most of this work is lock-free, once the work starts, it is impossible
 // to safely revert.
 //
-// If the sender is unable to process a send operation, then the the curren
+// If the sender is unable to process a send operation, then the the current
 // task is parked and the handle is sent on the parked task queue.
 //
 // Note that the implementation guarantees that the channel capacity will never
@@ -457,6 +457,25 @@ impl<T> Sender<T> {
         // queue
         let state = decode_state(self.inner.state.load(SeqCst));
         self.maybe_parked = state.is_open;
+    }
+
+    /// Polls the channel to determine if there is guaranteed to be capacity to send at least one
+    /// item without waiting.
+    ///
+    /// Returns `Ok(Async::Ready(_))` if there is sufficient capacity, or returns
+    /// `Ok(Async::NotReady)` if the channel is not guaranteed to have capacity. Returns
+    /// `Err(SendError(_))` if the receiver has been dropped.
+    ///
+    /// # Panics
+    ///
+    /// This method will panic if called from outside the context of a task or future.
+    pub fn poll_ready(&mut self) -> Poll<(), SendError<()>> {
+        let state = decode_state(self.inner.state.load(SeqCst));
+        if !state.is_open {
+            return Err(SendError(()));
+        }
+
+        Ok(self.poll_unparked())
     }
 
     fn poll_unparked(&mut self) -> Async<()> {
--- a/tests/mpsc.rs
+++ b/tests/mpsc.rs
@@ -37,11 +37,12 @@ fn send_recv_no_buffer() {
     // Run on a task context
     lazy(move || {
         assert!(tx.poll_complete().unwrap().is_ready());
+        assert!(tx.poll_ready().unwrap().is_ready());
 
         // Send first message
-
         let res = tx.start_send(1).unwrap();
         assert!(is_ready(&res));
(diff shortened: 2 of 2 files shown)


danburkert:
Failed due to using struct initialization shorthand syntax in the test; squashed a fix into the stress test commit.

src/sync/mpsc/mod.rs
```diff
@@ -459,6 +459,23 @@ impl<T> Sender<T> {
...
 
+    /// Polls the channel to determine if there is guaranteed to be capacity to send at least one
+    /// item without waiting.
+    ///
+    /// Returns `Ok(Async::Ready(_))` if there is sufficient capacity, or returns
+    /// `Ok(Async::NotReady)` if the channel is not guaranteed to have capacity. Returns `Err(_)`
+    /// if the receiver has been dropped.
+    ///
+    /// This method may only be called from inside the context of a task or future.
+    pub fn poll_ready(&mut self) -> Poll<(), ()> {
```
alexcrichton: Perhaps the error here could be some form of `SendError` to have roughly the same error type as the signatures below?

src/sync/mpsc/mod.rs
```diff
@@ -459,6 +459,23 @@ impl<T> Sender<T> {
...
     }
 
+    /// Polls the channel to determine if there is guaranteed to be capacity to send at least one
+    /// item without waiting.
+    ///
+    /// Returns `Ok(Async::Ready(_))` if there is sufficient capacity, or returns
+    /// `Ok(Async::NotReady)` if the channel is not guaranteed to have capacity. Returns `Err(_)`
+    /// if the receiver has been dropped.
+    ///
+    /// This method may only be called from inside the context of a task or future.
```
alexcrichton: Could this be under a `# Panics` section in the documentation?

danburkert:
> Thanks! Could this also have a clause in the documentation that due to the multiple nature of senders that the return value does not guarantee that the next call to start_send will succeed?

This is actually guaranteed; `poll_ready` will only indicate ready when the sender's reserved slot is vacant.  Without this guarantee the method would be much less useful (at least for the use-case I have in mind for it).  The attached stress test verifies that the guarantee holds by way of asserting on `start_send` returning ready.

src/sync/mpsc/mod.rs
danburkert: Done.

src/sync/mpsc/mod.rs
danburkert: Yes, the signature could change to be `pub fn poll_ready(&mut self) -> Poll<(), SendError<()>>;`, I went with plain `()` to keep it simple, since `SendError<()>` is pretty much equivalent, but if you prefer `SendError<()>` I can switch it over.

carllerche:
What @danburkert says is correct. If `poll_ready` returns `Ready`, then the next call to `start_send` is guaranteed to succeed **unless** the receive half closed.

alexcrichton:
re `()` vs `SendError<()>` -- my main hesitation with using `()` as an error is that elsewhere in `futures` we use `()` as "this can never happen", which while we want to change that may cause confusion if there's multiple definitions of it.

re the guarantees here, is this something we want to commit to long-term? Such a guarantee may preclude future optimizations and/or implementations? Is it really necessary to provide such a guarantee?

carllerche:
@alexcrichton I think returning `SendError<()>` is good (I wasn't arguing against it.

Re: making the guarantee, I think that it should. Being able to provide this guarantee is a pretty fundamental part of the design of the mpsc channel. Also, being able to rely on the guarantee is a requirement for providing sane back pressure support at the `Service` layer.

I think it's fine to have this structure provide the guarantee and if there is an alternate design that can provide different trade-offs, it can be added as a separate structure.

alexcrichton:
Ok sounds like we should leave the documentation guarantee and @danburkert mind changing to return `SendError<()>`?

FWIW The `()` in `SendError<()>` I agree isn't exactly ideal, but I'm hoping `SendError<??>` is better than `()`

danburkert:
> mind changing to return SendError<()>?

Done.

> FWIW The () in SendError<()> I agree isn't exactly ideal, but I'm hoping SendError<??> is better than ()

I agree but I think your argument about `()` indicating non-fallible is persuasive, I've definitely come around to your side of the argument.

alexcrichton:
👍

PR #578, opened 2017-09-11
Add `mpsc::Sender::try_send`

### LINKED ISSUE #478: MPSC: Send unless blocks

carllerche:
This is an API which serves for sending a message along a mpsc::Sender
without attempting to park the current task, making it safe to invoke
from a helper thread or a non-futures-aware thread.

Even though `try_send` doesn't "park" the current task, it still needs
to internally perform the parking logic. This is due to the fact that
parking is really tracking that a given sender instance is full.

Doing this required refactoring the internals to decouple tracking a
parked sender from having a task::Task value to notify. After this
patch, `SenderTask` has a `is_parked` field which is the authority when
it comes to determining if a sender is blocked or not. The task field
contains the task, if any, that should be notified on unpark.

Closes #478

### DIFF
--- a/src/sync/mpsc/mod.rs
+++ b/src/sync/mpsc/mod.rs
@@ -96,7 +96,7 @@ pub struct Sender<T> {
     // Handle to the task that is blocked on this sender. This handle is sent
     // to the receiver half in order to be notified when the sender becomes
     // unblocked.
-    sender_task: SenderTask,
+    sender_task: Arc<Mutex<SenderTask>>,
 
     // True if the sender might be blocked. This is an optimization to avoid
     // having to lock the mutex most of the time.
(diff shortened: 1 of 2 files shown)


src/sync/mpsc/mod.rs
```diff
@@ -224,7 +295,28 @@ const MAX_CAPACITY: usize = !(OPEN_MASK);
...
 // Sent to the consumer to wake up blocked producers
-type SenderTask = Arc<Mutex<Option<Task>>>;
+#[derive(Debug)]
+struct SenderTask {
+    task: Option<Task>,
+    is_parked: bool,
+}
+
+impl SenderTask {
+    pub fn new() -> Self {
```
alexcrichton: I think the `pub` can be avoided here, right?

carllerche:
Rebased... waiting for CI to pass.

PR #580, opened 2017-09-14
either: add a split_err method

mathstuf:


alexcrichton:
Thanks! I was wondering, though, should this incorporate `Result` into the return value as well?

mathstuf:
Maybe `split_err` isn't so interesting a name as `Either` doesn't exactly deal in errors and results. Maybe `split_b` (adding a new `split_a` for the existing method and deprecating it)?

If `Result` was wanted at the beginning, I'd rather `select2` just make one itself rather than having special cases for `Either`.

It might be interesting to also have an `impl<T> From<Either<T, T>> for T` to help with the any collapsing.

mathstuf:
Hrm, `split_b` also isn't good. Is there a problem with just calling it `split` also? Or does that need `T != U` assertions (is that possible?)?

alexcrichton:
Yeah unfortunately `split` can't work here due to how methods work in Rust, but something like `split_a` and `split_b` sounds pretty reasonable to me! I think we can't quite have that `From` impl as well but we could have an inherent method which does the conversion.

mathstuf:
I used `split_left` and `split_right` since the `A` and `B` aren't really the same (it's taking the left side of each tuple possibility). I also added a `collapse` method for `Either<T, T> -> T`.

alexcrichton:
Sorry for the delay! My only hesitation with the left/right naming is that it doesn't scale beyond either-with-two-type-parameters. If we ever add `Either3`, for example, that may cause inconsistencies in naming?

I'm also not entirely thrilled with the name `collapse` but I can't think of a better name, so seems fine by me!

mathstuf:
The Haskell names for `(a, b) -> a` and `(a, b) -> b` are `fst` and `snd`. The three-tuple variants are `fst3`, `snd3`, and `thd3`.

I don't think I've ever seen a meaningful wrapper around a 3-way `Either`, but `split_center` would work there.

alexcrichton:
My point though is that `split_a` and `split_b` I think have roughly the same connotation, and are also scalable to more variants if we ever add them?

alexcrichton:
I think it's ok to have the factored-out type on the far-left of the tuple (.0)

alexcrichton:
Ok I'm going to close this due to inactivity, but feel free to resubmit!

mathstuf:
I've updated the source branch. Do you want to reopen this or is a separate PR preferred?

mathstuf:
Yeah, it will. We can't add these methods to the external `Either` type. Maybe they can become free-standing functions instead?

alexcrichton:
If merged, could the methods be added to the `either` crate?

mathstuf:
I don't know. There may be some more generic code for it. I'll open an issue over there.

mathstuf:
See bluss/either#17

mathstuf:
Looks like `either` is now a dependency. Let's see what happens on the `either` side.

Issue #529, opened 2017-07-04
Presumed deadlock with sync::mpsc


alexcrichton:
https://travis-ci.org/alexcrichton/futures-rs/jobs/249262983

```
     Running target/debug/deps/mpsc-8f8a597072297523

running 12 tests

test recv_close_gets_none ... ok

test send_recv ... ok

test send_recv_no_buffer ... ok

test send_recv_threads ... ok

test send_shared_recv ... ok

test send_recv_threads_no_capacity ... ok

test stress_drop_sender ... ok

test stress_receiver_multi_task_bounded_hard ... ok

test stress_shared_bounded_hard ... ok

test stress_shared_unbounded ... ok

test tx_close_gets_none ... ok

test stress_close_receiver has been running for over 60 seconds

No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.

Check the details on how to adjust your build configuration on: https://docs.travis-ci.com/user/common-build-problems/#Build-times-out-because-no-output-was-received

The build has been terminated
```

I doubt that's spurious...

carllerche:
linux 64 bit I assume?

alexcrichton:
That specific deadlock was x86_64-apple-darwin, and after a bunch of runs I couldn't reproduce on Linux, although that doesn't necessarily mean it's OSX-specific.

alexcrichton:
Closing due to https://github.com/alexcrichton/futures-rs/pull/582

jpastuszek:
I am actually observing problem with `futures::sync::mpsc::Receiver` not waking up `Sender` in another thread. I can see that thread is still waiting on `mio::sys::unix::kqueue::{{impl}}::select` (`libsystem_kernel.dylib'kevent`) after I call `close()` or drop the `Receiver` in another thread on MacOS in some circumstances.

In my program I have a thread running `core.run()` with future that `send_all()` data read (and split by line) from named pipe (using tokio-file-unix crate). To exit the thread I `close()` the `Receiver` end. This works in case when I call `close()` first and then kill the process at the other end of the pipe. It will fail to unblock `kevent`/`core.run()` if I kill the process first and then `close()` the `Receiver`.
To work it around I have added `oneshot` pair that I `select2()` with that future and after `close()` call on `Receiver` I also `send(())` the `oneshot` channel to make sure the `core.run()` will exit on this future.
[...]

PR #582, opened 2017-09-18
Tweak MPSC stress test

carllerche:
The test ran 10k iterations, and for each iteration it spawned a new thread. However, it did not wait for the thread to complete before starting the next iteration. This can result in a situation where
spawning the threads can run faster than the threads closing and overload the system.

This could have been the cause for #529, though I can't be sure. I was able to reproduce the test hanging by running the test on my laptop 8x in parallel. It ended up never completing due to threads being spawned faster than they could complete. Once I added the join in each iteration, the test completed in a reasonable amount of time.

I would recommend closing #529 and if the issue shows up again, we can investigate further.


alexcrichton:
Awesome, thanks!

PR #583, opened 2017-09-18
add is_canceled to unsync Sender

seanmonstar:
This method exists on the `sync` oneshot, but was missing from `unsync`.


alexcrichton:
👍

PR #584, opened 2017-09-19
Fix `filter` example to use the correct divisor

joepie91:
It seems the divisor and the remainder got swapped around; `x % 0 == 2` doesn't produce a stream of even numbers, and is in fact not valid Rust (due to the divisor being zero).

This PR fixes that by swapping around the divisor and remainder to `x % 2 == 0` :)


alexcrichton:
Heh, thanks!

PR #590, opened 2017-09-21
allow call close on sink::Wait

BusyJay:
There seems to be no way to do "graceful shutdown" after calling wait on sink.


alexcrichton:
Thanks!

PR #593, opened 2017-09-27
undeprecate stream::iter as stream::iter_result

### LINKED ISSUE #576: Reasoning for stream::iter deprecation?
I've been relying on this in some tests, specifically the ability to test Err's in the Stream. It looks like the replacement iter_ok is the replacement, but it only yields Ok, is this correct?

sid0:
While this is less common than `stream::iter_ok`, iterators of `Result`s
do show up in the wild. For example, `std::fs::read_dir` can fail with
an `io::Error` in the middle of reading the directory.

Bring back `stream::iter` as `stream::iter_result`. Keep `stream::iter`
deprecated with the expectation that it will eventually become the same
as `stream::iter_ok`.

Closes #576.

### DIFF
--- a/src/stream/iter_result.rs
+++ b/src/stream/iter_result.rs
@@ -1,15 +1,14 @@
-#![deprecated(note = "implemention moved to `iter_ok` and `iter_result`")]
-#![allow(deprecated)]
-
-use Poll;
-use stream::{iter_result, IterResult, Stream};
+use {Async, Poll};
+use stream::Stream;
 
 /// A stream which is just a shim over an underlying instance of `Iterator`.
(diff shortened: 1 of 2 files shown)


sid0:
Though note that you can get the same behavior with `stream::iter_ok(v).and_then(|x| x)`

alexcrichton:
Thanks! Could the rustfmt be held off until later?

sid0:
Done, thanks!

sid0:
(I did a force push because otherwise blame would be somewhat ruined)

alexcrichton:
👍

PR #597, opened 2017-10-01
Improve thread notify

carllerche:
This depends on #596.

This PR improves `ThreadNotify` performance over 2x in some cases by doing two things:

* Avoiding an `Arc` per call to `wait`. Instead, a TLS is used to cache a `ThreadNotify` handle.
* Use an atomic to avoid locks / signaling when not needed.

### DIFF
--- a/src/task_impl/std/mod.rs
+++ b/src/task_impl/std/mod.rs
@@ -238,14 +238,15 @@ impl<F: Future> Spawn<F> {
     /// to complete. When a future cannot make progress it will use
     /// `thread::park` to block the current thread.
     pub fn wait_future(&mut self) -> Result<F::Item, F::Error> {
-        let unpark = Arc::new(ThreadNotify::new(thread::current()));
+        ThreadNotify::with_current(|notify| {
 
-        loop {
-            match self.poll_future_notify(&unpark, 0)? {
-                Async::NotReady => unpark.park(),
-                Async::Ready(e) => return Ok(e),
+            loop {
+                match self.poll_future_notify(notify, 0)? {
+                    Async::NotReady => notify.park(),
+                    Async::Ready(e) => return Ok(e),
+                }
             }
-        }
+        })
     }
 
     /// A specialized function to request running a future to completion on the
@@ -296,15 +297,17 @@ impl<S: Stream> Spawn<S> {
     /// Like `wait_future`, except only waits for the next element to arrive on
     /// the underlying stream.
     pub fn wait_stream(&mut self) -> Option<Result<S::Item, S::Error>> {
-        let unpark = Arc::new(ThreadNotify::new(thread::current()));
-        loop {
-            match self.poll_stream_notify(&unpark, 0) {
-                Ok(Async::NotReady) => unpark.park(),
-                Ok(Async::Ready(Some(e))) => return Some(Ok(e)),
-                Ok(Async::Ready(None)) => return None,
-                Err(e) => return Some(Err(e)),
+        ThreadNotify::with_current(|notify| {
+
+            loop {
+                match self.poll_stream_notify(notify, 0) {
+                    Ok(Async::NotReady) => notify.park(),
+                    Ok(Async::Ready(Some(e))) => return Some(Ok(e)),
+                    Ok(Async::Ready(None)) => return None,
+                    Err(e) => return Some(Err(e)),
+                }
             }
-        }
+        })
     }
 }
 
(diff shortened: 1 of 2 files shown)


alexcrichton:
Is the improvement over thread park/unpark here critical to land for now? Ideally we'd leave it as is if it's otherwise a "nice to have" and instead upstream the changes to libstd I think?

carllerche:
Avoiding the Arc on each call is pretty important and unrelated to std. Given that change, using a mutex / condvar avoids the extra indirection to 'std::Thread'.

alexcrichton:
Oh sorry yeah the arc caching is fine, but I was wondering if we could avoid adding the atomic optimizations here and instead upstream them.

carllerche:
If you want to pull this into rust proper, feel free to do so. For now, I'll just leave this PR here and it can be pulled in whenever you want to get the perf increase.

alexcrichton:
Was this needed for a particular project? I'm just thinking it's probably best to land the park/unpark improvements upstream and land the arc caching here, but if the park/unpark improvements are needed now then seems fine to land here and upstream in parallel

carllerche:
Yes, this change is required when going from async -> sync in a perf sensitive scenario. Specifically, using a channel to send telemetry info to a sync thread that processes it.

That said, I am currently using a custom waiter for this scenario until perf improvements are land upstream.

src/task_impl/std/mod.rs
```diff
@@ -505,16 +510,63 @@ impl ThreadNotify {
...
+            }
+            IDLE => {},
+            _ => unreachable!(),
+        }
+
+        // Loop until we've been notified
+        loop {
+            m = self.condvar.wait(m).unwrap();
+
+            // Transition back to idle, loop otherwise
```
twmb: Why would this ever not be `NOTIFY`?  Can this be just an atomic store and return?

src/task_impl/std/mod.rs
```diff
@@ -505,16 +510,63 @@ impl ThreadNotify {
...
+        // First, try transitioning from IDLE -> NOTIFY, this does not require a
+        // lock.
+        match self.state.compare_and_swap(IDLE, NOTIFY, Ordering::SeqCst) {
+            IDLE | NOTIFY => return,
+            SLEEP => {}
+            _ => unreachable!(),
+        }
+
+        // The other half is sleeping, this requires a lock
+        let _m = self.mutex.lock().unwrap();
```
twmb: Would it be worthwhile to have a separate `notifier_mutex`? This would allow something like...
```rust
let _nm = match self.notifier_mutex.try_lock() {
    Ok(g) => g,
    Err(e) => match e {
        TryLockResult::Poisoned(e) => panic!(e),
        TryLockResult::WouldBlock => { return ; }
    }
}
let _m = self.mutex.lock().unwrap();
...
```
which would avoid simultaneous notifications sitting on a mutex to notify a thread that will only need the first.

src/task_impl/std/mod.rs
```diff
@@ -505,16 +510,63 @@ impl ThreadNotify {
...
+
+        // Transition from SLEEP -> NOTIFY
+        match self.state.compare_and_swap(SLEEP, NOTIFY, Ordering::SeqCst) {
```
twmb: I think this cas would just need to be a store with notifier_mutex.

src/task_impl/std/mod.rs
twmb: Eh nvm - this would still require keeping one notifier on deck, which `TryLock` does not provide.

src/task_impl/std/mod.rs
carllerche: condvars can wakeup spuriously

twmb: Ah right, I forgot that actual OS condvars can do that - Go's sync.Cond does not. Apologies!

alexcrichton:
@carllerche ok sounds good to me! I'll send a PR to libstd with these changes so we can eventually move back to thread park/unpark

carllerche:
Thanks.

Incidentally, the entire impl and thread local could be avoided if we could convert `std::thread::Thread` to and from `*mut UnsafeNotify`. This should be possible as `std::Thread` is just a wrapper around `Arc<Inner>` but the necessary APIs aren't present.

I don't really have much insight into what `std` would want to provide to support this.

carllerche:
I mean, we could technically do it w/ `mem::transmute`, but that seems unwise.

alexcrichton:
Nah yeah I think we don't want to assume the size of `Thread` just yet, but we could consider into/from usize upstream!

PR #598, opened 2017-10-02
Provide TestHarness to make testing future / combinator implementations easier.

carllerche:
Depends on #597

This PR adds a new type: `executor::TestHarness`. This type is very similar to `Spawn` but is intended to be used from the context of tests. As such, it provides a slightly different API. While this functionality could be provided directly by `Spawn`, `Spawn` already has a fairly busy API surface and some of the capabilities of `TestHarness` doen't make as much sense when used in "real code".

This PR only provides the ability to wrap a `Future` for now, but at a later time, support for `Stream` and `Sink` can be added.

The idea would be for `TestHarness<impl Stream>` to provide:

* `impl Iterator`
* `poll_next(&mut self) -> Poll<Option<Self::Item>, Self::Error>`
* `wait_next_timeout(&mut self) -> Result<Option<Self::Item>, TimeoutError<Self::Error>>`

And for `TestHarness<impl Sink>`:

* `try_send(&mut self, item: T::SinkItem) -> StartSend<T::SinkItem, T::SinkError>`
* `send_timeout(&mut self, item: T::SinkItem) -> Result<(), SendTimeoutError<T::SinkItem, T::SinkError>>`

### DIFF
--- a/benches/thread_notify.rs
+++ b/benches/thread_notify.rs
@@ -1,0 +1,114 @@
+#![feature(test)]
+
+extern crate futures;
+extern crate test;
+
+use futures::{Future, Poll, Async};
+use futures::task::{self, Task};
+
+use test::Bencher;
+
+#[bench]
+fn thread_yield_single_thread_one_wait(b: &mut Bencher) {
+    const NUM: usize = 10_000;
+
+    struct Yield {
+        rem: usize,
+    }
+
+    impl Future for Yield {
+        type Item = ();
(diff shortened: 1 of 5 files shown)


alexcrichton:
Hm I'd imagine that if we want to make futures-based tests easier to write we could probably go farther? Maybe we could try to collect test-related helpers in a `futures-test` crate or something like that? (avoid exposing more surface area here)

PR #599, opened 2017-10-03
Oneshot example, switch variable names

raggi:
The example I think was intending for `p = producer, c = consumer`, but had `p` and `c` roles switched.


carllerche:
Thanks

PR #600, opened 2017-10-05
Clarify the relationship between `Stream::and_then` and `Stream::for_each`.

jimmycuadra:
When I was learning this crate, I had a lot of trouble figuring out when to use `Stream::and_then` and when to use `Stream::for_each`, which seemed very similar. This patch adds one sentence to the docs for each method clarifying when you'd want to use the other method instead. The key takeaway is that one returns a stream and one returns a future.


alexcrichton:
Thanks!

Issue #586, opened 2017-09-19
The `slot` abstraction again


tailhook:
Hi,

Thinking more about #585 and #570 I'm not sure I was quite right.

I think I need a `poll_cancel` in a majority of my use cases to avoid memory leaks.

I think the channel can be made single-producer - single-consumer, with `poll_cancel`, because that would make implementation much simpler. Then restoring current behavior can be done by user by wrapping a sender into an `Arc` (because `swap` method takes a readonly reference).

Alternatively, is there a viable implementation of `poll_cancel` with multi-producer channel?

Note: we should decide this before futures 0.1.17 comes out :)

alexcrichton:
This sounds plausible to me, albeit this still may be relatively niche compared to the existing mpsc channel. Currently the only real implementation of `poll_cancel` AFAIK is to just have a vector of tasks, which isn't great, unfortunately.

@carllerche do you have thoughts here?

tailhook:
> Currently the only real implementation of poll_cancel AFAIK is to just have a vector of tasks, which isn't great, unfortunately.

Is there a way to deduplicate the tasks in vector? Otherwise, I think memory leak is more serious than just not using `poll_cancel` :)

>  still may be relatively niche compared to the existing mpsc channel

My use cases are: resolving (and updating) a DNS name, or any other information synced from a remote location, publishing progress of an operation. From this list, only the latter could perhaps be multi-producer, but it's required too.

alexcrichton:
You could give each sender an ID and ensure that each sender has at most one task blocked, but you'd still need a slot for each sender.

carllerche:
I actually wanted `mpsc::Sender::is_canceled` recently, which is significantly different than poll_cancel of course.

That said, the `Sender` already has a way to [track its task task](https://github.com/alexcrichton/futures-rs/blob/master/src/sync/mpsc/mod.rs#L99). Implementing `poll_cancel` would require pushing that node in a "notify me on cancel" queue. Of course, as @alexcrichton hinted at, if that queue gets large, dropping the RX could be expensive as it would need to notify that entire queue...

alexcrichton:
Closing due to https://github.com/alexcrichton/futures-rs/pull/621

PR #604, opened 2017-10-07
Single producer slot

tailhook:
As I've described in #586 this removes `Clone` from `Sender` making `slot` a single producer single consumer stream. The points are:

1. The previous functionality can be restored with `Arc<Sender>` and using `swap` instead of `Sink::start_send`.
2. It now has `poll_cancel` method which allows deallocating resources on sender side when no receiver is present anymore.

The `Sink` functionality on `Arc<Sender>` can be restored by making a wrapper type, if this is crucial for some users. Also, this PR only updates the `sync` implementation, I can update the `unsync` implementation too if conceptually it's accepted.

### DIFF
--- a/src/sync/slot.rs
+++ b/src/sync/slot.rs
@@ -145,6 +145,13 @@ impl<T> Stream for Receiver<T> {
     }
 }
 
+impl<T> SendError<T> {
+    /// Returns the message that was attempted to be sent but failed.
+    pub fn into_inner(self) -> T {
+        self.0
+    }
+}
+
 /// Creates an in-memory Stream which only preserves last value
 ///
 /// This method is somewhat similar to `channel(1)` but instead of preserving
--- a/src/sync/slot.rs
+++ b/src/sync/slot.rs
@@ -35,7 +35,8 @@ pub struct SendError<T>(T);
 #[derive(Debug)]
 struct Inner<T> {
     value: Option<T>,
-    task: Option<Task>,
+    read_task: Option<Task>,
+    cancel_task: Option<Task>,
 }
 
 trait AssertKindsSender: Send + Sync {}
@@ -72,7 +73,7 @@ impl<T> Sender<T> {
                 let mut inner = lock.lock().unwrap();
                 result = inner.value.take();
                 inner.value = Some(value);
-                inner.task.take()
+                inner.read_task.take()
             } else {
                 return Err(SendError(value));
             }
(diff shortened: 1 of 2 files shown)


tailhook:
The last commit makes `poll_cancel` as `&self` (immutable). This makes certain scenarios easier (such as keeping `Arc<Sender>`) while this still clearly contains only a single task to wake up on cancel. This commit is much more controversial, but at least solves my use case better (otherwise I need to put a Sender into an excessive mutex), still not sure if it might prevent later optimizations.

carllerche:
Generally speaking, poll_* fns must take &mut self in order to ensure they aren't called from multiple tasks. 

So, a poll_cancel would not work if multiple senders attempted to share a Sender in an Arc.

tailhook:
Okay, that makes sense. Rewritten last commit to barely contain `is_canceled` method.

carllerche:
@tailhook what was the reason to remove `Clone` in the first place? Having `poll_cancel` seems useful?

tailhook:
@carllerche, sure. Without `poll_cancel` I'll get a memory leak in my use case.

carllerche:
Yeah... and a `poll_cancel` w/ `Clone` would make things more complicated...

I don't have a huge opinion about this either way.

tailhook:
Well, for now, I'm [vendoring it](https://github.com/tailhook/ns-router/blob/master/src/slot.rs) and it seems to work for that use case (the code using it is quite early and not extensively tested, though).

alexcrichton:
Are there enough questions about this abstraction that we should perhaps move it to a crate for now instead of putting it in `futures` the crate just yet?

carllerche:
Yes, probably

alexcrichton:
@tailhook would you agree as well? And if so, mind sending a PR?

tailhook:
Well, I don't see unsolved questions here. The implementation as is in current PR is clearly superior to the one in the current master. And is simple enough comparing to multi-producer implementation that could support `poll_cancel`.

alexcrichton:
@tailhook I'm somewhat uneasy as right after the inital implementaiton landed new questions we raised, and it sounds like in solving those questions it raises more on this PR? It seems overall that there's not a rock-solid consensus on what to do here, in which case this is somewhat experimental, so it sounds like it may be best to start in an external crate?

tailhook:
Well, the only question brought here is should `poll_cancel` have `mut self` or immutable self. The only `poll_*` method where immutable self is used is in `BiLock` which seems to be [wrong decision too](https://github.com/alexcrichton/futures-rs/pull/606). So I percept it more like my mistake than a controversial design issue.

Technically nothing stops to put it into another crate, but it's much more hard to discover needed kind of channel in another crate. And this abstraction is good for the following reasons:
1. Previously I've used unbounded channel for the task which is obviously more memory hungry (it can't be replaced by a bounded channel because that would get values lost)
2. There is only one channel that has `poll_cancel` support -- `oneshot`, and no `Stream` channels like that. Which means certain patterns isn't possible to implement without a memory leak.

PR #606, opened 2017-10-08
Alternative bi-lock implementation

Diggsey:
This PR replaces the implementation of `BiLock` with one which does not perform allocation, ~and is slightly simpler~.

This introduces a backwards-incompatible change to the API as it requires a mutable reference when locking/unlocking a `BiLock`.

In general, lock-free data structures require space proportional to the number of concurrent accessors, so taking mutable references to `self` should be preferred as it allows more efficient implementations.

Implementation details:
- The `BiLock` is initialised with three "tokens", numbered 0, 1, 2.
- At all times, each side of the `BiLock` owns one token, and the remaining token is "free".
- Only one thread-safe operation is supported by a `BiLock`: atomically swapping its own token with the "free" token.
- Owning a token gives you exclusive access to part of the data structure: tokens 0, 1 gives access to the corresponding indices in the "tasks" array, while token 2 gives access to the value protected by the lock.

`poll_lock` operation:
- The current token must be 0 of 1, or else we'd already have the lock
- The current task can be stored in the `tasks` array at the element we have access to
- Atomically swap our token with the free token
- If our token is now "2", we own the lock
- Otherwise, the free token now corresponds to our own task

`unlock` operation:
- The current token must be 2
- Atomically swap our token with the free token
- We get back either the 0 or 1 token, which gives us access to an element in the tasks array
- Wake up the task we obtained from the token

Tests:
I had to wrap the "concurrent" tests in a lazy future, because they called `poll_lock` outside a task, which used to work despite the documentation saying that the method should panic when called off-task.

### DIFF
--- a/src/sync/bilock.rs
+++ b/src/sync/bilock.rs
@@ -132,7 +132,6 @@ impl<T> Inner<T> {
         assert!(caller_state != 2, "Lock already held");
 
         // Fast path
-        *self.tasks[caller_state].get() = None;
         caller_state = self.free_state.swap(caller_state, AcqRel);
 
         if caller_state != 2 {
--- a/src/sync/bilock.rs
+++ b/src/sync/bilock.rs
@@ -6,7 +6,7 @@ use std::mem;
 use std::ops::{Deref, DerefMut};
 use std::sync::Arc;
 use std::sync::atomic::AtomicUsize;
-use std::sync::atomic::Ordering::AcqRel;
+use std::sync::atomic::Ordering::{AcqRel, Relaxed};
 
 use {Async, Future, Poll};
 use task::{self, Task};
@@ -32,13 +32,12 @@ use task::{self, Task};
 #[derive(Debug)]
 pub struct BiLock<T> {
     inner: Arc<Inner<T>>,
-    index: usize
+    index: bool
 }
 
 #[derive(Debug)]
 struct Inner<T> {
-    free_state: AtomicUsize,
-    used_states: [UnsafeCell<usize>; 2],
+    state: AtomicUsize,
     tasks: [UnsafeCell<Option<Task>>; 2],
     value: Option<UnsafeCell<T>>,
 }
@@ -54,13 +53,12 @@ impl<T> BiLock<T> {
     /// tasks to be managed there.
     pub fn new(t: T) -> (BiLock<T>, BiLock<T>) {
         let inner = Arc::new(Inner {
-            free_state: AtomicUsize::new(2),
-            used_states: [UnsafeCell::new(0), UnsafeCell::new(1)],
+            state: AtomicUsize::new(3),
             tasks: Default::default(),
             value: Some(UnsafeCell::new(t)),
         });
 
-        (BiLock { inner: inner.clone(), index: 0 }, BiLock { inner: inner, index: 1 })
+        (BiLock { inner: inner.clone(), index: false }, BiLock { inner: inner, index: true })
     }
 
     /// Attempt to acquire this lock, returning `NotReady` if it can't be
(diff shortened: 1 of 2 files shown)


Diggsey:
If this is of interest, a similar approach can be used for many of the other kinds of lock free data structure.

Diggsey:
Benchmarks show marginally better performance for the contended case, but with significantly lower variation, which is about what I was hoping for by eliminating the allocation.

The contention-free case is showing exactly a 2x slow-down though, which is not good! I'm still investigating the cause, but initial evidence seems to point to a performance cliff (possibly a missed optimisation?) rather than an incremental slowdown caused by false sharing/increased size of `Inner`, so I'm hoping it's possible to fix!

```
Before:
test contended   ... bench:     177,635 ns/iter (+/- 52,813)
test lock_unlock ... bench:      59,665 ns/iter (+/- 4,362)

After:
test contended   ... bench:     177,250 ns/iter (+/- 13,123)
test lock_unlock ... bench:     117,321 ns/iter (+/- 8,443)
```

Diggsey:
I discovered the major cause of slowdown was caused by storing the (mutable) token alongside the handle, rather than inside the data structure. Adding an extra layer of indirection improved the performance quite a bit:
```
Before:
test contended   ... bench:     177,635 ns/iter (+/- 52,813)
test lock_unlock ... bench:      59,665 ns/iter (+/- 4,362)

After:
test contended   ... bench:     144,758 ns/iter (+/- 12,908)
test lock_unlock ... bench:      77,714 ns/iter (+/- 6,243)
```

I may add some multi-threaded benchmarks, because at the moment all access is single-threaded and perfectly regular.

omni-viral:
I believe state may be defined by single `AtomicUsize`
There are 6 states possible.
Here `left` and `right` are two owners.
```
s - [left,right,free]
0 - [0,1,2]
1 - [1,0,2]
2 - [2,0,1]
3 - [0,2,1]
4 - [1,2,0]
5 - [2,1,0]
```
Swap operation for `left` owner is `state.fetch_xor(1)`
for `right` is `state.fetch_add(3)`
to get current token id `STATE_MAP[state.load() % 6]` (and then `[0]` for `left`, `[1]` for `right`)

Diggsey:
@omni-viral that's a really cool idea - I implemented it, but unfortunately it did not affect the performance. It should reduce the memory usage though.

omni-viral:
@Diggsey replacing one read-modify-write operation with another shouldn't affect performance.
Atomic swap may be even faster than atomic add or atomic or on some platforms.

Diggsey:
@omni-viral In the single-threaded test the bottle-neck was actually other factors than that atomic operations, possibly including the extra space, so I though it might help there.

I've added a concurrent test, and the changes do seem to be a significant improvement:
```
Before:
test concurrent  ... bench:   1,047,437 ns/iter (+/- 146,489)
test contended   ... bench:     161,603 ns/iter (+/- 11,326)
test lock_unlock ... bench:      54,580 ns/iter (+/- 4,425)

After:
test concurrent  ... bench:     695,118 ns/iter (+/- 118,187)
test contended   ... bench:     143,350 ns/iter (+/- 36,404)
test lock_unlock ... bench:      73,727 ns/iter (+/- 23,065)
```

omni-viral:
@Diggsey I you don't mind can you run benchmark with state in single atomic and in separate and share results?

alexcrichton:
Since this is a breaking change, would it be best to be in an external crate for now? One we could consider pulling in for the 0.2 release?

Diggsey:
@alexcrichton I'm not sure `BiLock` has enough usage outside of the futures/tokio libraries themselves for there to be much benefit to publishing it in an external crate (in so far as it probably wouldn't get much extra testing prior to the 0.2 release). Was there another reason you suggested publishing in an external crate?

I'm in no particular rush to get this merged, but I do think using `&mut self` for these data structures is a change that should be made with 0.2 to avoid breakage down the line. (You can always wrap it in a `RefCell` to get back the previous semantics)

alexcrichton:
Nah that was it, otherwise this'll just sit here until 0.2 which doesn't currently have a timeline.

arthurprs:
Interesting approach.

I wonder though if this is optimizing for the right use case? It appears to regress the single executor bench by ~38%, which is probably the most common due to split tcp sockets.

Diggsey:
I'm not sure the lock/unlock bench is particularly representative of the common use-case either - in my experiments the performance was very dependant on LLVM's ability to optimise out the allocation. Even slight changes to the code which made no *semantic* difference could cause the performance to drop by as much as 2x, whereas the allocation-free version was much more consistent.

carllerche:
Out of curiosity, has anyone actually compared `BiLock` with a plain ol' `Mutex`? It would be interesting to add that to the benchmarks.

Obviously a mutex blocks the thread, but if the critical section is small, that might end up just being the right solution.

arthurprs:
@Diggsey I don't understand, the Box can't be optimized out. It must happen in order to return NotReady, but that's juts not the case for split sockets on the same executor.

cramertj:
@Diggsey Would you be interested in rebasing this change and getting it in as part of the 0.2 release?

Diggsey:
@cramertj I don't think this is ready to go in - I need to do some more bench-marking as there's definitely something unexplained with the performance characteristics.

However, the question of whether `self` should be mutable or immutable is important to answer for 0.2 - especially for `BiLock` I think it should take a mutable reference - it's very surprising that one side of a `BiLock` can be polled concurrently by multiple tasks, essentially making it a full (but unfair) mutex.

cramertj:
I'm going to close this for now, but please reopen when you've had a chance to fix it up! Bilock [isn't reexported anymore](https://github.com/rust-lang-nursery/futures-rs/blob/master/futures-channel/src/lib.rs#L23), so we can fix this in the future.

cramertj:
@Diggsey Hey! I wanted to ping and see if you'd given any thought to picking this back up.

PR #607, opened 2017-10-09
Add full URL to deprecation warnings

dbrgn:
When the issue id is shown in a project that depends on futures indirectly, it's not obvious in which issue tracker that issue can be found.


alexcrichton:
Thanks!

PR #609, opened 2017-10-11
Allow Spawn to be unsized

cramertj:
This would allow me to save an allocation in a very performance-sensitive area.

### DIFF
--- a/src/task_impl/mod.rs
+++ b/src/task_impl/mod.rs
@@ -242,17 +242,13 @@ impl<T: ?Sized> Spawn<T> {
     /// Get a mutable reference to the object the Spawn is wrapping.
     pub fn get_mut(&mut self) -> &mut T {
         &mut self.obj
-    } 
-}
+    }
 
-impl<T> Spawn<T> {
     /// Consume the Spawn, returning its inner object
-    pub fn into_inner(self) -> T {
+    pub fn into_inner(self) -> T where T: Sized {
         self.obj
     }
-}
 
-impl<F: Future + ?Sized> Spawn<F> {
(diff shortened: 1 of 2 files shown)


src/task_impl/mod.rs
```diff
@@ -242,15 +242,17 @@ impl<T> Spawn<T> {
     /// Get a mutable reference to the object the Spawn is wrapping.
     pub fn get_mut(&mut self) -> &mut T {
         &mut self.obj
-    }
+    } 
+}
 
+impl<T> Spawn<T> {
```
alexcrichton: Mind leaving these in the same impl block, but adding `where Self: Sized` to this method?

alexcrichton: If you're super enterprising, you could collapse all the below impl blocks into one with the same strategy as well :)

(no need to happen in this PR though)

tests/all.rs
```diff
@@ -359,3 +359,17 @@ fn option() {
...
 }
+
+#[test]
+fn spawn_does_unsize() {
+    #[derive(Clone, Copy)]
+    struct EmptyNotify;
+    impl executor::Notify for EmptyNotify {
+        fn notify(&self, _: usize) { panic!("Cannot notify"); }
+    }
+    static EMPTY: &EmptyNotify = &EmptyNotify;
```
alexcrichton: Mind throwing a `'static` in here to fix the compile on 1.15?

cramertj:
@alexcrichton Did you want any other changes here?

alexcrichton:
Nah that was it, looks great, thanks!

PR #610, opened 2017-10-11
Futures cpupool debug impls

fitzgen:
r? @alexcrichton 

Want me to do the version bump commit too, or do you want to do that when publishing the next version (can you please publish a new version? 😺 )


alexcrichton:
Thanks! I'll publish a new version right now too

PR #613, opened 2017-10-16
Forward directly to underlying sink if Buffer capacity is 0

### LINKED ISSUE #608: sink::Buffered accepts 1 more than capacity
In `start_send`, it looks like`item` will be buffered if `self.buf.len() == self.cap`:

```
fn start_send(&mut self, item: Self::SinkItem) -> StartSend<Self::SinkItem, Self::SinkError> {
    self.try_empty_buffer()?;
    if self.buf.len() > self.cap {
        return Ok(AsyncSink::NotReady(item));
    }
    self.buf.push_back(item);
    Ok(AsyncSink::Ready)
}
```

If this is the case, `self.buf` will contain `amt + 1` items.

Not sure what the best fix is. To avoid a re-allocation, we should allocate `self.buf` with capacity `amt + 1`. If we want to guarantee that `self.buf` never contains more than `amt` items, it seems like it will make the logic in start_send more complicated, since it will need to attempt to send `item` to avoid hanging.

casey:
Fixes #608

### DIFF
--- a/src/sink/buffer.rs
+++ b/src/sink/buffer.rs
@@ -75,8 +75,12 @@ impl<S: Sink> Sink for Buffer<S> {
     type SinkError = S::SinkError;
 
     fn start_send(&mut self, item: Self::SinkItem) -> StartSend<Self::SinkItem, Self::SinkError> {
+        if self.cap == 0 {
+            return self.sink.start_send(item);
+        }
+
         self.try_empty_buffer()?;
-        if self.buf.len() > self.cap {
+        if self.buf.len() == self.cap {
             return Ok(AsyncSink::NotReady(item));
         }
         self.buf.push_back(item);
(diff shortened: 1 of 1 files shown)


src/sink/buffer.rs
```diff
@@ -75,21 +75,33 @@ impl<S: Sink> Sink for Buffer<S> {
...
+        }
+
         try_ready!(self.try_empty_buffer());
         debug_assert!(self.buf.is_empty());
         self.sink.poll_complete()
     }
 
     fn close(&mut self) -> Poll<(), Self::SinkError> {
+        if self.cap == 0 {
+          return self.sink.close();
```
alexcrichton: Looks like indentation may be off here?

casey:
Fixed the indentation. The test `buffer_noop` has a case where it instantiates a buffer with 0 capacity. Were you thinking of something more extensive?

alexcrichton:
Ah ok nah it's fine if we already have a test, thanks!

Issue #605, opened 2017-10-08
Visit all futures in FuturesUnordered


tailhook:
My use case is: I have a set of futures each following its own stream which it processes and sends result over its own channel. But when the configuration is updated I need to visit every future and check if everything in the future's state is up to date.

The proposed API:
```rust
impl FuturesUnordered<F> {
  fn visit<G: FnMut(&mut F)>(&mut self, f: G) { ... }
}
```
Or alternatively, just allow to recover futures from the collection:
```rust
impl FuturesUnordered<F> {
  fn drain(&mut self) -> impl Iterator<F> { ... }
}
```
(note the latter would require `mem::replace` for `FuturesUnordered` for the use case, but is probably acceptable too)

Currently, the simplest solution I could find is:
1. Put a `Shared<()>` into each future
2. When `Shared` is ready, yield internal state as a result of the future
3. Receive state from `FuturesUnordered::poll` and reconstruct the future with a new `Shared<()>` future (waiting for next update)

The alternative strategy might be:
1. Add an id for each future and an input channel for config
2. Add a collection of outgoing channels by id
3. Wrap the collection (2) into an `Arc<Mutex<>>` so the future can remove itself from the collection

Both strategies look overly complicated and have large overhead compared to a visitor over a collection we already have.

Another thing to note: both strategies actually represent pub-sub kind of pattern. Which would be useful for futures on its own. But presumably, it will always have additional overhead.

What do you think?

carllerche:
Is there a reason why adding an iterator that iterates over the inner futures doesn't work (vs a callback)?

tailhook:
Ah, yes, an iterator over mutable references should work too.

alexcrichton:
Various iterators (iter_mut/into_iter/drain/etc) sound like a good idea to me!

srijs:
@tailhook Just wanted to let you know that `FuturesUnordered::iter_mut()` just landed in master via #618.

tailhook:
@srijs, thank you! Now I'm waiting for a release :) This ticket might be closed indeed.

tailhook:
Oh, sorry. I've missed one detail when reading comments: the original proposal was to use a closure because everything done in that closure **must wake up an inner future** not just the parent future holding the FuturesUnordered object. (The drain proposal would work because it would push futures back to the new collection).

So while `iter_mut` might be actually useful, it doesn't solve my task.

srijs:
[...]

PR #618, opened 2017-10-22
Implement FuturesUnordered::iter_mut

srijs:
Implementing `FuturesUnordered::iter_mut` as suggested in #605. Generally it should be said that I don't have a lot of experience with unsafe rust yet, so I would appreciate a thorough review!

I did not go for the other suggested ones (`into_iter`/`drain`), since I lack the insight into how to implement this safely without leaking memory.

### DIFF
--- a/src/stream/futures_unordered.rs
+++ b/src/stream/futures_unordered.rs
@@ -1,3 +1,5 @@
+//! An unbounded set of futures.
+
 use std::cell::UnsafeCell;
 use std::fmt::{self, Debug};
 use std::iter::FromIterator;
@@ -9,7 +11,7 @@ use std::sync::atomic::{AtomicPtr, AtomicBool};
 use std::sync::{Arc, Weak};
 use std::usize;
 
-use {task, Stream, Future, Poll, Async, IntoFuture};
+use {task, Stream, Future, Poll, Async};
 use executor::{Notify, UnsafeNotify, NotifyHandle};
 use task_impl::{self, AtomicTask};
 
@@ -50,29 +52,6 @@ pub struct FuturesUnordered<F> {
 
 unsafe impl<T: Send> Send for FuturesUnordered<T> {}
 unsafe impl<T: Sync> Sync for FuturesUnordered<T> {}
-
-/// Converts a list of futures into a `Stream` of results from the futures.
-///
-/// This function will take an list of futures (e.g. a vector, an iterator,
-/// etc), and return a stream. The stream will yield items as they become
-/// available on the futures internally, in the order that they become
-/// available. This function is similar to `buffer_unordered` in that it may
-/// return items in a different order than in the list specified.
-///
-/// Note that the returned set can also be used to dynamically push more
-/// futures into the set as they become available.
-pub fn futures_unordered<I>(futures: I) -> FuturesUnordered<<I::Item as IntoFuture>::Future>
-    where I: IntoIterator,
-          I::Item: IntoFuture
-{
-    let mut set = FuturesUnordered::new();
-
-    for future in futures {
-        set.push(future.into_future());
-    }
-
-    return set
-}
 
 // FuturesUnordered is implemented using two linked lists. One which links all
 // futures managed by a `FuturesUnordered` and one that tracks futures that have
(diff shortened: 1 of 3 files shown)


src/stream/futures_unordered.rs
```diff
@@ -427,6 +436,37 @@ impl<T> Drop for FuturesUnordered<T> {
     }
 }
 
+#[derive(Debug)]
+/// Mutable iterator over all futures in the unordered set.
+pub struct FuturesUnorderedIterMut<'a, F: 'a> {
```
carllerche: I would probably just name this `IterMut` as it already is in the `futures_unordered` module.

src/stream/mod.rs
```diff
@@ -112,7 +112,7 @@ if_std! {
     pub use self::collect::Collect;
     pub use self::wait::Wait;
     pub use self::split::{SplitStream, SplitSink};
-    pub use self::futures_unordered::{futures_unordered, FuturesUnordered};
+    pub use self::futures_unordered::{futures_unordered, FuturesUnordered, FuturesUnorderedIterMut};
```
carllerche: I probably wouldn't re-export this type. Instead, maybe make the `futures_unordered` module pub (and ensure no priv types are accidentally exported).

srijs:
@carllerche Thanks for the review, addressed all feedback!

alexcrichton:
Inside this module there's a public `futures_unordered` function which if we make the module public will also make that function reachable (but it's not intentional to have that function reachable. I think I'm ok making the module public for the iterator name (albeit it's a bit of a wonky setup) but can we ensure that the free function isn't exposed?

src/stream/futures_unordered.rs
```diff
@@ -440,6 +451,37 @@ impl<F: Future> FromIterator<F> for FuturesUnordered<F> {
     }
 }
 
+#[derive(Debug)]
+/// Mutable iterator over all futures in the unordered set.
+pub struct IterMut<'a, F: 'a> {
+    node: *const Node<F>,
+    len: usize,
+    _marker: PhantomData<&'a F>
```
alexcrichton: I think this may best be `&'a mut FuturesUnordered<F>` perhaps?

carllerche:
Which release of rust included `pub(crate)`? I would guess that we should be able to use it soon if not now.

alexcrichton:
Unsure, but the function can also just be moved

src/stream/futures_unordered.rs
srijs: Apologies -- can you go into more details why that would be?

srijs:
According to [this](https://github.com/rust-lang/rust/issues/32409#issuecomment-297080960), `pub(crate)` was stabilised in 1.18.

@alexcrichton the `futures_unordered` function is exported, according to the docs: https://docs.rs/futures/0.1.16/futures/stream/fn.futures_unordered.html

alexcrichton:
@srijs right yeah but as is this would also export `futures::stream::futures_unordered::futures_unordered` (that is, a function in the module). Ideally only `futures::stream::futures_unordered` would be the exported function.

src/stream/futures_unordered.rs
alexcrichton: Long long ago in Rust there were subtle distinctions between how `&mut T` and `&T` worked, and I forget if any of them are still true today. Basically because this is a mutable iterator it's "better" to have a phantom data of `&mut ...` instead of `&...` to ensure this information is conveyed to the compiler. 

Additionally I figure it's best to store `FuturesUnordered<F>` as that's effectively what's happening here (vs just `F`). This can also cause subtle differences but probably not as important as `&mut` vs not

srijs: Thanks for explaining that, changed it!

srijs:
@alexcrichton Now that we have `FromIterator` for `FuturesUnordered`, could it be an option to deprecate the free-standing function, or do still see a use-case for it which is not covered by `from_iter`?

src/stream/mod.rs
```diff
@@ -103,7 +103,7 @@ if_std! {
     mod wait;
     mod channel;
     mod split;
-    mod futures_unordered;
+    mod futures_unordered_impl;
```
alexcrichton: Oh I was thinking that we'd still want a `futures_unordered` module which contained the `FuturesUnordered` type, for example. I think we'll just want to move the definition of the constructor, `futures_unordered`, from the module to this `stream` module

alexcrichton:
Ah yeah so for now I personally like the free function for symmetry with the rest of the crate, but eventual deprecation is of course possible!

src/stream/mod.rs
srijs: No problem, changed!

srijs:
@alexcrichton Feedback addressed, let me know if you'd like me to change anything else!

alexcrichton:
Awesome, thanks!

PR #585, opened 2017-09-19
Add an `unsync` version of the `slot` channel

alexcrichton:
Also tweak the implementation slightly of both the sync/unsync versions.

cc @tailhook

PR #621, opened 2017-10-23
Remove slot abstraction as it deemed uncertain

tailhook:
Closes #604, #586. Reverts #570, #585 (note it reverts rust version to 1.13.0 too)

Okay, I'm not sure if my argumentation was strong enough in #604, but let's remove it so we don't have to block a release anymore.


alexcrichton:
Thanks @tailhook! Sorry for being a little slow to reply on these, I've been pretty busy lately :(

carllerche:
@tailhook Do you plan on creating this in a repo as I think that I'd be interested in using it personally :)

tailhook:
@carllerche Yes. Just released [async-slot](https://crates.io/crates/async-slot) crate.

carllerche:
Thanks, I actually realized what I need is multi consumer. So I will try implementing something for that.

carllerche:
I pushed my code here: [futures-watch](https://github.com/carllerche/better-future/tree/master/futures-watch).

PR #624, opened 2017-10-26
Fix race condition in oneshot

### LINKED ISSUE #620: Oneshot may panic if the receiver is closed
The implementation of oneshot is racy, which results in a random chance of it panicking when the receiver is closed and the sender tries to send at the same time.

I wrote a test to demonstrate the issue: https://github.com/Diggsey/futures-rs/commit/aa70fbce690cd5b92bd7e8e59661a20061ecf142

The issue is with the assertion made in this comment: https://github.com/alexcrichton/futures-rs/blob/master/src/sync/oneshot.rs#L123 which is not true, because the receiver sets the complete flag also when it is closed.

A correct lock-free implementation is essentially impossible without either:
- Reserving space in the data structure for at least two values of type `T`
- Making `close()` consume `self` and return any "in transit" value directly, instead of having the caller poll the future.

Diggsey:
Fixes #620


alexcrichton:
I'm sort of confused, [this patch](https://gist.github.com/alexcrichton/502de8e4b617258d37a977a2da3d824e) fixes the panic for me locally, but does it not fix it for you?

Diggsey:
@alexcrichton It fixes the panic, but it breaks the behaviour of close. It's essential for `oneshot` that `poll()` *can* return an item after the oneshot has been closed (see the documentation for the close method) as this is the only way to avoid losing items.

Diggsey:
I can try to add a test for this behaviour (never losing items), although it will be somewhat probabilistic.

alexcrichton:
That's true yes but I don't really understand why this requires an unsafe rewrite. Can the `unwrap` just not be removed and instead return an `Err`?

Diggsey:
I'm not sure which `unwrap` you are referring to, do you mean the one in `send()`? The problem there is not the failure case - it's the case where it successfully obtains the lock, but *after* or *while* the receiver is calling `close()`.

If that happens, the sender successfully stores the value, and then `poll()` (with your patch) returns `Canceled`, and so the value is lost in the oneshot. Without your patch, `poll()` may fail to take the lock (let's say the sender hasn't released it yet) and then it has no choice but to return `NotReady`, which again violates the contract specified by the oneshot documentation.

The rewrite is needed because I couldn't see a way to implement the contract while using these "pollable locks" - ultimately, if the sender is holding the lock whilst the receiver calls `close()` and then `poll()`, there's no way for the receiver to proceed without waiting until the sender has released the lock. And waiting on a `Lock` is not supported.

alexcrichton:
Yes I was referring to the one in `send`. 

The problem is that [this comment](https://github.com/alexcrichton/futures-rs/blob/6d861eeed14a8158cad86ea242c04e8c825f363e/src/sync/oneshot.rs#L123-L125) is not true. The `complete` flag can be set after it's witnessed to be `false` notably by `close`. This means that I think we could fix this via:

* In `poll` if the receiver was closed then we try to grab the `data` lock. If we get the lock then we figure out what to return. If we fail to acquire the lock then a `send` is in progress which will fail, so we returned `Canceled` as well.
* In `send` if we avoid `unwrap` on the lock and instead take a look at the return. If we failed to grab the lock then a receiver must have closed the channel and is looking at the data. At that point we fail the send.
* If the `send` branch successfully acquires the lokc then it knows that it was not canceled coming in and the data is stored. The branch could then re-check the `canceled` flag on the way out. If it's set to `false` then the data was "sent" and if it's `true` then the data is pulled back out. If the lock is acquired the data is returned (if it's still there), and if the lock fails then the receiver must be pulling it out so the data was successfully sent.

Does that not fix the issue?

Diggsey:
@alexcrichton The sender cannot check both the `canceled` flag *and* release the lock simultaneously. If either happens first, there's room for `poll()` and/or `close()` to occur between them. Both cases cause immediate problems.

There may be further modifications you could make to handle those cases too, but at this point, the number of possible inter-leavings is so complicated that I don't think anyone could be confident about correctness...

alexcrichton:
Yes it cannot be done atomically, I don't believe it's necessary to do so. What's the execution trace you're worried about?

Diggsey:
I'll cover both cases:
1) Check `cancelled` and then unlock
=> `cancelled` may be set between checking and unlocking, and then `poll()` could fail to take the lock
2) Unlock and then check `cancelled`
=> `poll()` may take the lock before the send has a chance to pull the value back out again, so when you try to take the lock again to extract the value, it could fail. This you can presumably treat as success. Alternatively, it could see that it was cancelled, take the lock successfully, but the receiver could have already pulled out the value, and again, you can probably just treat that as a successful send.

Assuming all of those edge cases are handled correctly, I can't immediately see any further issues, but then I haven't even thought much about task notification yet. I also have no real *reason* to believe it's correct: with the implementation in this PR, I can write down all the possible states, their transitions, and the possible operations in each state. From that, it's fairly easy to check that no two conflicting operations can occur in any given state. With `Lock`s, I can be sure that it's not *unsafe*, but as far as correctness goes, I'm just hoping to have thought of every possible ordering, and checked that under each ordering, no values are lost, and no tasks go un-notified.

alexcrichton:
I will not have time to dig in to this much for the next few weeks. I don't mind quickly reviewing a patch that doesn't add any `unsafe` code but I'd much rather have a slightly buggy-via-panics implementation than one that may have lurking memory safety violations. It will take me a significant amount of time to review this unsafe code and actually conclude there's no memory safety issues.

Diggsey:
I implemented the quick fix in #641 and rebased this PR on top of that.

cramertj:
@Diggsey are you still wanting a review on this?

PR #630, opened 2017-11-02
Adding state to streams

valarauca:
Not having a stateful operators proved pretty painful when writing prod GRPC systems.

The common use case is the first(_ish_) message in a `Stream` will contain auth data. Then instead of re-transmitting it with every message the server is expected to hold all that identity information for the duration of the `Stream`, using that data to supplement and/or override data incoming. 

One could also think of a middleware acting in the same fashion.
 
---

Included a basic `Buffer` operator as there was a `TODO` mentioning it, and it roughly just seemed like `Scanner` but with an `Option<T>` instead of `T`.

### DIFF
--- a/src/stream/mod.rs
+++ b/src/stream/mod.rs
@@ -60,6 +60,8 @@ mod then;
 mod unfold;
 mod zip;
 mod forward;
+mod scanner;
+mod buffer;
 pub use self::and_then::AndThen;
 pub use self::chain::Chain;
 pub use self::concat::{Concat, Concat2};
@@ -90,6 +92,8 @@ pub use self::then::Then;
 pub use self::unfold::{Unfold, unfold};
 pub use self::zip::Zip;
 pub use self::forward::Forward;
+pub use self::scanner::Scanner;
+pub use self::buffer::Buffer;
 use sink::{Sink};
 
 if_std! {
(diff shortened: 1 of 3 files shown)


src/stream/mod.rs
```diff
@@ -669,6 +673,82 @@ pub trait Stream {
...
+    ///     stream::iter::<_,_,()>((1..7usize).map(Ok))
+    ///       .scanner(&mut state, lambda)
+    ///       .fold(0,|acc,x| future::ok(acc+x))
+    ///       .wait()
+    /// };
+    /// assert_eq!(sum, Ok(21));
+    /// assert_eq!(state, 21);
+    /// ```
+    fn scanner<F, T, Fut>(self, init: T, f: F) -> Scanner<Self, T, F, Fut>
+        where F: FnMut(&mut T, Self::Item) -> Fut,
```
alexcrichton: Just taking a look at the signature here, isn't this the same as `and_then`? The closure can capture over `T` as an argument so I'm not sure what this is buying over `and_then`

src/stream/mod.rs
```diff
@@ -669,6 +673,82 @@ pub trait Stream {
...
+    ///     stream::iter::<_,_,()>(data.into_iter().map(Ok))
+    ///       .buffer(&mut state, lambda)
+    ///       .fold(0,|acc,x| future::ok(acc+x))
+    ///       .wait()
+    /// };
+    /// assert_eq!(sum, Ok(10));
+    /// assert_eq!(state, 6);
+    /// ```
+    fn buffer<F, T, Fut>(self, init: T, f: F) -> Buffer<Self, T, F, Fut> 
+        where F: FnMut(&mut T, Self::Item) -> Option<Fut>,
```
alexcrichton: This similarly seems pretty similar to something like `filter` or `filter_map`?

src/stream/mod.rs
valarauca: You are correct, but I'll still push for this change.

Accepting a value as an argument vs capturing it in a closure makes this patterns more explicit. While your suggestion is function, it is implicit, and not demonstrated in an example. Furthermore it makes managing move semantics easier if the stream/future chain is getting moved to another thread, as capturing part of move can lead to headaches. 

Basically I feel this makes the interface simple. 

By your logic why does [scan](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.scan) exist in [Iterators](https://doc.rust-lang.org/std/iter/trait.Iterator.html)?

alexcrichton: Hm I'm sort of confused in that case? It seems like the main advantage here is discoverability? Otherwise it seems less ergonomic to call and work with to me at least.

Could we add examples to `and_then` or the documentation in general to encourage this pattern?

valarauca: >It seems like the main advantage here is discoverability?

Yes for the most part, No for a different reason. 

>Otherwise it seems less ergonomic to call and work with to me at least.

I guess this really depends on the definition of ergonomic. 

Lambda capturing and move semantics are _kind of_ implicit. Especially when dealing with lambdas that cross thread barriers, or capturing objects that hold lifetimes.

By removing the implicit capture semantics, and moving to explicit argument semantics this encodes that contract more clearly. Avoiding confusion for lexiconal borrow checker, and the end developer.

Lastly. `Scan` in `std::iter::Iterator` would have a nearly identical interface, which helps discoverability and ergonomics even if it is slightly redundant.

alexcrichton: I don't think though that a local choice about how to write a contract more clearly should pollute the `Stream` trait itself though. I also don't think that `Scan` on the iterator trait is the direct equivalent of another method (unlike these methods)

valarauca: Cool closing then.

valarauca:
Rejected

src/stream/mod.rs
jgrund: FWIW, I found this while looking for `scan` which is a well-known combinator that emits the current state of it's reduction for each turn through the closure.


I think a combinator that did this, (as I can't find anything else that matches without capturing state in a closure) would be a welcome addition to the streams lib.

PR #631, opened 2017-11-04
chmod -x tests/support/mod.rs

ignatenkobrain:
Having executable bit on file with shebang (which is actually Rust's

Signed-off-by: Igor Gnatenko <i.gnatenko.brain@gmail.com>


alexcrichton:
👍

PR #632, opened 2017-11-05
Implement Sink::fanout

srijs:
Implements a sink combinator that is able to fan-out writes to multiple underlying sinks.

In my use-case, I'm dealing with an incoming stream of bytes, and my goal is to both forward this stream to an upstream destination, as well as do various stats/aggregation computations in-process, such as calculating a hash digest over the bitstream.

Being able to fan-out to multiple sinks allows me to implement each consumer of the stream as a sink, and then forward the incoming stream to the combined sink.

I'm unsure about the name in particular, maybe you have other suggestions/ideas?

Let me know what you think!

### DIFF
--- a/src/sink/fanout.rs
+++ b/src/sink/fanout.rs
@@ -1,0 +1,135 @@
+use core::fmt::{Debug, Formatter, Result as FmtResult};
+use core::mem::replace;
+
+use {Async, AsyncSink, Poll, Sink, StartSend};
+
+/// Sink that clones incoming items and forwards them to two sinks at the same time.
+///
+/// Backpressure from any downstream sink propagates up, which means that this sink
+/// can only process items as fast as its _slowest_ downstream sink.
+pub struct Fanout<A: Sink, B: Sink> {
+    left: Downstream<A>,
+    right: Downstream<B>
+}
+
+impl<A: Sink, B: Sink> Fanout<A, B> {
+    /// Consumes this combinator, returning the underlying sinks.
+    ///
+    /// Note that this may discard intermediate state of this combinator,
+    /// so care should be taken to avoid losing resources when this is called.
+    pub fn into_inner(self) -> (A, B) {
+        (self.left.sink, self.right.sink)
+    }
+}
+
+impl<A: Sink + Debug, B: Sink + Debug> Debug for Fanout<A, B>
+    where A::SinkItem: Debug,
+          B::SinkItem: Debug
+{
+    fn fmt(&self, f: &mut Formatter) -> FmtResult {
+        f.debug_struct("Fanout")
+            .field("left", &self.left)
+            .field("right", &self.right)
+            .finish()
+    }
+}
+
+pub fn new<A: Sink, B: Sink>(left: A, right: B) -> Fanout<A, B> {
+    Fanout {
+        left: Downstream::new(left),
+        right: Downstream::new(right)
+    }
+}
+
+impl<A, B> Sink for Fanout<A, B>
+    where A: Sink,
+          A::SinkItem: Clone,
+          B: Sink<SinkItem=A::SinkItem, SinkError=A::SinkError>
+{
+    type SinkItem = A::SinkItem;
+    type SinkError = A::SinkError;
+
+    fn start_send(
+        &mut self, 
+        item: Self::SinkItem
+    ) -> StartSend<Self::SinkItem, Self::SinkError> {
+        // Attempt to complete processing any outstanding requests.
+        self.left.keep_flushing()?;
+        self.right.keep_flushing()?;
+        // Only if both downstream sinks are ready, start sending the next item.
+        if self.left.is_ready() && self.right.is_ready() {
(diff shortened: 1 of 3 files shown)


carllerche:
Sink fanout is tricky in that it requires handling lagging target sinks. A single blocked destination sink could stall the entire fan out unless it is explicitly dealt with.

There are a few strategies.

a) Allow lagging sinks to stall dispatching.
b) Add buffering (infinite?)
c) Drop messages to stalled sinks.

Could you describe a bit how you handled slow sinks. Also, given that there is no "one size fits all" strategy here, I would rather not add a function on the `Stream` trait.

srijs:
Hi @carllerche!

I agree with you on the challenges behind a fan-out mechanism.

The sink combinator that is implemented in this PR will propagate backpressure from the underlying sinks upwards (your option A). I can call this out more explicitly in the docs later.

My reasoning for this is that the backpressure-propagating version is able to function as a building block for the two other versions that are resilient to backpressure.

Given this basic combinator, backpressure resilience can be added in a modular way to the downstream sinks, for instance via the `Buffer` combinator:

```
left_sink.buffer(100).fanout(right_sink.buffer(100))
```

One could also imagine a combinator that discard items when observing backpressure from down stream, which would be composed in a similar way.

Decomposing it like this gives us two main advantages IMO:

1. New ways of coping with backpressure can be easily composed with a single, well tested fan-out implementation
2. It is possible to pick different coping mechanisms for different downstream sinks, depending on the resilience and consistency requirements

Curious to hear your thoughts on this!

carllerche:
So, while it isn't the best possible strategy for all cases, I think I agree w/ your argument and it is an acceptable strategy for most cases.

I'm OK w/ this conceptually if @alexcrichton wants to pull it in.

alexcrichton:
Thanks for the PR @srijs! This all sounds great to me. I agree that this strategy of "wait for everything to be ready" seems like the lowest common denominator. Buffering can be added (like you've mentioned with the `buffer` combinator) and I imagine we could add an "always ready" combinator which is always ready by dropping message that can't be sent to a lower layer.

@carllerche did you want to request changes to this? Or were you ok with this as-is?

src/sink/fanout.rs
```diff
@@ -0,0 +1,128 @@
...
+    }
+
+    fn keep_flushing(&mut self) -> Result<(), S::SinkError> {
+        if let AsyncSink::NotReady(item) = replace(&mut self.state, AsyncSink::Ready) {
+            self.state = self.sink.start_send(item)?;
+        }
+        Ok(())
+    }
+
+    fn poll_complete(&mut self) -> Poll<(), S::SinkError> {
```
carllerche: This doesn't seem like a correct implementation? `keep_flushing` might not actually send the value, then the next line might return `ready`.

src/sink/fanout.rs
```diff
@@ -0,0 +1,128 @@
...
+        self.keep_flushing()?;
+        self.sink.poll_complete()
+    }
+
+    fn close(&mut self) -> Poll<(), S::SinkError> {
+        self.keep_flushing()?;
```
carllerche: Same comment as `poll_complete`.

src/sink/fanout.rs
```diff
@@ -0,0 +1,128 @@
...
+            Ok(Async::NotReady)
+        }
+    }
+
+    fn close(&mut self) -> Poll<(), Self::SinkError> {
+        let left_async = self.left.close()?;
+        let right_async = self.right.close()?;
+        // Only if both downstream sinks are ready, signal readiness.
+        if left_async.is_ready() && right_async.is_ready() {
+            Ok(Async::Ready(()))
```
carllerche: This shouldn't return `ready` until both calls to `close` return ready.

srijs: That should be what's checked on line 84. Or am I misunderstanding you?

src/sink/fanout.rs
srijs: Good catch, I'll add a check here!

src/sink/fanout.rs
srijs: The check on the line below (`self.state.is_ready()`) provides the desired check here, as far as I can see.

srijs:
@carllerche Addressed your feedback, looking forward to a second review!

alexcrichton:
Thanks @srijs!

src/sink/fanout.rs
```diff
@@ -0,0 +1,135 @@
...
+        }
+        Ok(())
+    }
+
+    fn poll_complete(&mut self) -> Poll<(), S::SinkError> {
+        self.keep_flushing()?;
+        let async = self.sink.poll_complete()?;
+        // Only if all values have been sent _and_ the underlying
+        // sink is completely flushed, signal readiness.
+        if self.state.is_ready() && async.is_ready() {
```
carllerche: You shouldn't have to check `self.state.is_ready()` here as the `keep_flushing` check sit.

PR #634, opened 2017-11-08
Deprecate Future::wait.

carllerche:
This commit deprecates `Future::wait`. Instead of having a trait function, a wrapper is provided at `future::Blocking`. `Blocking` wraps a `Future` value and provides thread-blocking operation on it.

The deprecation strategy differs from that proposed in tokio-rs/tokio-rfcs#3. The intent is that there still will be current-thread executor functions described in the PR but those will come after.

This PR proposes adding `future::Blocking` which wraps a `T: Future` and provides thread-blocking operations. Currently, only `wait` and `poll` are provided, but `wait_timeout` could be added easily later.

I opted for this strategy because in [@alex's branch](https://github.com/alexcrichton/futures-rs/tree/spawn), he provided both `BlockingStream` and `BlockingSink`, yet `BlockingFuture` was omitted. If `BlockingSink` and `BlockingStream` are worth having (and I think they are), then `BlockingFuture` should also be provided. This proves to be a natural place to add any functions for using a future outside the scope of an executor.

I believe that this solves the problems being encountered with `wait` by not being a trait function (you won't find this function by looking at what the `Future` trait provides). Also, in the future, thread-blocking operations will panic when invoked from the context of an executor.

This PR opts to name the type `future::Blocking`.

If this PR is accepted, I will apply a similar treatment to `Sink` and `Stream`.

This PR is based against the `tokio-reform` branch, and not master.

## Remaining

* [x] Deprecate `Wait` types
* [x] Deprecate functions on `Spawn`.
* [x] Check docs
* [x] Add test for new functionality.

### DIFF
--- a/src/future/blocking.rs
+++ b/src/future/blocking.rs
@@ -1,6 +1,6 @@
 use Async;
 use future::Future;
-use executor;
+use executor::{self, NotifyHandle};
 use task_impl::ThreadNotify;
 
 /// Provides thread-blocking operations on a future.
@@ -34,14 +34,12 @@ impl<T: Future> Blocking<T> {
     /// the inner future is not in a ready state.
     ///
     /// This function will return immediately if the inner future is not ready.
-    pub fn poll(&mut self) -> Option<Result<T::Item, T::Error>> {
-        ThreadNotify::with_current(|notify| {
-            match self.inner.poll_future_notify(notify, 0) {
-                Ok(Async::NotReady) => None,
-                Ok(Async::Ready(v)) => Some(Ok(v)),
-                Err(e) => Some(Err(e)),
-            }
-        })
+    pub fn try_take(&mut self) -> Option<Result<T::Item, T::Error>> {
+        match self.inner.poll_future_notify(&NotifyHandle::noop(), 0) {
+            Ok(Async::NotReady) => None,
+            Ok(Async::Ready(v)) => Some(Ok(v)),
+            Err(e) => Some(Err(e)),
+        }
     }
 
     /// Block the current thread until this future is resolved.
--- a/src/future/blocking.rs
+++ b/src/future/blocking.rs
@@ -86,7 +86,7 @@ impl<T> Blocking<T> {
         self.inner.get_mut()
     }
 
-    /// Consume the `Blocking`, returning its inner inner future.
+    /// Consume the `Blocking`, returning its inner future.
     pub fn into_inner(self) -> T {
         self.inner.into_inner()
     }
--- a/src/future/blocking.rs
+++ b/src/future/blocking.rs
@@ -7,6 +7,7 @@ use task_impl::ThreadNotify;
 ///
 /// See [`blocking`](fn.blocking.html) documentation for more details.
 #[derive(Debug)]
+#[must_use = "futures do nothing unless used"]
 pub struct Blocking<T> {
     inner: executor::Spawn<T>,
 }
--- a/src/future/blocking.rs
+++ b/src/future/blocking.rs
@@ -5,7 +5,7 @@ use task_impl::ThreadNotify;
 
 /// Provides thread-blocking operations on a future.
 ///
-/// See [`blocking`] documentation for more details.
+/// See [`blocking`](fn.blocking.html) documentation for more details.
 #[derive(Debug)]
 pub struct Blocking<T> {
     inner: executor::Spawn<T>,
@@ -24,7 +24,7 @@ pub struct Blocking<T> {
 /// block the task executor's progress.
 pub fn blocking<T: Future>(future: T) -> Blocking<T> {
     let inner = executor::spawn(future);
-    Blocking { inner }
+    Blocking { inner: inner }
 }
 
 impl<T: Future> Blocking<T> {
--- a/src/future/blocking.rs
+++ b/src/future/blocking.rs
@@ -1,0 +1,95 @@
+use Async;
+use future::Future;
+use executor;
+use task_impl::ThreadNotify;
+
+/// Provides thread-blocking operations on a future.
+///
+/// See [`blocking`] documentation for more details.
+#[derive(Debug)]
+pub struct Blocking<T> {
+    inner: executor::Spawn<T>,
+}
+
+/// Provides thread-blocking operations on a future.
+///
+/// `blocking` consumes ownership of `future`, returning a `Blocking` backed by
+/// the future. The `Blocking` value exposes thread-blocking operations that
+/// allow getting the realized value of the future. For example,
+/// `Blocking::wait` will block the current thread until the inner future has
+/// completed and return the completed value.
+///
+/// **These operations will block the current thread**. This means that they
+/// should not be called while in the context of a task executor as this will
+/// block the task executor's progress.
+pub fn blocking<T: Future>(future: T) -> Blocking<T> {
+    let inner = executor::spawn(future);
+    Blocking { inner }
+}
+
+impl<T: Future> Blocking<T> {
+    /// Query the inner future to see if its value has become available.
+    ///
+    /// Unlike `Future::poll`, this function does **not** register interest if
+    /// the inner future is not in a ready state.
+    ///
+    /// This function will return immediately if the inner future is not ready.
+    pub fn poll(&mut self) -> Option<Result<T::Item, T::Error>> {
+        ThreadNotify::with_current(|notify| {
+            match self.inner.poll_future_notify(notify, 0) {
+                Ok(Async::NotReady) => None,
+                Ok(Async::Ready(v)) => Some(Ok(v)),
+                Err(e) => Some(Err(e)),
+            }
+        })
+    }
+
+    /// Block the current thread until this future is resolved.
+    ///
+    /// This method will drive the inner future to completion via
+    /// `Future::poll`. **The current thread will be blocked** until the future
+    /// transitions to a ready state. Once the future is complete, the result of
+    /// this future is returned.
+    ///
+    /// > **Note:** This method is not appropriate to call on event loops or
+    /// >           similar I/O situations because it will prevent the event
+    /// >           loop from making progress (this blocks the thread). This
+    /// >           method should only be called when it's guaranteed that the
+    /// >           blocking work associated with this future will be completed
+    /// >           by another thread.
+    ///
+    /// This method is only available when the `use_std` feature of this
+    /// library is activated, and it is activated by default.
+    ///
+    /// # Panics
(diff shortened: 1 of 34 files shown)


carllerche:
Also worth noting, calling `future:::Blocking::wait` will *only* drive the current future and *not* the current thread executor. This avoids the problem of accidentally dropping tasks that never completed.

src/future/blocking.rs
```diff
@@ -0,0 +1,95 @@
...
+}
+
+impl<T: Future> Blocking<T> {
+    /// Query the inner future to see if its value has become available.
+    ///
+    /// Unlike `Future::poll`, this function does **not** register interest if
+    /// the inner future is not in a ready state.
+    ///
+    /// This function will return immediately if the inner future is not ready.
+    pub fn poll(&mut self) -> Option<Result<T::Item, T::Error>> {
```
seanmonstar: The name `poll` returning something that isn't `Poll` seems confusing, when every other future or stream `poll` will. Maybe something like `peek` or `tap`?

src/future/blocking.rs
```diff
@@ -0,0 +1,95 @@
+use Async;
+use future::Future;
+use executor;
+use task_impl::ThreadNotify;
+
+/// Provides thread-blocking operations on a future.
+///
+/// See [`blocking`] documentation for more details.
```
Thomasdezeeuw: This links to nothing, missing "(fn.blocking.html)"?

src/future/blocking.rs
carllerche: I'm open to change the name if a good alternative is proposed.

* `peek` implies the value remains in the future (and there is a peek combinator that does this on Future).
* `tap` I don't understand how tap applies to the behavior (I am not familiar w/ tap as a name here).

Re confusion, the return type is different, so at least it won't compile if you are expecting `Poll`.

carllerche:
I would also want to add some tests for `Blocking::poll` (naming TBD?) before merging. I am waiting to hear if this is PR represents a good strategy.

aturon:
@carllerche It'd be helpful to see the rationale for this API surface as an addition beyond what's in the RFC (maybe I missed the motivation somewhere?) How should one think about having both this and the current-thread executor?

> Also worth noting, calling `future:::Blocking::wait` will *only* drive the current future and *not* the current thread executor. This avoids the problem of accidentally dropping tasks that never completed.

I'm also curious to understand how this fits into the RFC design, which was also attempting to solve this problem.

carllerche:
@aturon A few things:

* The RFC doesn't touch on what to do with `Stream::wait` and `Sink::wait`.
* The latest version of the RFC doesn't actually provide a way to block the thread waiting until a single future completes.

@alexcrichton's PR deals with `Stream::wait` and `Sink:wait` by providing `BlockingStream` and `BlockingSink`, but there is no way in the PR to block waiting for a single future to complete. This fact is apparent when trying to update the tests.

It doesn't make sense to include `BlockingStream` and `BlockingSink` without including `BlockingFuture`. So, either those two types should go or a `BlockingFuture` equivalent should be added.

> I'm also curious to understand how this fits into the RFC design, which was also attempting to solve this problem.

The goal of the RFC was to prevent being able to spawn current-thread tasks and having them be implicitly dropped. This PR maintains this goal.

* `Blocking::wait` does not "enter" the current-thread executor, as such a current-thread spawn will panic.

aturon:
@carllerche Got it, thanks! And yeah, an earlier version of the RFC had something like this (though it was less elaborate; it wasn't a wrapper), which was lost in the shuffle.

I'm good with this approach!

src/future/blocking.rs
aturon: I want to second @seanmonstar's point about `poll`; we've used that pretty universally to designate that interest *will* be registered.

It could match cases with blocking data structures in `std` and be called `try_wait`, but that's not awesome.

Also, it's a *tad* confusing for this to be part of the `Blocking` wrapper, since the point here is precisely that it *doesn't* block :-)

carllerche: It could be named something besides `Blocking`. The real goal of the type is "Use this future from the edge outside of the future task system", but I went with `Blocking` to make that behavior very explicit.

I considered `try_wait`, but that doesn't really make sense as it isn't trying to wait...

carllerche: The only other name that I thought of would be `try_take`.

aturon: Oh I'm not complaining about the name `Blocking` (which I like), I'm saying that `Blocking::poll` is a bit confusing because it's *not* in fact blocking. If you're wanting to poll the future outside of an executor, why would you think to reach for `Blocking` to do so?

But anyway, I'm more concerned that we find a name other than `poll`.

aturon: re: `try_take`, I was thinking along similar lines. Seems good?

carllerche: Does anyone hate `try_take` instead of `poll`?

twmb: I'm confused as to how this avoids registering interest. My understanding of futures is that they register with the current task before returning `NotReady`. With `poll_future_notify` below, isn't the interest is still registered?

twmb:
Is there a reason for `try_take` to exist at the moment? I realize that a no-op notifier is the only way to internally implement it, but w.r.t. the inability to avoid actually registering interest, it seems like fitting a square peg into a round hole. Also, I'm not sure why I would wrap a future in `Blocking` and then have the desire to check it in a non-blocking manner. This also extends down into the `get_ref`, `get_mut`, and `into_inner` functions - if the only purpose of `Blocking` is truly to block until the future returns, there is no need to access a `Blocking`'s internals.

Additionally, I feel as if `blocking`, and its corresponding type type `Blocking`, feel ergonomically awkward to read and I think that feeling would translate to use as well. I recognize that a pattern in the futures crate is to name wrapper types after the functions that produce them (`and_then` -> `AndThen`, etc.), but those function names are all verbs that extend off a `Future`, and the types are really only used in in return types (something that should go away once `impl Trait` is stable).

I'd prefer something closer to `blocking::wrap(f)` or `block::wrap(f)` - this _reads_ to me as something that provides blocking is wrapping a future, which is exactly what `blocking(f)` does. Assuming that `.wait()` will be chained on that 99% of the time, `blocking::wrap(f).wait()` still gives me a bit more context as to what is happening as opposed to `blocking(f).wait()`. Alternatively, `block_on(f)`, which would directly replace `wait` as a purely blocking function call).

In terms of the type `Blocking` itself - this feels the weirdest, because the -ing suffix makes me read it as a verb, but the word itself is a [verbal noun](https://forum.wordreference.com/threads/deverbal-and-verbal-nouns.2689882/#post-13577342). The lack of a sentence as context makes the isolation weird. This is nitpicky, and feel free to ignore this and the prior two paragraphs. I bring this up (and will add more in other tickets / prs / rfc#3) because I had a difficult time reading and understanding the futures api and source due to small semantic things such as this.

carllerche:
@twmb 

`try_take` is very useful in tests.

> if the only purpose of Blocking is truly to block until the future returns, there is no need to access a Blocking's internals.

I don't understand why you believe this to be true. A future is very stateful, there is plenty of reason to want to access the inner T before or after blocking. Also, at some point there will be a `wait_timeout`.

Re naming, the only thing that I care about is that it is succinct, and clearly communicates that the type can block the thread.

twmb:
> I don't understand why you believe this to be true.

I meant that only if `try_take` were to be removed and the only option were to be `wait` - `wait_timeout` is more compelling. I'm still a bit skeptical on providing a public function that is mostly useful for tests, but that's not really a showstopper for me.

re: Re naming, if anybody else has input, that would be great as well. `Blocking` and `blocking` are likely perfectly fine, but I figured I would throw that out there.

src/future/blocking.rs
```diff
@@ -0,0 +1,94 @@
...
+    }
+
+    /// Block the current thread until this future is resolved.
+    ///
+    /// This method will drive the inner future to completion via
+    /// `Future::poll`. **The current thread will be blocked** until the future
+    /// transitions to a ready state. Once the future is complete, the result of
+    /// this future is returned.
+    ///
+    /// > **Note:** This method is not appropriate to call on event loops or
```
aturon: We should make the function *panic* in those cases, similarly to what was proposed in the RFC. Basically, just need to assert that there is not currently an executor. But maybe you're planning on doing that once we get the `Enter` stuff in?

src/sink/blocking.rs
```diff
@@ -0,0 +1,114 @@
...
+/// `blocking` consumes ownership of `sink`, returning a `Blocking` backed by
+/// the sink. The `Blocking` value exposes thread-blocking operations that allow
+/// sending values into the sink. For example, `Blocking::send` will block the
+/// current thread until the inner sink has capacity to accept the item being
+/// sent.
+///
+/// **These operations will block the current thread**. This means that they
+/// should not be called while in the context of a task executor as this will
+/// block the task executor's progress. Also, note that **this value will block
+/// the current thread on drop**. Droping `Blocking` will ensure that the inner
```
aturon: Might be worth mentioning that this is analogous to dropping a `BufWriter` in `std`.

src/sink/blocking.rs
```diff
@@ -0,0 +1,114 @@
...
+use executor;
+use task_impl::ThreadNotify;
+
+/// Provides thread-blocking operations on a sink.
+///
+/// See [`blocking`](fn.blocking.html) documentation for more details.
+#[derive(Debug)]
+#[must_use = "sinks do nothing unless used"]
+pub struct Blocking<T: Sink> {
+    inner: Option<executor::Spawn<T>>,
```
aturon: I can't figure out why this is an `Option`; AFAICT it's always `Some` (and everything except the dtor asserts that).

src/future/blocking.rs
carllerche: Yes, this step will happen in future PRs.

src/sink/blocking.rs
carllerche: It is an `Option` to support `into_inner` (which has to be able to extract the inner type without triggering the dtor).

aturon: Ah yep, missed that. :+1:

carllerche:
@alexcrichton could you enable "squash & merge" and "rebase and merge" for this repo?

stbuehler:
I'm curious why `blocking` is not a trait method; using `blocking` `Future`s and `Stream`s in the same module feels unnecessary complicated.

Also there are a lot of doc tests (and some normal unit tests) using the old `wait` functions; they should be converted to `blocking`.

PR #639, opened 2017-11-10
Current thread executor

carllerche:
Provides a current thread executor allowing users to spawn tasks that are guaranteed to run on the current thread.

This is based on the Tokio reform RFC but made some API tweaks.

## Remaining

* [x] Get CI passing.
* [x] More tests.
  * [x] Fairness
  * [x] Cancellation
* [x] Benchmark.
* [x] Docs
* [x] API bikeshedding.


src/executor/current_thread.rs
```diff
@@ -0,0 +1,361 @@
...
+use std::sync::Arc;
+
+/// Executes tasks on the current thread.
+#[derive(Debug, Clone)]
+pub struct CurrentThread {
+    // Prevent the handle from moving across threads.
+    _p: ::std::marker::PhantomData<Rc<()>>,
+}
+
+/// Executes dameonized tasks on the current thread.
```
cramertj: Here and elsewhere it would be good to clarify what it means for a task to be a "daemon"-- i.e. it executes indefinitely, and it does not prevent functions like block_on_all from returning.

src/executor/current_thread.rs
```diff
@@ -0,0 +1,361 @@
...
+
+    /// Returns a new handle that spawns daemonized tasks on the current thread.
+    pub fn daemon_executor(&self) -> DaemonExecutor {
+        DaemonExecutor {
+            _p: ::std::marker::PhantomData,
+        }
+    }
+
+    /// Execute the given future *synchronously* on the current thread, blocking
+    /// until it (and all spawned tasks) completes and returning its result.
```
cramertj: *all non-daemon tasks.

src/executor/current_thread.rs
```diff
@@ -0,0 +1,361 @@
...
+    ///
+    /// Note that there is no `'static` or `Send` requirement on the future.
+    pub fn block_on_all<F: Future>(future: F) -> Result<F::Item, F::Error> {
+        TaskRunner::enter(|| future)
+    }
+
+    /// Execute the given closure, then block until all spawned tasks complete.
+    ///
+    /// In more detail, this function will block until:
+    /// - All spawned tasks are complete, or
```
cramertj: *non-daemon

src/executor/current_thread.rs
```diff
@@ -0,0 +1,361 @@
...
+    where F: Future<Item = (), Error = ()> + 'static
+    {
+        spawn(future, false).unwrap_or_else(|_| {
+            panic!("cannot call `spawn` unless your thread is already \
+                    in the context of a call to `block_on_all` or \
+                    `block_with_init`")
+        })
+    }
+
+    /// Spawns a daemon, which does *not* block the pending `block_on_all` call.
```
cramertj: ...which is *a task* that does not block....

src/scheduler.rs
```diff
@@ -0,0 +1,663 @@
...
+/// This is used both by `FuturesUnordered` and the current-thread executor.
+pub struct Scheduler<T, W> {
+    inner: Arc<Inner<T, W>>,
+    len: usize,
+    head_all: *const Node<T, W>,
+}
+
+pub struct Notify<'a, T: 'a, W: 'a>(&'a Arc<Node<T, W>>);
+
+/// Wakeup a sleeper
```
cramertj: Probably worth clarifying what a "sleeper" is. My understanding is its something that is in charge of polling the scheduler (either a Task in the FuturesUnordered case or a thread in the current-thread executor).

src/scheduler.rs
```diff
@@ -0,0 +1,663 @@
...
+use std::fmt::{self, Debug};
+use std::marker::PhantomData;
+use std::mem;
+use std::ptr;
+use std::sync::atomic::Ordering::{Relaxed, SeqCst, Acquire, Release, AcqRel};
+use std::sync::atomic::{AtomicPtr, AtomicBool};
+use std::sync::{Arc, Weak};
+use std::usize;
+
+/// A generic task-aware scheduler.
```
cramertj: Could you add a more in-depth comment? Something to the effect of: Scheduler is used to run many child tasks inside of another thing that can be woken up via a "Wakeup".

src/scheduler.rs
```diff
@@ -0,0 +1,663 @@
...
+
+    unsafe fn drop_raw(&self) {
+        let mut me: *const ArcNode<T, W> = self;
+        let me = &mut me as *mut *const ArcNode<T, W> as *mut Arc<Node<T, W>>;
+        ptr::drop_in_place(me);
+    }
+}
+
+unsafe fn hide_lt<T, W: Wakeup>(p: *mut ArcNode<T, W>) -> *mut UnsafeNotify {
+    mem::transmute(p as *mut UnsafeNotify)
```
cramertj: I think this `transmute` call isn't actually doing anything?

carllerche: It is hiding the lifetime :)

That said, I think that `W` needs to be `'static`.

src/scheduler.rs
carllerche: I would definitely appreciate a patch that improves on the docs 👍

src/executor/current_thread.rs
```diff
@@ -0,0 +1,361 @@
...
+    /// Return a new `TaskRunner`
+    fn new(thread_notify: Arc<ThreadNotify>) -> TaskRunner {
+        TaskRunner {
+            non_daemons: 0,
+            scheduler: scheduler::Scheduler::new(thread_notify),
+        }
+    }
+
+    /// Enter a new `TaskRunner` context
+    fn enter<F, A>(f: F) -> Result<A::Item, A::Error>
```
aturon: Would appreciate a bit more descriptive comments on this and the next couple top-level functions.

src/executor/current_thread.rs
```diff
@@ -0,0 +1,361 @@
...
+    {
+        let mut result = None;
+        let mut future = Some(executor::spawn(future));
+
+        while future.is_some() || self.non_daemons > 0 {
+            if current.cancel.get() {
+                // TODO: This probably can be improved
+                current.cancel.set(false);
+
+                debug_assert!(current.scheduler.get().is_null());
```
aturon: This assert initially caught me by surprise, because I hadn't caught that the scheduler was only temporarily established in `enter`. Can we make this flow more clear?

src/scheduler.rs
```diff
@@ -0,0 +1,653 @@
...
+        // polling here.
+        self.inner.enqueue(ptr);
+    }
+
+    /// Advance the scheduler state.
+    ///
+    /// This function should be called whenever the caller is notified via a
+    /// wakeup.
+    pub fn tick<F, R>(&mut self, mut f: F) -> Tick<R>
+    where F: FnMut(&mut Self, &mut T) -> Async<R>
```
aturon: Very clever to thread the `&mut` issue by passing it back to the callback.

src/executor/current_thread.rs
```diff
@@ -0,0 +1,361 @@
...
+        result.unwrap()
+    }
+
+    fn poll_all(&mut self, current: &CurrentRunner) {
+        use scheduler::Tick;
+
+        loop {
+            let res = self.scheduler.tick(|scheduler, spawned, notify| {
+                current.set_scheduler(scheduler, || {
+                    match spawned.inner.0.poll_future_notify(notify, 0) {
```
aturon: I'd love some commentary here about the fact that `tick` gives us `&mut` access back to the scheduler, which provides some amount of safety checking against re-entrancy issues.

carllerche:
@aturon thanks for the review. I’m currently out of town but will try to address shortly. Docs should definitely be improved before merging. I just wanted to get the code up ASAP for initial review. I should have included a doc checkbox in the original Pr.

carllerche:
I'm going to remove `CurrentThread::block_on_all` for now. It is pretty similar to `block_with_init`, so I think we should be conservative for now with the API.

The original reason for `block_on_all` was to drive a single future to completion, but that has been handled with `future::blocking(fut).wait()`.

Either way, it can be added back later.

carllerche:
I don't love the name `cancel_all_spawned` as I think we have mixed up our terminology here.

carllerche:
In general, we have "executor" and "execute", but `CurrentThread::spawn`. We probably should converge to execute and move away from `spawn`

cramertj:
> In general, we have "executor" and "execute", but CurrentThread::spawn. We probably should converge to execute and move away from spawn

If I saw `execute` i'd personally expect it to be a call to the method on the [Executor](http://alexcrichton.com/futures-rs/futures/executor/trait.Executor.html) trait.

carllerche:
@cramertj That is basically what it is.

carllerche:
I have made a number of API changes in this PR compared to the RFC:

**`CurrentThread` instead of free functions**

The PR just proposes a bucket of free functions. Instead, I put all of these functions on `executor::CurrentThread`. This type also is the implementor of the `Executor` trait.

I assume that at some point, there will also be `executor::Pool` as a facade over a customizable thread pool implementation.

**Naming polish**

I renamed `spawn` -> `execute` to bring it inline with the `Executor` trait. There are probably more tweaks besides this.

**Merged `block_for_all` and `block_with_init`.

Now there is only one blocking function. The original split makes more sense without the `future::blocking` helper.

**Pass `&mut current_thread::Context` to the `block_with_init` closure.

The RFC calls for passing `&mut executor::Enter` directly, however, this isn't very future proof. Instead, I define a new type that will eventually contain `&mut executor::Enter`. This way, if the current thread executor wants to eventually provide more context, it is able to do so without breaking changes.

I personally just picked `Context` without putting much thought into it, so if there are other ideas for naming, please post them.

src/executor/current_thread.rs
```diff
@@ -1,4 +1,69 @@
...
+//! have not yet completed are immediately dropped.
+//!
+//! A daemon task can be spawned with [`CurrentThread::spawn_daemon`].
+//!
+//! [here]: https://tokio.rs/docs/going-deeper-futures/tasks/
+//! [`CurrentThread`]: struct.CurrentThread.html
+//! [`block_with_init`]: struct.CurrentThread.html#method.block_with_init
+//! [`block_on_all`]: struct.CurrentThread.html#method.block_on_all
+//! [`CurrentThread::spawn`]: struct.CurrentThread.html#method.spawn
+//! [`CurrentThread::spawn_daemon`]: struct.CurrentThread.html#method.spawn_daemon
```
aturon: Docs look great :+1:

src/executor/current_thread.rs
```diff
@@ -14,13 +79,25 @@ use std::rc::Rc;
 use std::sync::Arc;
 
 /// Executes tasks on the current thread.
+///
+/// All tasks spawned using this executor will be executed on the current thread
```
aturon: This is confusing given the inherent methods (which can spawn daemons). Maybe you want to say "when used as an `Executor`?

aturon:
Other than the spawn/execute issue, the other departures from the RFC are largely cosmetic and are fine by me.

carllerche:
@aturon I personally prefer the `spawn` terminology as well, however I think that having disjoint terminology is worse. Precise and consistent terminology is valuable. I made note of the naming issue in a tracking PR (#645).

carllerche:
@aturon OK, this PR should be good to go (assuming CI passes). Please provide final PR review.

PR #641, opened 2017-11-13
Oneshot race condition quick fix

Diggsey:
Should be easier to review than #624, although I still consider that a superior implementation overall, as it does a lot less work on each operation and has weaker memory ordering requirements due to synchronisation being via a single atomic variable.

cc @alexcrichton


alexcrichton:
Thanks!

PR #647, opened 2017-11-17
Implement executor::Enter

carllerche:
This implements executor::Enter as described by the Tokio reform RFC. This helps prevent accidentally blocking an executor thread.

This does not update CpuPool to use enter as the strategy by which to do this is not clear. Specifically, calling `executor::enter` from CpuPool threads is easy, but there is no way to expose the `&Enter` handle to the user (this also probably shouldn't be exposed directly).

### DIFF
--- a/src/executor/enter.rs
+++ b/src/executor/enter.rs
@@ -1,0 +1,89 @@
+use std::prelude::v1::*;
+use std::cell::{RefCell, Cell};
+use std::fmt;
+
+thread_local!(static ENTERED: Cell<bool> = Cell::new(false));
+
+/// Represents an executor context.
+///
+/// For more details, see [`enter` documentation](fn.enter.html)
+pub struct Enter {
+    on_exit: RefCell<Vec<Box<Callback>>>,
+    permanent: bool,
+}
+
+/// Marks the current thread as being within the dynamic extent of an
+/// executor.
+///
+/// Executor implementations should call this function before blocking the
+/// thread. If `None` is returned, the executor should fail by panicking or
+/// taking some other action without blocking the current thread. This prevents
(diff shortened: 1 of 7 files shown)


src/executor/enter.rs
```diff
@@ -0,0 +1,89 @@
...
+        self.on_exit.borrow_mut().push(Box::new(f));
+    }
+
+    /// Treat the remainder of execution on this thread as part of an
+    /// executor; used mostly for thread pool worker threads.
+    ///
+    /// All registered `on_exit` callbacks are *dropped* without being
+    /// invoked.
+    pub fn make_permanent(mut self) {
+        self.permanent = true;
```
aturon: This needs to actually drop the callbacks (and no further ones should be accepted).

aturon:
> This does not update CpuPool to use enter as the strategy by which to do this is not clear. Specifically, calling `executor::enter` from CpuPool threads is easy, but there is no way to expose the `&Enter` handle to the user (this also probably shouldn't be exposed directly).

So the intent, in my mind, was that `&Enter` should only be exposed during the initialization phase of an executor (e.g., it may allow the client to register an initialization hook). I believe some parts of the RFC depend on you not being able to get your hands on an `&Enter` arbitrarily.

src/executor/enter.rs
carllerche: Isn't that handled by the fact that it takes `self` via ownership?

aturon: Missed that. You're right.

carllerche:
@aturon either way, updating CpuPool doesn't need to block this PR. It can be tracked in the meta issue. So, let me know if you have any other change requests. If not, plz give the +1.

aturon:
Approved.

PR #648, opened 2017-11-18
add a new api: `any` to  `Stream`

nooberfsh:
it is  same with  `for_each` except that `any` will not  be "blocked" or "exited" by the results of the generated futures. 

for example:
```rust
    let mut core = Core::new().unwrap();
    let handle = core.handle();

    // Bind the server's socket
    let addr = "127.0.0.1:12345".parse().unwrap();
    let tcp = TcpListener::bind(&addr, &handle).unwrap();

    // use `any` instead of `for_each`
    let server = tcp.incoming().any(|(tcp, _)| {
        // Split up the read and write halves
        let (reader, writer) = tcp.split();

        // Future of the copy
        let bytes_copied = io::copy(reader, writer);

        // ... after which we'll print what happened
        bytes_copied.map(|(n, _, _)| {
            println!("wrote {} bytes", n)
        }).map_err(|err| {
            println!("IO error {:?}", err)
        })

        // do not need any more.
        //handle.spawn(handle_conn);
        //Ok(())
    });

```

### DIFF
--- a/src/stream/any.rs
+++ b/src/stream/any.rs
@@ -1,0 +1,66 @@
+use {Async, Future, IntoFuture, Poll};
+use stream::{Stream, FuturesUnordered};
+
+
+/// A stream combinator which executes a unit closure over each item on a
+/// stream.
+///
+/// This structure is returned by the `Stream::any` method.
+#[derive(Debug)]
+#[must_use = "streams do nothing unless polled"]
+pub struct Any<S, F, U> where U: IntoFuture {
+    stream: S,
+    eof: bool,
+    f: F,
+    futures: FuturesUnordered<U::Future>,
+}
+
+pub fn new<S, F, U>(s: S, f: F) -> Any<S, F, U>
+    where S: Stream,
+          F: FnMut(S::Item) -> U,
+          U: IntoFuture,
+{
+    Any {
+        stream: s,
+        eof: false,
+        f: f,
+        futures: FuturesUnordered::new(),
+    }
+}
+
+impl<S, F, U> Future for Any<S, F, U>
+    where S: Stream,
+          F: FnMut(S::Item) -> U,
+          U: IntoFuture,
+{
+    type Item = ();
+    type Error = S::Error;
+
+    fn poll(&mut self) -> Poll<(), S::Error> {
+        while !self.eof {
+            match self.stream.poll()? {
+                Async::Ready(Some(d)) => {
+                    let f = (self.f)(d).into_future();
+                    self.futures.push(f);
+                }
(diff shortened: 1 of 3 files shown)


carllerche:
Thanks. This looks similar to buffer_unordered. Could you clarify the differences.

nooberfsh:
`BufferUnordered` is a `Stream`, `Any` is a `Future`.
`BufferUnordered`  is a buffered version of `AndThen`
`Any` is a buffered version of `ForEach`.
the main difference is that `Any` just poll the internel futures and  does not care about the  
result of the future.  It will return Error only when the internal stream failes.

alexcrichton:
Thanks for the PR!

I'm a little worried about this wrt `Iterator::any` and the semantics here. The `any` method typically takes a list of bools and returns if any of them were true, but it looks like this is pretty different? It also looks like this is discarding all the results (even errors!) from the internal futures. For these precise semantics I'd be more comfortable with something like:


```rust
stream.for_each(|x| x.then(|_| Ok(())))
````

that way you're explicitly ignoring the error I think?

nooberfsh:
consider a such situation：
1: you do not have a handle of the excutor.
2: you do not want the stream to be blocked because of the current item return not ready.

any solves this problem.  In the tokio example, use any can be less error prone and more ergonomic.

alexcrichton:
@nooberfsh my point though is that I don't think this is the behavior you want most of the time (ignoring errors etc) and I believe that this combinator is equivalent to `stream.for_each(|x| x.then(|_| Ok(())))` which is quite short. In that sense it doesn't quite justify itself to me as existing as a standalone combinator.

stbuehler:
I think `stream.any(f)` is equivalent to `stream.map(|x| f(x).then(|_| Ok(()))).buffer_unordered(!0usize).for_each(|x| x)`, i.e. ignoring results apart from errors, run in parallel without limits, then pick the first error or return `Ok(())` when the stream ends.

`stream.map(f).buffer_unordered(!0usize).for_each(|x| x.then(|_| Ok(()))` is similar (but might keep the intermediate results a little bit longer), but still longer than `stream.for_each(|x| x.then(|_| Ok(())))` :)

I think `buffer_unordered(!0usize)` (or equivalent) should not be hidden - the buffer should be limited, and the limit should be exposed.

Also I think combinators shouldn't ignore inner results - i.e. `any` should require `U::Future::Item = ()`; then it basically becomes `stream.map(f).buffer_unordered(!0usize).for_each(|x| x)`.

I too don't see the need to make this a standalone combinator.

alexcrichton:
Ok I'm going to close this for now, but let me know if this is in error!

PR #654, opened 2017-11-21
CpuPool: add `stack_size` option on `Builder`.

hicqu:


alexcrichton:
Thanks!

futures-cpupool/src/lib.rs
```diff
@@ -411,6 +419,9 @@ impl Builder {
             if let Some(ref name_prefix) = self.name_prefix {
                 thread_builder = thread_builder.name(format!("{}{}", name_prefix, counter));
             }
+            if self.stack_size > 0 {
```
tailhook: Em... it isn't very rust'y. Don't we use options for such things?

PR #658, opened 2017-11-24
test: make sure mpsc::spawn streams get dropped when receiver is dropped

stbuehler:
`mpsc::spawn` says:

> The `stream` will be canceled if the `SpawnHandle` is dropped.

But actually the `stream` will only get dropped (my reading of "canceled") after it produced an item that couldn't be sent over the (now dead) channel.

This pull request adds two tests (one for `sync`, one for `unsync`) to make sure the `stream` actually gets dropped; the tests should be failing right now.

### DIFF
--- a/src/sync/mpsc/mod.rs
+++ b/src/sync/mpsc/mod.rs
@@ -77,6 +77,7 @@ use std::thread;
 use std::usize;
 
 use sync::mpsc::queue::{Queue, PopResult};
+use sync::oneshot;
 use task::{self, Task};
 use future::Executor;
 use sink::SendAll;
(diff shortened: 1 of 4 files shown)


alexcrichton:
Hm the tests are passing?

stbuehler:
That's because I'm stupid :) I wanted to see whether they actually fail and added `#[should_panic]` - removed those lines, so they really should fail now.

stbuehler:
I'll try.

stbuehler:
At first I though it might be possible to query the "open" state of the sinks in the `Execute` futures. It is, but if the sink isn't parked (due to a pending send), the task won't get notified when the receiver is closed.  It probably isn't a good idea to modify this behavior or clone all those types just to add it.

I still have another idea :)

alexcrichton:
I think implementation-wise this may be difficult to do as the same notification mechanism present on oneshots isn't present on channels, but it may still be possible via various trickery.

stbuehler:
Sorry for abusing the same branch and pull request for a new feature :)

Basically `job` is a generic way of canceling futures by dropping a handle.  If there is a simpler way to do this I'm happy to look into it.

alexcrichton:
Hm ok, thanks for the fix!

I wonder, though, the job idea here seems very similar to the oneshot-style concept, no? Would it be possible to perhaps leverage those existing concepts and have something like this bake on crates.io first before moving into this crate?

stbuehler:
Ok, so here is another approach with `oneshot`.

alexcrichton:
Awesome, thanks!

PR #659, opened 2017-11-24
Better error message when polling CpuFuture twice

tailhook:
This should help debugging. Because previous message `shouldn't be canceled: Canceled` look like a bug in cpupool at a glance.


alexcrichton:
Thanks!

PR #663, opened 2017-11-27
deprecate unneeded types, fix usage of deprecated API

stbuehler:


stbuehler:
Regarding the failed test:

```
thread '<unnamed>' panicked at 'assertion failed: tx.start_send("fail").unwrap().is_not_ready()', tests\mpsc.rs:456:12
[...]
---- try_send_2 stdout ----
	thread 'try_send_2' panicked at 'assertion failed: `(left == right)`
  left: `Some(Ok("fail"))`,
 right: `Some(Ok("goodbye"))`', tests\mpsc.rs:469:4
```

My guess is that 300ms were not enough for the background thread to fail sending, but I don't think this is my fault :)

alexcrichton:
Awesome, thanks!

PR #664, opened 2017-11-30
convert Run to Notify API

stbuehler:
`Run` (created by `Spawn::execute`) still used the deprecated `Unpark` API; use `Notify` instead. (`Run` and friends are still required by `futures-cpupool`, so they probably shouldn't be marked deprecated just yet).

After this no public non-deprecated feature uses deprecated features, and the second commit will add `#[cfg(feature = "with-deprecated")]` to all `deprecated` and `allow(deprecated)`  items.

Without `with-deprecated` it will always use the stub `UnparkEvents` and `BorrowedEvents` from the `core` implementation.

This should improve benchmarks without `with-deprecated` overall, but `unbounded_uncontended` and `thread_yield_multi_thread` seem to get slower by this...

### DIFF
--- a/src/task_impl/std/mod.rs
+++ b/src/task_impl/std/mod.rs
@@ -438,7 +438,6 @@ struct RunInner {
 impl Run {
     /// Actually run the task (invoking `poll` on its future) on the current
     /// thread.
-    #[allow(deprecated)]
     pub fn run(self) {
         let Run { mut spawn, inner } = self;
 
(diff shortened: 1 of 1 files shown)


alexcrichton:
Thanks for the PR! The `with-deprecated` feature though was sort of a hack added awhile back which we unfortunately can't actually expand like this in a backwards compatible fashion. For example if a crate disabled default features and then didn't enable `with-deprecated` this commit could run the risk of "deleting" used apis for that crate.

Could this for now just include the `execute` changes? We can hopefully eventually just remove the `with-deprecated` business entirely!

stbuehler:
Right, I see.  So there could theoretically be a feature the other way round (`disable-deprecated`) which one would have to enable manually, instead of one enabled by default. I still think this would be nice to have :)

But I'll remove the second commit from this PR, and we can discuss in a new one.

alexcrichton:
Ok that sounds fine by me. I'm not sure I'd like to stick with the `with-deprecated` (or `disable-deprecated`) business though as it's typically only temporarily useful but can be a big burden to maintain over time (huge `#[cfg]` explosion)

stbuehler:
Removed the commit with the changes to `with-deprecated`.

alexcrichton:
Thanks!

PR #665, opened 2017-11-30
Sleep trait

carllerche:
Introduce a `Sleep` trait that executors can use to customize the strategy used to put the current thread to sleep. This will be used by Tokio to enable running an instance of `CurrentThread` and an instance of the Tokio reactor on a single thread.

This essentially makes the `CurrentThread` implementation generic over `W: Wakeup`. This introduces some trickiness as the thread-local state that allows spawning new futures into the scheduler cannot be generic over the `W`.

Handling this required some restructuring of `scheduler`. This work was done incrementally in each commit and each commit includes a commit message that describes the step.

/cc @aturon


src/executor/sleep.rs
```diff
@@ -0,0 +1,22 @@
+use std::time::Duration;
+
+/// Puts the current thread to sleep.
```
aturon: ... and allows waking it back up.

src/executor/current_thread.rs
```diff
@@ -458,37 +464,17 @@ impl CurrentRunner {
...
-    /// within the closure passed to `set_scheduler`. This function also takes a
-    /// &mut reference to the scheduler, which is essentially holding a "lock"
-    /// on that reference, preventing any other location in the code from
-    /// also getting that &mut reference.
-    ///
-    /// When `set_scheduler` returns, the thread-local variable containing the
-    /// mut reference is set to null. This is done even if the closure panics.
-    ///
-    /// This reduces the odds of introducing pointer aliasing.
-    fn set_scheduler<F, R>(&self, scheduler: &mut Scheduler, f: F) -> R
```
aturon: There are various comments in the file still referring to this function/it's implications.

src/executor/current_thread.rs
```diff
@@ -343,58 +359,58 @@ impl TaskRunner {
...
 
-                // Execute the runner.
-                //
-                // This function will not return until either
-                //
-                // a) All non daemon futures have completed execution
-                // b) `cancel_all_executing` is called, forcing the executor to
-                // return.
-                runner.run(thread_notify, current);
+                println!("~~~ DONE ENTER ~~~");
```
aturon: Needs to be cleaned out.

src/scheduler.rs
```diff
@@ -22,6 +22,34 @@ pub struct Scheduler<T, W> {
...
+/// generic must be set at compile time. This prevents letting the user run a
+/// scheduler with a custom wake up generic (for example, a Tokio reactor).
+///
+/// `Push` sets the node's `W` generic to `Void`, a type that cannot be
+/// instantiated. Then, when the node is added to a scheduler, it is transmuted
+/// to the correct `W` type.
+///
+/// This is safe for the following reasons
+///
+/// 1) The only usage of `W` in `Node` is `Option<Arc<Inner<_, W>>>`. When used
```
aturon: I feel pretty uneasy about this approach; it's "probably fine" but we don't have enough representation guarantees to be sure.

I'd be a lot more comfortable using a `SmallVec` or something similar to queue up tasks, creating nodes afterward; assuming that most polling will result in at most a couple of new tasks, this is likely to be extremely small overhead and can be done in safe code.

carllerche:
@aturon 

I will revert the removal of `Weak`. Your assessment is correct. I had forgotten that the mpsc channel required a strong ref. Incidentally, this is why I personally don't like punting issues in complex bits of code until later. It is hard to fully reload the context.

I am worried that the weak ref will cause contention on NUMA systems, but we can punt until this is demonstrated. I will add additional comments and hopefully I won't forget the context again in the future.

Re: the `W` transmute, I think I have another completely different idea that should be safe and not require a vec. I will explore.

carllerche:
@aturon the latest commit should address your feedback.

aturon:
Thanks, the changes look great! Merge when ready.

carllerche:
Ok, I fixed CI + updated impl comments.

PR #666, opened 2017-11-30
Fix Select2 RustDoc

hawkw:
The `Select2` struct's RustDoc incorrectly refers to the no longer extant `merge` combinator. I've updated the RustDoc to refer to the `select2` combinator instead, and added a link to that method.


alexcrichton:
Thanks!

PR #679, opened 2017-12-13
Implement From<Result<T, E>> for FutureResult

P-E-Meunier:
I need this to do the following: I'd like to require associated types for a trait to implement `Future`, but I also want to provide default implementations for some methods of the trait.

Now, in many simple cases, in particular at the beginning of a project, the associated types will essentially be `FutureResult`s, which is why this would be useful.


alexcrichton:
Sure!

PR #680, opened 2017-12-18
Add standard stream accessors to Forward future

YetAnotherMinion:
I saw from https://github.com/alexcrichton/futures-rs/issues/572 that adding accessors would be welcome in a PR. I need to access the inside of the Forward future so that I can use a tokio::net::TcpStream as the sink and call shutdown on it if the Forward future returns Err from poll().

I only need `sink_mut`, but for the sake of completeness I added the shared reference versions as well.

I did not have a clear vision for implementing into_inner because forward has an internal buffer. The business value for me is only in the sink_mut method, so I wanted to maximize the chances of a quick upstreaming by leaving out anything contentious from this PR.

### DIFF
--- a/src/stream/forward.rs
+++ b/src/stream/forward.rs
@@ -30,14 +30,28 @@ impl<T, U> Forward<T, U>
           T: Stream,
           T::Error: From<U::SinkError>,
 {
-    fn sink_mut(&mut self) -> &mut U {
-        self.sink.as_mut().take()
-            .expect("Attempted to poll Forward after completion")
+    /// Get a shared reference to the inner sink.
+    /// If this combintator has already been polled to completetion, None will be returned.
+    pub fn sink_ref(&self) -> Option<&U> {
+        self.sink.as_ref()
     }
 
-    fn stream_mut(&mut self) -> &mut Fuse<T> {
-        self.stream.as_mut().take()
-            .expect("Attempted to poll Forward after completion")
+    /// Get a mutable reference to the inner sink.
+    /// If this combintator has already been polled to completetion, None will be returned.
+    pub fn sink_mut(&mut self) -> Option<&mut U> {
+        self.sink.as_mut()
+    }
+
+    /// Get a shared reference to the inner stream.
+    /// If this combintator has already been polled to completetion, None will be returned.
+    pub fn stream_ref(&self) -> Option<&T> {
+        self.stream.as_ref().map(|x| x.get_ref())
+    }
+
+    /// Get a mutable reference to the inner stream.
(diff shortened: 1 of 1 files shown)


alexcrichton:
Thanks for the PR! This is an unfortunate case though of it's not clear to me how we'll want to do this. Most of the existing accessor methods are infallible and don't deal with sink/stream distinctions but rather just one item. I think the `sink_ref` and `stream_ref` terminology here is a good idea though for accessing multiple inner items!

I'd avoid exposing the fact that `Fuse` is used for the stream though and instead just return references.

For the panicking I'd personally prefer to return an `Option` in this case and defer the unwrap to happen at a different layer. That'll also serve as a heads up that the return value can be `None` and we can document why `None` is being returned as well perhaps?

YetAnotherMinion:
I updated PR with requested changes and squashed it.

alexcrichton:
Thanks!

YetAnotherMinion:
Hey, I made some spelling errors in the documentation of the methods,
```diff
diff --git a/src/stream/forward.rs b/src/stream/forward.rs
index 718f694..2ecde10 100644
--- a/src/stream/forward.rs
+++ b/src/stream/forward.rs
@@ -31,25 +31,25 @@ impl<T, U> Forward<T, U>
           T::Error: From<U::SinkError>,
 {
     /// Get a shared reference to the inner sink.
-    /// If this combintator has already been polled to completetion, None will be returned.
+    /// If this combinator has already been polled to completion, None will be returned.
     pub fn sink_ref(&self) -> Option<&U> {
         self.sink.as_ref()
     }
 
     /// Get a mutable reference to the inner sink.
-    /// If this combintator has already been polled to completetion, None will be returned.
+    /// If this combinator has already been polled to completion, None will be returned.
     pub fn sink_mut(&mut self) -> Option<&mut U> {
         self.sink.as_mut()
     }
 
     /// Get a shared reference to the inner stream.
-    /// If this combintator has already been polled to completetion, None will be returned.
+    /// If this combinator has already been polled to completion, None will be returned.
     pub fn stream_ref(&self) -> Option<&T> {
         self.stream.as_ref().map(|x| x.get_ref())
     }
 
     /// Get a mutable reference to the inner stream.
-    /// If this combintator has already been polled to completetion, None will be returned.
+    /// If this combinator has already been polled to completion, None will be returned.
     pub fn stream_mut(&mut self) -> Option<&mut T> {
         self.stream.as_mut().map(|x| x.get_mut())
     }
```

alexcrichton:
Oh no worries! @YetAnotherMinion want to send a PR?

PR #686, opened 2017-12-30
Implement `Error` and `Display` for `SharedError` when possible

lo48576:
`impl Error for SharedError<E>` enables us to use some useful
functionalities such as
`impl<E: Error+Send+Sync> From<E> for Box<Error+Send+Sync>`.

`impl Display for SharedError<E>` is required by `Error` trait.


alexcrichton:
Thanks!

Issue #670, opened 2017-12-06
No way to represent a set of tasks


canndrew:
I want to implement a future which can be cloned so that multiple tasks can be blocked waiting for the same external event. This doesn't seem to be possible to implement. `Task` doesn't implement `Hash` or `Eq` so I can't keep the set of pending tasks in a `HashSet` or even a `Vec` because there's no way to de-duplicate the tasks. If someone keeps `poll()`ing my future, or keeps cloning new ones, the `Vec` would grow indefinitely.

Could we give `Task` an implementation of `Hash` or `Ord + Eq` or something? Alternatively, could we add a `TaskSet` type for this use-case?

canndrew:
No, but looking at how it's implemented did. So thanks for that.

I might leave this open though since any of my suggestions would have made this easier to solve.

carllerche:
I’m not sure how providing a task set would help. I can’t really comment more without knowing what api you are trying to provide, but my initial impression is that using something like a TaskSet would result in a buggy impl.

canndrew:
@carllerche It might result in some spurious wakeups if an object that needs to be notified keeps moving between different tasks. I think it would still be more efficient (in the general case) than the allocation and duplication involved in keeping tasks in a hash map and giving every object it's own key (aka what `Future::shared` does). And I don't see how it could actually be buggy.

carllerche:
@candrew what happens if there are two clones being used on the same task and one drops? Would it not deregister all handles on that same task?

alexcrichton:
@canndrew for the original question (something like hash/eq/ord for `Task`) I don't think we'll be able to do that as we don't really have an underlying id and it's often quite difficult to equate two handles (as they've got trait objects and such behind the hood). In that sense I'd recommend tracking ids manually in what you're working on and use that for deduplication of task handles. Would that work for you?

carllerche:
I'm going to close this as I don't think that there is anything to do. Please let me know if I am wrong.

AljoschaMeyer:
Came here for exactly the same question (hash, ord, or at least eq for tasks). Literally the first thing I saw after clicking on the "src" button for the `Task` struct was `pub struct Task {id: usize, ...}`. Which does not look like "[...] we don't really have an underlying id" 😸.

[...]

PR #689, opened 2018-01-04
Explain why `Task` does not impl `Eq + Hash`.

carllerche:
Attempt to clarify the problem with using a set structure to store tasks.

This question has come up a bunch; e.g. #670.

### DIFF
--- a/src/task_impl/mod.rs
+++ b/src/task_impl/mod.rs
@@ -54,6 +54,31 @@ fn with<F: FnOnce(&BorrowedTask) -> R, R>(f: F) -> R {
 /// the future as notifications arrive, until the future terminates.
 ///
 /// This is obtained by the `task::current` function.
+///
+/// # FAQ
+///
+/// ### Why does `Task` not implement `Eq` and `Hash`?
+///
+/// A valid use case for `Task` to implement these two traits has not been
+/// encountered.
+///
+/// Usually, this question is asked by someone who wants to store a `Task`
+/// instance in a `HashSet`. This seems like an obvious way to implement a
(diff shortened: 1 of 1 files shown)


alexcrichton:
Hm the API docs seem like it's not quite the right location for something like this (an FAQ), maybe something like the website and/or somewhere on this repo?

carllerche:
I mean, the API docs is where people are going to come by here and see it. If you want to move the snippet to a different location, feel free to do so. I just see this question coming up repeatedly so I am trying to address the confusion.

seanmonstar:
Instead of putting this in the documentation, I'd just leave this as comments in the code, so anyone showing to try adding it can see why. (To clarify, as `//` instead of `///` comments.)

alexcrichton:
sounds reasonable to me!

burdges:
I can't comment specifically on where to draw the line in the current situation, but I like it when API docs give a bit more insight into the design and correct usage, like say https://docs.rs/curve25519-dalek/0.14.0/curve25519_dalek/ristretto/index.html

ahmedcharles:
This is effectively a documenting rationale. There's precedence for having documentation with rationale (all Boost libraries are required to document rationale for important design decisions... there's even a "rationale rationale"): http://www.boost.org/development/requirements.html#Rationale

I.e. add it and your users will thank you later (though, change FAQ to Rationale).

aturon:
@carllerche want to update this to attach it to `Waker`?

carllerche:
@aturon Update it how? afaik, the PR was not accepted?

carllerche:
Continued here: #1236

Issue #612, opened 2017-10-14
RFC - Rename types to avoid confusion


yazaddaruvala:
Edit: After some discussion, I've better formulated my point. It may be better to start reading at [this comment](https://github.com/rust-lang-nursery/futures-rs/issues/612#issuecomment-363212079).

Original:
Not understanding you need to execute a Future is such a common confusion. This might stem from the fact that `Future`s in other languages are promises that a value will show up later. i.e. as soon as the Future type is created, you expect a value to show up later. Maybe there should be a bit of renaming, to avoid this confusion?

Current -> Proposed
`Future` -> `TaskBuilder`
`Reactor` -> `TaskRunner`
`Task` -> `Task`
`Future` -> `TaskHandle`

A TaskBuilder is a collection of closures/structs which create the state machine for a Task. A TaskRunner creates a Task to execute a TaskBuilder (and its entire state machine inside the TaskRunner's thread). However, Tasks are an internal construct of the runner, so the TaskRunner should output a TaskHandle. TaskHandles are poll-able, and can chain closures, exactly like today's Futures.

Compared to today, only the names would be different (I'm open to other names as well), but it seems to me there should be a lot less confusion around the whole async task system.

Future could then be written anew as an ergonomic sibling for TaskHandle, i.e. a Condvar around the result of the Task so that Futures are similar to every other language which has them.

This would mean that the async task system is identical to today, for people who need the performance and flexibility but Futures would be more ergonomic for most common usecases.

It is likely too late in the story to change these names, but thoughts?

alexcrichton:
Thanks for the report! Have you seen tokio-rs/tokio-rfcs#3? Would that solve the problems you're worried about here?

yazaddaruvala:
Sorry, for the slow response. While that RFC definitely helps the story as a whole, I don't think it relates to my point.

Very succinctly, the implementation of "Futures" in this crate and as such Rust, is very unique. Should we make the vocabulary unique as well? (examples in my original post)

To elaborate a bit more, people from other languages are familiar with the way a "Future" works. While it is not an ideal implementation, a "Future" takes a callback and runs it once a value exists. This crate challenging that familiarity is causing unneeded friction. Additionally, in no other language does the Futures concept also have this "task api" concept, and especially given the "task api" is hidden (for good reason), it ends up being very jarring to realize that it exists and is arguably  understanding it is crucial to the use of Futures, which should have been a familiar concept.

[...]

PR #691, opened 2018-01-05
Add new function task::in_task

brain0:
This function can be used to determine if a function is run within the context
of a task.

### DIFF
--- a/src/task_impl/mod.rs
+++ b/src/task_impl/mod.rs
@@ -38,12 +38,24 @@ fn fresh_task_id() -> usize {
     id
 }
 
+fn try_with<F: FnOnce(Option<&BorrowedTask>) -> R, R>(f: F) -> R {
+    let task = unsafe {
+        match get_ptr() {
+            Some(ptr) => {
+                if ptr.is_null() {
+                    None
+                } else {
+                    Some(&*(ptr as *const BorrowedTask))
+                }
+            },
+            None => None
+        }
+    };
+    f(task)
+}
+
 fn with<F: FnOnce(&BorrowedTask) -> R, R>(f: F) -> R {
-    unsafe {
-        let task = get_ptr().expect("no Task is currently running");
-        assert!(!task.is_null(), "no Task is currently running");
-        f(&*(task as *const BorrowedTask))
-    }
+    try_with(|task| f(task.expect("no Task is currently running")))
 }
 
 /// A handle to a "task", which represents a single lightweight "thread" of
@@ -103,6 +115,13 @@ pub fn current() -> Task {
 #[deprecated(note = "renamed to `current`")]
 pub fn park() -> Task {
     current()
+}
+
+/// Determines if a task is currently being executed.
+///
+/// If this function returns false, `current` and `Task::will_notify_current` will panic.
+pub fn in_task() -> bool {
+    try_with(|task| task.is_some())
 }
 
 impl Task {
--- a/src/task.rs
+++ b/src/task.rs
@@ -33,7 +33,7 @@
 #[allow(deprecated)]
 pub use task_impl::{Spawn, spawn, Unpark, Executor, Run, park};
 
-pub use task_impl::{Task, AtomicTask, current, init};
+pub use task_impl::{Task, AtomicTask, current, in_task, init};
 
 #[allow(deprecated)]
 #[cfg(feature = "use_std")]


src/task_impl/mod.rs
```diff
@@ -105,6 +105,17 @@ pub fn park() -> Task {
     current()
 }
 
+/// Determines if a task is currently being executed.
+///
+/// If this function returns false, `current` and `Task::will_notify_current` will panic.
+pub fn has_current() -> bool {
```
Thomasdezeeuw: Has current doesn't really mean much to me. Maybe `is_task_running`, or `in_task` is a better name.

brain0: I tend to agree, but my use case is this:

I am using tokio-core to interface an external library that needs to tie into a main loop. Naturally, that library has callbacks into rust code. As long as the main loop runs, these calls are all made from within tasks and there is no problem.

However, once the mainloop shuts down and I call the cleanup methods in the external library, they are not called from within a task. The external library makes callback to rust code during cleanup (unregister events, ...). Since I have no way of knowing that the main loop has stopped, those methods will likely panic.

brain0:
I tend to agree, but my use case is this:

I am using tokio-core to interface an external library that needs to tie into a main loop. Naturally, that library has callbacks into rust code. As long as the main loop runs, these calls are all made from within tasks and there is no problem.

However, once the mainloop shuts down and I call the cleanup methods in the external library, they are not called from within a task. The external library makes callback to rust code during cleanup (unregister events, ...). Since I have no way of knowing that the main loop has stopped, those methods will likely panic.

alexcrichton:
Thanks for the PR!

This has been somewhat contentious in the past though which has stopped us from adding it historically. Would you mind providing a few more details about your use case? For example what work is being done in destructors that ends up using the task system?

brain0:
Okay, so let's be a bit less generic. I am trying to build a rust abstraction for the asynchronous pulseaudio API. To do this, libpulse requires a [mainloop abstraction](https://freedesktop.org/software/pulseaudio/doxygen/async.html#mainloop_sec). An implementation of a mainloop abstraction needs to provide [this structure](https://freedesktop.org/software/pulseaudio/doxygen/structpa__mainloop__api.html).

I want to use tokio-core, futures and mio because that allows combining existing asynchronous rust code with pulseaudio. Using pulseaudio's own mainloop implementation would not enable me to safely and conveniently get data from other rust code into pulseaudio. I also do not want to run pulseaudio in a separate thread, but asynchronously in an existing mainloop.

Using the mainloop API, libpulse registers and unregisters events. It also registers cleanup functions for its events, which the mainloop calls when an event is freed. Now things become tricky. As far as I can see, an event handler might free its own event. For this to be safe, the mainloop only marks the event as dead and cleans it up later, asynchronously. (At least that is how pulseaudio's own implementation behaves, therefore it would be unwise to behave differently here.)

Here comes the problem: Whenever I exit the mainloop, and exit the main function, rust begins cleaning up my objects. Eventually, it calls pa_context_unref() and libpulse starts cleaning up the context. The context then calls into the mainloop abstraction to clean up its event registrations. As seen in the last paragraph, I perform some asynchronous work now: start spawning futures into the mainloop, potentially waking up some tasks as well (for an IO event for example, I wake up its stream so it sees the event is no longer needed and returns Async::Ready(None)). Now, this will all fail (and in some cases panic), since the mainloop has already exited. Thus, I need to do cleanup work synchronously, including calling the registered cleanup functions.

I hope this writeup was not too confusing. I just [uploaded my code to github](https://github.com/brain0/pulseaudio-rs). Note that this is all proof-of-concept for now and that it does nothing useful. If you have a Linux system with pulseaudio running, cargo run should output something like this:

    New state: Connecting
    New state: Authorizing
    New state: SettingName
    New state: Ready
    <Pause for 5 seconds>
    Disconnecting
    New state: Terminated
    Exiting with success!

To observe the cleanup behaviour I described above, you need to exit without a clean disconnect (try the unclean_exit branch). This will cause the code to enter the code paths where task::in_task() returns false.

I hope this helps understand the motivation for this PR.

alexcrichton:
Hm fascinating, thanks for the detailed background info @brain0! So I think you *could* perhaps structure this differently to not require an "is a current task api" but I'm not sure whether that's the best solution here. What I'm thinking is that you'd have something like a weak pointer to the "main context" which while the event loop is running you can upgrade and use to defer destruction. Once the event loop has exited though you chould ensure that the strong pointer goes away before event loop is destroyed, and then a failure to upgrade could signal "ok, synchronously destroy this"

We've gotten enough requests over time though that it may be best to just add a method? I'd just want to be sure that this would indeed cover your use case as I know destructors are especially tricky with then begin run at various times and such. For example it looks like in your usage today you're spawning a task to run a destructor, or otherwise running the destructor immediately. What happens, though, if the future doesn't run? (the core exits prematurely or finishes otherwise)

brain0:
That would work, but: The only way to do something "when the main loop has exited" is doing it in the future that drives the core. That would mean that my crate cannot be used with any core existing core, but would require specific adjustments to that main future. I would prefer to not require that (right now, all that I need is a Handle to a Core passed to PaMainLoopApi::new()).

As to your other question: If the future that would drive the destructor never runs, but the core exits before, that future is dropped. However, at some point, the last clone of the PaMainLoopApi struct is dropped (which is probably either owned by some future, or sits on the stack of the function that ran the core). Then, the [drop implementation of its inner struct will perform all the remaining cleanup work synchronously](https://github.com/brain0/pulseaudio-rs/blob/6f8768768728027411efa8819b4f1618a87b58fa/src/tokio_mainloop_api/api_impl/mod.rs#L37).

Also note this: While preparing my previous answer and trying to reproduce all this again, I noticed that if I disconnect the PaContext cleanly before exiting the core, all cleanup happens while the core is still running. These problems only occur if I either exit the core prematurely, or if I unwind due to a panic. However, panicking while unwinding is especially undesirable.

brain0:
If I knew whether the Core is active or not, that would help, yes. This would probably be within the scope of tokio-core, not futures.

yazaddaruvala:
@brain0 I don't know how feasible it is, but instead of an API `in_task()`, I would far prefer if this information could be codified using the type system (I've mentioned something similar [here](https://github.com/rust-lang-nursery/futures-rs/issues/612)).

It would be interesting if the `Context` being discussed in [this RFC](https://github.com/rust-lang-nursery/futures-rfcs/pull/2) is an associated type, then for certain implementations of Future it could be `()`, and you would know there is no task.

`Future<..., ..., task::Context>` vs `Future<..., ..., ()>`

Thomasdezeeuw:
Maybe take a look at https://github.com/rust-lang-nursery/futures-rfcs/pull/2, a current debate is going on to to add the `Task` to the `poll` method of `Future`.

aturon:
I'm going to close this PR out, since with futures 0.2 the task context will be passed in explicitly (rather than via TLS).

PR #692, opened 2018-01-05
Minor update to spawn rustdoc to avoid confusion from usage of "fuse".

echochamber:
When reading the docs, the word "fuse" made me think there was some relation between this function and [futures::future::Future::fuse](https://docs.rs/futures/0.1/futures/future/trait.Future.html#method.fuse).


alexcrichton:
Thanks!

PR #696, opened 2018-01-08
Add killable future combinator

### LINKED ISSUE #693: Consider Including Cancellable Combinator
I find myself wanting this occasionally. [Example here](https://play.rust-lang.org/?gist=d49b79a00aba4f935ff42dfe5ef5c65e&version=stable).

cramertj:
Closes #693

### DIFF
--- a/src/future/killable.rs
+++ b/src/future/killable.rs
@@ -1,0 +1,117 @@
+use {Future, Poll, Async};
+use task::AtomicTask;
+use std::sync::Arc;
+use std::sync::atomic::{AtomicBool, Ordering};
+
+/// A future which can be cancelled using a `KillHandle`.
+#[derive(Debug, Clone)]
+#[must_use = "futures do nothing unless polled"]
+pub struct Killable<T> {
+    future: T,
+    inner: Arc<KillInner>,
+}
+
+impl<T> Killable<T> where T: Future {
+    /// Creates a new `Killable` future using an existing `KillRegistration`.
+    /// `KillRegistration`s can be acquired through `KillHandle::new`.
+    ///
+    /// When `kill` is called on `handle`, this future will complete
+    /// immediately. If `kill` has already been called, the future
+    /// will complete immediately without making any further progress.
+    pub fn new(future: T, reg: KillRegistration) -> Self {
+        Killable {
+            future: future,
+            inner: reg.inner,
+        }
+    }
+}
+
+/// A registration handle for a `Killable` future.
+/// Values of this type can be acquired from `KillHandle::new` and are used
+/// in calls to `Killable::new`.
+#[derive(Debug, Clone)]
+pub struct KillRegistration {
+    inner: Arc<KillInner>,
+}
+
+/// A handle to a `Killable` future.
+#[derive(Debug, Clone)]
+pub struct KillHandle {
+    inner: Arc<KillInner>,
+}
+
+impl KillHandle {
+    /// Creates a new `KillHandle` which can be used to kill a running future.
+    /// This function also returns a `KillRegistration` which is used to
+    /// pair the handle with a `Killable` future.
+    ///
+    /// This function is usually paired with a call to `Killable::new`.
+    pub fn new() -> (Self, KillRegistration) {
+        let inner = Arc::new(KillInner {
+            task: AtomicTask::new(),
+            cancel: AtomicBool::new(false),
+        });
+
+        (
+            KillHandle {
+                inner: inner.clone(),
+            },
+            KillRegistration {
+                inner: inner,
+            },
+        )
+    }
+}
+
+// Inner type storing the task to awaken and a bool indicating that it
+// should be cancelled.
+#[derive(Debug)]
+struct KillInner {
+    task: AtomicTask,
+    cancel: AtomicBool,
+}
+
+/// Creates a new `Killable` future and a `KillHandle` which can be used to stop it.
+///
+/// This function is a convenient (but less flexible) alternative to calling 
+/// `KillHandle::new` and `Killable::new` manually.
+pub fn killable<T>(future: T) -> (Killable<T>, KillHandle)
+    where T: Future
+{
+    let (handle, reg) = KillHandle::new();
+    (
+        Killable::new(future, reg),
+        handle,
+    )
+}
+
+/// Indicator that the `Killable` future was killed.
+#[derive(Copy, Clone, Debug, Eq, PartialEq)]
+pub struct Killed;
+
+impl<T> Future for Killable<T> where T: Future {
+    type Item = Result<T::Item, Killed>;
+    type Error = T::Error;
+    fn poll(&mut self) -> Poll<Self::Item, Self::Error> {
+        if self.inner.cancel.load(Ordering::Acquire) {
+            Ok(Async::Ready(Err(Killed)))
+        } else {
+            match self.future.poll() {
+                Ok(Async::NotReady) => {
+                    self.inner.task.register();
+                    Ok(Async::NotReady)
+                },
+                Ok(Async::Ready(x)) => Ok(Async::Ready(Ok(x))),
+                Err(e) => Err(e),
+            }
+        }
+    }
+}
+
+impl KillHandle {
+    /// Kill the `Killable` future associated with this handle.
+    pub fn kill(&self) {
+        self.inner.cancel.store(true, Ordering::Release);
+        self.inner.task.notify();
+    }
+}
--- a/src/future/mod.rs
+++ b/src/future/mod.rs
@@ -80,11 +80,13 @@ pub use self::inspect::Inspect;
 if_std! {
     mod catch_unwind;
     mod join_all;
+    mod killable;
     mod select_all;
     mod select_ok;
     mod shared;
     pub use self::catch_unwind::CatchUnwind;
     pub use self::join_all::{join_all, JoinAll};
+    pub use self::killable::{killable, Killable, Killed, KillHandle, KillRegistration};
     pub use self::select_all::{SelectAll, SelectAllNext, select_all};
     pub use self::select_ok::{SelectOk, select_ok};
     pub use self::shared::{Shared, SharedItem, SharedError};
--- a/tests/killable.rs
+++ b/tests/killable.rs
@@ -1,0 +1,41 @@
+extern crate futures;
+
+use futures::future::{killable, Killed};
+use futures::sync::oneshot;
+
+use std::thread;
+use std::time::Duration;
+
+#[test]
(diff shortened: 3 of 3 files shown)


cramertj:
If you think it's worth including a non-sync version as well, let me know and I'll add that. IMO the tiny perf difference is probably not worth the additional complexity.

alexcrichton:
This seems pretty slick to me! This form of cancellation feels like it's common enough in other libraries that it seems worthwhile to include here as well. I do think though yeah if we have a `sync` version it'd be wortwhile to have an `unsync` version too. Before that though I'd be curious to hear others' thoughts on a combinator like this, @aturon perhaps?

carllerche:
At a high level.

* I'm +1 on the feature.
* I don't love the name... maybe `ShortCircuit` or `Abort`. Generally, it seems that '*able' is not a common naming idiom and "kill" sounds violent!
* Why does it require the inner future to be of `()`? Also, the fact that the inner future was aborted is not exposed.
* I'm not a fan of the proliferation of of unsync duplication, but that ship has sailed, so ¯\_(ツ)_/¯.

Going in more detail, I'm not sure about either the implementation or the API. The implication of `Killable` taking `&KillHandle` is that the same `KillHandle` can be used to back multiple `Killable` instances. However, the implementation does not support this.

I would say either the implementation should change or the API should not allow constructing a `Killhandle` standalone.

cramertj:
@carllerche I need to be able to construct a `KillHandle` standalone so that I can cancel an outer future from deep inside-- think a request handler wanting to kill a service. If you only provide a `KillHandle` after constructing the top-level `Killable`, there's no way to do this.

I do agree, though, that it's misleading and user might think they could use one `KillHandle` for multiple futures. Perhaps make a constructor that returns both a `KillHandle` and a `KillRegistration`, and the `KillRegistration` must be moved into `Killable::new`?

cramertj:
@carllerche 
> Why does it require the inner future to be of ()? Also, the fact that the inner future was aborted is not exposed.

How would you support other types? The cancellation path only knows how to return `()`. I guess you could also allow the user to supply a default if-cancelled value, but then the API quickly gets more complex. Most of the use-cases I have for this are things that are being `spawn`ed or are otherwise top-level side-effect-y futures.

carllerche:
> How would you support other types?

An enum.

> Most of the use-cases I have for this are things that are being spawned or are otherwise top-level side-effect-y futures.

Should this API be tied to the executor then? It could be an additional fn on the `Executor` trait that returns the cancel handle.

I don't have strong opinions either way. I'm mostly just brainstorming here.

cramertj:
> An enum.

`Cancellable<T>` impls `Future<Item=Option<T>>`? Seems fine to me. Just means I'll be adding one more `.map(|_| ())`.

carllerche:
To be clear, I don't necessarily think my idea to support more than `Item=()` inner futures is a good idea, but I am proposing it for discussion to see what others think of the pros / cons trade off.

cramertj:
Yeah it seems like a reasonable option.

tailhook:
Another question is whether future has to be canceled when KillHandle/CancelHandle is dropped.

This is useful if you keep a handle in some structure just to cancel the future when not needed anymore. And that structure may occasionally drop unexpectedly (e.g. on panic). Currently, I use often use `oneshot` with `select` and it's handy that it cancels future on dropping `oneshot::Sender` too. So this use case is quite real.

cramertj:
@tailhook I'd personally be opposed to cancelling on on drop of the `CancelHandle`-- it seems like surprising behavior to me and I'd have to be careful to avoid it.

tailhook:
> @tailhook I'd personally be opposed to cancelling on on drop of the CancelHandle-- it seems like surprising behavior to me and I'd have to be careful to avoid it.

Well, you have to be careful both ways. You're likely to have a memory leak if CancelHandle is dropped without cancelation. Can you describe your real life use case?

cramertj:
> You're likely to have a memory leak if CancelHandle is dropped without cancelation.

Nah-- I just use this to kill futures early or to kill a future from another future deep inside without plumbing through some kind of shared flag. In the first case, the future would still be expected to complete normally, and in the second case, it doesn't make any sense why an inner future dropping should kill the outer one.

alexcrichton:
@cramertj I wonder if all this is evidence that we may want to let this bake externally first? That way we could iterate on it elsewhere and bring it in once it's stabilized?

cramertj:
@alexcrichton I'd be happy to throw it up in its own repo somewhere, but I'm not sure we'll get more useful feedback from having it there than we would by continuing to discuss here (or on IRC or something).

carllerche:
IMO we should fix the "out of crate" discover ability problem.

I would suggest taking the following two steps:

* Add a github wiki page that includes an organized list of (somewhat curated) third party utilities.
* Add a link to the github page in the futures-rs documentation explaining that there are additional utilities provided in third party crates.

src/future/killable.rs
```diff
@@ -0,0 +1,94 @@
...
+    inner: Arc<KillInner>,
+}
+
+impl<T> Killable<T> where T: Future<Item = ()> {
+	/// Creates a new `Killable` future using an existing `KillHandle`.
+	///
+	/// When `kill` is called on `handle`, this future will complete
+	/// immediately. If `kill` has already been called, the future
+	/// will complete immediately without making any further progress.
+    pub fn new(future: T, handle: &KillHandle) -> Self {
```
rphmeier: docs here are tabs

src/future/killable.rs
```diff
@@ -0,0 +1,94 @@
...
+#[derive(Debug, Clone)]
+pub struct KillHandle {
+    inner: Arc<KillInner>,
+}
+
+impl KillHandle {
+	/// Creates a new `KillHandle` which can be used to kill a running future.
+	///
+	/// This function is usually paired with one or more calls to `Killable::new`.
+    pub fn new() -> Self {
```
rphmeier: and here

cramertj:
Updated in response to comments.

src/future/killable.rs
```diff
@@ -0,0 +1,112 @@
...
+{
+    let (handle, reg) = KillHandle::new();
+    (
+        Killable::new(future, reg),
+        handle,
+    )
+}
+
+impl<T> Future for Killable<T> where T: Future<Item = ()> {
+    type Item = ();
```
rphmeier: `Item` could be `Option<T::Item>` where when canceled returns `None`. A separate type to convey more semantic meaning makes sense to me as well, but would be missing basically all the useful stuff from `Option`.

cramertj: All of my use cases for this have been with futures that were side-effect only, usually being given to a `spawn` or similar. However, it seems that both you and @carllerche have other use cases in mind, so I'll change to use `Result<T::Item, Killed>`.

rphmeier: I can't think of something in particular but there isn't much reason not to provide the generality

cramertj: I suppose at that point we can also provide a `Stream` impl.

alexcrichton:
@cramertj would you still be looking to merge in this repo? Or would you be ok having it out of repo for a bit?

cramertj:
@alexcrichton Up to you-- my preference would be to try and get in good shape to be merged into `future`s since it seems like a fairly common thing to want to do. If you'd prefer to have it available through a separate crate, I'll need someone else to make the repo (Google reasons). @aturon previously suggested starting a `tokio-tools` crate as a proving ground for combinators-- that could be a place for this as well.

carllerche:
@cramertj well, it seems like discoverability is a general problem. Do you not think that my suggestion above would aid in that regards?

cramertj:
@carllerche I disagree that discoverability is the main problem here. I think the main problem is that we need a place to incubate standard, well-motivated combinators without tying them to futures/tokio stability guarantees. I think a better solution to this would be to have a dedicated futures/tokio combinator library that can frequently release breaking changes. Breaking releases to combinators won't result in the interoperability problems that frequent breaking changes to the `futures` library will because they won't break the interface exposed by libraries and applications or cause interoperability problems. Libraries/applications could freely use multiple versions of the combinator libraries.

FWIW, I think this applies as well to methods like `map`, `and_then`, etc. on the actual `Future` and `Stream` types. Over time there've been a number of problems discovered with various combinator methods that would require breaking changes. If these combinators had been implemented via an extension trait in a separate library, it would be easy to make the appropriate fixes. However, the fact that they're in the main `futures` repo and even the main `Future` trait means that we can't make changes without breaking the fundamental interface of the library.

alexcrichton:
@cramertj I think you hit the nail on the head there with the problem of tying combinators to the stability of the `Future` trait itself (and how breakage in the combinators is totally ok, it's just `Future` which can't break.

This sounds a lot like `itertools`, though? It definitely seems to me like we should start a `futuretools` (to make clear it's not Tokio-specific) crate and probably consider moving a number of standard combinators to the crate if we ever do a major release of `futures` itself.

cramertj:
@alexcrichton Yup, +1 for `futuretools` or similar.

carllerche:
Yes, I'm definitely for creating another crate and moving stuff there. IMO the core traits already have too many fns on them, which causes problems for iteration.

I didn't know moving stuff to another crate was on the table, so this makes me happy.

alexcrichton:
@cramertj would you be willing to maintain such a crate? I'm not sure we have a great place for it right now...

cramertj:
Note for folks reading: [futures-tools](https://github.com/alexcrichton/futures-tools) has been created.

carllerche:
Is there a reason the crate isn't included in this git repository? The cargo workspace feature would make it easy.

It seems that if the plan is to migrate code between the two repos (move a bunch of stuff from here into -tools in futures 0.2 then incrementally "promote" as APIs stabilize) it would be easiest to do so in one repository.

cramertj:
@carllerche That approach sounds good to me.

carllerche:
Perhaps discussion for a futures-util crate should be promoted to a dedicated issue at this point :)

cramertj:
@carllerche @alexcrichton @aturon I've opened https://github.com/alexcrichton/futures-rs/issues/711 for discussion.

alexcrichton:
@cramertj want to retarget at 0.2 to merge?

aturon:
@cramertj what's the status here?

tikue:
Is `AtomicTask` used to make it safe to move the future to different tasks? What about the case where I'm just doing `tokio::spawn(killable_future)`? It can't change tasks then, right? So

Keruspe:
Any chance to get this in for 0.3?

cramertj:
Closing, opened #1079 targeting 0.3.

Issue #698, opened 2018-01-09
Difficulty translating existing benchmarking to using `CurrentThread`


alexcrichton:
In tokio-core there's some existing benchmarking code that looks like:

```rust
let reactor = ...;
let handle = reactor.handle();

handle.spawn(/* some TCP listener task */);

bh.iter(|| {
    core.run(some_local_task);
});
```

The idea here is that there's a daemon/background task which is being used to process some data (like accepting TCP sockets). This persistent daemon task, thougho, is paired up with synchronously running futures in lockstep (specifically benchmarking the timing of each iteration in this case).

Unfortunately though this represents a pattern which isn't translateable to `CurrentThread`. Notably you can't build a "context" which is persistently used amongst several blocking scopes. For example every time a `CurrentThread` scope exits and reenters there's no data that's preserved across the boundaries (like the necessary daemon task in this case).

Is this something we want to accomodate? Say "don't do this"? I'm not sure how often this would come up in other contexts, although I suspect it will become quite relevant for client-side integration of futures which blocks much more maybe?

carllerche:
Setting up something like this has been made more difficult by design. By spawning daemon tasks and then intermittently calling `core.run` while mixing in other potentially blocking logic is a hazard. When the core is not running, the daemon task cannot make progress.

One option today would be to spawn the listener task on a thread pool.

One reason I added `&mut Context` as an argument to provide the ability to re-add similar functionality at a later time. The `&mut Context` can have a `yield_to_executor(fut)` function that allows driving a single future to completion similar to the setup you have above.

I did not yet propose such a function as it carries the same hazards mentioned above.

carllerche:
I can add a `block_on(fut)` to `Context` if we like that idea.

I also don't have a great idea for a name, but `block_on` seems ok.

aturon:
@carllerche IMO, let's not add anything right now, and instead wait for feedback.

@alexcrichton, are you still planning to add unstable features? That can be another route for experimentation with new APIs like this.

alexcrichton:
@aturon I think it's fine for us to add that on an as-needed basis (although I'd like to originally populate it with `Sleep` if we plan on keeping it for the first release). 

For `block_on` though I'd personally like to see that in the first release rather than behind an unstable feature to start out with.

aturon:
Closing in favor of the [executor RFC](https://github.com/rust-lang-nursery/futures-rfcs/pull/3).

Issue #697, opened 2018-01-08
Unclear how to use CurrentThread and a "main future"


alexcrichton:
A typical pattern I've seen with `tokio-core` is to do something like:

```rust
let srv = ...;

handle.spawn(another_future);
handle.spawn(another_future2);
// ...

let result = core.run(srv);

// work with `result`
```

The `Handle::spawn` functionality is getting folded into `CurrentThread` but unfortunately I'm not sure how to recover the `let result = core.run(srv)` functionality from `CurrentThread`. Currently the closures in `CurrentThread` both return a bare `R`, the value returned by the closure, and the `future::blocking` module doesn't allow usage of `CurrentThread`. As a result, I don't think it's currently easily possible to construct a server like this, where you've got one "main future" you're interested in and it may spawn (or already have spawned) some sub-futures it was interested in.

The current state of https://github.com/tokio-rs/tokio-rfcs/pull/3 solves this use case with:

```rust
pub fn block_on_all<F: Future>(f: F)  -> Result<F::Item, F::Error>;
```

and the RFC's current text could be tweaked a bit to accommodate an initialization closure as well:

```rust
pub fn block_with_init<F, R>(f: F) -> Result<R::Item, R::Error> 
    where F: FnOnce(&Enter) -> R,
          R: IntoFuture;
```

carllerche:
Similar to #698.

My impression based on confusion that has been encountered is that your initial example is an anti-pattern. The primary hazard is that `another_future` and `another_future2` are spawned onto the executor, then the executor is run until `srv` completes, yielding the value back. At this point, the executor does not run anymore and `another_future` and `another_future2` will be stalled.

`block_on_all` as it was proposed is an attempt to solve this. As described in the RFC, the supplied future is driven to completion, then the `block_with_init` function blocks until all spawned futures complete (i.e. to quiescence). The issue being that, until these futures complete, you cannot process the result. This seems like a hazard to me.

So, the way I see it, there are two scenarios:

1) You have a single root future that you want to drive to completion and get the result immediately once it is ready.
2) You have a set of futures you want to manage concurrently on a single thread.

In the first case, the answer is to use `blocking::future`. In the second, the answer is to spawn a bunch of futures onto `CurrentThread`, all of which are peers, and wait for them all to complete.

So, in your case, you would process the result in an `and_then` block and be done with it.

[...]

PR #699, opened 2018-01-11
Move `executor::CurrentThread` to `current_thread`

alexcrichton:
This commit moves the `CurrentThread` type to a top-level `current_thread`
module. Along the way a number of changes were made as well:

* Functions with "execute" in the name are tweaked to have "spawn" instead
  (e.g. `execute_daemon` becomes `spawn_daemon`).
* The `CurrentThread` type was renamed to `TaskExecutor` and
  `CurrentThread::current` was renamed to `task_executor`.
* The `run` method was renamed to `block_with_init`.

This commit is currently following the [proposed design][rfc] and will make room
soon for a `block_on_all` method as well. For now though it's primarily just
restruturing without any other deeper changes.

[rfc]: https://github.com/aturon/tokio-rfcs/blob/tokio-reform/tokio-reform.md#changes-to-the-futures-crate

### DIFF
--- a/src/current_thread.rs
+++ b/src/current_thread.rs
@@ -1,25 +1,24 @@
 //! Execute tasks on the current thread
 //!
-//! [`CurrentThread`] provides an executor that keeps futures on the same thread
+//! This module implements an executor that keeps futures on the same thread
 //! that they are submitted on. This allows it to execute futures that are
-//! `!Send`. For more details on general executor concepts, like executing
+//! not `Send`. For more details on general executor concepts, like executing
 //! futures, see [here].
 //!
-//! Before being able to execute futures onto [`CurrentThread`], an executor
-//! context must be setup. This is done by calling [`run`].  From within that
-//! context, [`CurrentThread::execute`] may be called with the future to run in
-//! the background.
+//! Before being able to spawn futures with this module, an executor
+//! context must be setup by calling [`run`]. From within that context [`spawn`]
+//! or [`spawn_daemon`] may be called with the future to run in the background.
 //!
 //! ```
-//! # use futures::executor::current_thread::*;
+//! # use futures::current_thread::*;
 //! use futures::future::lazy;
 //!
 //! // Calling execute here results in a panic
-//! // CurrentThread::execute(my_future);
+//! // spawn(my_future);
 //!
-//! CurrentThread::run(|_| {
+//! run(|_| {
 //!     // The execution context is setup, futures may be executed.
-//!     CurrentThread::execute(lazy(|| {
+//!     spawn(lazy(|| {
 //!         println!("called from the current thread executor");
 //!         Ok(())
 //!     }));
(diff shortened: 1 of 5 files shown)


carllerche:
For context, the RFC originally introduced the `block_with_*` naming scheme as the current thread executor was intended to be a general means to block the thread, drive a future to completion, and get the result, i.e. a replacement for `Future::wait`. Given this original intent, there was a hazard that users call this function while in the context of an existing executor, potentially blocking the thread.

However, the goal of the current thread executor has since diverged (though #697 is still open). Given this, I am no longer sure what the value of calling the function that runs the executor `block_with_init`. It seems to me a more appropriate terminology would be `run` or some variant.

Would you mind explaining the rational for the naming choice?

aturon:
@carllerche Reading #697, I think there's a bit of a miscommunication going on. Let's see if we can settle on something that meets all the constraints.

I think that what's missing from `CurrentThread` today is a means of easily getting a value out of a "main" future *while blocking on everything it spawns* (unlike `Core::run`), which is a fairly common thing to want to do. You need to use `CurrentThread` to spawn thread-locally, but then in order to extract the final value, you have to do something like set up a one shot and use `blocking` *after* `run` returns.

OTOH, @carllerche I believe part of your concern is that if we provide an API that singles out a particular future, it will be unclear whether you end up blocking on *just* that future.

This was originally addressed through naming, and I'd like to reconsider that route. What I'd personally like to see is for today's `run` (and the `block_with_init` in this PR) to instead be:

```rust
fn block_on_all<F, Fut>(f: F) -> Result<Fut::Item, Fut::Error>> where
    F: FnOnce(&mut Context) -> Fut,
    Fut: IntoFuture
```

I believe this has the following properties:

- Makes clear that blocking is occurring (you end up writing `current_thread::block_on_all`), and further makes clear that you're blocking on *all* tasks, not just the given future.
- Allows for initialization.
- Allows for using non-`'static` futures with `current_thread`.
- Makes it easy to extract a value from a "main" future when using the executor.

wdyt?

aturon:
On a separate note: @alexcrichton I noticed that you are moving `current_thread` out of `executor`, was curious why?

carllerche:
@aturon 

> which is a fairly common thing to want to do

This is probably the crux of the issue. My current feel for the matter is that it is an anti-pattern. Specifically, spawning peer futures, blocking for *all* futures to complete, and returning the result.

So, there are two options:

* This is a majority case pattern.
* This is an edge case pattern.

You are suggesting that this is a majority case pattern, and as such the API supporting it should be made the primary API. My feel is that it is an edge case pattern.

Specifically, I think that situations that came up in the past are better served by other options.

So, could you perhaps give me an example (as concrete as possible) that you think is best served w/ `block_on_all`?

carllerche:
Also, I'll try to write up a more in depth reply either tonight or tomorrow.

aturon:
Thanks @carllerche!

I'm curious why you think this is an anti-pattern, given daemon tasks? IOW, what downsides/pitfalls are you envisioning?

I'm not sure whether I think it's a majority case, but it's a pretty typical thing to want to do when you're e.g. setting up a server. There's usually a "main" task taking connections and spawning subtasks, some form of shutdown, and a desire to read out the final result.

Naturally, the shutdown piece here is the key, and does raise questions of how forced shutdown interacts with the "special" future.

Do you have thoughts on how to ergonomically pull out a result, synchronously, after `run`? I don't feel strongly that `block_on_all` needs to be the default, just that we should have an easy way to pull out a main result.

aturon:
FWIW, I'm fine with keeping the current function as-is and adding additional API surface later after things are more clear.

That said, we should bikeshed `run` a bit; can we give a more descriptive name? `block_on_tasks`?

alexcrichton:
I personally share @aturon's thoughts and have a hunch that "extract the value of a future" is a pretty common thing to do. I don't see this sort of desire come up much when writing servers where the current `CurrentThread::run` API is sufficient almost always, but rather this I think comes up a lot when writing clients or otherwise have an asynchronous/synchronous bridge.

For example if you're in a synchronous context and you delegate to an asynchronous one, how do you extract the result? One option is to use `future::blocking` but the problem with that is then you can't spawn futures (can't use today's `CurrentThread`). If we expect libraries to use the ability to spawn futures (and document that they need to) then it seems like the desire to extract a result synchronously from a future would rule out using these libraries?

So to me this boils down two two constraints:

* A synchronous context is invoking an asynchronous library or function and needs the result synchronously. This probably comes up in the context of a client most of the time. 
* The asynchronous functionality being invoked wants to spawn sub-tasks or sub-futures throughout the execution.

Today's `Core::run` does indeed carry the footgun that it doesn't run all the other futures to completion but this is what I believe `block_on_all` (or otherwise returning a future) here will solve. I'm not sure how much https://github.com/alexcrichton/futures-rs/issues/698 will come into play though where a client may want a persistent daemon task across all asynchronous invocations.

src/current_thread.rs
```diff
@@ -1,25 +1,25 @@
...
-//! `!Send`. For more details on general executor concepts, like executing
+//! not `Send`. For more details on general executor concepts, like executing
 //! futures, see [here].
 //!
-//! Before being able to execute futures onto [`CurrentThread`], an executor
-//! context must be setup. This is done by calling [`run`].  From within that
-//! context, [`CurrentThread::execute`] may be called with the future to run in
-//! the background.
+//! Before being able to spawn futures with this module, an executor
+//! context must be setup. This is done by calling either [`block_on_all`] or
```
carllerche: There is no block_on_all fn in the module. This fn is referenced a number of other spots as well.

carllerche:
Re: name of `block_on_all`, my main issue w/ the name is that nothing in the name reflect the main purpose of the function, which is to run the executor. `block_on_all` makes it seem that the function's primary job is to block the thread (probably doing no work) and wait for all tasks to complete "somewhere else".

Re: extracting the future, the case described by @alexcrichton is basically the only situation I can think of where you would want to extract the value of a single future. Unfortunately, no matter how you structure the API there are perils. This is why, the two step process as I described in #698 is my preferred API:

```rust
let res = current_thread::run_executor_blocking_current_thread_until_all_tasks_are_complete(|handle| {
    // In executor context, make sure to not block the thread
    let res = handle.block_on_future_while_running_executor(fut);

    // Do some more work with the result

    res
})
```

In this snippet, the current executor context is setup, and the closure is executed. The context handle can be used to block the thread waiting for future is done, while at the same time driving the executor.

In this case, the `block_on_future_while_running_executor` fn can **return immediately** once the future is complete. The result can be returned from `run_executor_blocking_current_thread_until_all_tasks_are_complete`.

Splitting it up makes it clearer IMO how blocking happens. One function drives the executor to completion until all tasks are complete and the second function is responsible for blocking until the future is complete. Because `context` is passed into the closure w/ `&mut`, it should not be possible to nest calls to `block_on_future_while_running_executor`.

carllerche:
My comment above is all I have for this PR. I want to figure out how to make the current thread executor pluggable to solve #675 and generally provide an escape hatch for cases that the default executor does not handle. That work can happen in another PR and I can take a stab at it once this lands.

alexcrichton:
cc @cramertj curious how you feel about this as well!

Also @aturon how would you feel about this PR with the tweak of renaming `block_with_init` back to `run`, and following up with a method to execute a future on `&mut Context` (the argument to the closure passed to `run`) to completion?

aturon:
@alexcrichton I'm good with that!

cramertj:
@alexcrichton Sounds good to me!

alexcrichton:
@carllerche to confirm are you opposed to the renamings like `executor::CurrentThread => current_thread` or `execute => spawn`? If not then I think this can be merged.

carllerche:
Yes, I'm fine with these changes.

I would also be ok w/ `run_executor` instead of `run`, but either is OK.

carllerche:
Feel free to merge when CI passes.

PR #705, opened 2018-01-15
Add `Flush::into_inner` method

canndrew:
Hello. I need this method.


alexcrichton:
Thanks! Ideally this'd return an `Option` instead of panicking but it looks like we forgot to do that for the other functions! I'll file an issue.

PR #708, opened 2018-01-16
For consideration: round-robin polling for JoinAll and JoinX

rphmeier:
This gives each future in a `Join` or `JoinAll` combinator the same priority for polling over time, preventing starvation of the futures towards the end.

### DIFF
--- a/src/future/join.rs
+++ b/src/future/join.rs
@@ -8,7 +8,7 @@ use {Future, Poll, IntoFuture, Async};
 macro_rules! generate {
     ($(
         $(#[$doc:meta])*
-        ($Join:ident, $new:ident, <A, $($B:ident),*>),
+        ($Join:ident, $count: expr, $new:ident, <A, $($B:ident),*>),
     )*) => ($(
         $(#[$doc])*
         #[must_use = "futures do nothing unless polled"]
@@ -16,6 +16,7 @@ macro_rules! generate {
             where A: Future,
                   $($B: Future<Error=A::Error>),*
         {
+            poll_idx: usize,
             a: MaybeDone<A>,
             $($B: MaybeDone<$B>,)*
         }
@@ -41,6 +42,7 @@ macro_rules! generate {
                   $($B: Future<Error=A::Error>),*
         {
             $Join {
+                poll_idx: 0,
                 a: MaybeDone::NotYet(a),
                 $($B: MaybeDone::NotYet($B)),*
             }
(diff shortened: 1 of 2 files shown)


rphmeier:
I've added another commit for round-robin polling for the `JoinX` combinators, but I am not sure how efficient is is. You end up with a loop that looks like this for `Join5`:

```rust
for i in (self.poll_idx..5).chain(0..self.poll_idx) {
    let mut tracked = 0;
    if i == 0 {
        // poll A
    }
    
    tracked += 1;
    if i == tracked { // poll B1 }

    tracked += 1;
    if i == tracked { // poll B2 }

    tracked += 1;
    if i == tracked { // poll B3 }

    tracked += 1;
    if i == tracked { // poll B4 }
}

// increment poll_idx
self.poll_idx = (self.poll_idx + 1) % 5;
```

rphmeier:
Travis is taking a while -- anyone know why?

I think in the `JoinAll` case the additional overhead of the `usize` is not really an issue, but in the `JoinX` case it may be. We could also put the `poll_idx` into a `task_local!` value, but the overhead associated with getting/setting it could be too high.

alexcrichton:
Thanks! Yeah my main worry here would be the overhead associated with a strategy like this, mind doing a bit of profiling/analysis to make sure it all compiles away?

rphmeier:
https://play.rust-lang.org/?gist=a0e704c7ed73f60db7df7a9596c0acb8&version=stable

shows that the weird loop on `JoinX` compiles down to the same assembly as a regular match
```
playground::Items::match_weird:
	pushq	%rax
.Lcfi3:
	cmpq	$4, %rsi
	ja	.LBB4_7
	leaq	.LJTI4_0(%rip), %rax
	movslq	(%rax,%rsi,4), %rcx
	addq	%rax, %rcx
	jmpq	*%rcx
.LBB4_2:
	addq	$8, %rdi
	jmp	.LBB4_6
.LBB4_3:
	addq	$16, %rdi
	jmp	.LBB4_6
.LBB4_5:
	addq	$32, %rdi
	jmp	.LBB4_6
.LBB4_4:
	addq	$24, %rdi
.LBB4_6:
	movq	(%rdi), %rax
	popq	%rcx
	retq
.LBB4_7:
	leaq	ref.2(%rip), %rdi
	callq	std::panicking::begin_panic
	ud2
.Lfunc_end4:
.LJTI4_0:
```

I'm not sure what a good way to profile this would be. For starters `chain` might end up slower than a single loop, and I know that the extra `usize` will have a space overhead.

We could also add a `Fair` combinator to those -- it's only really useful when you have lots of streams that yield values often.

carllerche:
> preventing starvation of the futures towards the end.

Join will (should) always poll all futures that are not yet ready. Could you clarify the starvation concern?

rphmeier:
I guess it's really more about fair ordering of futures to poll. I had a few futures in a join, one of which is almost always ready to make progress but had a fair amount of work when polled. When one of the later futures was ready to be polled it always had to wait for the first one.

carllerche:
> but had a fair amount of work when polled

I'm not sure what a "fair amount" is here, but it is also possible that this might better be suited by running this future on a thread pool or at least another task and then joining on the result.

In theory, the order that futures are polled on in the join should not matter.

alexcrichton:
Ok I'm going to close this now, but @rphmeier if this is still an issue please feel free to let me know!

PR #713, opened 2018-01-23
Update README.md

cssivision:
maybe use "0.1" instead of the latest version "0.1.17" is better?


alexcrichton:
Eh either's fine, 0.1.17 should be good for now!

Issue #461, opened 2017-04-26
select_all for streams


rushmorem:
It will be awesome to have `join_all` for streams as well.

alexcrichton:
Just to make sure I understand, you mean something that basically takes a `Vec<Stream<T>>` and gives you back a `Stream<Vec<T>>`?

rushmorem:
No. Something that takes a `Vec<Stream<T>>` and gives back a `Stream<T>`. Instead of waiting for one stream to finish before moving on to the next, the steams should all run concurrently so that the receiver can just pull values one at a time from whichever stream is ready.

Stebalien:
Ah, you mean `select_all` (a streaming version of [`future::select_all`][https://docs.rs/futures/0.1.13/futures/future/fn.select_all.html]).

rushmorem:
@Stebalien Ah, yes that one! Can we have it for streams?

alexcrichton:
Thanks for the clarification @rushmorem!

I think this is definitely a plausible combinator to have for streams. We'll likely want an implementation similar to `futures_unordered` using unpark events to know precisely which stream needs to be polled.

swizard0:
I've recently implemented such combinator on top of `Stream::select`: https://crates.io/crates/futures-stream-select-all 
You could use it as a temporary solution.

0xcaff:
This seems to be in 0.2.x versions but disabled in later versions. What's up with `select_all` for streams?

https://github.com/rust-lang-nursery/futures-rs/blob/d040a57521ca09d7eb162840bfb084d90d72db03/futures-util/src/stream/disabled/select_all.rs

Nemo157:
At a guess it was just never fully updated for 0.3, I don't see any reason to not have the same basic API so a PR to re-add it would probably be accepted.

taiki-e:
In 0.3, this seems to be currently implemented. https://rust-lang-nursery.github.io/futures-api-docs/0.3.0-alpha.13/futures/stream/fn.select_all.html

PR #721, opened 2018-01-29
streams: implement select_all based on futures_unordered (#461)

steffengy:
Since this is a very basic functionality, I decided to attempt to upstream this.
A common usecase would be some case of event stream from multiple sources.
Currently you would have trouble just merging those streams and using them as one.

This internally utilities futures_unordered to combine multiple
streams into one pollable stream, that yields the items in the order
they become available.
This does not expose any internals regarding futures_unordered,
so the internal implementation can be swapped out without breaking
backwards compatibility.

I first considered basing FuturesUnordered on SelectAll (and moving the implementation
to SelectAll) to save these implicit future conversions using `IntoFuture`.
This seemed not very beneficial, since the future is "popped" every time anyways,
for safety reasons (see Bomb), so I decided to stay with the more conservative implementation
for now.

### DIFF
--- a/src/stream/mod.rs
+++ b/src/stream/mod.rs
@@ -106,6 +106,7 @@ if_std! {
     mod collect;
     mod wait;
     mod channel;
+    mod select_all;
     mod split;
     pub mod futures_unordered;
     mod futures_ordered;
@@ -115,6 +116,7 @@ if_std! {
     pub use self::chunks::Chunks;
     pub use self::collect::Collect;
     pub use self::wait::Wait;
+    pub use self::select_all::SelectAll;
     pub use self::split::{SplitStream, SplitSink, ReuniteError};
     pub use self::futures_unordered::FuturesUnordered;
     pub use self::futures_ordered::{futures_ordered, FuturesOrdered};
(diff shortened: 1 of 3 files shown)


steffengy:
@alexcrichton 
any comments on the general approach taken/landing this as is?
Should this be targeted against 0.2 instead?

alexcrichton:
Egad sorry about that, this fell by the wayside! The implementation here seems fine to me and seems like a nifty combinator to have! It may not be the most optimal implementation but we can always figure that out later too :)

@cramertj how do you feel about combinators like this?

cramertj:
I think it's great! I've combined `FuturesUnordered` and `StreamFuture` by hand in the past to get this exact same result, so it'd be handy to have it as a build-in combinator!

alexcrichton:
Ok! @steffengy if it's ok with you I'll work on getting this on the 0.2 branch and we can avoid it in 0.1? If you'd like though I'd also be fine merging this into 0.1

steffengy:
@alexcrichton Yeah sure, 0.2 is fine.

alexcrichton:
Ok merged to 0.2!

Issue #633, opened 2017-11-07
Add into_inner/get methods to more types


canndrew:
I often want to deconstruct combinator futures/streams back into their components or at least be able to access their components. For example, I want to use `my_stream.map(...)` for a while, then remove the `map` and use `my_stream` again. Or I want implement a `Future` extension trait on types like `AndThen` and `Select` but I need to access their components in order to do it.

Is the lack of methods like `into_inner` and `get_inner` on these types an oversight or is this sort of thing intentionally discouraged (or will we be making these types abstract in the future)?

carllerche:
`get_ref` / `get_mut` / `into_inner` should be on all relevant types unless there is a reason for it to not be.

If there is an obvious impl, it is most likely an oversight and PRs appreciated.

canndrew:
Really, we could just expose the internals of a lot of these types. `Select2`, for example, has an obvious implementation. It's basically:

```rust
enum Select2<A, B> {
    Running(A, B),
    Done,
}
```

It's never going to *not* be implemented like this, so there's no risk of painting ourselves into a corner by exposing its implementation details as part of its public API. It's the same story with all the really basic combinators.

alexcrichton:
I'd prefer to avoid exposing any internals for now, but it's true that we can just add accessors where necessary if they're still missing.

carllerche:
👍 to not exposing internals.

canndrew:
The reason I suggested exposing internals is that I tried implementing the `get_ref`/`get_mut`/`into_inner` methods for `AndThen`, and they look like this:

```rust
impl<A, B, F> AndThen<A, B, F>
where
    A: Future,
    B: IntoFuture,
{
    fn get_ref<'a>(&'a self) -> AndThenInnerRef<'a, A, B, F> {
        match self.state {
            Chain::First(ref a, ref c) => AndThenInnerRef::RunningFirst(a, c),
            Chain::Second(ref b) => AndThenInnerRef::RunningSecond(b),
            Chain::Done => AndThenInnerRef::Done,
        }
    }

    fn get_mut<'a>(&'a mut self) -> AndThenInnerMut<'a, A, B, F> {
        match self.state {
            Chain::First(ref mut a, ref mut c) => AndThenInnerMut::RunningFirst(a, c),
            Chain::Second(ref mut b) => AndThenInnerMut::RunningSecond(b),
            Chain::Done => AndThenInnerMut::Done,
        }
    }

[...]

PR #722, opened 2018-01-30
add into_inner and similar methods to StreamFuture

tdyas:
When selecting on a StreamFuture and other futures, it is useful to be able to recover the underlying Stream if one of the other futures resolves first.

See https://github.com/rust-lang-nursery/futures-rs/issues/633#issuecomment-360932386 for more information.

### DIFF
--- a/src/stream/future.rs
+++ b/src/stream/future.rs
@@ -14,6 +14,46 @@ pub fn new<S: Stream>(s: S) -> StreamFuture<S> {
     StreamFuture { stream: Some(s) }
 }
 
+impl<S> StreamFuture<S> {
+    /// Acquires a reference to the underlying stream that this combinator is
+    /// pulling from.
+    ///
+    /// This method returns an `Option` to account for the fact that `StreamFuture`'s
+    /// implementation of `Future::poll` consumes the underlying stream during polling 
+    /// in order to return it to the caller of `Future::poll` if the stream yieleded
+    /// an element.
+    pub fn get_ref(&self) -> Option<&S> {
+        self.stream.as_ref()
+    }
+
+    /// Acquires a mutable reference to the underlying stream that this
+    /// combinator is pulling from.
+    ///
+    /// Note that care must be taken to avoid tampering with the state of the
+    /// stream which may otherwise confuse this combinator.
+    ///
+    /// This method returns an `Option` to account for the fact that `StreamFuture`'s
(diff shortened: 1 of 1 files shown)


alexcrichton:
Thanks! Looks like CI may be failing?

tdyas:
Pushed a fix. Helps if I had run `cargo test` in the correct shell window ...

alexcrichton:
Thanks! Could these functions actually return `Option` though with documentation indicating when `None` is returned?

tdyas:
They could. That's an API design choice that is yours to make. From what I understand, the `self.stream` `Option` should be` Some(S)` unless the `StreamFuture` is being actively polled. So I see two choices:

1. Document that the methods can return `None` if the future is being actively polled. The developer has the responsibility to ensure that polling has stopped (such as when the `StreamFuture` is returned to the closure of some other combinator). This choice won't panic of its own accord, but will if the developer tries to unilaterally unwrap the Option.

2. Keep the implementation I wrote which will panic if the `StreamFuture` is in the midst of being polled.

Pushing the chance of panic to the developer seems like the better solution, although the fact that the stream is stored in an Option internally seems like an internal implementation detail of `StreamFuture`. I would lean to #1 but up to you.

alexcrichton:
Yeah the convention so far that we're trying to stick to is to push the panics outwards, e.g. give you a chance to handle the case where the inner stream is gone.

tdyas:
The latest change as the Option return types and applicable documentation.

alexcrichton:
Thanks!

src/stream/future.rs
```diff
@@ -14,6 +14,46 @@ pub fn new<S: Stream>(s: S) -> StreamFuture<S> {
     StreamFuture { stream: Some(s) }
 }
 
+impl<S> StreamFuture<S> {
+    /// Acquires a reference to the underlying stream that this combinator is
+    /// pulling from.
+    ///
+    /// This method returns an `Option` to account for the fact that `StreamFuture`'s
+    /// implementation of `Future::poll` consumes the underlying stream during polling 
+    /// in order to return it to the caller of `Future::poll` if the stream yieleded
```
discosultan: _yielded_ 😉

tdyas: PR for fix: https://github.com/rust-lang-nursery/futures-rs/pull/723

tdyas: And thanks. :)

PR #728, opened 2018-02-05
Change Stream::filter predicate to return a Future

### LINKED ISSUE #311: Inconsistency between `Stream::filter` and `Stream::skip_while`
The predicate of `Stream::filter` returns a plain `bool`, while the predicate of `skip_while` returns an `IntoFuture<Item=bool, Error=Self::Error>`.
I think this should be made consistent.

cramertj:
This makes Stream::filter and Stream::skip_while
consistent by making both predicates return a
Future<Item=bool> rather than a plain bool.

@alexcrichton 

Closes #311

### DIFF
--- a/src/stream/filter.rs
+++ b/src/stream/filter.rs
@@ -1,4 +1,4 @@
-use {Async, Poll};
+use {Async, Future, IntoFuture, Poll, Sink};
 use stream::Stream;
 
 /// A stream combinator used to filter the results of a stream and only yield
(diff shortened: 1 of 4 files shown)


carllerche:
IMO this adds unnecessary noise.

The vast majority of use cases for filter & skip_while are passing a predicate that has no asynchronicity.

An asynchronous variant could always be added w/ `filter_async`.

cramertj:
@carllerche IMO `Ok` wrapping is less overhead than needing to know that the `_async` variants exist and when to use them, but we can iterate on this once the combinators are in a separate crate.

carllerche:
The question is also how often you have to do each one. If it was 50/50, sure. But I doubt it will be common at all to have to reach for the async version.

PR #729, opened 2018-02-05
Replace the Future impl for Option with IntoFuture

cramertj:
@alexcrichton

### DIFF
--- a/src/future/option.rs
+++ b/src/future/option.rs
@@ -1,13 +1,33 @@
 //! Definition of the `Option` (optional step) combinator
 
-use {Future, Poll, Async};
+use {Future, IntoFuture, Poll, Async};
+use std::option;
 
-impl<F, T, E> Future for Option<F> where F: Future<Item=T, Error=E> {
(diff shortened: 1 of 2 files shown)


jonhoo:
Hmm, isn't this a backwards-incompatible change? Specifically:
```rust
let x: Option<impl Future<...>> = ...;
core.run(x)
```
no longer compiles (see https://github.com/jonhoo/fantoccini/issues/8).

jonhoo:
And as far as I can tell, there isn't a forwards-compatible way to change this either, since `Option<F>` doesn't implement `IntoFuture` in `0.1.18` :/

luser:
This change broke people doing `cargo install sccache` because cargo doesn't respect Cargo.lock in that case.

cramertj:
Yes, this change was backwards-incompatible and was only supposed to appear in the 0.2 release.

cramertj:
I've opened https://github.com/rust-lang-nursery/futures-rs/pull/914 with the revert and yanked 0.1.19.

PR #731, opened 2018-02-05
Remove deprecated functionality

cramertj:
@alexcrichton


src/task_impl/mod.rs
```diff
@@ -58,7 +57,6 @@ fn with<F: FnOnce(&BorrowedTask) -> R, R>(f: F) -> R {
 pub struct Task {
     id: usize,
```
alexcrichton: Weird I *thought* this was deprecated but we may still be using it somewhere? Can always figure that out later of course.

src/task_impl/std/mod.rs
```diff
@@ -8,21 +8,17 @@ use std::ptr;
 use std::sync::{Arc, Mutex, Condvar, Once, ONCE_INIT};
 use std::sync::atomic::{AtomicUsize, Ordering};
 
-use {Future, Stream, Sink, Poll, Async, StartSend, AsyncSink};
+use {Future, Stream, Sink, Async, AsyncSink};
 use super::core;
 use super::{BorrowedTask, NotifyHandle, Spawn, spawn, Notify, UnsafeNotify};
+pub use super::core::{BorrowedUnpark, TaskUnpark};
```
alexcrichton: Now that there's no need for unpark events this can (in the future of course if you'd prefer) be cleaned up a lot and not have this sd/core business. These types can move from the `core` module directly into the parent module here to directly make up `Task` and remove a few layers of indirection in the meantime.

Cargo.toml
```diff
@@ -22,8 +22,7 @@ appveyor = { repository = "alexcrichton/futures-rs" }
 
 [features]
 use_std = []
```
alexcrichton: While we're at it I think the conventional name for this today is just `std`, so we could change that here too

cramertj: Done.

src/task_impl/mod.rs
carllerche: We never deprecated the actual `id` but all usage of it has been deprecated.

The `id` **should** be removed in 0.2.

cramertj: Isn't `id` still used in calls to `with_notify`, so that it will be passed down into the inner call to `notify`?

alexcrichton: It is indeed! That's, confusingly, a [different id though](https://github.com/rust-lang-nursery/futures-rs/blob/b11684d43be2d77e84324d309fa3e1d8b15aa076/src/task_impl/core.rs#L25), which indeed can't be removed unless we remove the ids from the `Notify` trait.

cramertj: Ah, gotcha. Removed.

PR #734, opened 2018-02-06
Separate futures-rs into workspace crates

cramertj:
The tests, `futures-cpupool`, and `futures-unordered`/executor related things are not working / commented out for now-- we can fix them when `futures-executor` gets integrated back in and the `task::Context` changes are made (or decided against).


futures-sink/Cargo.toml
```diff
@@ -0,0 +1,19 @@
+[package]
+name = "futures-sink"
+version = "0.1.0"
+authors = ["Alex Crichton <alex@alexcrichton.com>"]
+license = "MIT/Apache-2.0"
+repository = "https://github.com/alexcrichton/futures-rs"
+homepage = "https://github.com/alexcrichton/futures-rs"
+documentation = "https://docs.rs/futures-core"
```
alexcrichton: These documentation links will likely want to get updated, and we may want to update the repo links to the specific folders (but we can do that later of course)

futures-core/src/task_impl/mod.rs
```diff
@@ -218,57 +217,16 @@ impl<T: ?Sized> Spawn<T> {
...
-    ///
-    /// If the underlying operation returns `NotReady` then the `notify` value
-    /// passed in will receive a notification when the operation is ready to be
-    /// attempted again.
-    pub fn close_notify<N>(&mut self,
-                           notify: &N,
-                           id: usize)
-                           -> Poll<(), T::SinkError>
+    /// Invokes the function with the provided `notify` in the task context.
+    pub fn poll_notify<N, F, R>(&mut self, notify: &N, id: usize, f: F) -> R
```
alexcrichton: Just to confirm, was this intended to be `pub`? (not sure if it needs to be)

futures-core/src/stream/mod.rs
```diff
@@ -0,0 +1,120 @@
...
+//! ready as well.
+//!
+//! You can find more information/tutorials about streams [online at
+//! https://tokio.rs][online]
+//!
+//! [online]: https://tokio.rs/docs/getting-started/streams-and-sinks/
+
+use Poll;
+
+if_std! {
```
alexcrichton: tiny nit: could this come after the trait?

futures-core/src/lib.rs
```diff
@@ -0,0 +1,33 @@
...
+#![no_std]
+#![deny(missing_docs, missing_debug_implementations)]
+#![doc(html_root_url = "https://docs.rs/futures/0.2")]
+
+#[macro_use]
+#[cfg(feature = "std")]
+extern crate std;
+
+#[doc(hidden)]
+#[macro_export]
```
alexcrichton: Since this is so tiny, could it be duplicated where needed?

futures-core/src/lib.rs
```diff
@@ -0,0 +1,33 @@
...
+        $i
+    )*)
+}
+
+#[macro_use]
+mod poll;
+pub use poll::{Async, Poll};
+
+pub mod future;
+pub use future::{Future, IntoFuture};
```
alexcrichton: This makes me think, could we avoid this for now?

Maybe each crate could have its own prelude which slowly builds over time?

futures-core/src/future/mod.rs
```diff
@@ -0,0 +1,229 @@
...
+            self.0.poll()
+        }
+    }
+}
+
+/// Class of types which can be converted into a future.
+///
+/// This trait is very similar to the `IntoIterator` trait and is intended to be
+/// used in a very similar fashion.
+pub trait IntoFuture {
```
alexcrichton: I can't seem to find the blanket impl for `IntoFuture`... that wasn't lost by accident was it?

futures-channel/src/oneshot.rs
```diff
@@ -412,158 +411,3 @@ impl<T> Drop for Receiver<T> {
...
-///
-/// This handle is a future representing the completion of a different future on
-/// a separate executor. Created through the `oneshot::spawn` function this
-/// handle will resolve when the future provided to `spawn` resolves on the
-/// `Executor` instance provided to that function.
-///
-/// If this handle is dropped then the future will automatically no longer be
-/// polled and is scheduled to be dropped. This can be canceled with the
-/// `forget` function, however.
-pub struct SpawnHandle<T, E> {
```
alexcrichton: Just to confirm, was this removed for "permanent removal" or just temporarily for executors to shake out?

futures-channel/src/lib.rs
```diff
@@ -0,0 +1,66 @@
...
+//!
+//! This crate provides channels which can be used to communicate between futures.
+
+#![deny(missing_docs, missing_debug_implementations)]
+#![doc(html_root_url = "https://docs.rs/futures/0.2")]
+
+#[macro_use]
+extern crate futures_core;
+
+if_std! {
```
alexcrichton: I think since this whole crate requires std it's fine to not bother with `if_std!` here

futures-channel/src/lib.rs
```diff
@@ -0,0 +1,66 @@
...
+    pub mod mpsc;
+    pub mod oneshot;
+}
+
+// TODO: these types should be moved to futures-sink when it stabilizes
+
+#[doc(hidden)]
+/// The result of an asynchronous attempt to send a value to a sink.
+#[derive(Copy, Clone, Debug, PartialEq)]
+pub enum AsyncSink<T> {
```
alexcrichton: Urgh this is indeed a bummer.

Could a weird combination of things like `Result<Option<T>, E>` be used instead though?

futures-sink/Cargo.toml
```diff
@@ -0,0 +1,19 @@
...
+The asynchronous `Sink` trait for the futures-rs library.
+"""
+
+[features]
+std = ["futures-core/std"]
+default = ["std"]
+
+[dependencies]
+futures-core = { path = "../futures-core", version = "0.1.0" }
+futures-channel = { path = "../futures-channel", version = "0.1.0" }
```
alexcrichton: This may actually be a good optional dependency?

futures-util/Cargo.toml
```diff
@@ -0,0 +1,21 @@
...
+
+[features]
+std = ["futures-core/std", "futures-channel/std", "futures-executor/std", "futures-sink/std"]
+default = ["std"]
+
+[dependencies]
+futures-core = { path = "../futures-core", version = "0.1.0" }
+futures-channel = { path = "../futures-channel", version = "0.1.0" }
+futures-executor = { path = "../futures-executor", version = "0.1.0" }
+futures-sink = { path = "../futures-sink", version = "0.1.0" }
```
alexcrichton: Ah now this would definitely be a good optional dependency (if you don't otherwise use sinks/channels/etc)

alexcrichton: Although in this case of a util crate it may be a bummer b/c "util" tends to be "wait but I want this to be easy"

futures-util/src/future/mod.rs
```diff
@@ -960,140 +762,3 @@ fn assert_future<A, B, F>(t: F) -> F
...
-    /// The item that the future may resolve with.
-    type Item;
-    /// The error that the future may resolve with.
-    type Error;
-
-    /// Consumes this object and produces a future.
-    fn into_future(self) -> Self::Future;
-}
-
-impl<F: Future> IntoFuture for F {
```
alexcrichton: Aha here's that missing impl!

futures-util/src/lib.rs
```diff
@@ -0,0 +1,30 @@
...
+use futures_core::{Async, Future, IntoFuture, Poll, Stream};
+use futures_core::task;
+use futures_sink::{AsyncSink, Sink, StartSend};
+
+#[macro_use]
+#[cfg(feature = "std")]
+extern crate std;
+
+#[cfg(feature = "std")]
+pub mod lock;
```
alexcrichton: I think it's probably ok to avoid exporting this module, right?

futures-util/src/lib.rs
```diff
@@ -0,0 +1,30 @@
...
+
+pub mod future;
+pub use future::FutureExt;
```
alexcrichton: Hm... we may wanna chat on IRC briefly about preludes/exports maybe to sort this out, I'm not so sure my prelude-per-crate idea will pan out

futures-util/src/stream/catch_unwind.rs
```diff
@@ -52,20 +52,11 @@ impl<S> Stream for CatchUnwind<S>
...
+                    Ok(Async::Pending) => Ok(Async::Pending),
                     Ok(Async::Ready(Some(r))) => Ok(Async::Ready(Some(Ok(r)))),
                     Ok(Async::Ready(None)) => Ok(Async::Ready(None)),
                 }
             }
         }
     }
 }
-
-impl<S: Stream> Stream for AssertUnwindSafe<S> {
```
alexcrichton: I think this didn't make it through to the other end?

futures-core/src/task_impl/mod.rs
cramertj: Yeah-- I was trying to leave a way for people to `poll` a `Sink` now that this crate doesn't include the `Sink` trait.

futures-sink/Cargo.toml
cramertj: Agreed, we'll need to do a docs pass once everything is in the right place.

futures-channel/src/oneshot.rs
cramertj: I think something like this should definitely exist, but it's going to depend on what happens w/ executors. Do you want me to leave a note about it somewhere?

futures-channel/src/lib.rs
cramertj: I've switched it to `Result<Result<(), T>, E>`, but it's quite funky...

futures-core/src/future/mod.rs
cramertj: Fixed.

futures-core/src/lib.rs
cramertj: Done.

futures-core/src/stream/mod.rs
cramertj: Done

futures-util/src/stream/catch_unwind.rs
cramertj: It had to get moved into `futures-core` for coherence: https://github.com/rust-lang-nursery/futures-rs/pull/734/files/304d0bc9d07ac3d7b5d95bf691b1031f641e290a#diff-c3412131fd6221fd427f2ed71eca84abR30

futures-channel/src/oneshot.rs
alexcrichton: Ok yeah that's fine by me, want to add like a `TODO.md` or something like that in the repo which can be updated? (and mention this)

futures-util/src/lib.rs
cramertj: This is the old `bilock` stuff that was under `sync`. I'm fine w/ leaving it out if you are.

alexcrichton: Oh oops! Yeah seems fine to leave here.

futures-cpupool/src/lib.rs
```diff
@@ -430,6 +436,62 @@ impl Builder {
     }
 }
 
+/// Units of work submitted to an `Executor`, currently only created
+/// internally.
+pub struct Run {
```
alexcrichton: I think it's ok to drop the `pub` here and on the `Run::run` below

futures/Cargo.toml
```diff
@@ -20,10 +20,11 @@ appveyor = { repository = "alexcrichton/futures-rs" }
...
 [dependencies]
 futures-core = { path = "../futures-core", version = "0.1.0" }
+futures-channel = { path = "../futures-channel", version = "0.1.0" }
 futures-executor = { path = "../futures-executor", version = "0.1.0" }
 futures-sink = { path = "../futures-sink", version = "0.1.0" }
 futures-util = { path = "../futures-util", version = "0.1.0" }
 
 [features]
-std = []
+std = ["futures-core/std", "futures-executor/std", "futures-sink/std", "futures-util/std"]
```
alexcrichton: Oh hm this may not actually have much effect unless all the above dependencies are with `default-features = false` as well...

cramertj: ah, you're probably right.

futures-cpupool/src/lib.rs
cramertj: Done

futures-channel/src/oneshot.rs
cramertj: Done.

futures-core/src/lib.rs
cramertj: Discussed on IRC-- decision was to keep individual preludes for crates and not to reexport items from dependencies.

futures-util/src/future/mod.rs
cramertj: Fixed.

futures-util/src/lib.rs
cramertj: Discussed on IRC-- decision was to keep individual preludes for crates and not to reexport items from dependencies.

Issue #735, opened 2018-02-07
More Flexible FuturesUnordered


cramertj:
Right now, `FuturesUnordered` allocates a new node to store each future. If the plan is to continue with this behavior forever, `FuturesUnordered` should allow inserting unsized types (`Future` trait objects and similar).

Alternatively, `FuturesUnordered` could offer a behavior that allows the user to re-use allocations from completed `Future`s.

Either of these improvements would preclude the other, but IMO one of them should be offered.

PR #736, opened 2018-02-07
Initial pass at a futures-executor crate

alexcrichton:


futures-executor/src/current_thread.rs
```diff
@@ -0,0 +1,158 @@
...
+///
+/// # Panics
+///
+/// This function can only be invoked from the context of a
+/// `run` call; any other use will result in a panic.
+pub fn spawn<F>(future: F)
+    where F: Future<Item = (), Error = ()> + 'static
+{
+    _spawn(future, false).unwrap_or_else(|_| {
+        panic!("cannot call `execute` unless the thread is already \
```
cramertj: Remind me: is there a reason for `unwrap_or_else(|_| panic!(...))` instead of `expect`?

futures-executor/src/current_thread.rs
```diff
@@ -0,0 +1,158 @@
...
+///
+/// This cancels both daemon and non-daemon futures.
+///
+/// # Panics
+///
+/// This function can only be invoked from the context of a
+/// `run` call; any other use will result in a panic.
+pub fn cancel_all_executing() {
+    with_current(|current| {
+        match current {
```
cramertj: Again WRT `expect` question-- is there a reason why this isn't `current.expect("...").cancel_all_executing();`?

futures-executor/src/task_runner.rs
```diff
@@ -0,0 +1,301 @@
...
+///
+/// Note that `TaskRunner` is likely a low-level detail you shouldn't concern
+/// yourself with, when in doubt use `CurrentThread` instead. If you're looking
+/// to spawn a list of futures within a future you're likely going to want
+/// `FuturesUnordered`. Using `TaskRunner` typically implies that you're
+/// integraing the `futures` crate into a foreign event loop (aka not Tokio
+/// which already knows about `futures`).
+pub struct TaskRunner {
+    inner: Rc<Inner>,
+    futures: Spawn<FuturesUnordered<SpawnedFuture>>,
```
cramertj: Related: I opened https://github.com/rust-lang-nursery/futures-rs/issues/735 to track this, but is there a reason why you have to box futures to put them in `FuturesUnordered` (which puts them in an `Arc` internally)? Should `FuturesUnordered` just allow `F: ?Sized`?

cramertj:
LGTM, just left a couple quick comments.

futures-executor/src/current_thread.rs
alexcrichton: Ah mostly just for a nicer error message as this is one we're expecting will come up in practice, but no reason in general.

futures-executor/src/task_runner.rs
alexcrichton: Nah for now that's just how it is, but I think it'd be quite plausible for us to change it!

PR #739, opened 2018-02-08
Fill out `futures-executor`, fix some tests

alexcrichton:
This implements the `Enter` type in the `futures-executor` crate and starts migrating and fixing a lot of tests. Not all tests are working just yet though!


alexcrichton:
I think all the remaining tests after this in the `tests` folder need to get moved to the `futures-util` crate, but other than that the other crates should be good to go for testing!

futures-channel/Cargo.toml
```diff
@@ -10,8 +10,16 @@ description = """
 Channels for asynchronous communication using futures-rs.
 """
 
+[features]
+std = []
```
cramertj: Do we want an `std` feature here? In my last PR, I removed it from `futures-channel` since everything in `futures-channel` requires `std`.

futures-core/src/future/result.rs
```diff
@@ -29,7 +29,7 @@ impl<T, E> IntoFuture for result::Result<T, E> {
 /// # Examples
 ///
 /// ```
-/// use futures::future::*;
+/// use futures_core::future::*;
```
cramertj: One thing to think about: these doc tests will show up in the crate docs for `futures`, so people using `futures` directly might be confused when they see references to the individual crates in the docs.

futures-cpupool/src/lib.rs
```diff
@@ -179,22 +181,18 @@ impl CpuPool {
         Builder::new().create()
     }
 
-    fn execute_run(&self, run: Run) {
```
cramertj: Clarification: there are no functional changes to the behavior of this file, right? It's just refactoring?

futures-util/src/future/mod.rs
```diff
@@ -71,34 +71,6 @@ impl<T: ?Sized> FutureExt for T where T: Future {}
...
-    ///
-    /// This method is only available when the `std` feature of this
-    /// library is activated, and it is activated by default.
-    ///
-    /// # Panics
-    ///
-    /// This function does not attempt to catch panics. If the `poll` function
-    /// of this future panics, panics will be propagated to the caller.
-    #[cfg(feature = "std")]
-    fn wait(self) -> result::Result<Self::Item, Self::Error>
```
cramertj: :cry:

futures-executor/src/enter.rs
```diff
@@ -0,0 +1,96 @@
...
+                permanent: false,
+            })
+        }
+    })
+}
+
+impl Enter {
+    /// Register a callback to be invoked if and when the thread
+    /// ceased to act as an executor.
+    pub fn on_exit<F>(&self, f: F) where F: FnOnce() + 'static {
```
cramertj: This seems really fragile to me-- do you have a use-case in mind for it?

futures-sink/Cargo.toml
```diff
@@ -11,9 +11,9 @@ The asynchronous `Sink` trait for the futures-rs library.
 """
 
 [features]
-std = ["futures-core/std"]
+std = ["futures-core/std", "futures-channel/std"]
```
cramertj: Ditto question about `futures-channel/std`-- this crate currently doesn't have any non-std behavior. What should we do here?

cramertj:
@alexcrichton LGTM-- we should figure out what the story is for crates which don't expose any `no_std` features, such as `futures-channel`. Also, I don't understand the point of `Enter::on_exit`, especially since it isn't usable in combination with thread pools (or anything that calls `make_permanent`).

futures-channel/Cargo.toml
alexcrichton: Oh I recalled you saying we'd just have `std` everywhere for consistency, but yeah I can remove this.

futures-cpupool/src/lib.rs
alexcrichton: Ah the change here was that each instance of `Run` contained a `CpuPool` and it was switched to an `Arc<Inner>` because the destructor of `CpuPool` was significant for shutting it down. Other than that though no big changes!

futures-channel/Cargo.toml
cramertj: Yeah, I said that's why I did it originally and then I removed it haha

futures-util/src/future/mod.rs
alexcrichton: Do you think this should stick around for now? (not sure if we're ready to remove?)

futures-channel/Cargo.toml
cramertj: I'm fine w/ either. If we're including the feature, we should if_std out the body of futures_channel

futures-util/src/future/mod.rs
cramertj: We can always add it back in some form or another. Up to you how you want to handle it vs. the churn in the tests.

futures-executor/src/enter.rs
alexcrichton: I believe this was intended for things along the lines of `handle.set_default_for(&enter)` where you'd have a handle to some "thing" (like an event loop, timer, etc) and you, as the executor, would want to configure these global resources for your particular scope. For example you'd say something like "use this timer when you're on this executor" or "use that event loop". 

The `on_exit` here was a way for libraries to use `Enter` to communicate this sort of "per executor" state where you'd have a function that takes an `Enter` which sets up some state for your thread and is then cleaned up in `Enter::on_exit`. An example of this would be you'd register an event loop handle in a thread local, and then when the executor finishes you'd pull out that event loop handle and release it.

futures-sink/Cargo.toml
alexcrichton: Hm this actually makes me think that we may want to have a `std` feature even for the `futures-channel` crate (which makes the crate compile to nothing). That way crates like `futures-sink` which are actually somewhat `no_std` compatible would have a way of saying "you just don't get channel things if you turn off `std`".

How's that sound?

cramertj: Sounds good to me!

PR #744, opened 2018-02-09
[WIP] Explicit task::Context argument to poll.

ahmedcharles:
Only `futures-core` has been changed so far. Feedback welcome.

#129 https://github.com/rust-lang-nursery/futures-rfcs/pull/2

### DIFF
--- a/futures-channel/tests/mpsc.rs
+++ b/futures-channel/tests/mpsc.rs
@@ -34,29 +34,29 @@ fn send_recv_no_buffer() {
 
     // Run on a task context
     let f = lazy(move || {
-        assert!(tx.flush().unwrap().is_ready());
-        assert!(tx.poll_ready().unwrap().is_ready());
+        assert!(tx.flush(&mut TaskContext).unwrap().is_ready());
+        assert!(tx.poll_ready(&mut TaskContext).unwrap().is_ready());
 
         // Send first message
-        let res = tx.start_send(1).unwrap();
+        let res = tx.start_send(&mut TaskContext, 1).unwrap();
         assert!(res.is_ok());
-        assert!(tx.poll_ready().unwrap().is_not_ready());
+        assert!(tx.poll_ready(&mut TaskContext).unwrap().is_not_ready());
 
         // Send second message
-        let res = tx.start_send(2).unwrap();
+        let res = tx.start_send(&mut TaskContext, 2).unwrap();
         assert!(res.is_err());
 
         // Take the value
-        assert_eq!(rx.poll().unwrap(), Async::Ready(Some(1)));
-        assert!(tx.poll_ready().unwrap().is_ready());
+        assert_eq!(rx.poll(&mut TaskContext).unwrap(), Async::Ready(Some(1)));
+        assert!(tx.poll_ready(&mut TaskContext).unwrap().is_ready());
 
-        let res = tx.start_send(2).unwrap();
+        let res = tx.start_send(&mut TaskContext, 2).unwrap();
         assert!(res.is_ok());
-        assert!(tx.poll_ready().unwrap().is_not_ready());
+        assert!(tx.poll_ready(&mut TaskContext).unwrap().is_not_ready());
 
         // Take the value
-        assert_eq!(rx.poll().unwrap(), Async::Ready(Some(2)));
-        assert!(tx.poll_ready().unwrap().is_ready());
+        assert_eq!(rx.poll(&mut TaskContext).unwrap(), Async::Ready(Some(2)));
+        assert!(tx.poll_ready(&mut TaskContext).unwrap().is_ready());
 
         Ok::<(), ()>(())
     });
@@ -125,8 +125,8 @@ fn recv_close_gets_none() {
     let f = lazy(move || {
         rx.close();
 
-        assert_eq!(rx.poll(), Ok(Async::Ready(None)));
-        assert!(tx.poll_ready().is_err());
+        assert_eq!(rx.poll(&mut TaskContext), Ok(Async::Ready(None)));
+        assert!(tx.poll_ready(&mut TaskContext).is_err());
 
         drop(tx);
 
@@ -142,8 +142,8 @@ fn tx_close_gets_none() {
 
     // Run on a task context
     let f = lazy(move || {
-        assert_eq!(rx.poll(), Ok(Async::Ready(None)));
-        assert_eq!(rx.poll(), Ok(Async::Ready(None)));
+        assert_eq!(rx.poll(&mut TaskContext), Ok(Async::Ready(None)));
+        assert_eq!(rx.poll(&mut TaskContext), Ok(Async::Ready(None)));
 
         Ok::<(), ()>(())
     });
@@ -322,7 +322,7 @@ fn stress_receiver_multi_task_bounded_hard() {
                         // Just poll
                         let n = n.clone();
                         let f = lazy(move || {
-                            let r = match rx.poll().unwrap() {
+                            let r = match rx.poll(&mut TaskContext).unwrap() {
                                 Async::Ready(Some(_)) => {
                                     n.fetch_add(1, Ordering::Relaxed);
                                     *lock = Some(rx);
@@ -443,12 +443,12 @@ fn stress_poll_ready() {
         type Item = ();
         type Error = ();
 
-        fn poll(&mut self) -> Poll<(), ()> {
+        fn poll(&mut self, ctx: &mut TaskContext) -> Poll<(), ()> {
             // In a loop, check if the channel is ready. If so, push an item onto the channel
             // (asserting that it doesn't attempt to block).
             while self.count > 0 {
-                try_ready!(self.sender.poll_ready().map_err(|_| ()));
-                assert!(self.sender.start_send(self.count).unwrap().is_ok());
+                try_ready!(self.sender.poll_ready(ctx).map_err(|_| ()));
+                assert!(self.sender.start_send(ctx, self.count).unwrap().is_ok());
                 self.count -= 1;
             }
             Ok(Async::Ready(()))
@@ -523,7 +523,7 @@ fn try_send_2() {
     let th = thread::spawn(|| {
         run(|c| {
             c.block_on(lazy(|| {
-                assert!(tx.start_send("fail").unwrap().is_err());
+                assert!(tx.start_send(&mut TaskContext, "fail").unwrap().is_err());
                 Ok::<_, ()>(())
             })).unwrap();
 
--- a/futures-channel/tests/mpsc.rs
+++ b/futures-channel/tests/mpsc.rs
@@ -34,29 +34,29 @@ fn send_recv_no_buffer() {
 
     // Run on a task context
     let f = lazy(move || {
-        assert!(tx.flush(&mut TaskContext).unwrap().is_ready());
-        assert!(tx.poll_ready(&mut TaskContext).unwrap().is_ready());
+        assert!(tx.flush(&mut TaskContext::panicking()).unwrap().is_ready());
+        assert!(tx.poll_ready(&mut TaskContext::panicking()).unwrap().is_ready());
 
         // Send first message
-        let res = tx.start_send(&mut TaskContext, 1).unwrap();
+        let res = tx.start_send(&mut TaskContext::panicking(), 1).unwrap();
         assert!(res.is_ok());
-        assert!(tx.poll_ready(&mut TaskContext).unwrap().is_not_ready());
+        assert!(tx.poll_ready(&mut TaskContext::panicking()).unwrap().is_not_ready());
 
         // Send second message
-        let res = tx.start_send(&mut TaskContext, 2).unwrap();
+        let res = tx.start_send(&mut TaskContext::panicking(), 2).unwrap();
         assert!(res.is_err());
 
         // Take the value
-        assert_eq!(rx.poll(&mut TaskContext).unwrap(), Async::Ready(Some(1)));
-        assert!(tx.poll_ready(&mut TaskContext).unwrap().is_ready());
+        assert_eq!(rx.poll(&mut TaskContext::panicking()).unwrap(), Async::Ready(Some(1)));
+        assert!(tx.poll_ready(&mut TaskContext::panicking()).unwrap().is_ready());
 
-        let res = tx.start_send(&mut TaskContext, 2).unwrap();
+        let res = tx.start_send(&mut TaskContext::panicking(), 2).unwrap();
         assert!(res.is_ok());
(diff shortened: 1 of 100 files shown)


udoprog:
Echoing this in the right place this time:
---
So I'm playing around converting the code base, and I have hit a roadblock with trying to implement `catch_unwind` ([See comment](https://github.com/rust-lang-nursery/futures-rfcs/pull/2#issuecomment-364218145)). Does anyone know how to work around this?

Basically, this:
```rust
let (res, future) = catch_unwind(|| (future.poll(), future))?;
```

Becomes:
```rust
let (res, future) = catch_unwind(|| (future.poll(ctx), future))?;
```

And `ctx` can't be safely passed into the closure since it's not `UnwindSafe`. (My context is based on `BorrowedTask`).

ahmedcharles:
I'm working on `futures-util` now. I'll update again after I get that to compile.

udoprog:
After reading up this, if we can reason that `task::Context` _is_ unwind safe, we can use [`AssertUnwindSafe`] to work around this:

```rust
let (res, future) = catch_unwind(panic::AssertUnwindSafe(|| (future.poll(ctx), future)))?;
```

As to whether they are unwind safe seems to depend on how a specific NotifyHandle is implemented, and how task-local storage works (is it still platform dependent in 0.2?). It would be nice to make sure this is documented for NotifyHandle implementors, and make sure that the task-local APIs we provide guarantee unwind safety.

[`AssertUnwindSafe`]: https://doc.rust-lang.org/std/panic/struct.AssertUnwindSafe.html

ahmedcharles:
My goal so far has been to make the 'interesting' commit as short as possible, so it's easier to discuss/review.

It currently compiles, though the tests don't pass (well, more accurately, the tests have regressed compared to the penultimate commit.

futures-executor/src/task_runner.rs
```diff
@@ -298,6 +298,7 @@ impl Future for SpawnedFuture {
     type Error = bool;
 
     fn poll(&mut self, ctx: &mut task::Context) -> Poll<bool, bool> {
-        self.inner.with_task_data(|f| f.poll(ctx))
+        //self.inner.with_task_data(|f| f.poll(ctx))
+        self.inner.get_mut().poll(ctx)
```
ahmedcharles: I'm not sure I understand why this was implemented this way to begin with. It seems like it would be a noop, but I have no idea where the current task gets set.

futures-util/src/stream/catch_unwind.rs
```diff
@@ -45,7 +45,6 @@ impl<S> Stream for CatchUnwind<S>
             }
             CatchUnwindState::Stream(stream) => stream,
         };
-        // FIXME: fix passing the actual context here.
         let res = catch_unwind(|| (stream.poll(&mut task::Context::panicking()), stream));
```
ahmedcharles: This needs more thought.

futures-util/src/future/catch_unwind.rs
```diff
@@ -30,7 +30,6 @@ impl<F> Future for CatchUnwind<F>
 
     fn poll(&mut self, _ctx: &mut task::Context) -> Poll<Self::Item, Self::Error> {
         let mut future = self.future.take().expect("cannot poll twice");
-        // FIXME: fix passing the actual context here.
         let (res, future) = catch_unwind(|| (future.poll(&mut task::Context::panicking()), future))?;
```
ahmedcharles: This needs more thought.

futures-core/src/task_impl/mod.rs
```diff
@@ -266,30 +199,16 @@ impl<T: ?Sized> Spawn<T> {
         self.enter(BorrowedUnpark::new(&mk, id), f)
     }
 
-    /// TODO: dox
-    pub fn with_task_data<F, R>(&mut self, f: F) -> R
```
ahmedcharles: The RFC doesn't actually state what the changes to all of the API's should be, but this one doesn't seem useful anymore, given that the context explicitly contains task data.

futures-core/src/task_impl/mod.rs
```diff
@@ -15,106 +15,64 @@ pub use self::std::*;
...
+/// The poll function accepts a context, which is usually provided by the executor.
+pub struct Context<'a>(Option<ContextInner<'a>>);
+
+impl<'a> Context<'a> {
     /// Returns a `Context` which will panic when `waker` is called.
     /// This is useful when calling `poll` on a `Future` which doesn't
     /// access the context.
-    pub fn panicking() -> Context {
-        Context { waker: None }
+    pub fn panicking() -> Context<'a> {
```
ahmedcharles: This function seems questionable, but it did help migrate the tests for futures which don't access the context at all.

futures-core/src/task_impl/mod.rs
```diff
@@ -15,106 +15,64 @@ pub use self::std::*;
...
+///
+/// A task represents a single lightweight "thread" of execution driving a
+/// future to completion.
+///
+/// In general, futures are composed into large units of work, which are then
+/// spawned as tasks onto an *executor*. The executor is responsible for polling
+/// the future as notifications arrive, until the future terminates.
+///
+/// The poll function accepts a context, which is usually provided by the executor.
+pub struct Context<'a>(Option<ContextInner<'a>>);
```
ahmedcharles: The interaction between `Spawn` and `Context` make having `Context` own the task-local data difficult, hence the lifetime.

futures-util/src/stream/catch_unwind.rs
udoprog: As mentioned in [this comment], unwind safety seems to have been deferred to implementors of NotifyHandle and the platform-specific task-local storage implementation. This also was the case for TLS-based contexts. So it seem ignoring this using `panic::AssertUnwindSafe` is the way to go, and making sure panic safety is a documented requirement for `NotifyHandle`, `UnsafeNotify`, and task-local.

[this comment]: https://github.com/rust-lang-nursery/futures-rs/pull/744#issuecomment-364449525

futures-core/src/task_impl/mod.rs
udoprog: no-op contexts seems to be supported by providing an empty `LocalMap` plus a no-op `NotifyHandle` (or collecting, for testing).

The use for `Context(None)` seems to be to implement the panicking context. Could this be accomplished with an empty `LocalMap` and panicking `NotifyHandle` instead to avoid the `Option`?

futures-core/src/task_impl/mod.rs
udoprog: 👍

futures-core/src/task_impl/mod.rs
ahmedcharles: Probably, though I don't think it's important to get it perfect in the first PR, since it's an implementation detail that can be changed locally at any time.

futures-channel/tests/mpsc.rs
```diff
@@ -34,29 +34,29 @@ fn send_recv_no_buffer() {
 
     // Run on a task context
     let f = lazy(move || {
-        assert!(tx.flush().unwrap().is_ready());
-        assert!(tx.poll_ready().unwrap().is_ready());
+        assert!(tx.flush(&mut TaskContext).unwrap().is_ready());
```
cramertj: `TaskContext` isn't the name of a variable in this scope-- is this just a WIP artifact?

futures-core/src/task_impl/mod.rs
```diff
@@ -15,69 +15,64 @@ pub use self::std::*;
...
+/// A task represents a single lightweight "thread" of execution driving a
+/// future to completion.
 ///
 /// In general, futures are composed into large units of work, which are then
 /// spawned as tasks onto an *executor*. The executor is responsible for polling
 /// the future as notifications arrive, until the future terminates.
 ///
-/// This is obtained by the `task::current` function.
+/// The poll function accepts a context, which is usually provided by the executor.
+pub struct Context<'a>(Option<ContextInner<'a>>);
```
alexcrichton: This is certainly quite interesting! I think that for efficiency we will probably want the `'a` lifetime parameter on `Context`, but I could also see how that's not always desired (and it doesn't follow 100% for me in terms of what I might naively expect).

@cramertj do you have thoughts on this? Do you think we should avoid this entirely? Or in general, do you think that the concrete implementation of `Context` has an "obvious" implementation we should otherwise be doing?

futures-core/src/task_impl/mod.rs
```diff
@@ -217,42 +187,28 @@ impl<T: ?Sized> Spawn<T> {
...
-            let new_task = BorrowedTask {
-                unpark: task.unpark,
-                map: data,
-            };
-
-            set(&new_task, || f(obj))
-        })
-    }
-
     fn enter<F, R>(&mut self, unpark: BorrowedUnpark, f: F) -> R
```
alexcrichton: The whole `Spawn` type I think may not make 100% sense in the world where contexts are explicitly passed around. I'd sort of naively expect that we'd have something like:

```rust
struct Spawn<F> {
    inner: F,
    context: Context,
}
```

but that unfortunately doesn't work well with the lifetime argument on `Context` right now. In theory though the task-local data should be stored on the `Context` itself for now rather than in `Spawn` I think?

I wonder if we could do something like:

```rust
struct Context {
    data: TaskLocalDataMap, // I forget the actual name
    waker: NotifyHandle, // how to do a wakeup
}

impl Context {
    fn new<T: Into<NotifyHandle>>(inner: T) -> Context;
   
    // variuos accessors for `data`

    fn set_waker(&mut self, ...); // maybe?
}
```

One problem I think is that right now we make heavy usage of the lifetime parameters and stack closures and whatnot. In that sense we sort of want there to be a `Context` type (with no lifetime parameters maybe?) as well as a `ContextRef<'a>` type which all futures take. Or at least that matches that current scheme... I'm sort of losing myself here anyway

futures-executor/src/task_runner.rs
```diff
@@ -297,7 +297,8 @@ impl Future for SpawnedFuture {
     type Item = bool;
     type Error = bool;
 
-    fn poll(&mut self) -> Poll<bool, bool> {
-        self.inner.with_task_data(|f| f.poll())
+    fn poll(&mut self, ctx: &mut task::Context) -> Poll<bool, bool> {
```
alexcrichton: At least stylistically for this repo I'd personally prefer `cx` over `ctx`, but I may also be too used to rustc

futures/src/lib.rs
```diff
@@ -238,6 +238,8 @@ pub mod prelude {
         Poll,
     };
 
+    pub use futures_core::task::Context as TaskContext;
```
alexcrichton: @cramertj I think you followed the RFC more closely than I, but is this what the RFC concluded?

futures-core/src/task_impl/mod.rs
cramertj: I had tossed around a few ideas here: I think the actual function signature we want is `mut ctx: Context<'a>`. `Context` does need to take the locally provided impls of `Spawn` etc. by reference, as well as the "super" `Context`, so we do need a lifetime parameter.

alexcrichton: Agreed!

futures/src/lib.rs
cramertj: When the RFC concluded, I said we'd leave it as `task::Context`, and that further prelude-import-renamings could be addressed via followup discussions/proposals. I'd be inclined to leave this out for now, and instead import `task`.

futures-executor/src/task_runner.rs
carllerche: For what it is worth, `ctx` commonly stands for "context" in programming contexts (can be verified by google search "ctx programming" vs. "cx programming").

alexcrichton:
Ok so I think this could be a great opportunity to jettison `Spawn`-the-type! I wonder if we could get away with:

```rust
impl Context<'a> {
    pub fn new<F>(f: F, data: &'a TaskData) -> Context<'a> 
        where F: FnMut() -> NotifyHandle + 'a;
}
```

and I think via that we could eliminate `Spawn` entirely (the type) along with `spawn` the function and still have all the information needed for all `poll` calls, right?

carllerche:
@alexcrichton  I think this is good. As you pointed out, `Context` is the only thing you need to poll a task. There should be no need for `Spawn` anymore.

alexcrichton:
Hm so with some more thinking, I think we've got two choices of signatures for `Future`:

```rust
trait Future {
    // option 1
    fn poll(&mut self, cx: Context<'_>) -> Poll<T, E>;

    // option 2
    fn poll(&mut self, cx: &mut Context<'_>) -> Poll<T, E>;
}
```

If we were to go with option 1 I think it also requires a function like:

```rust
impl Context<'a> {
    fn reborrow(&'b mut self) -> Context<'b>;
}
```

to enable use cases like:

```rust
impl Future for Select {
    fn poll(&mut self, cx: Context<'_>) -> Poll<T, E> {
        let res1 = self.a.poll(cx);
        let res2 = self.b.poll(cx); //~ ERROR: use of moved value `cx`
        // ...
    }
}
```

I would personally lean towards `&mut Context<'_>` I think?

carllerche:
I lean very heavily towards `&mut Context`. Adding reborrowing fns does not feel great. There also should always be a place on the stack for the `Context` value (before calling `poll`).

futures-core/src/task_impl/mod.rs
ahmedcharles: I did try an iteration of this code with `context` as a member of `Spawn`. It makes panic safety harder when trying to change the notify handle temporarily for the task because you have to have a way to unset it. But you need mut access to the future and the context at the same time (which rust doesn't like).

futures/src/lib.rs
ahmedcharles: That's fine. It's an easy change.

futures-channel/tests/mpsc.rs
ahmedcharles: This is using the prelude (I don't really like preludes, but I figured I'd use it... maybe it would confuse someone and convince them preludes were bad? j/k).

futures-executor/src/task_runner.rs
cramertj: +1 for ctx on the bikeshed, if for no other reason than that it will be obvious to folks coming from golang.

ahmedcharles: It's probably not worth much, but I'm for `ctx` since I picked it in the first place and changing it would mean I'd have to change it. :)

alexcrichton:
Ok thanks again for starting this out @ahmedcharles! With https://github.com/rust-lang-nursery/futures-rs/pull/758 extracted from this and merged I think I've finished up most of the rest at https://github.com/rust-lang-nursery/futures-rs/pull/762 so I'm going to close this for now

PR #745, opened 2018-02-09
Don't close sinks in send_all/forward

### LINKED ISSUE #462: Can sink be reused after `send_all`?
Hi, I notice that `SendAll` struct will call `close` method when polling, doesn't it mean that though `SendAll` will be resolved back to Sink eventually, but it can't be reused anymore? And I don't find any combinator except `SendAll` will call close under the hook. So if I want to close a sink asynchronously, do I have to either reimplement a Future or just use `SendAll`?

alexcrichton:
Closes #462
Closes #485

### DIFF
--- a/futures-util/src/sink/mod.rs
+++ b/futures-util/src/sink/mod.rs
@@ -192,13 +192,12 @@ pub trait SinkExt: Sink {
     ///
     /// This future will drive the stream to keep producing items until it is
     /// exhausted, sending each item to the sink. It will complete once both the
-    /// stream is exhausted, the sink has received all items, the sink has been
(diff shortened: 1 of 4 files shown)


cramertj:
:shipit:

Ekleog:
Hmm, would it make sense to backport to 0.1? tokio uses 0.1, and so is stuck with this issue :/

cramertj:
@Ekleog Unfortunately this is a breaking change, and so is only available with futures 0.2.

Ekleog:
Oh :/ well, so be it, then, thanks!

PR #746, opened 2018-02-09
Increate unfold's flexibility

### LINKED ISSUE #260: Increase the flexibility of unfold
```
16:58:55 <aweinstock> Is there a compelling reason for unfold's step function to be (State -> Option<IntoFuture<Item=(StreamItem, State)>> rather than (State -> IntoFuture<Item=Option<(StreamItem, State)>>)?
16:59:43 <aweinstock> The flexibility to choose whether or not to yield the next item in the continuation of a future is helpful (perhaps necessary?) in some cases
16:59:58 <acrichto> hm yeah good point, makes sense to me!
17:00:02 <acrichto> want to open an issue w/ that?
17:00:12 <acrichto> unfortunately it'll have to wait until 0.2 at this point, but I can be sure we include it
17:00:32 <aweinstock> will do
17:00:50 <aweinstock> (I'm currently using my own version of unfold in https://github.com/aweinstock314/kingr4_weinsa_distributedsystems_project2/blob/master/src/main.rs#L302-L322 due to this)
17:01:38 <acrichto> yeah putting the option in the future makes sense to me
```

alexcrichton:
Move the `Option` to the result of the future, rather than immediately returning
an option of a future.

Closes #260

### DIFF
--- a/futures-util/src/stream/unfold.rs
+++ b/futures-util/src/stream/unfold.rs
@@ -53,8 +53,8 @@ use futures_core::{Future, IntoFuture, Async, Poll, Stream};
 /// # }
 /// ```
 pub fn unfold<T, F, Fut, It>(init: T, f: F) -> Unfold<T, F, Fut>
-    where F: FnMut(T) -> Option<Fut>,
-          Fut: IntoFuture<Item = (It, T)>,
+    where F: FnMut(T) -> Fut,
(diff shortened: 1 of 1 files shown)


lnicola:
@cramertj it looks like this was reverted in `futures 0.1.21`, but the commit wasn't pushed to GitHub. Will it come back in a future version?

PR #747, opened 2018-02-09
Rename `Sink::poll_complete` to `flush`

### LINKED ISSUE #379: Rename `poll_complete` to `poll_flush`
Clarify the semantics of `poll_complete` by renaming to `poll_flush`.

alexcrichton:
Along with a few other changes:

* Move the `Sink::flush` combinator to a free `flush` function
* Add a free `close` combinator

Closes #379

### DIFF
--- a/futures-util/src/sink/close.rs
+++ b/futures-util/src/sink/close.rs
@@ -1,0 +1,52 @@
+use futures_core::{Poll, Async, Future};
+use futures_sink::Sink;
+
+/// Future for the `close` combinator, which polls the sink until all data has
+/// been closeed.
+#[derive(Debug)]
+#[must_use = "futures do nothing unless polled"]
+pub struct Close<S> {
+    sink: Option<S>,
+}
+
+/// A future that completes when the sink has finished closing.
+///
+/// The sink itself is returned after closeing is complete.
+pub fn close<S: Sink>(sink: S) -> Close<S> {
+    Close { sink: Some(sink) }
+}
+
+impl<S: Sink> Close<S> {
+    /// Get a shared reference to the inner sink.
+    /// Returns `None` if the sink has already been closeed.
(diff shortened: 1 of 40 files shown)


futures-util/src/sink/close.rs
```diff
@@ -0,0 +1,52 @@
...
+/// The sink itself is returned after closeing is complete.
+pub fn close<S: Sink>(sink: S) -> Close<S> {
+    Close { sink: Some(sink) }
+}
+
+impl<S: Sink> Close<S> {
+    /// Get a shared reference to the inner sink.
+    /// Returns `None` if the sink has already been closeed.
+    pub fn get_ref(&self) -> Option<&S> {
+        self.sink.as_ref()
```
ahmedcharles: Some combinators return `&S` here instead and panic when sink is `None`. Is there a reason why there are two patterns?

cramertj: All of the `as_ref`/`get_ref`/`get_mut` style combinators return `Option`. `Future` implementations panic if they are polled after completion.

alexcrichton: Indeed! And to clarify, @ahmedcharles if you find inconsistencies where we panic then that's a bug we should fix!

ahmedcharles: https://github.com/rust-lang-nursery/futures-rs/blob/304978c09292390c17c03ad7d249b496dd0c3f9e/futures-util/src/sink/send.rs#L20-L38

It's not quite the same but I think it's what I remember seeing.

alexcrichton: Ok thanks, I fixed that in https://github.com/rust-lang-nursery/futures-rs/commit/8689770aa7ab74c17b03999fddb58b4ab34b20be

PR #757, opened 2018-02-13
First draft of stable futures. [DO NOT MERGE]

withoutboats:
This branch adds support for "pinned" futures, which can contain
internal self-references. It is currently based on the API exposed by
the anchor_experiment crate.

It significantly adjusts the definition of future, creating two traits:

```
trait Future {
    type Item;
    type Error;

    unsafe fn poll_unsafe(&mut self) -> Poll<Self::Item, Self::Error>;
}

trait FutureMove: Future {
     fn poll_move(&mut self) -> Poll<Self::Item, Self::Error>;
}
```

All existing futures are `FutureMove`, and their `poll_unsafe` methods
can delegate to `poll_move`. There are some complications for
combinators and other futures that are generic over underlying futures.

Today, the crate provides a `poll_safe!()` helper, to generate a safe
implementation of the `poll_unsafe` method if your type implements
`FutureMove`.

An even more ideal adapter would be something that looks like this:

```
impl_future_safely! {
    impl Future for MyType {
        type Item = MyItem;
        type Error = MyError;
            // ....
        }
    }
}
```

Then, users would just be able to wrap `impl_future_safely!` around
their existing implementations of `Future` and have upgraded.

This is not intended to be the final API of the Future trait. In
particular, we want to replace `poll_unsafe` with a safe method. That
API would look like this:

```
trait Future {
    type Item;
    type Error;

    fn poll(self: PinMut<Self>) -> Poll<Self::Item, Self::Error>;

}
```

However, that requires `arbitrary_self_types`, which cannot be used
currently for these reasons:

1. `futures` needs to run on stable. Its not clear how to cfg out the
   method with a `PinMut` receiver and still have everything work.
2. Arbitrary self types are not object safe yet, even on nightly. Its
   essential that the `Future` trait be object safe

These implementations exist to make it safe to call a self-referential
future:

- `impl<'a, F: ?Sized + Future> FutureMove for PinMut<'a, F>`
- `impl<F: ?Sized + Future> FutureMove for Anchor<Box<F>>`

This commit is not finished. In particular, the `futures-util` crate has
not been adapted to compile against this new version of `futures-core`.

### DIFF
--- a/futures-core/src/task_impl/mod.rs
+++ b/futures-core/src/task_impl/mod.rs
@@ -1,7 +1,7 @@
 use core::fmt;
 use core::marker::PhantomData;
 
-use {Poll, Future, Stream};
+use {Poll, FutureMove, Stream};
 
 mod atomic_task;
 pub use self::atomic_task::AtomicTask;
@@ -204,9 +204,9 @@ impl<T: ?Sized> Spawn<T> {
                                  notify: &N,
                                  id: usize) -> Poll<T::Item, T::Error>
         where N: Clone + Into<NotifyHandle>,
-              T: Future,
+              T: FutureMove,
     {
-        self.poll_notify(notify, id, |s| s.poll())
+        self.poll_notify(notify, id, |s| s.poll_move())
     }
 
     /// Like `poll_future_notify`, except polls the underlying stream.
--- a/futures-channel/src/lib.rs
+++ b/futures-channel/src/lib.rs
@@ -9,6 +9,7 @@
 #[cfg(feature = "std")]
 extern crate std;
 
+#[macro_use]
 extern crate futures_core;
 
 macro_rules! if_std {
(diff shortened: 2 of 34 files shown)


futures-core/src/task_impl/mod.rs
```diff
@@ -204,9 +204,9 @@ impl<T: ?Sized> Spawn<T> {
                                  notify: &N,
                                  id: usize) -> Poll<T::Item, T::Error>
         where N: Clone + Into<NotifyHandle>,
-              T: Future,
+              T: FutureMove,
     {
-        self.poll_notify(notify, id, |s| s.poll())
+        self.poll_notify(notify, id, |s| s.poll_move())
```
withoutboats: I don't know enough about the task system. Does it need to be bound `FutureMove`, or is it safe to bound it `Future`. Can we make changes to it, and then it will support `Future`?

withoutboats:
Closing in favor of the strategy taken in alexcrichton/futures-await#63

PR #758, opened 2018-02-13
Add task::Context to poll, etc.

ahmedcharles:
This doesn't change the functionality and the Context is `()`.
This allows having this commit be purely mechanical.
Also add some missing `#[must_use]`'s to futures and streams.

The primary goal here is to avoid merge conflicts. The changes here should be obvious given the RFC.


alexcrichton:
Ok I've merged this in and touched it up a bit, thanks @ahmedcharles!

PR #759, opened 2018-02-13
 Add Context argument to lazy.

ahmedcharles:
It seems quite common to use lazy to get into a task context. Make it explicit.

### DIFF
--- a/futures-channel/benches/sync_mpsc.rs
+++ b/futures-channel/benches/sync_mpsc.rs
@@ -81,7 +81,7 @@ fn unbounded_uncontended(b: &mut Bencher) {
         for i in 0..1000 {
             UnboundedSender::unbounded_send(&tx, i).expect("send");
             // No need to create a task, because poll is not going to park.
-            assert_eq!(Ok(Async::Ready(Some(i))), rx.poll());
+            assert_eq!(Ok(Async::Ready(Some(i))), rx.poll(&mut task::Context));
         }
     })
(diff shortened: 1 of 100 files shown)


ahmedcharles:
Depends on #758

alexcrichton:
Thanks! I think though that we may want to leave this to `poll_fn` perhaps? I see `lazy` as akin to things like `and_then` where I think we definitely don't want a context argument, right? (lots of the tests are pretty old and don't necessarily reflect idiomatic usage)

ahmedcharles:
Technically `poll_fn` doesn't work because it's `FnMut` and not `FnOnce`... Well, it works sometimes. The tests use `lazy` because it's easy and programmers are lazy (pun intended). I tried moving some of the tests over to using existing combinators, but that made me realize that most of the time, combinators don't let you do what you want to test. Then I figured I'd write a custom impl of `Future` for each test, but that seemed silly when I could use `poll_fn`, but then I used that and realized it's `FnMut`... anyways, I'm sure you see where this is going. :)

I considered adding a `lazy_poll` or similar, which took an argument while leaving the existing `lazy` alone, but then I figured that it's easy to ignore the parameter if it's unneeded.

alexcrichton:
Hm ok, I think I rewrote a number of tests to work in https://github.com/rust-lang-nursery/futures-rs/pull/762 so the rationale there may not be too pressing any more?

PR #761, opened 2018-02-13
Remove the `futures::task::current` function

alexcrichton:
This currently just makes it private and adds a new `Context::waker` function
which will call the original private version.


alexcrichton:
Closing in favor of https://github.com/rust-lang-nursery/futures-rs/pull/762

PR #762, opened 2018-02-13
Reorganize the `futures_core::task` module

alexcrichton:
This commit is a large reorganization of the `task` module in accordance with
RFC 2 which added an explicit `Context` argument to all futures-related
functions. This argument is now explicitly constructed when necessary and
otherwise passed around in implementations of `poll`.


futures-core/src/task/mod.rs
```diff
@@ -1,35 +1,142 @@
+//! Tasks used to drive a future computation
+//!
+//! It's intended over time a particular operation (such as servicing an HTTP
+//! request) will involve many futures. This entire operation, however, can be
+//! thought of as one unit, as the entire result is essentially just moving
+//! through one large state machine.
+//!
+//! A "task" is the unit of abstraction for what is driving this state machine
```
cramertj: @aturon Is this still how we want to talk about "tasks"? My intuition always lead me to think that a "task" was the top-level future driven by an executor -- we'd even talked about `trait Task = Future<Item = (), Error = ()>;`.

futures-core/src/task/mod.rs
```diff
@@ -1,35 +1,142 @@
...
-
 mod atomic_task;
 pub use self::atomic_task::AtomicTask;
 
-mod core;
+/// The context for the currently running task.
+pub struct Context<'a> {
+    handle_factory: &'a (Fn() -> NotifyHandle + 'a),
+    id: usize,
+    map: &'a LocalMap,
```
cramertj: Why is this an immutable reference, rather than a mutable one? We are mutating the `LocalMap`, and it should be shared with anyone else at the time, should it?

futures-core/src/task/mod.rs
```diff
@@ -1,35 +1,142 @@
...
+    /// It's sometimes necessary to pass extra information to the task when
+    /// unparking it, so that the task knows something about *why* it was woken.
+    /// See the `FutureQueue` documentation for details on how to do this.
+    ///
+    /// # Panics
+    ///
+    /// This function will panic if a task is not currently being executed.
+    /// That is, this method can be dangerous to call outside of an
+    /// implementation of `poll`.
+    pub fn waker(&mut self) -> Task {
```
cramertj: TODO: make this return a `Waker` rather than a `Task`.

futures-core/src/task/mod.rs
cramertj: Similarly, is there a reason the `handle_factory` is an `&'a Fn()` instead of an `&'a mut FnMut`?

futures-core/src/task/mod.rs
alexcrichton: Oh right yeah so I've just copied/pasted doc stuff around, I'm thinking we'll want docs to be a future PR. Right now it's all likely horribly outdated.

futures-core/src/task/mod.rs
alexcrichton: An excellent point! Just a carryover from before, I'll change it to a `mut` everywhere.

futures-core/src/task/mod.rs
alexcrichton: Sure yeah, I can do that in this PR as well.

futures-executor/src/current_thread.rs
```diff
@@ -122,12 +122,12 @@ impl<'a> Context<'a> {
...
     /// not.
-    pub fn block_on<F>(&mut self, future: F) -> Result<F::Item, F::Error>
+    pub fn block_on<F>(&mut self, mut future: F) -> Result<F::Item, F::Error>
         where F: Future,
     {
-        let mut future = task::spawn(future);
+        let mut map = LocalMap::new();
         loop {
-            match future.poll_future_notify(&::IntoNotifyHandle(self.handle), 0) {
+            match future.poll(&mut task::Context::new(&mut map, 0, self.handle)) {
```
cramertj: IMO this method is a bit confusing-- if i call `task::Context::new(&mut map, 0, handle).block_on(fut)` manually, it seems like i could reasonably expect it to use the `LocalMap` I provided, rather than generating a new one for this specific task. I guess this function will probably go away (or at least look different) in the new executor design, so maybe it's not worth worrying about now. What do you think?

aturon: Yeah I think let's revisit then.

aturon:
Gonna go ahead and merge this and start hacking on the executor stuff and the `Sink` refactor.

PR #764, opened 2018-02-14
Remove unneccessary poll_complete when trying to empty buffered Sink

### LINKED ISSUE #752: Potential issue with flushing a buffered Sink
I think this might be related to the recent Tokio reform, as I have only noticed on code that uses the new `tokio` rather than `tokio-core`, but tracing the code it appears to be a bug when flushing a buffered Sink.

The symptom was that the final few messages being written to a Sink were not being flushed, even when using `send_all` to feed the Sink, which should have flushed the sink before completing.  Instead, the `send_all` was never completing and the buffered messages never flushed.

After following the chain of function calls, the issue appears to be here: https://github.com/rust-lang-nursery/futures-rs/blob/master/src/sink/buffer.rs#L47-L61 (although I'm not familiar enough with the codebase to be sure).

```rust
    fn try_empty_buffer(&mut self) -> Poll<(), S::SinkError> {
        while let Some(item) = self.buf.pop_front() {
            if let AsyncSink::NotReady(item) = self.sink.start_send(item)? {
                self.buf.push_front(item);

                // ensure that we attempt to complete any pushes we've started
                self.sink.poll_complete()?;

                return Ok(Async::NotReady);
            }
        }

        Ok(Async::Ready(()))
    }
```

The result of the call to `self.sink.poll_complete()` was `Ok(Async::Ready(()))` because the underlying Sink had flushed successfully, but this function returns `Ok(Async::NotReady)` as this buffer was not fully flushed.

However, since the underlying Sink was flushed successfully, the future was not scheduled for further execution, hence why no more flushes were attempted and my system ground to a halt.

I verified this by creating a custom `send_all` within my client code, which explicitly rescheduled the future in the event of a `Async::NotReady` from the `poll_complete` function of the buffer.  Which did fix the problem, but this is obviously inefficient as it will re-poll the future at the earliest opportunity, rather than when the bottom-level indicates it is ready (in this case a `tokio` TcpStream).

The simplistic fix would be to check the result of `self.sink.poll_complete()?` and, if the result was `Async::Ready(())` to continue looping through the buffer until either the whole buffer is accounted for or we get a `Async::NotReady` from the underlying Sink.  But when I attempted this, the `buffer` test went into an infinite loop, so either the test needs to be updated, or (much more likely) I'm missing information and the simplistic fix isn't a fix after-all.

benashford:
This fixes #752 

Summary of the bug: buffered Sinks would sometimes get stuck - i.e. the task was never triggered to be polled again - because the call to `poll_complete` on line 53 would return `Async::Ready(())` even though the buffer still had data that needed to be written.

@alexcrichton pointed out that the call to `poll_complete` was actually unnecessary as `start_send` should (if implemented correctly) call `poll_complete` before returning, and therefore could be removed.

Removing the call did indeed fix the issue for me.  And all the futures-rs tests still pass.  So unless there are any side-effects I haven't anticipated, this should be a safe change.

### DIFF
--- a/src/sink/buffer.rs
+++ b/src/sink/buffer.rs
@@ -49,9 +49,6 @@ impl<S: Sink> Buffer<S> {
             if let AsyncSink::NotReady(item) = self.sink.start_send(item)? {
                 self.buf.push_front(item);
 
-                // ensure that we attempt to complete any pushes we've started
-                self.sink.poll_complete()?;
-
                 return Ok(Async::NotReady);
             }
         }


alexcrichton:
Thanks!

PR #765, opened 2018-02-14
Refactor Sink trait

### LINKED ISSUE #751: Consider an alternate naming scheme for `Sink` fns
#747 recently renamed `poll_complete` -> `flush`, dropping the `poll_` prefix.

I would suggest to do the opposite. Keep `poll_flush` and rename `close` -> `poll_close`.

The primary reason is to reserve `flush` and `close` for futures that perform these operations, especially for async / await. This would allow:

```rust
await sink.flush();
```

and

```rust
await sink.close();
```

I have other thoughts on `Sink` based on experience implementing and using it, but I am assuming that there will be an RFC at some point to go into more depth on the `Sink` API.

cramertj:
Fix #751

### DIFF
--- a/futures-sink/src/channel_impls.rs
+++ b/futures-sink/src/channel_impls.rs
@@ -1,39 +1,17 @@
 use {Async, Sink, Poll};
 use futures_core::task;
-use futures_channel::mpsc::{Sender, SendError, UnboundedSender};
-
-/// The error type of `<Sender<T> as Sink>`
-///
-/// It will contain a value of type `T` if one was passed to `start_send`
-/// after the channel was closed.
-#[derive(Debug)]
-pub struct ChannelClosed<T>(Option<T>);
-
-impl<T> ChannelClosed<T> {
-    /// Extract the inner message.
-    pub fn into_inner(self) -> Option<T> {
-        self.0
-    }
-
-    fn from_send_error(err: SendError<T>) -> Self {
-        ChannelClosed(Some(err.into_inner()))
-    }
-
-    fn from_empty_send_error(_: SendError<()>) -> Self {
-        ChannelClosed(None)
-    }
-}
+use futures_channel::mpsc::{Sender, ChannelClosed, UnboundedSender};
 
 impl<T> Sink for Sender<T> {
     type SinkItem = T;
     type SinkError = ChannelClosed<T>;
 
     fn poll_ready(&mut self, cx: &mut task::Context) -> Poll<(), Self::SinkError> {
-        self.poll_ready(cx).map_err(ChannelClosed::from_empty_send_error)
+        self.poll_ready(cx)
     }
 
     fn start_send(&mut self, msg: T) -> Result<(), Self::SinkError> {
-        self.start_send(msg).map_err(ChannelClosed::from_send_error)
+        self.start_send(msg)
     }
 
     fn start_close(&mut self) -> Result<(), Self::SinkError> {
@@ -50,11 +28,11 @@ impl<T> Sink for UnboundedSender<T> {
     type SinkError = ChannelClosed<T>;
 
     fn poll_ready(&mut self, cx: &mut task::Context) -> Poll<(), Self::SinkError> {
-        self.poll_ready(cx).map_err(ChannelClosed::from_empty_send_error)
+        self.poll_ready(cx)
     }
 
     fn start_send(&mut self, msg: T) -> Result<(), Self::SinkError> {
-        self.start_send(msg).map_err(ChannelClosed::from_send_error)
+        self.start_send(msg)
     }
 
     fn start_close(&mut self) -> Result<(), Self::SinkError> {
@@ -68,7 +46,7 @@ impl<T> Sink for UnboundedSender<T> {
 
 impl<'a, T> Sink for &'a UnboundedSender<T> {
     type SinkItem = T;
-    type SinkError = SendError<T>;
+    type SinkError = ChannelClosed<T>;
 
     fn poll_ready(&mut self, _: &mut task::Context) -> Poll<(), Self::SinkError> {
         Ok(Async::Ready(()))
--- a/futures-sink/src/channel_impls.rs
+++ b/futures-sink/src/channel_impls.rs
@@ -1,44 +1,67 @@
-use {Async, Sink, AsyncSink, StartSend, Poll};
+use {Async, Sink, Poll};
 use futures_core::task;
 use futures_channel::mpsc::{Sender, SendError, UnboundedSender};
 
-fn res_to_async_sink<T>(res: Result<(), T>) -> AsyncSink<T> {
-    match res {
-        Ok(()) => AsyncSink::Ready,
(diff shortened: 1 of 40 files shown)


futures-sink/src/channel_impls.rs
```diff
@@ -1,44 +1,67 @@
...
-fn res_to_async_sink<T>(res: Result<(), T>) -> AsyncSink<T> {
-    match res {
-        Ok(()) => AsyncSink::Ready,
-        Err(x) => AsyncSink::Pending(x),
+/// The error type of `<Sender<T> as Sink>`
+///
+/// It will contain a value of type `T` if one was passed to `start_send`
+/// after the channel was closed.
+#[derive(Debug)]
+pub struct ChannelClosed<T>(Option<T>);
```
alexcrichton: How come this was needed?

cramertj: Previously, it was always the case that an error contained a `T`. That's not the case anymore, since `poll_ready` doesn't have a `T` to respond with (previously, the only function that could error was `start_send`, which has a `T`).

alexcrichton: Hm could we mirror the `Option` in the original futures-channel crate as well though? (to avoid the introduction of another type)

cramertj: Sure-- i debated about this, since it's sort of misleading to the end user, because some functions will only ever return `Some` or `None`. In the `Sink` trait that's not preventable, but it is if you're directly using functions on `Channel`. I think I agree that the extra complexity isn't worth it, though.

alexcrichton: Eh I'd probably be fine either way, I understand now what's going on at least :)

cramertj:
@alexcrichton I've pushed another commit fixing up futures-util.

futures-sink/src/channel_impls.rs
```diff
@@ -1,62 +1,66 @@
...
 
-    fn flush(&mut self, _: &mut task::Context) -> Poll<(), SendError<T>> {
-        Ok(Async::Ready(()))
+    fn start_send(&mut self, msg: T) -> Result<(), Self::SinkError> {
+        self.start_send(msg)
     }
 
-    fn close(&mut self, _: &mut task::Context) -> Poll<(), SendError<T>> {
+    fn start_close(&mut self) -> Result<(), Self::SinkError> {
+        Ok(())
```
seanmonstar: This has confused me in the past. I would expect to be able to try to close a channel from the sender side, even if I have multiple senders. Should this instead try to close the channel, instead of being a noop?

futures-sink/src/channel_impls.rs
```diff
@@ -1,62 +1,66 @@
...
     }
 }
 
 impl<'a, T> Sink for &'a UnboundedSender<T> {
     type SinkItem = T;
-    type SinkError = SendError<T>;
+    type SinkError = ChannelClosed<T>;
+
+    fn poll_ready(&mut self, _: &mut task::Context) -> Poll<(), Self::SinkError> {
+        Ok(Async::Ready(()))
```
seanmonstar: Shouldn't this also check the inherent `poll_ready`? An unbounded channel can still be closed.

futures-sink/src/channel_impls.rs
cramertj: Can you open an issue for this? I'd prefer to keep this PR scoped to just refactoring changes, not behavioral ones.

futures-sink/src/channel_impls.rs
cramertj: If it's closed, it will throw an error in `start_send`. However, I think I agree with you that it would be good to allow the client to stop producing values in these cases.

cramertj: (note that to accomplish this, i'll be changing the inherent `poll_ready` from `&mut self` to `&self`)

cramertj: oh, it seems I can't actually do that, because `poll_ready` calls `poll_unparked` internally, which mutates `self.maybe_parked`.

seanmonstar: As a user, I find it surprising that passing `tx` would let `poll_ready` tell me it was closed, but `&tx` wouldn't.

cramertj: I don't disagree. Can you open an issue for this?

futures-sink/src/channel_impls.rs
aturon: I guess the key question to me is whether we can expect `start_send` to *always* be able to give you the value back on an error, in which case the `Sink` trait should probably continue to use a wrapper around the error type.

cc @carllerche

futures-sink/src/channel_impls.rs
aturon: So just to confirm, we're gonna tackle this in a follow-up PR?

futures-sink/src/lib.rs
```diff
@@ -21,75 +21,32 @@ macro_rules! if_std {
     )*)
 }
 
-use futures_core::Poll;
-use futures_core::task;
-
-/// The result of an asynchronous attempt to send a value to a sink.
-#[derive(Copy, Clone, Debug, PartialEq)]
-pub enum AsyncSink<T> {
```
aturon: ahhhhhh so nice

futures-sink/src/channel_impls.rs
cramertj: Yes, I think that's a good idea.

futures-sink/src/lib.rs
cramertj: \o/

cramertj:
@aturon 
> I guess the key question to me is whether we can expect start_send to always be able to give you the value back on an error, in which case the Sink trait should probably continue to use a wrapper around the error type.

@carllerche said in #751:

> IMO the current start_send API is pretty unusable in practice. The primary reason is that it is impossible to return the send item in all cases that are non trivial. The poll_ready strategy has been what I've been using and has been working quite well in practice.

aturon:
@cramertj ok, that's good enough for me :-)

aturon:
Merged!

carllerche:
@aturon @cramertj 

yes, it should definitely not be assumed that, once a value is sent, it can **always** be returned on error. `poll_ready` lets you have a pretty good idea that the value is accepted, however there is is always the possibility of a race. For example, poll_ready() returns `Ready`, the channel closes, then `start_send` is called. In this case, `start_send` will error.

Now, **some** `Sink` implementations are able to always guarantee that the value can be returned on error. In those cases, the implementations are free to include the send value.

```rust
enum SendError<T> {
    Closed(Option<T>), // Must be option to handle returning this on `poll_ready`.
    // other variants
}

impl<T> Sink for Sender<T> {
    type Item = T;
    type Error = SendError<T>;
}
```

This is probably roughly what @cramertj did (but I didn't look in detail).

The implementation can document that `T` will always be returned when the error is generated from `start_send`, but the main point is that the decision is left up to the implementation.

PR #768, opened 2018-02-15
Fix if condition in Sender::drop()

### LINKED ISSUE #766: sync::mpsc vs. unsync::mpsc
I have two futures running on the same tokio core.
The first future processes a stream and forwards the items into a `mpsc`.
The second future reads from that `mpsc`.
Everything works fine with `sync::mpsc`.
But when I switch to `unsync::mpsc` then the second future will apparently not know when the `Sender` side is closed and continue running.
No `unsafe` code at all.

Before I dig deeper and try to extract a minimal not-working example, I wanted to ask: Is there anything special about the `unsync::mpsc`?

dominikwerder:
The if clause was never reached because there was always at least one weak
reference remaining, which is our own.

Closes #766

### DIFF
--- a/src/unsync/mpsc.rs
+++ b/src/unsync/mpsc.rs
@@ -110,7 +110,11 @@ impl<T> Drop for Sender<T> {
             Some(shared) => shared,
             None => return,
         };
-        if Rc::weak_count(&shared) == 0 {
+        // The number of existing `Weak` indicates if we are possibly the last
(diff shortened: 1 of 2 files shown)


dominikwerder:
Travis CI fails on `master` currently anyway, also before this PR.

dominikwerder:
I'll push an extended test case later.

alexcrichton:
Awesome, thanks!

PR #769, opened 2018-02-15
futures_[un]ordered cleanup

imp:
I was experimenting recently with some variations of `futures_ordered` and while doing so cleaned it and its cousin `futures_unordered` a little bit. Thought I'd share it.

Each commit addresses single change, which should be pretty straightforward.


alexcrichton:
Thanks!

PR #772, opened 2018-02-15
Simplify JoinAll

matklad:
Hi!

It looks like `JoinAll` is overly complicated, b/c it is parametrized over `T: IntoFuture`, while a `F: Future` would do.


matklad:
BTW, this I think is technically a breaking change, so, if we want to move forward with this, we might want to do it before 0.2 is released?

PR #773, opened 2018-02-15
`.collect()` API as an alternative to `join_all`.

matklad:
Note: this also includes changes from https://github.com/rust-lang-nursery/futures-rs/pull/772.

Allow using collect to turn a vec of futures into a future of vec
 
In today's Rust, it is possible to go from `Vec<Option<T>>` to 
`Option<Vec<T>>` and from `Vec<Result<T, E>>` to `Result<Vec<T>, E>`:

    let xs = vec![Some(1), Some(2), Some(3)];
    let xs: Option<Vec<_>> = xs.into_iter().collect();
    assert_eq!(xs, Some(vec![1, 2, 3]));

    let xs = vec![Ok(1), Ok(2), Err(92)];
    let xs: Result<Vec<_>, _> = xs.into_iter().collect();
    assert_eq!(xs, Err(92));

These are the instances of a general phenomenon, which allows to pull 
the inner monad outside of the list; see [1] for theory behind this.

As futures are a monad, it makes sense to provide a similar api: a way 
to go from `Vec<Future<T, E>>` to `Future<Vec<T>, E>` via `collect`. 
However in Rust future is not a single type, but a trait, so we need to
implement `FromIterator` for some particular future, and it is `JoinAll`
in our case. Note that the user will have to provide `JoinAll` type
on call-site via ::<> or type ascription.    

[1]: https://hackage.haskell.org/package/base-4.10.1.0/docs/Data-Traversable.html#v:sequenceA

I ... am not actually sure we really *need* this: it just seems "proper" to have `.collect` working for futures, the same way it works for options and results.

### DIFF
--- a/futures-util/src/future/join_all.rs
+++ b/futures-util/src/future/join_all.rs
@@ -10,9 +10,9 @@ use futures_core::{Future, IntoFuture, Poll, Async};
 use futures_core::task;
 
 #[derive(Debug)]
-enum ElemState<T> where T: Future {
-    Pending(T),
-    Done(T::Item),
+enum ElemState<F> where F: Future {
+    Pending(F),
+    Done(F::Item),
 }
 
 /// A future which takes a list of futures and resolves with a vector of the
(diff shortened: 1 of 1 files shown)


futures-util/src/future/join_all.rs
```diff
@@ -137,3 +137,28 @@ impl<Item> Future for JoinAll<Item>
...
 }
+
+impl<F: IntoFuture> FromIterator<F> for JoinAll<F::Future> {
+    fn from_iter<T: IntoIterator<Item=F>>(iter: T) -> Self {
+        join_all(iter)
+    }
+}
+
+#[test]
+fn join_all_from_iter() {
```
matklad: This test should go to `/tests/all.rs` I think, but those tests are not tested at all in the 0.2 branch.

futures-util/src/future/join_all.rs
```diff
@@ -137,3 +137,28 @@ impl<Item> Future for JoinAll<Item>
         }
     }
 }
+
+impl<F: IntoFuture> FromIterator<F> for JoinAll<F::Future> {
```
cramertj: :heart:

alexcrichton:
Thanks!

alexcrichton:
Ah no worries! We're still working on getting 0.2 tests working again and once they're up and running a follow-up PR would be more than appreciated!

PR #777, opened 2018-02-19
Fix `unsync::mpsc::Sender`'s Drop impl.

goffrie:
This was testing `Rc::weak_count(&self.shared) == 0`, which is never the case
because `self.shared` still exists. This would cause a missed wakeup. Instead
test `weak_count == 1`.


alexcrichton:
Thanks! I think this was covered by https://github.com/rust-lang-nursery/futures-rs/pull/768 though so I'm going to close this

PR #779, opened 2018-02-20
Introduce top-level `spawn` combinators

aturon:
Adds a `spawn` helper to `Context`, and top-level `spawn` and `spawn_with_handle` functions, each of which return futures that complete a spawn.

The `_with_handle` variant gives you a "join handle" to the spawned task, allowing you to wait on its completion and retrieve its final value. The code was taken from the old `CpuPool` code and made executor-agnostic.

### DIFF
--- a/futures-executor/src/spawn.rs
+++ b/futures-executor/src/spawn.rs
@@ -1,0 +1,137 @@
+use futures_core::{Future, Async, Poll};
+use futures_core::never::Never;
+use futures_core::task::{self, Context};
+use futures_channel::oneshot::{channel, Sender, Receiver};
+use futures_util::FutureExt;
+
+use std::thread;
+use std::sync::Arc;
+use std::sync::atomic::Ordering;
+use std::panic::{self, AssertUnwindSafe};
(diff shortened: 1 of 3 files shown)


futures-executor/src/spawn.rs
```diff
@@ -0,0 +1,137 @@
...
+        Ok(Async::Ready(()))
+    }
+}
+
+/// dox
+#[derive(Debug)]
+pub struct SpawnWithHandle<F>(Option<F>);
+
+/// TODO: dox
+pub fn spawn_with_handle<F>(f: F) -> SpawnWithHandle<F>
```
cramertj: Can we also offer `spawn_with_handle` on `task::Context`?

aturon: Problem is, the implementation relies on a bunch of the other crates. We could of course offer an extension trait here and put it in the prelude, but didn't seem worth it right now.

aturon:
Merging for now -- we can follow up on a `Context`-based `with_handle` variant later if need be.

cramertj:
SGTM!

PR #780, opened 2018-02-20
Add futures-io

cramertj:
Currently, a few combinators have been disabled because they rely on `BufRead` capabilities. We should probably add an `AsyncBufRead` in order to bring back this functionality, but I have not yet done so.

I have also not yet ported the tests from the original repo. I plan to add these in a future commit.

### DIFF
--- a/futures-io/src/lib.rs
+++ b/futures-io/src/lib.rs
@@ -97,7 +97,7 @@ if_std! {
         /// If reading would block, this function returns `Ok(Async::Pending)`
         /// and arranges for `cx.waker()` to receive a notification when the
         /// object becomes readable or is closed.
-        fn poll_read(&mut self, buf: &mut [u8], cx: &mut task::Context)
+        fn poll_read(&mut self, cx: &mut task::Context, buf: &mut [u8])
             -> Poll<usize, Error>;
 
         /// Attempt to read from the `AsyncRead` into `vec` using vectored
@@ -113,11 +113,11 @@ if_std! {
         /// If reading would block, this function returns `Ok(Async::Pending)`
         /// and arranges for `cx.waker()` to receive a notification when the
         /// object becomes readable or is closed.
-        fn poll_vectored_read(&mut self, vec: &mut [&mut IoVec], cx: &mut task::Context)
+        fn poll_vectored_read(&mut self, cx: &mut task::Context, vec: &mut [&mut IoVec])
             -> Poll<usize, Error>
         {
-            if let Some(first_iovec) = vec.get_mut(0) {
-                self.poll_read(&mut *first_iovec, cx)
+            if let Some(ref mut first_iovec) = vec.get_mut(0) {
+                self.poll_read(cx, first_iovec)
             } else {
                 // `vec` is empty.
                 return Ok(Async::Ready(0));
@@ -134,7 +134,7 @@ if_std! {
         /// If writing would block, this function returns `Ok(Async::Pending)`
         /// and arranges for `cx.waker()` to receive a notification when the
         /// the object becomes writable or is closed.
-        fn poll_write(&mut self, buf: &[u8], cx: &mut task::Context)
+        fn poll_write(&mut self, cx: &mut task::Context, buf: &[u8])
             -> Poll<usize, Error>;
 
         /// Attempt to write bytes from `vec` into the object using vectored
@@ -150,11 +150,11 @@ if_std! {
         /// If writing would block, this function returns `Ok(Async::Pending)`
         /// and arranges for `cx.waker()` to receive a notification when the
         /// object becomes writable or is closed.
-        fn poll_vectored_write(&mut self, vec: &[&IoVec], cx: &mut task::Context)
+        fn poll_vectored_write(&mut self, cx: &mut task::Context, vec: &[&IoVec])
             -> Poll<usize, Error>
         {
-            if let Some(first_iovec) = vec.get(0) {
-                self.poll_write(&*first_iovec, cx)
+            if let Some(ref first_iovec) = vec.get(0) {
+                self.poll_write(cx, &*first_iovec)
             } else {
                 // `vec` is empty.
                 return Ok(Async::Ready(0));
@@ -187,16 +187,16 @@ if_std! {
                 (**self).initializer()
             }
 
-            fn poll_read(&mut self, buf: &mut [u8], cx: &mut task::Context)
+            fn poll_read(&mut self, cx: &mut task::Context, buf: &mut [u8])
                 -> Poll<usize, Error>
             {
-                (**self).poll_read(buf, cx)
+                (**self).poll_read(cx, buf)
             }
 
-            fn poll_vectored_read(&mut self, vec: &mut [&mut IoVec], cx: &mut task::Context)
+            fn poll_vectored_read(&mut self, cx: &mut task::Context, vec: &mut [&mut IoVec])
                 -> Poll<usize, Error>
             {
-                (**self).poll_vectored_read(vec, cx)
+                (**self).poll_vectored_read(cx, vec)
             }
         }
     }
@@ -217,7 +217,7 @@ if_std! {
                 Initializer::nop()
             }
 
-            fn poll_read(&mut self, buf: &mut [u8], _: &mut task::Context)
+            fn poll_read(&mut self, _: &mut task::Context, buf: &mut [u8])
                 -> Poll<usize, Error>
             {
                 Ok(Async::Ready(StdIo::Read::read(self, buf)?))
@@ -239,16 +239,16 @@ if_std! {
 
     macro_rules! deref_async_write {
         () => {
-            fn poll_write(&mut self, buf: &[u8], cx: &mut task::Context)
+            fn poll_write(&mut self, cx: &mut task::Context, buf: &[u8])
                 -> Poll<usize, Error>
             {
-                (**self).poll_write(buf, cx)
+                (**self).poll_write(cx, buf)
             }
 
-            fn poll_vectored_write(&mut self, vec: &[&IoVec], cx: &mut task::Context)
+            fn poll_vectored_write(&mut self, cx: &mut task::Context, vec: &[&IoVec])
                 -> Poll<usize, Error>
             {
-                (**self).poll_vectored_write(vec, cx)
+                (**self).poll_vectored_write(cx, vec)
             }
 
             fn poll_flush(&mut self, cx: &mut task::Context) -> Poll<(), Error> {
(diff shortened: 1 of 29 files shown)


futures-io/src/lib.rs
```diff
@@ -0,0 +1,305 @@
...
+        /// Returns a new `Initializer` which will not zero out buffers.
+        ///
+        /// # Safety
+        ///
+        /// This method may only be called by `AsyncRead`ers which guarantee
+        /// that they will not read from the buffers passed to `AsyncRead`
+        /// methods, and that the return value of the method accurately reflects
+        /// the number of bytes that have been written to the head of the buffer.
+        #[inline]
+        pub unsafe fn nop() -> Initializer {
```
seanmonstar: What is `nop`? Why not use a descriptive name, like the alternate constructor, `zeroing`?

futures-io/src/lib.rs
```diff
@@ -0,0 +1,305 @@
...
+        /// The default implementation returns an initializer which will zero
+        /// buffers.
+        ///
+        /// # Safety
+        ///
+        /// This method is `unsafe` because and `AsyncRead`er could otherwise
+        /// return a non-zeroing `Initializer` from another `AsyncRead` type
+        /// without an `unsafe` block.
+        #[inline]
+        unsafe fn initializer(&self) -> Initializer {
```
seanmonstar: I thought the point of the `Initializer` having an unsafe constructor was so that this method didn't need to be `unsafe`?

futures-util/src/io/codec.rs
```diff
@@ -0,0 +1,373 @@
+//! Utilities for encoding and decoding frames.
```
seanmonstar: All of the codec stuff seems out of scope.

futures-util/src/io/codecs.rs
```diff
@@ -0,0 +1,124 @@
+use bytes::{Bytes, BufMut, BytesMut};
```
seanmonstar: This codecs stuff also seems out of scope.

futures-util/src/io/framed.rs
```diff
@@ -0,0 +1,257 @@
+use std::io::{self, Read, Write};
```
seanmonstar: This framed/codec stuff seems out of scope.

futures-util/src/io/mod.rs
```diff
@@ -0,0 +1,150 @@
...
+    /// the number of bytes written.
+    fn write_buf<B: Buf>(&mut self, buf: &mut B, cx: &mut task::Context)
+        -> Poll<usize, std_io::Error>
+        where Self: Sized,
+    {
+        if !buf.has_remaining() {
+            return Ok(Async::Ready(0));
+        }
+
+        let n = try_ready!(self.poll_write(buf.bytes(), cx));
```
seanmonstar: Seems the default implementation here could make use of `self.poll_vectored_write`.

futures-io/src/lib.rs
```diff
@@ -0,0 +1,305 @@
...
+    pub use iovec::IoVec;
+
+    // Re-export io::Error so that users don't have to deal
+    // with conflicts when `use`ing `futures::io` and `std::io`.
+    pub use StdIo::Error as Error;
+
+    /// A type used to conditionally initialize buffers passed to `AsyncRead`
+    /// methods.
+    #[derive(Debug)]
+    pub struct Initializer(bool);
```
seanmonstar: Looking at the [std Read::initializer tracking issue](https://github.com/rust-lang/rust/issues/42788#issuecomment-325837435), it seems there is doubt that this is the correct API. In std, the `unstable` attribute protects them.

It seems odd to be using that API here if it's thought to be inferior, especially since we can't rely on `unstable` to protect against breaking changes.

futures-io/src/lib.rs
cramertj: This is done to match the current (unstable) std API. When that API is released, this `Iniitalizer` will be changed to a type alias to that one: https://doc.rust-lang.org/std/io/struct.Initializer.html

futures-io/src/lib.rs
cramertj: `Initializer::zeroing()` is safe, `Initializer::nop()` is `unsafe`. It is safe to implement this method using `Initializer::zeroing()`, but unsafe to implement it using `Initializer::nop()`.

This function is made `unsafe` so that users cannot implement it safely by delegating to the internally-unsafe implementation of the `initializer` function on another type.

All of this is copied from the std::io API. If you have comments, questions, or ideas about how to improve this design, please leave them [in the tracking issue](https://github.com/rust-lang/rust/issues/42788). If that API changes, we will update this one as necessary.

futures-io/src/lib.rs
cramertj: It's true that some people have doubts about the API. I personally think that it's the best-designed interface for this that I've seen, and it's the only one I've seen that is sound in the presence of https://github.com/rust-lang/rfcs/pull/2316.

carllerche:
For the record, Tokio will continue to provide it's own `AsyncRead` / `AsyncWrite` traits as well as all the utilities that were copied from the Tokio source.

I/O is fundamental to Tokio as such it needs to maintain its own traits (they will probably diverge some).

futures-util/src/io/mod.rs
cramertj: Done! Thanks for the reminder :)

aturon:
@carllerche I continue to see that as a rather unfortunate decision, since a basic goal here is to provide interoperation between different backing I/O systems, and basic implementation-agnostic conveniences -- exactly akin to `std::io`. I'm not aware of anything on a technical level that would prevent Tokio from using these traits. But I suspect it's fruitless to try to argue this point further.

In any case, the good news is that Rust provides ample means for "external" interoperation (through newtype wrappers and the like), which means the Tokio team's stance here will merely lead to increased friction, rather than total inability to interoperate.

futures-io/src/lib.rs
seanmonstar: Quoting from the 0.2 RFC:

> This resolves the issue of the `unsafe` on `prepare_uninitialized_buffer` being unsafe to implement, not unsafe to use (the more common, and arguably the correct meaning of `unsafe fn`).

The implementation doesn't seem to match, since even if `Initializer::zeroing()` is returned from `fn initializer()`, it doesn't matter, because it's unsafe to call `initializer()`.

I'm leaving comments here because this API is being proposed in futures, a library that I rely on, and a place where stability markers cannot protect the it. The tracking issue in std hasn't seen much traction otherwise.

futures-io/src/lib.rs
seanmonstar: The name `nop` seems unfortunate. Regardless of the previous naming choice of an unstable std API, it seems better to pick a better name in a new library.

carllerche:
@aturon Tokio already provides traits for interop that are decoupled from the reactor. I'm not sure what the value is in renaming the lib from `tokio-io` to `futures-io` and creating an additional standard instead of making the changes in the existing tokio-io crate, especially since it looks like you are copying the contents.

But yes, as you mentioned, one will always be able to write adapters.

futures-io/src/lib.rs
cramertj: This PR does indeed address the issue I raised in the text you quoted. However, I think the working I used may have been confusing.

Currently, `unsafe fn prepare_uninitialized_buffer` is `unsafe` to implement, but not `unsafe` to call. `unsafe` on functions is intended to mean "unsafe to call", not "unsafe to implement." If an RFC like https://github.com/rust-lang/rfcs/pull/2316 were to be approved, the `prepare_uninitialized_buffer` API would be unsound.

This PR fixes this problem by moving the unsafety to the call to `Initializer::nop()`. Even if https://github.com/rust-lang/rfcs/pull/2316 were approved, this API would be sound. The `initializer` function itself is still `unsafe`, but the reason is different: it isn't that `initializer` is `unsafe` to implement, but that *calling* `initializer` is `unsafe` because it could allow you to produce values of type `Initializer::nop()` even when your `AsyncRead` implementation read from the supplied buffer.

seanmonstar: > Currently, unsafe fn prepare_uninitialized_buffer is unsafe to implement, but not unsafe to call.

I assume you mean that is the intent, since it definitely *is* unsafe to call the function with the current compilers.

I find the reasoning that you could get an `Initializer::nop()` back a little bit arbitrary: you don't find this sort of `unsafe` "poisoning" elsewhere. For instance, you can get a string back with `str::from_utf8_unchecked()`, but that shouldn't mean that any function that happens to wrap that is inherently unsafe.

cramertj: Consider this implementation:

```rust
struct MyUnsafePerformantRead;

impl AsyncRead for MyUnsafePerformantRead {
    fn initializer(&self) -> Initializer {
        unsafe { Initializer::nop() }
    }
    fn poll_read.... // Doesn't read from the buffer
}

struct MySafeButStillUninitRead;

impl AsyncRead for MySafeButStillUninitRead {
    fn initializer&self) -> Initializer {
        MyUnsafePerformantRead.initializer()
    }
    fn poll_read.... // Does read from the buffer (uninitialized memory), causes UB
}
```

seanmonstar: Yes, I imagined that is the concern, but I don't see anything that adding an `unsafe` to the function will help with. In fact, adding the `unsafe` keywords doesn't protect it at all. A wrong implementation is still wrong.

cramertj: `unsafe` prevents the unsound implementation of `MySafeButStillUninitRead`. If `MyUnsafeButPerformantRead::initializer()` is unsafe, then, `MySafeButStillUninitRead::initializer` would have to use `unsafe` in order to trigger the UB.

seanmonstar: I'm even more unconvinced now, since you can easily grab an `Initializer` from a completely unrelated reader, and carry it around, and safely use it via `unrelated_initializer.initialize(my_buf)` everywhere else, and trigger the UB.

As a not-arbitrary example, consider `Chain` implementation takes `Box<AsyncRead>`s, and pushes them into a vec. The naive implementation calls `self.reads[0].initializer()`, and saves it, and then uses it in later reads, instead of remembering to ask for a new one.

I think the std API gives more an illusion of protection against UB, while still quite easily allowing it to happen, even by mistake.

cramertj: > since you can easily grab an Initializer from a completely unrelated reader, and carry it around

Not without using `unsafe`. If you're making `unsafe` calls and then randomly passing around the result, in explicit violation of the invariants stated in the docs, then you're right, it is easy to cause UB.

futures-io/src/lib.rs
```diff
@@ -0,0 +1,305 @@
...
+        }
+
+        /// Attempt to read from the `AsyncRead` into `buf`.
+        ///
+        /// On success, returns `Ok(Async::Ready(num_bytes_read))`.
+        ///
+        /// If reading would block, this function returns `Ok(Async::Pending)`
+        /// and arranges for `cx.waker()` to receive a notification when the
+        /// object becomes readable or is closed.
+        fn poll_read(&mut self, buf: &mut [u8], cx: &mut task::Context)
```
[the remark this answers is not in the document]
cramertj: Done.

PR #783, opened 2018-02-21
Sink refactor take 2

cramertj:
cc https://github.com/rust-lang-nursery/futures-rs/issues/751

In this case, `poll_close` should both flush *and* close the `Sink` before returning `Async::Ready`.

### DIFF
--- a/futures-util/src/io/framed_write.rs
+++ b/futures-util/src/io/framed_write.rs
@@ -37,7 +37,6 @@ pub struct FramedWrite<T, E> {
 pub struct FramedWrite2<T> {
     inner: T,
     buffer: BytesMut,
-    do_close: bool,
 }
 
 const INITIAL_CAPACITY: usize = 8 * 1024;
@@ -136,7 +135,6 @@ pub fn framed_write2<T>(inner: T) -> FramedWrite2<T> {
     FramedWrite2 {
         inner,
         buffer: BytesMut::with_capacity(INITIAL_CAPACITY),
-        do_close: false,
     }
 }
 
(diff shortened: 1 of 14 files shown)


futures-util/src/io/framed_write.rs
```diff
@@ -170,6 +167,33 @@ impl<T> FramedWrite2<T> {
...
+                return Err(io::Error::new(io::ErrorKind::WriteZero, "failed to
+                                          write frame to transport").into());
+            }
+
+            // TODO: Add a way to `bytes` to do this w/o returning the drained
+            // data.
+            let _ = self.buffer.split_to(n);
+        }
+
+        if do_close {
```
aturon: seems like this could just be inlined into the callers to get rid of the flag

futures-util/src/stream/split.rs
```diff
@@ -82,27 +78,31 @@ impl<S: Sink> Sink for SplitSink<S> {
...
                 inner.poll_flush(cx)
             }
             Async::Pending => Ok(Async::Pending),
         }
     }
+
+    fn poll_close(&mut self, cx: &mut task::Context) -> Poll<(), S::SinkError> {
+        match self.lock.poll_lock(cx) {
+            Async::Ready(mut inner) => {
+                if self.slot.is_some() {
```
aturon: tiny thing, but this can be rewritten as `if let Some(val) = self.slot.take()`

cramertj: Unfortunately not-- the try_ready! may cause an early return, in which case we don't want to remove the item from the slot.

futures-util/src/io/framed_write.rs
cramertj: Done

PR #784, opened 2018-02-21
Simplify task-local access

cramertj:
Here's one possible option for simplifying task-local data.

I'm also curious what people think about the idea of making `LocalMap` into a trait so that users can provide their own implementations, especially for no_std contexts. However, it would increase the complexity of the API (add to that that we'd probably need fallible entries to support no_std). WDYT?


aturon:
This PR seems great! I'd prefer not to explore a trait at this point.

PR #786, opened 2018-02-21
Roll back mentions of traits in intra-rustdoc links to avoid ICE

aturon:
Work around https://github.com/rust-lang/rust/issues/48414


aturon:
Merge once travis is happy.

PR #787, opened 2018-02-22
Recover, infallible spawn

Ralith:
I'd appreciate feedback on the details of `recover` before I fill in the documentation.


futures-executor/src/thread_pool.rs
```diff
@@ -315,8 +316,8 @@ impl Task {
                 };
                 match res {
                     Ok(Async::Pending) => {}
-                    Ok(Async::Ready(())) |
-                    Err(()) => return wake_handle.mutex.complete(),
+                    Ok(Async::Ready(())) => return wake_handle.mutex.complete(),
+                    Err(_) => unreachable!(), // FIXME: This case shouldn't be needed at all!
```
Ralith: I was under the impression that rust no longer required patterns containing uninhabited types to be matched, but it seems like that isn't the case yet?

srijs: Have you tried `Err(never) => match never {}` here?

futures-util/src/future/recover.rs
```diff
@@ -0,0 +1,33 @@
...
+impl<A, B, F> Future for Recover<A, B, F>
+    where A: Future,
+          F: FnOnce(A::Error) -> A::Item,
+{
+    type Item = A::Item;
+    type Error = B;
+
+    fn poll(&mut self, cx: &mut task::Context) -> Poll<A::Item, B> {
+        match self.inner.poll(cx) {
+            Err(e) => Ok(Async::Ready((self.f.take().unwrap())(e))),
```
cramertj: Nit: `except("Polled future::Recover after completion")`?

futures-util/src/future/recover.rs
```diff
@@ -0,0 +1,33 @@
+use std::marker::PhantomData;
+
+use futures_core::{Future, Poll, Async};
+use futures_core::task;
+
+#[must_use = "futures do nothing unless polled"]
+#[derive(Debug)]
+pub struct Recover<A, B, F> {
```
cramertj: Nit: `E` would probably be clearer than `B` as to the intent.

futures-executor/src/thread_pool.rs
cramertj: @Ralith That change got rolled back and is being separately feature gated from the `!` type due to backcompat concerns. IMO it'll probably still go through at some point, but it isn't available now. For the time being, I'd use `match never {}` as @srijs suggested.

cramertj:
LGTM so far-- you'll also want to add a version for `Stream` (where the recovery function will need to be `FnMut`, not just `FnOnce`).

aturon:
LGTM!

Ralith:
I chose to make the a `Stream` version return `Option<Self::Item>` to provide the option to terminate the stream on error.

Are there any weird variance concerns with the specific use of `PhantomData`?

cramertj:
Nope, standard implied variance (covariance) is fine.

futures-util/src/stream/recover.rs
```diff
@@ -0,0 +1,35 @@
+use std::marker::PhantomData;
+
+use futures_core::{Stream, Poll, Async};
+use futures_core::task;
+
+/// Future for the `recover` combinator, handling errors by converting them into
+/// an `Option<Item>`, compatible with any error type of the caller's choosing.
```
cramertj: Can you add more documentation around the `None` behavior? Specifically, that `None` causes the stream to end, rather than just skipping to the next element.

futures-util/src/stream/recover.rs
```diff
@@ -0,0 +1,35 @@
...
+#[must_use = "futures do nothing unless polled"]
+#[derive(Debug)]
+pub struct Recover<A, F, E> {
+    inner: A,
+    f: Option<F>,
```
cramertj: You can just use `F` here (rather than `Option<F>`, since the closure is `FnMut`, and we want to be able to `recover` from multiple errors.

futures-util/src/stream/recover.rs
```diff
@@ -0,0 +1,35 @@
...
+impl<A, F, E> Stream for Recover<A, F, E>
+    where A: Stream,
+          F: FnMut(A::Error) -> Option<A::Item>,
+{
+    type Item = A::Item;
+    type Error = E;
+
+    fn poll_next(&mut self, cx: &mut task::Context) -> Poll<Option<A::Item>, E> {
+        match self.inner.poll_next(cx) {
+            Err(e) => Ok(Async::Ready((self.f.take().expect("Polled stream::Recover after completion"))(e))),
```
cramertj: The `take` here will cause the stream to panic if multiple errors occur. Changing from `Option<F> ` to `F` as I suggested above will fix this.

Ralith: Whoops, good catch!

cramertj:
LGTM once travis is happy. [Right now it's complaining](https://travis-ci.org/rust-lang-nursery/futures-rs/jobs/345781074) because some of your imports are only used when "std" is enabled. Can you `cfg` the imports?

cramertj:
@Ralith looks like you still need to use `core` in some places, rather than `std`.

PR #789, opened 2018-02-22
Util refactor

cramertj:


futures-util/src/io/mod.rs
```diff
@@ -114,6 +136,48 @@ pub trait AsyncReadExt: AsyncRead {
         framed::framed(self, codec)
     }
 
+    /// Tries to read some bytes directly into the given `buf` in asynchronous
+    /// manner, returning a future type.
+    ///
+    /// The returned future will resolve to both the I/O stream and the buffer
+    /// as well as the number of bytes read once the read operation is completed.
+    fn read<T>(self, buf: T) -> Read<Self, T>
```
seanmonstar: I haven't thought this all the way through, but with the generators working with borrows, could these methods eventually exist as mini futures that borrow self instead of consuming and returning self?

So, there's `poll_read`, and with await support, a `read` method could return a future borrowing self. This would allow this pattern, I think:

```rust
let n = await io.read(ctx, buf)?;
await io.write_all(ctx, &but[..n])?;
// bonus if await can insert context automatically
```

If so, then all these methods would conflict with that...

cramertj: Yeah, I'd love to see these borrow self in the future rather than moving everything back and forth! Current implementations (and all future plans I've heard of) for `await` drive an existing `Future` type to completion, yielding on `Async::Pending`. This means that the `read` function would still return a `Read` future, but rather than taking `self` by value, it'd bound the future's lifetime to the lifetime of the mutable reference:

```rust
fn read<T>(&'a mut self, buf: T) -> Read<'a, T>
// or perhaps `-> Read<'a, Self, T>` if you want monomorphization
```

As far as having the existing methods conflict with the future versions, there are a couple reasons I don't think that will be an issue:
1. These combinators have been moved to `futures-util` specifically because they're intended to be less stable and evolve faster than the core crate. By keeping them versioned separately from the `futures-core`/`futures-io`/`futures-sink` traits, we can release breaking changes to these combinators without breaking the public APIs of async libraries in the ecosystem.
2. The eventual plans for final async/await will probably involve adding a `Pin` self-type to the `Future::poll` and `Stream::poll_next` functions. This is a breaking change, and will require us to  release an 0.3, at which time we can make these transitions as necessary. There will probably be a lot of these sorts of transitions around changing from owned to borrowed `Future`s.

Does that seem reasonable to you?

futures-util/src/io/mod.rs
```diff
@@ -91,6 +96,23 @@ pub trait AsyncReadExt: AsyncRead {
...
+    ///
+    /// The returned future will copy all the bytes read from this `AsyncRead` into the
+    /// `writer` specified. This future will only complete once the `reader` has hit
+    /// EOF and all bytes have been written to and flushed from the `writer`
+    /// provided.
+    ///
+    /// On success the number of bytes is returned and this `AsyncRead` and `writer` are
+    /// consumed. On error the error is returned and the I/O objects are consumed as
+    /// well.
+    fn copy<W>(self, writer: W) -> Copy<Self, W>
```
aturon: `copy_into`?

futures-util/src/io/mod.rs
aturon: I'm fine with the plan here.

seanmonstar: Ya, I thought something like that works become available, and all these `poll_*` methods everywhere could grow sister methods without the `poll_` prefix:

```rust
fn ready(&mut self) -> impl Future<...> {
    poll_fn(|| self.poll_ready())
}
```

It's my personal preference to not set up users to have to make breaking changes later, and being free functions seems like an easy compromise.

futures-util/src/io/mod.rs
cramertj: Done.

PR #791, opened 2018-02-22
Update futures-sink docs

cramertj:


futures-sink/src/lib.rs
```diff
@@ -114,101 +114,70 @@ pub trait Sink {
...
     /// As the name suggests, this method only *begins* the process of sending
     /// the item. If the sink employs buffering, the item isn't fully processed
     /// until the buffer is fully flushed. Since sinks are designed to work with
     /// asynchronous I/O, the process of actually writing out the data to an
-    /// underlying object takes place asynchronously. **You *must* use
-    /// `flush` in order to drive completion of a send**. In particular,
-    /// `start_send` does not begin the flushing process
-    ///
-    /// # Return value
+    /// underlying object takes place asynchronously.**You *must* use
```
aturon: Missing a space: `.**You *must*`

futures-sink/src/lib.rs
```diff
@@ -114,101 +114,70 @@ pub trait Sink {
...
+    /// `poll_flush` or `poll_close` in order to guarantee completion of a
+    /// send**.
     ///
     /// This method returns `AsyncSink::Ready` if the sink was able to start
     /// sending `item`. In that case, you *must* ensure that you call
```
aturon: This is feeling a bit repetitive, since you just said (and highlighted!) the flushing requirement. Also this only talks about flush rather than also including close as above.

futures-sink/src/lib.rs
```diff
@@ -114,101 +114,70 @@ pub trait Sink {
...
-    /// `flush` to process the sent item to completion. Note, however,
+    /// `poll_flush` to process the sent item to completion. Note, however,
     /// that several calls to `start_send` can be made prior to calling
```
aturon: I think this is missing a subtle, but important point: a lot of the time, the right way to use a sink is just to continually send into it, *which will often involve flushing behind the scenes*. In practice it works much like buffered writing in the synchronous world. 

The main point is that you need to:

- Call `poll_flush` to completion if you need to ensure data is actually sent at a given time.
- Call `poll_close` to completion before dropping the sink.

futures-sink/src/lib.rs
```diff
@@ -114,101 +114,70 @@ pub trait Sink {
...
+    /// case the current task is scheduled (via `cx.waker()`) to wake up when
+    /// `poll_flush` should be called again.
     ///
-    /// - It is called outside of the context of a task.
-    /// - A previous call to `start_send` or `flush` yielded an error.
+    /// In most cases, if the sink encounters an error, the sink will
+    /// permanently be unable to receive items.
+    fn poll_flush(&mut self, cx: &mut task::Context) -> Poll<(), Self::SinkError>;
+
+    /// Flush any remaining output and close this sink, if necessary.
```
aturon: the "if necessary" here is a bit confusing

futures-sink/src/lib.rs
cramertj: Done

futures-sink/src/lib.rs
cramertj: Done

futures-sink/src/lib.rs
cramertj: Added some text about this.

futures-sink/src/lib.rs
cramertj: Done.

PR #792, opened 2018-02-23
Add `run` to `ThreadPool`

aturon:
Also simplifies `ThreadPool` construction given that we have a builder.

This makes simple app setup look like: `ThreadPool::new().run(my_app)`


cramertj:
LGTM! :)

PR #794, opened 2018-02-23
Add docs and otherwise clean up the facade crate

aturon:
Note that I expanded the glob re-exports because otherwise rustdoc will not actually show the contents of the modules.

### DIFF
--- a/futures/src/lib.rs
+++ b/futures/src/lib.rs
@@ -103,8 +103,8 @@ pub mod channel {
 pub mod executor {
     //! Task execution.
     //!
-    //! All asynchronous computation occurs within an *executor*, which is
-    //! capable of "spawning" futures as *tasks*. This module provides several
+    //! All asynchronous computation occurs within an executor, which is
+    //! capable of spawning futures as tasks. This module provides several
     //! built-in executors, as well as tools for building your own.
     //!
     //! # Using a thread pool (M:N task scheduling)
@@ -136,7 +136,8 @@ pub mod executor {
     //! There are two ways to spawn a task:
     //!
     //! - Spawn onto a "default" execuctor by calling the top-level
-    //! [`spawn`](::executor::spawn) function.
+    //! [`spawn`](::executor::spawn) function or [pulling the executor from the
+    //! task context](::task::Context::executor).
     //!
     //! - Spawn onto a specific executor by calling its
     //! [`spawn`](::executor::Executor::spawn) method directly.
@@ -179,7 +180,7 @@ pub mod future {
     //! - The [`FutureExt`](::future::FutureExt) trait, which provides adapters
     //! for chaining and composing futures.
     //!
-    //! - Top-level future contructors like [`lazy`](::future::lazy) which
+    //! - Top-level future combinators like [`lazy`](::future::lazy) which
     //! creates a future from a closure that defines its return value, and
     //! [`result`](::future::result), which constructs a future with an
     //! immediate defined value.
@@ -188,7 +189,7 @@ pub mod future {
         Result, Future, FutureFrom, IntoFuture, err, ok, result
     };
     pub use futures_util::future::{
-        AndThen, Empty, Flatten, FlattenStream, FromErr, Fuse,
+        AndThen, Empty, Flatten, FlattenStream, ErrInto, Fuse,
         Inspect, IntoStream, Join, Join3, Join4, Join5, Lazy, LoopFn,
         Map, MapErr, OrElse, PollFn, Select, Then, Either, Loop, FutureExt, empty,
         lazy, loop_fn, poll_fn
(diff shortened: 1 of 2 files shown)


aturon:
Also, these docs will eventually want to have more specific pointers into the book throughout, but I'd like to do that in a separate pass once the book has more content.

futures/src/lib.rs
```diff
@@ -1,164 +1,28 @@
...
+//! - [Futures](::Future) (sometimes called promises), which represent a single
+//! asychronous computation that may result in a final value or an error.
 //!
-//! You can find extensive tutorials and documentations at [https://tokio.rs]
-//! for both this crate (asynchronous programming in general) as well as the
-//! Tokio stack to perform async I/O with.
+//! - [Streams](::Stream), which represent a series of values or errors produced asynchronously.
 //!
-//! [https://tokio.rs]: https://tokio.rs
+//! - [Sinks](::Sink), which represent asynchronous writing of data.
```
cramertj: "represent asynchronous writing of data" doesn't sound quite right to me.
"which represent asynchonous pipes into which data can be written" is too wordy, but maybe something along those lines?

futures/src/lib.rs
```diff
@@ -1,164 +1,28 @@
...
+//! [cross-task communication](::channel).
 //!
-//! ```toml
-//! [dependencies]
-//! futures = "0.1"
-//! ```
+//! Underlying all of this is the *task system*, which is a form of lightweight
+//! threading. Large asynchronous computations are built up using futures,
+//! streams and sinks, and then spawned as independent, asynchronous tasks which
+//! run to completion, but *do not block* the thread running them.
```
cramertj: [are] run to completion

futures/src/lib.rs
```diff
@@ -1,164 +1,28 @@
...
-//! # fn main() {}
-//! ```
-//!
-//! Some more information can also be found in the [README] for now, but
-//! otherwise feel free to jump in to the docs below!
-//!
-//! [README]: https://github.com/alexcrichton/futures-rs#futures-rs
+//! **The best way to learn about this crate is through the [Asynchronous
+//! Programming in Rust](https://aturon.github.io/apr/) book**, which provides a
+//! comprehensive introduction. The docs within this crate, by contract, are
```
cramertj: "by contract" seems unnecessary

futures/src/lib.rs
```diff
@@ -192,62 +56,185 @@ macro_rules! try_ready {
     })
 }
 
-// TODO: task_local macro
+/// A macro to create a `static` of type `LocalKey`
+///
+/// This macro is intentionally similar to the `thread_local!`, and creates a
+/// `static` which has a `with` method to access the data on a task.
```
cramertj: `with` was changed to [`get_mut`](https://github.com/rust-lang-nursery/futures-rs/pull/784/files#diff-48e5eeeacb6dc31a3e5b5ed5134314caR98).

futures/src/lib.rs
aturon: 😆 that was supposed to be "by *contrast*"

cramertj: lol :)

aturon:
Updated with new commit.

futures/src/lib.rs
```diff
@@ -192,62 +56,185 @@ macro_rules! try_ready {
...
+
+    pub use futures_channel::{oneshot, mpsc};
 }
 
+#[cfg(feature = "std")]
 pub mod executor {
-    //! Executors
+    //! Task execution.
+    //!
+    //! All asynchronous computation occurs within an *executor*, which is
```
cramertj: Lots of *italics* and "scare quotes" here. :woman_shrugging:

futures/src/lib.rs
```diff
@@ -192,62 +56,185 @@ macro_rules! try_ready {
...
+    //! The call to [`run`](::executor::ThreadPool::run) will block the current
+    //! thread until the future defined by `my_app` completes, and will return
+    //! the result of that future.
+    //!
+    //! # Spawning additional tasks
+    //!
+    //! There are two ways to spawn a task:
+    //!
+    //! - Spawn onto a "default" execuctor by calling the top-level
+    //! [`spawn`](::executor::spawn) function.
```
cramertj: Possibly worth mentioning `cx.spawn(fut)` can be used when implementing futures manually. (maybe this belongs in the docs of `spawn`, though, rather than these docs?)

futures/src/lib.rs
```diff
@@ -192,62 +56,185 @@ macro_rules! try_ready {
...
+    //!
+    //! This module contains:
     //!
-    //! This module contains the `Future` trait and adapters for this trait.
+    //! - The [`Future` trait](::Future).
+    //!
+    //! - The [`FutureExt`](::future::FutureExt) trait, which provides adapters
+    //! for chaining and composing futures.
+    //!
+    //! - Top-level future contructors like [`lazy`](::future::lazy) which
```
cramertj: s/contructors/constructors

We use the term "combinators" elsewhere

futures/src/lib.rs
```diff
@@ -192,62 +56,185 @@ macro_rules! try_ready {
...
-    //! as a number of combinators and extensions for using them.
-    pub use futures_io::*;
-    pub use futures_util::io::*;
+    // TODO: document utility functions once the re-org is done
+
+    pub use futures_io::{
+        Error, Initializer, IoVec, ErrorKind, AsyncRead, AsyncWrite, Result
+    };
+    pub use futures_util::io::{
+        codec, io, AsyncReadExt, AsyncWriteExt
```
cramertj: I think this nested `io::io` module should be gone now

aturon:
Updated with all feedback addressed.

cramertj:
:shipit:

PR #797, opened 2018-02-24
Document the -executor crate

aturon:


futures-executor/src/enter.rs
```diff
@@ -22,11 +22,13 @@ pub struct EnterError {
 /// executor.
 ///
 /// Executor implementations should call this function before blocking the
-/// thread.
+/// thread. Doing so ensures that executors aren't accidentally invoked in a
```
cramertj: "before blocking the thread [for the first time]."

cramertj: Also, it would be good to include a comment about keeping the `Enter` result around until you're *done* blocking the task.

futures-executor/src/enter.rs
```diff
@@ -21,12 +21,14 @@ pub struct EnterError {
 /// Marks the current thread as being within the dynamic extent of an
 /// executor.
 ///
-/// Executor implementations should call this function before blocking the
-/// thread.
+/// Executor implementations should call this function before beginning to
+/// execute any tasks. Doing so ensures that executors aren't accidentally
+/// invoked in a nested fashion.
 ///
 /// # Error
```
cramertj: It would be good to include a comment about keeping the `Enter` result around until you're *done* blocking the task, and not holding onto it after you've finished blocking.

cramertj:
LGTM aside from the `Enter` questions. IMO the thing that would help the most there is an example showing `let enter = enter().expect("..."); /* run task */ drop(enter);`.

aturon:
Rebased and comments addressed; ready to merge when Travis is happy.

PR #799, opened 2018-02-24
Use `either` crate

Ralith:
I went in planning to remove the internal `Either` type entirely, but obviously we can't implement `Future` for an external type in an external crate. There's a case to be made for removing the internal `Either` type and instead submitting a (feature-guarded?) `Future` impl to the external `Either` crate, but I'm not sure how I feel about that.

I'm also not sure how I feel about the small change here that remains if we're not removing the internal Either`; let's discuss.


Ralith:
For one thing, the `Either` `Future` impl will be considerably less important once async/await is readily usable, since you'll be able to write branches naturally.

cramertj:
We actually can implement `Future` for the `Either` type in a separate crate-- the impl just has to be inside `futures-core`. I think I'd be +1 for this, but it should be under an `either` feature flag, and we should re-export the type in `futures-util` so that users don't have to manually depend on the `either` crate. WDYT?

Ralith:
I suppose I don't see any reason not to put the impl in `-core`. Would the feature flag be disabled by default?

cramertj:
@Ralith Hm... I think i'd probably disable by default inside futures-core, but make futures-util enable it by default. This way we keep futures-core dependencies as minimal as possible.

futures-util/Cargo.toml
```diff
@@ -12,7 +12,7 @@ Common utilities and extension traits for the futures-rs library.
 
 [features]
 std = ["futures-core/std", "futures-io/std", "futures-sink/std", "either/use_std"]
-default = ["std"]
+default = ["std", "futures-core/either"]
```
Ralith: Is this sufficient, or does the `futures-core` feature need to be handled in other ways?

cramertj: Seems fine to me!

futures-util/src/future/mod.rs
```diff
@@ -46,9 +45,10 @@ pub use self::err_into::ErrInto;
 pub use self::or_else::OrElse;
 pub use self::select::Select;
 pub use self::then::Then;
-pub use self::either::Either;
 pub use self::inspect::Inspect;
 
+pub use either::Either;
```
cramertj: Can you also re-export from the `futures` facade crate? Once that's done, I think we can ship this.

futures-util/src/lib.rs
```diff
@@ -9,6 +9,7 @@
 extern crate futures_core;
 extern crate futures_io;
 extern crate futures_sink;
+extern crate either;
```
kpp: `#[cfg(feature = "std")]`?

Ralith: Shouldn't be needed; I'm already carefully dispatching to `either`'s own `std` feature.

futures-util/src/future/mod.rs
Ralith: Isn't there [a pre-existing export](https://github.com/rust-lang-nursery/futures-rs/blob/master/futures/src/lib.rs#L189)?

futures-util/src/future/either.rs
```diff
@@ -1,55 +0,0 @@
...
-    /// Second branch of the type
-    B(B),
-}
-
-impl<T, A, B> Either<(T, A), (T, B)> {
-    /// Splits out the homogeneous type from an either of tuples.
-    ///
-    /// This method is typically useful when combined with the `Future::select2`
-    /// combinator.
-    pub fn split(self) -> (T, Either<A, B>) {
```
mathstuf: See #580 for modifications related to this. I don't know if `either` is inclined to add something similar, but if not, the method is useful enough that free-standing functions may still be useful.

futures-util/src/future/mod.rs
cramertj: Oh, you're right! Sorry, I that didn't occur to me.

cramertj:
Rebased and reopened https://github.com/rust-lang-nursery/futures-rs/pull/828 targeting master.

PR #804, opened 2018-02-26
Futures util docs update

cramertj:


aturon:
LGTM -- I assume you also checked over the docs on the extension trait methods?

cramertj:
Yep!

PR #805, opened 2018-02-26
Reexport future::Option

cramertj:


aturon:
I feel queasy about the name clash with `std` but that's a separate concern from this PR.

r+ after rebase

cramertj:
@aturon we do the same thing for `future::Result`, which is actually `future::FutureResult` in `0.1` but was changed in the transition to `0.2`. Do you have a preferred naming scheme?

aturon:
I generally disprefer shadowing names from `std`, especially ones that are in the prelude. `FutureResult` and `FutureOption` seem good to me...

cramertj:
Haha. The reasoning for the name change is that (I think) most people `use futures::future;` and then refer to the types like `future::Result`, in which case `future::FutureResult` causes stuttering.

aturon:
yeah that's usually how this kind of thing goes. I've personally come to prefer the stuttering to the confusion that can otherwise arise -- but only for prelude-level types.

carllerche:
I personally rarely (if ever?) refer to future types scoped by the module. Only the free fns.

cramertj:
Okay! i'll send a PR to change both.

cramertj:
see https://github.com/rust-lang-nursery/futures-rs/pull/807

PR #808, opened 2018-02-26
Remove BoxSink alias

cramertj:
cc @alexcrichton


peku33:
Why exactly BoxSink was removed?

ernestask:
I’m four days early, but why exactly was `BoxSink` removed?

PR #809, opened 2018-02-26
Touch up futures-io docs

aturon:


futures-io/src/lib.rs
```diff
@@ -35,7 +39,7 @@ if_std! {
     pub use StdIo::Result as Result;
 
     /// A type used to conditionally initialize buffers passed to `AsyncRead`
-    /// methods.
+    /// methods, modeled after `std`.
```
cramertj: Nit: seems weird to say that it's modeled after `std` when the `std` feature is unstable. When it becomes stable, I'd expect we'd just reuse it here.

PR #811, opened 2018-02-27
Add WakeArc trait

cramertj:
I encountered the need for this when attempting to transition the Fuchsia executor over to the new `Wake` trait. In order to place an `Arc<Task>` onto a queue in the impl of `wake` for `Task`, I needed access to the `Arc`, not just the `Task` inside. I could implement `Wake` for `MyWrapper(Arc<Task>)`, but then I have to use `Arc<MyWrapper>`, which requires double boxing. After popping the `Arc<Task>` off of the task queue, I need to create an `Arc<MyWrapper>` again in order to poll the underlying future, which means I have to re-`Arc` it (causing an allocation every time). If this sounds like a mess, that's because it is. This trait would make it much easier :)

### DIFF
--- a/futures-core/src/task/mod.rs
+++ b/futures-core/src/task/mod.rs
@@ -143,7 +143,10 @@ impl fmt::Debug for LocalMap {
 /// A way of waking up a specific task.
 ///
 /// Any task executor must provide a way of signaling that a task it owns
-/// is ready to be `poll`ed again. Executors do so by implementing this trait.
+/// is ready to be `poll`ed again. Executors do so by implementing this trait
+/// or the `WakeArc` trait. Types which implement `Wake` will automatically
+/// receive implementaitons of `WakeArc`, so `Wake` impls should be
+/// preferred where possible.
 ///
 /// Note that, rather than working directly with `Wake` trait objects, this
 /// library instead uses a custom [`Waker`](::task::Waker) to allow for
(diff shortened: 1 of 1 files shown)


carllerche:
Is this not what `UnsafeWake` is for? Basically, if you want to avoid double boxing, you can drop down a level and use the unsafe trait which gives you full flexibility to do whatever.

cramertj:
Sure, I could manually implement `UnsafeWake`, but it seems like a waste of a perfectly good safe abstraction to copy-paste around a bunch of `unsafe` code. I anticipate that many executor implementations will want this functionality, so why not offer it via a safe interface? (especially considering how convoluted the `unsafe` trickery has to be in order to achieve the same goal)

aturon:
I agree that it would be good to provide functionality like this, but I would strongly prefer not to add *another* trait to the mix (which is already complicated).

I'm wondering whether we could instead do the following:

- Sever the inheritance relationship between `UnsafeWake` and `Wake`, so that the former provides its own `wake` function.
- Make `Wake` look more like `WakeArc` in this PR -- tie it more to `Arc`.

Basically, make the "happy path" fully `Arc`-based, and then use `UnsafeWake` for everything else.

wdyt?

cramertj:
@aturon That sounds good to me! i'll close this and open another PR.

carllerche:
@aturon I agree w/ this stratregy. IMO UnsafeWake should not be related to the other trait. This allows for additional flexibility (for example, maybe bringing back the `id` component 😉 )

PR #812, opened 2018-02-27
Change Wake trait to take an Arc (now std-only)

cramertj:
Continuation of #811

cc @carllerche


carllerche:
If this works, seems good to me. I personally have never used `Wake` and have always favored the unsafe version. Maybe this will help make the safe one more useful.

alexcrichton:
r=me, but looks like CI is failing?

aturon:
:+1:

leodasvacas:
I thought `impl<T: Wake> From<&'static T> for Waker` was cute but I guess nobody used it.

PR #813, opened 2018-02-27
Register Waker with AtomicWaker instead of context

carllerche:
This patch changes the AtomicWaker::register API to take a `Waker`
instance instead of `&mut Context`.

The existing implementation only needed the context to get a waker
handle. Switching the API to take the waker explicitly increases the
flexibility of `AtomicWaker`. There are cases where one wishes to
register a waker that is not associated with the current context.

### DIFF
--- a/futures-core/src/task/atomic_waker.rs
+++ b/futures-core/src/task/atomic_waker.rs
@@ -5,7 +5,7 @@ use core::cell::UnsafeCell;
 use core::sync::atomic::AtomicUsize;
 use core::sync::atomic::Ordering::{Acquire, Release};
 
-use task::{self, Waker};
+use task::Waker;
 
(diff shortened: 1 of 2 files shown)


aturon:
Rebased cleanly; awaiting Travis.

cramertj:
Similarly, should `LocalKey::get_mut` take an `&mut LocalMap`? I had been using `cx` purely for ease-of-use, but if we're going for maximum flexibility it's technically unnecessary.

aturon:
@cramertj No, that would regress ergonomics and I think doesn't have a clear-cut use case. Whereas having a `Waker` without a corresponding `Context` definitely happens.

cramertj:
Okay! That seems fine to me.

cramertj:
@carllerche can you reopen this PR targeting master?

PR #814, opened 2018-02-28
Remove references to APR book

aturon:
I've mostly updated the book, but also realized that I want to make some substantial changes to the early presentation, and figured better to just remove it for now and unblock 0.2.


cramertj:
LGTM-- I'm sad that people won't get to read APR during their initial investigations, though.

PR #820, opened 2018-02-28
Remove erroneous backtick from documentation

PigeonF:
Before: `Err(err)`` if the future is finished but resolved to an error `err`.

After: `Err(err)` if the future is finished but resolved to an error `err`.

I hope merging into master is okay, if not, just tell me.


alexcrichton:
Thanks!

PR #829, opened 2018-03-02
Add Either combinators on FutureExt and StreamExt

cramertj:


futures-util/src/future/either.rs
```diff
@@ -1,58 +0,0 @@
-use futures_core::{Future, Poll, Stream};
```
cramertj: This should've been deleted in an older PR, but it got left around accidentally.

PR #832, opened 2018-03-02
change mpsc::Receiver's error to Never

seanmonstar:
This matches `oneshot::Receiver`, making it clearer that you don't need to handle the error case with mpsc channels.


cramertj:
Thanks!

PR #833, opened 2018-03-02
add Receiver::try_next method

seanmonstar:
Adds `try_next`, allowing popping values out of the channel when not in the context of a futures task. This is especially useful in combination with `Receiver::close()`.

My use case: I have a channel type that wraps `mpsc`'s channel. When the receiver drops, I need to drain it and signal all untouched values with a special error message. That'd happen in my `Drop` impl, where there is not `task::Context` available.

### DIFF
--- a/futures-channel/src/mpsc/mod.rs
+++ b/futures-channel/src/mpsc/mod.rs
@@ -151,6 +151,11 @@ enum SendErrorKind {
     Disconnected,
 }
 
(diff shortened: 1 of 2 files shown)


cramertj:
LGTM! Ping me when you've had a chance to rebase and I'll merge.

seanmonstar:
Rebased.

PR #834, opened 2018-03-02
split up the purpose for try_send vs start_send

seanmonstar:
- `try_send` will return the message back if there is an error.
- `start_send` will consume the message always.
- Change `TryChannelClosed` to `TrySendError`, since the former sounds
  like there was an error trying to close the channel.
- Change `ChannelClosed` to `SendError`, so that it may also notify in
  `start_send` if the queue is full.

If the user called `start_send` when the queue was full, they could
receive an error that the channel was disconnected instead, which is
misleading.

Additionally, when `ChannelClosed<T>` would have a `T` and when it
wasn't was not very clear, especially when it was the error value for
`poll_ready`, `poll_flush`, and `poll_close`.

Finally, there was bug that `start_send` would always push on the queue
even if the requested buffer size was met, essentially making all
channels unbounded unless you used `try_send`.

-----

Currently there is no real difference between `try_send` and `start_send`, other than they return different error types, one that always has the message, and one that sometimes does. With this new separation, it is clearer when to use which: if you want to try to get the value back, use `try_send`. Otherwise, use `start_send`.

### DIFF
--- a/futures-sink/src/channel_impls.rs
+++ b/futures-sink/src/channel_impls.rs
@@ -1,10 +1,10 @@
 use {Async, Sink, Poll};
 use futures_core::task;
-use futures_channel::mpsc::{Sender, ChannelClosed, UnboundedSender};
+use futures_channel::mpsc::{Sender, SendError, UnboundedSender};
 
 impl<T> Sink for Sender<T> {
     type SinkItem = T;
-    type SinkError = ChannelClosed<T>;
+    type SinkError = SendError;
 
     fn poll_ready(&mut self, cx: &mut task::Context) -> Poll<(), Self::SinkError> {
         self.poll_ready(cx)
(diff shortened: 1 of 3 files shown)


futures-sink/src/channel_impls.rs
```diff
@@ -25,7 +25,7 @@ impl<T> Sink for Sender<T> {
 
 impl<T> Sink for UnboundedSender<T> {
     type SinkItem = T;
-    type SinkError = ChannelClosed<T>;
+    type SinkError = SendError;
```
cramertj: With this new error type, you can no longer get the value back if send failed. I don't think this will be too much of a problem in practice, plus it was kind of a pain because it would only *sometimes* be present in the error type.

seanmonstar: That's true, but there is `Sender::try_send` and `UnboundedSender::unbounded_send` if you need that fuctionality, while cleaning up everything else.

cramertj: Okay-- that seems reasonable to me. Thanks for the PR!

cramertj:
@seanmonstar LGTM-- thanks for the PR! Ping me when you've had a chance to rebase and I'll merge this.

Issue #775, opened 2018-02-17
mpsc Channel and custom retry logic make for never-closing channel


xrl:
I am writing a AWS Kinesis client using the Rusoto AWS SDK and I have hit an issue where I cannot retry requests without cloning the futures::sync::mpsc::channel Sender and still have the right kind of drop semantics for closing the Receiver.

The code spawns a thread to handle receiving records, it batches them up, and will create `inflight` number of concurrent requests. This is great for throughput as Kinesis has relatively high latency (50ms) and small request size (500 records). These requests can fail for a number of reasons: the TCP connection sees an error, the kinesis stream is under-provisioned for the current message rate (bursts, etc), Kinesis has errors, etc. I need to simply retry requests when they fail.

My first stab at retrying requests has me cloning the Sender and moving that handle in to the Receiver thread. Now the issue is that when the original Sender is dropped the receiver does not automatically close -- my Receiver thread is now open forever and I am leaking threads in my code.

So my issue, is there a way to create a Sender which does NOT count for the life of the channel? Or is there a way to `close()` a Sender which includes closing all the other Senders? Either one of these solutions may yield data loss for requests which need to be retried. But perhaps it would be useful? Or, perhaps there is a better way to structure this retry logic? Could I do it in the `AndThen` to keep that initial request going?

Here is code which spawns the Receiver thread and attempts retries:

```
pub fn kinesis_tx(stream_name: String, chan_buf: usize, inflight: usize) -> RecordsChannel {
    info!("CREATING kinesis channel");
    use futures::sync::mpsc::channel;
    use futures::{Future, Sink, Stream};

    let (tx, rx) = channel(chan_buf);
    let client = Arc::new(KinesisClient::simple(Region::UsWest2));

    let mut retry_tx = tx.clone().wait();
    std::thread::spawn(move || {
        info!("STARTING kinesis batch put thread");
        let puts = rx.chunks(500)
            .map(|batch: Vec<PutRecordsRequestEntry>| {
                let input = PutRecordsInput {
                    records: batch,
                    stream_name: stream_name.clone(),
                };
                let chain = client.put_records(&input).then(|put_res| {
                    let retval: Result<
                        Result<PutRecordsOutput, (PutRecordsError, PutRecordsInput)>,
                        (),
                    > = match put_res {
                        Ok(res) => {
                            trace!("match put_res: it worked");
                            Ok(Ok(res))
                        }
                        Err(err) => {
                            trace!("match put_res: failed");
[...]

PR #836, opened 2018-03-02
add Sender::is_closed method

seanmonstar:
It useful to be able to check if the channel is closed without registered to a task.

I didn't do it in this PR because I'm unsure of the name, but it's related: as mention in https://github.com/rust-lang-nursery/futures-rs/issues/775, it would be useful to close a channel from a `Sender`. I'm not sure if it should be required to have a `Context`, like with `poll_close`, so that someone could signal a close in another thread, but I'm sure the name `close` is reserved to be the mini-future for `poll_close` that borrows self.

### DIFF
--- a/futures-channel/src/mpsc/mod.rs
+++ b/futures-channel/src/mpsc/mod.rs
@@ -601,6 +601,11 @@ impl<T> Sender<T> {
         Ok(self.poll_unparked(Some(cx)))
     }
 
+    /// Returns whether this channel is closed without needing a context.
+    pub fn is_closed(&self) -> bool {
+        !decode_state(self.inner.state.load(SeqCst)).is_open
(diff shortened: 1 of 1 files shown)


seanmonstar:
Any thoughts on the point in the description comment?

cramertj:
About the naming of close? Maybe `close_channel`?

cramertj:
Or you could keep with the pattern used in some of the other methods and have `poll_close` and `try_close`, where `try_close` doesn't take a `task::Context`.

seanmonstar:
Of course, closing an mpsc channel has no error case, so it could just be `fn close_channel(&mut self) {`.

cramertj:
Seems fine to me!

PR #837, opened 2018-03-02
change Async::is_not_ready() to Async::is_pending()

seanmonstar:
This makes the methods symmetrical with their variants, whereas before, it was less obvious that `is_not_ready` was for `Async::Pending`.


cramertj:
Thanks!

PR #841, opened 2018-03-05
Ensure building `futures-executor` on `no_std` environments works

Nemo157:
`num_cpus` is not capable of running in a `no_std` environment so just remove it as a dependency when building without `std`.


alexcrichton:
Thanks!

PR #843, opened 2018-03-05
fix travis badge in readme

srijs:
Fixes the travis badge in the README after the move into the `rust-lang-nursery` org.

The appveyor project that is linked is also wrong (when you navigate to https://ci.appveyor.com/project/alexcrichton/futures-rs you can see that the last build is over a month ago), but it seems this needs enabling on appveyor first.


aturon:
Thanks!

Issue #842, opened 2018-03-05
Not all targets support AtomicUsize


Nemo157:
https://github.com/rust-lang-nursery/futures-rs/blob/master/futures-core/src/task/atomic_waker.rs requires `AtomicUsize`, unfortunately some targets such as `thumbv6m-none-eabi` lack atomic primitives and so currently don't build `futures 0.2`.

Nemo157:
One solution is to use the `#[cfg(target_has_atomic = "ptr")]` check to avoid compiling the `futures_core::task::atomic_waker` module when not supported, that currently requires using nightly for the unstable `cfg_target_has_atomic` feature:

```diff
diff --git futures-core/src/lib.rs futures-core/src/lib.rs
index edef417d..a3335849 100644
--- futures-core/src/lib.rs
+++ futures-core/src/lib.rs
@@ -4,6 +4,8 @@
 #![deny(missing_docs, missing_debug_implementations, warnings)]
 #![doc(html_root_url = "https://docs.rs/futures-core/0.2")]

+#![feature(cfg_target_has_atomic)]
+
 #[macro_use]
 #[cfg(feature = "std")]
 extern crate std;
diff --git futures-core/src/task/mod.rs futures-core/src/task/mod.rs
index 8a6055af..bca90953 100644
--- futures-core/src/task/mod.rs
+++ futures-core/src/task/mod.rs
@@ -13,7 +13,9 @@ if_not_std! {
     type Exec<'a> = ();
 }

+#[cfg(target_has_atomic = "ptr")]
 mod atomic_waker;
+#[cfg(target_has_atomic = "ptr")]
 pub use self::atomic_waker::AtomicWaker;

 /// Information about the currently-running task.
diff --git futures/src/lib.rs futures/src/lib.rs
index 3096ffe0..409bd32a 100644
--- futures/src/lib.rs
+++ futures/src/lib.rs
@@ -22,6 +22,8 @@
 #![no_std]
 #![doc(html_root_url = "https://docs.rs/futures/0.2")]

+#![feature(cfg_target_has_atomic)]
+
 extern crate futures_core;
 extern crate futures_channel;
 extern crate futures_executor;
@@ -354,9 +356,12 @@ pub mod task {
     //! executors or dealing with synchronization issues around task wakeup.

     pub use futures_core::task::{
-        AtomicWaker, Context, LocalMap, Waker, UnsafeWake,
+        Context, LocalMap, Waker, UnsafeWake,
     };

[...]

PR #845, opened 2018-03-07
Only include AtomicWaker with std build

Nemo157:
Workaround for #842 

This is not my preferred solution, but it's the best solution I can think of that works on stable rust

### DIFF
--- a/futures-core/src/task/mod.rs
+++ b/futures-core/src/task/mod.rs
@@ -13,9 +13,6 @@ if_not_std! {
     type Exec<'a> = ();
 }
 
-mod atomic_waker;
-pub use self::atomic_waker::AtomicWaker;
-
 /// Information about the currently-running task.
 ///
 /// Contexts are always tied to the stack, since they are set up specifically
@@ -58,6 +55,9 @@ impl<'a> Context<'a> {
 }
 
 if_std! {
+    mod atomic_waker;
+    pub use self::atomic_waker::AtomicWaker;
+
     use std::boxed::Box;
     use Future;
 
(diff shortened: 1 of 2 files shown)


cramertj:
I'm not sure about this change-- do you know of specific platforms that don't support `AtomicUsize` which you could `cfg` out instead?

Nemo157:
The only platform I know of/care about right now is `thumbv6m-none-eabi`. I assume there are other embedded platforms that don’t, but I can start with just excluding that.

Nemo157:
So, `#[cfg(not(target = "thumbv6m-none-eabi"))]` doesn't seem to be supported (not listed [in the reference](https://doc.rust-lang.org/reference/attributes.html#conditional-compilation) and using it doesn't exclude the code from compiling).

`#[cfg(not(target_arch = "arm"))]` works, but is far too broad since (for example) that includes `armv7-apple-ios` which supports atomic instructions.

As mentioned in the linked ticket `#[cfg(target_has_atomic = "ptr")]` is the correct solution, but that requires the [unstable `cfg_target_has_atomic` feature](https://github.com/rust-lang/rust/issues/32976).

None of the other `cfg` options seem relevant.

Nemo157:
Since #854 introduced a `nightly` feature flags, I've pushed an [alternative branch](https://github.com/rust-lang-nursery/futures-rs/compare/master...Nemo157:target_has_atomic) that uses `#[cfg_attr(feature = "nightly", cfg(target_has_atomic = "ptr"))]` to disable `AtomicWaker` when building with the `nightly` feature for a target that doesn't support it. If you'd prefer that I can replace this PR with it.

cramertj:
@Nemo157 That sounds great to me!

Nemo157:
I have many other things tying me to nightly at the moment, so 🤷‍♂. I’ll wait for #871 to be merged then open a PR to avoid conflicts.

PR #847, opened 2018-03-07
hide From<NodeToHandle> for Waker from docs

seanmonstar:
It's not actually usable outside the crate, but rustdoc includes it on the `Waker` doc page.


cramertj:
Thanks!

PR #852, opened 2018-03-09
stream::select_all: replace "use_std" feature check with "std" check

steffengy:
This also moves the select_all fn into select_all.rs,
since that's where it should be (this uses if_std to
check for std, and prevents further issues with
not updating a feature name)

This e.g. leads to hiding this function from the current 0.2.0-alpha doc.


cramertj:
Thanks!

PR #854, opened 2018-03-09
Integrate futures-stable into futures crate.

withoutboats:
futures-stable provides adapters to work with futures that contain
internal self-references.

futures-stable only works on nightly. A new nightly feature is
added to futures to turn futures-stable on. The futures-stable
items are re-exported from the `stable` submodule.

Also on nightly only, Future and Stream are implemented for
PinBoxes of futures and streams.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -5,6 +5,7 @@ members = [
   "futures-channel",
   "futures-executor",
   "futures-io",
-  "futures-util",
   "futures-sink",
+  "futures-stable",
+  "futures-util",
 ]
--- a/futures-core/Cargo.toml
+++ b/futures-core/Cargo.toml
@@ -11,7 +11,7 @@ The core traits and types in for the `futures` library.
 """
 
 [dependencies]
-pin-api = { version = "0.1.2", default-features = false }
+pin-api = { version = "0.1.3", default-features = false }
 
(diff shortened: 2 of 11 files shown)


withoutboats:
cc @cramertj

cramertj:
This looks great! Thanks! i'll merge once travis is happy.

withoutboats:
Tests are failing because `cargo test --all` in the workspace root can't figure out that we want pin-api with default features turned off. Trying to figure out how to make that work.

withoutboats:
Turned nightly off by default in futures-stable, which should get tests to pass.

cramertj:
> this is unfortunate, since futures-stable does nothing unless nightly is turned on. But futures-stable is unfortunate in general

SGTM. There's a precedent for this in `futures-channel`, which is entirely cfg'd out when the `std` feature is disabled.

withoutboats:
> There's a precedent for this in futures-channel, which is entirely cfg'd out when the std feature is disabled.

Yea thats fine, the unfortunate part is that `nightly` is not a default feature in futures-stable. Ideally, if you depended directly on futures-stable, by default it would have all the nightly-only items in it.

cramertj:
Yeah, that is annoying.

PR #859, opened 2018-03-10
Derive PartialEq and Eq for Waker

nwoeanhinnogaehr:
These impls can be useful sometimes - I see no reason not to include them.


carllerche:
I don’t think this should be added. The implication is that a waker handle references a task and now one can test task identity using the waker handle.

Of course, executors might use the same pointer for all tasks. This adds an additional requirement for executor implementations that does not seem necessary.

nwoeanhinnogaehr:
Ah, I hadn't considered that executors might not differentiate between tasks.

But, I had completely missed the existence of the `will_wake` method, which seems to just do the same thing as Eq. So that requirement is already in place.

PR #863, opened 2018-03-12
Update `AtomicWaker` docstrings to refer to `wake` instead of `notify`

### LINKED ISSUE #862: Stale AtomicWaker documentation
`AtomicWaker` docs reference method `notify()`. However, there is no such method. It seems commit d3097569 has changed `notify()` to `wake()`, while leaving the documentation behind.

imp:
Fixes #862


cramertj:
Thanks!

PR #865, opened 2018-03-13
Fix bug where Senders were not unparked upon calling Receiver::try_next

### LINKED ISSUE #861: Surprising behavior from mpsc channels when only using `try_` methods
Once a bounded channel has been filled and cleared using only `try_send` and `try_next`, it remains in a broken state. From here calling `try_send` always reports that the queue is full, where as `try_next` always reports that it is empty. This behavior is not found with `std::sync::mpsc`. 

See the following example (0.2.0-alpha):
```rust
use futures::channel::mpsc;
fn main() {
    let (mut tx, mut rx) = mpsc::channel(1);
    println!("{:?}", tx.try_send("hello")); // Ok(())
    println!("{:?}", tx.try_send("hello")); // Ok(())
    println!("{:?}", rx.try_next()); // Ok(Some("hello"))
    println!("{:?}", rx.try_next()); // Ok(Some("hello"))
    println!("{:?}", tx.try_send("hello")); // Err(TrySendError { kind: Full })
    println!("{:?}", rx.try_next()); // Err(TryRecvError)
    println!("{:?}", tx.try_send("hello")); // Err(TrySendError { kind: Full })
    println!("{:?}", rx.try_next()); // Err(TryRecvError)
}
```

nwoeanhinnogaehr:
Fixed by moving calls to unpark_one and dec_num_messages into next_message so that calling them cannot be missed (as it was with try_next).

Closes #861

### DIFF
--- a/futures-channel/src/mpsc/mod.rs
+++ b/futures-channel/src/mpsc/mod.rs
@@ -832,7 +832,9 @@ impl<T> Receiver<T> {
     /// no longer empty.
     pub fn try_next(&mut self) -> Result<Option<T>, TryRecvError> {
         match self.next_message() {
-            Async::Ready(msg) => Ok(msg),
+            Async::Ready(msg) => {
+                Ok(msg)
+            },
             Async::Pending => Err(TryRecvError { _inner: () }),
         }
     }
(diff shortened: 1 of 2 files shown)


cramertj:
Thanks!

PR #867, opened 2018-03-13
Document error kinds AsyncRead/AsyncWrite don't allow

### LINKED ISSUE #803: AsyncRead / AsyncWrite EINTR handling
Continuation of https://github.com/tokio-rs/tokio-io/issues/37.

PR: https://github.com/tokio-rs/tokio/pull/136

This should be figured out before releasing.

Nemo157:
Closes #803

Also, drop unnecessary handling of ErrorKind::Interrupted

### DIFF
--- a/futures-io/src/lib.rs
+++ b/futures-io/src/lib.rs
@@ -107,6 +107,13 @@ if_std! {
         /// `Ok(Async::Pending)` and arranges for the current task (via
         /// `cx.waker()`) to receive a notification when the object becomes
         /// readable or is closed.
+        ///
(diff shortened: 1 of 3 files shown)


Nemo157:
As far as I'm aware the implementations in `futures-io/src/lib.rs` are safe from `ErrorKind::Interrupted` as they are all memory backed. Since `futures-util/src/io/allow_std.rs` could wrap an arbitrary type that might return `Interrupted` that has a simple retry loop to ignore that error.

cramertj:
This looks great to me! Thanks!

PR #868, opened 2018-03-13
Fix bug in AtomicWaker

carllerche:
The original implementation of `AtomicWaker` required calls to
`register` to be coordinated. As such, there could be no concurrent
updates to the waker.

This requirement was removed by adding some additional internal
coordination to allow concurrent calls to `register`. However, this
change left an instance of the waker handle being used outside of the
coordination logic.

This patch fixes the bug by moving the usage of the `waker` instance
inside the lock's critical section.

This patch also reduces spurious notifications by removing the waker
once it has been notified.

### DIFF
--- a/futures-core/src/task/atomic_waker.rs
+++ b/futures-core/src/task/atomic_waker.rs
@@ -110,6 +110,8 @@ impl AtomicWaker {
     ///     type Error = Never;
     ///
     ///     fn poll(&mut self, cx: &mut task::Context) -> Poll<(), Never> {
+    ///         // Register **before** checkin `set` to avoid a race condition
+    ///         // that would result in lost notifications.
     ///         self.waker.register(cx.waker());
     ///
     ///         if self.set.load(SeqCst) {
--- a/futures-core/src/task/atomic_waker.rs
+++ b/futures-core/src/task/atomic_waker.rs
@@ -77,7 +77,9 @@ impl AtomicWaker {
     /// The new task will take place of any previous tasks that were registered
     /// by previous calls to `register`. Any calls to `wake` that happen after
     /// a call to `register` (as defined by the memory ordering rules), will
-    /// notify the `register` caller's task.
+    /// notify the `register` caller's task and deregister the waker from future
+    /// notifications. Because of this, callers should ensure `register` gets
+    /// invoked with a new `Waker` **each** time they require a wakeup.
     ///
     /// It is safe to call `register` with multiple other threads concurrently
     /// calling `wake`. This will result in the `register` caller's current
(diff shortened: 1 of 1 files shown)


cramertj:
Thanks for the fix to that race!

> This patch also reduces spurious notifications by removing the waker
once it has been notified.

This breaks existing code that I've written in the past which assumes that a task will be re-notified without needing to reregister so long as the task hasn't changed. I think it's fine to sacrifice that optimization for the sake of preventing redundant wakeups, but we should definitely take care to document the behavior here. Would you mind adding some comments to the docs explaining the new behavior?

cc @alexcrichton have you written anything else that made the same assumption?

carllerche:
The current docs say:

> Consumers should call `register` before checking the result of a computation and producers should call `notify` after producing the computation (this differs from the usual `thread::park` pattern).

I agree this is isn't clear. what do you think would be clearer?

cramertj:
Maybe something like this? "Users should call `register` with a new `Waker` each time they require a wakeup. When `notify` is called, the current `Waker` is awoken and deregistered."

carllerche:
Updated w/ an example.

futures-core/src/task/atomic_waker.rs
```diff
@@ -87,6 +89,37 @@ impl AtomicWaker {
...
+    ///     waker: AtomicWaker,
+    ///     set: AtomicBool,
+    /// }
+    ///
+    /// impl Future for Flag {
+    ///     type Item = ();
+    ///     type Error = Never;
+    ///
+    ///     fn poll(&mut self, cx: &mut task::Context) -> Poll<(), Never> {
+    ///         self.waker.register(cx.waker());
```
cramertj: I'm not sure I understand this example, but shouldn't the `register` only be called in the `Pending` case?

carllerche: It has to be called *before* checking the value. So, if you were to do that, you'd need to check `set`, call `register`, then check `set` again.

cramertj: I see-- because otherwise you might miss a wakeup. Can you leave a comment explaining that here?

carllerche: Done

futures-core/src/task/atomic_waker.rs
```diff
@@ -110,6 +110,8 @@ impl AtomicWaker {
     ///     type Error = Never;
     ///
     ///     fn poll(&mut self, cx: &mut task::Context) -> Poll<(), Never> {
+    ///         // Register **before** checkin `set` to avoid a race condition
```
cramertj: s/checkin/checking/

carllerche: gd macbook keyboards... fixed

cramertj:
Thanks!

PR #871, opened 2018-03-13
0.2 rebased

cramertj:
Rebased https://github.com/rust-lang-nursery/futures-rs/pull/870

### DIFF
--- a/futures-core/src/future/mod.rs
+++ b/futures-core/src/future/mod.rs
@@ -141,6 +141,15 @@ if_std! {
         }
     }
 
+    #[cfg(feature = "nightly")]
+    impl<F: ?Sized + Future> Future for ::pin_api::PinBox<F> {
+        type Item = F::Item;
+        type Error = F::Error;
+
+        fn poll(&mut self, cx: &mut task::Context) -> Poll<Self::Item, Self::Error> {
+            unsafe { ::pin_api::PinMut::get_mut(&mut self.as_pin_mut()).poll(cx) }
+        }
+    }
 
     impl<F: Future> Future for ::std::panic::AssertUnwindSafe<F> {
         type Item = F::Item;
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -234,7 +234,7 @@ Notable deprecations in the 0.1.4 release that will be deleted in an eventual
 * The `failed` constructor is now `err`.
 * The `done` constructor is now `result`.
 
-As always, please report bugs to https://github.com/alexcrichton/futures-rs and
+As always, please report bugs to https://github.com/rust-lang-nursery/futures-rs and
 we always love feedback! If you've got situations we don't cover, combinators
 you'd like to see, or slow code, please let us know!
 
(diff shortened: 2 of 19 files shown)


withoutboats:
I'm actually becoming uncomfortable with how we've brought futures-stable into futures. I'd like to ship futures 0.2 without a dependency on pin-api.

cramertj:
@withoutboats can we ship with a dependency only when the `nightly` feature is enabled?

withoutboats:
Not if you want it to be no_std compatible, because there's no way to turn on the `pin-api/std` feature only when both `std` and `nightly` are turned on.

cramertj:
@withoutboats Couldn't the `pin-api` dependency be removed entirely when `nightly` is disabled?

withoutboats:
no because the std feature will add the dependency.

cramertj:
`nightly_std` -- yuck

Nemo157:
Maybe it would still make sense to bring the code into this repository if the basic design of `futures-stable` is set, but don’t expose it through the `futures` facade yet?

cramertj:
@Nemo157 The [futures-core](https://github.com/rust-lang-nursery/futures-rs/pull/871/files#diff-a8eaa3ebdb4a916ec664cc05d61127f2R145) impl of `Future` for `PinBox<StableFuture>` needs to be there, or none of this is usable in practice.

futures-core/src/future/mod.rs
```diff
@@ -141,6 +141,15 @@ if_std! {
         }
     }
 
+    #[cfg(feature = "nightly")]
+    impl<F: ?Sized + Future> Future for ::pin_api::PinBox<F> {
```
cramertj: @withoutboats Shouldn't this be `T: StableFuture`?

withoutboats: No. StableFuture isn't object safe, so `StableFuture::pin` returns a `PinBox<Future>`, using unsafe code.

cramertj: Gotcha. Thanks!

withoutboats: If it were able to be StableFuture, it could just live in futures-stable and there'd be no need for this to be in-tree other than convenience.

cramertj:
i'm going to merge this for now so that we can update futures-await and iterate on it. We can remove (or figure out a different feature gate arrangement) before publishing a final 0.2 if we can't ensure that it will be forwards-compatible with the std API to come.

cramertj:
@withoutboats np! 0.2 got pulled into master a bit before the alpha release. I should've deleted the branch to make it clear what was happening.

Nemo157:
Ah, yeah, I haven’t been following the `PinBox` stuff much since I’m stuck in`no_std` world. I guess that would allow avoiding the executor needing to explicitly know about `StableFuture`.

Actually, that makes me wonder if there should be an `impl<T: StableFuture> Future for PinMut<T>` as well to allow passing `StackPinned` stable futures into an executor without it having to know the stability. I’ll experiment with that and might open a PR.

Issue #858, opened 2018-03-10
Clone impl for LocalExecutor?


AljoschaMeyer:
futures 0.2.0

LocalPool's [`executor`](https://docs.rs/futures-executor/0.2.0-alpha/futures_executor/struct.LocalPool.html#method.executor) method promises a "clonable handle" to the pool. But LocalExecutor does not implement Clone.

AljoschaMeyer:
This is actually pretty bad. Say I have a superfuture that wants to spawn non-threadsafe subfutures. Since the Executor trait requires things to be `Send` for spawning, we can't use the default executor to spawn a non-send subfuture. So we have to somehow pass a `LocalExecutor` to the superfuture.

To spawn the superfuture via `spawn_local`, it needs to be `'static`, so it has to take ownership of the `LocalExecutor`. But by moving it into the superfuture, I can't use it anymore to run the `LocalPool`.

So unless I'm miss(understand)ing something, it is currently impossible to use futures 0.2.0 for non-trivial, non-threadsafe execution on a single thread.

cramertj:
@AljoschaMeyer Thanks for the report! I've opened https://github.com/rust-lang-nursery/futures-rs/pull/872 to fix this.

PR #872, opened 2018-03-13
Add and document Clone impls for executor handles

cramertj:
Fixes https://github.com/rust-lang-nursery/futures-rs/issues/858


aturon:
r+ after rebase + CI

PR #873, opened 2018-03-14
Disable atomics on targets that don't support them

Nemo157:
Fixes (but only with `nightly` feature for now) #842.


cramertj:
This looks great, thanks!

PR #877, opened 2018-03-17
Fix nightly feature

Nemo157:
I've added the nightly feature into the travis build matrix to try and avoid breakage of it in the future. If you don't want to guarantee that then I could drop that change from this PR.


cramertj:
Thanks!

PR #878, opened 2018-03-18
Expose task::Context to lazy

srijs:
I'm currently working on moving the hyper 0.12.x branch over to the futures 0.2 alpha release.

As part of this, I'm starting to notice a few places where I would like to get hold of `task::Context` without needing to write a special future for it.

Here is an example:

```rust
fut.and_then(move |res| {
    // how do I get access to `cx`?
    if let Ok(Async::Pending) = pooled.tx.poll_ready(cx) {
        // do stuff
    }

    Ok(Async::Ready(res))
});
```

I tried using `poll_fn`, since that gives me the `task::Context` as an argument, but since it has a `FnMut` it doesn't work for most of cases I've tried.

Therefore I would like to propose this change, introducing `task::Context` as an argument to the closure passed in to `lazy`, similar to how it works for `poll_fn`.

Let me know what you think!

### DIFF
--- a/futures-executor/tests/local_pool.rs
+++ b/futures-executor/tests/local_pool.rs
@@ -40,7 +40,7 @@ fn run_until_single_future() {
     {
         let mut pool = LocalPool::new();
         let mut exec = pool.executor();
-        let fut = lazy(|| {
+        let fut = lazy(|_| {
             cnt += 1;
             DONE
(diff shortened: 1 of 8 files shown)


cramertj:
Thanks for the PR! This makes sense to me.

PR #879, opened 2018-03-18
use `*const UnsafeWake` instead of `mut`

stbuehler:
Hi,

I couldn't find a good reason to use `*mut UnsafeWake` in the git history (it was introduced in 42f690993 as `pub struct NotifyHandle { inner: *mut UnsafeNotify, }`).

Given that the `UnsafeWake` trait methods take `&self` (i.e. don't need `mut`) and `Arc::{from,into}_raw` operates with `*const T` I suggest using `*const UnsafeWake` instead.


cramertj:
Makes sense to me, thanks!

PR #880, opened 2018-03-18
use `Arc::{into,from}_raw`

stbuehler:
`Arc` provides an API to convert into and from raw pointers (since rust 1.17). Replace `arc2ptr` and `ptr2arc` introduced in 684e2650e.


cramertj:
Thanks!

PR #881, opened 2018-03-19
Rewrite `AtomicTask`, fixing unsoundness

carllerche:
**This is a work in progress** I am creating the PR early in case anyone wants to take an early look.

The `AtomicTask` abstraction was initially added as an internal utility for a specific use case. It has since been included in the public API.

The original implementation was edited incrementally, unfortunately adding bugs. An attempt (#868) was made to fix an unsoundness issue as well as another bug, but this unfortunately introduced new unsoundness due to another pre-existing bug that just happened to be hidden by the bug that was fixed... i.e., it was a mess.

This PR attempts to resolve the issues by taking starting from scratch.

Preliminary testing finds that this PR fixes the issue tracked in https://github.com/tokio-rs/tokio/issues/236, but more thorough testing and reviewing is needed.

### Remaining

* [x] More extensive testing against Tokio
* [x] Document implementation.

### DIFF
--- a/src/task_impl/atomic_task.rs
+++ b/src/task_impl/atomic_task.rs
@@ -90,38 +90,38 @@ impl AtomicTask {
                     let mut curr = REGISTERING;
 
                     // If a task has to be notified, it will be set here.
-                    let mut notify = None;
+                    let mut notify: Option<Task> = None;
 
                     loop {
                         let res = self.state.compare_exchange(
-                            curr, WAITING, AcqRel, Acquire);
+                            curr, WAITING, Release, Acquire);
 
                         match res {
                             Ok(_) => {
-                                // The atomic exchange was successful, now break
-                                // out of the loop os that task stored in
-                                // `notify` can be notified (if set).
-                                break;
+                                // The atomic exchange was successful, now
+                                // notify the task (if set) and return.
+                                if let Some(task) = notify {
+                                    task.notify();
+                                }
+
+                                return;
                             }
                             Err(actual) => {
+                                // This branch can only be reached if a
+                                // concurrent thread called `notify`. In this
+                                // case, `actual` **must** be `REGISTERING |
+                                // `NOTIFYING`.
+                                debug_assert_eq!(curr, REGISTERING | NOTIFYING);
+
+                                // Take the task to notify once the atomic operation has
+                                // completed.
+                                notify = (*self.task.get()).take();
+
                                 // Update `curr` for the next iteration of the
                                 // loop
                                 curr = actual;
                             }
                         }
-
-                        // Since we aren't using the weak variant of the atomic
-                        // operation, the only possible option for `curr` is to
-                        // also include the `NOTIFYING` bit.
-                        debug_assert_eq!(curr, curr | NOTIFYING);
-
-                        // Take the task to notify once the atomic operation has
-                        // completed.
-                        notify = (*self.task.get()).take();
-                    }
-
-                    if let Some(task) = notify {
-                        task.notify();
                     }
                 }
             }
@@ -131,8 +131,17 @@ impl AtomicTask {
                 // So, we call notify on the new task handle
                 task.notify();
             }
-            _ => {
-                // TODO: Assert valid state
+            state => {
+                // In this case, a concurrent thread is holding the
+                // "registering" lock. This probably indicates a bug in the
+                // caller's code as racing to call `register` doesn't make much
+                // sense.
+                //
+                // We just want to maintain memory safety. It is ok to drop the
+                // call to `register`.
+                debug_assert!(
+                    state == REGISTERING ||
+                    state == REGISTERING | NOTIFYING);
             }
         }
     }
@@ -156,13 +165,18 @@ impl AtomicTask {
                     task.notify();
                 }
             }
-            _ => {
+            state => {
                 // There is a concurrent thread currently updating the
                 // associated task.
                 //
-                // Nothing more to do as the `NOTIFYING` bit has been set
+                // Nothing more to do as the `NOTIFYING` bit has been set. It
+                // doesn't matter if there are concurrent registering threads or
+                // not.
                 //
-                // TODO: Validate the state
+                debug_assert!(
+                    state == REGISTERING ||
+                    state == REGISTERING | NOTIFYING ||
+                    state == NOTIFYING);
             }
         }
     }
--- a/src/task_impl/atomic_task.rs
+++ b/src/task_impl/atomic_task.rs
@@ -29,6 +29,102 @@ pub struct AtomicTask {
     task: UnsafeCell<Option<Task>>,
 }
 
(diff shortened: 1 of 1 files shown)


src/task_impl/atomic_task.rs
```diff
@@ -86,90 +73,96 @@ impl AtomicTask {
...
+                    loop {
+                        let res = self.state.compare_exchange(
+                            curr, WAITING, AcqRel, Acquire);
+
+                        match res {
+                            Ok(_) => {
+                                // The atomic exchange was successful, now break
+                                // out of the loop os that task stored in
+                                // `notify` can be notified (if set).
+                                break;
```
cramertj: Nit: this loop is a bit hard for me to read w/ this `break` in it. Can you inline the `if let` below into the this branch and put a `return` here?

src/task_impl/atomic_task.rs
```diff
@@ -86,90 +73,96 @@ impl AtomicTask {
...
-                // is currently being called on the old task handle. So, we call
-                // notify on the new task handle
+            NOTIFYING => {
+                // Currently in the process of notifying the task, i.e.,
+                // `notify` is currently being called on the old task handle.
+                // So, we call notify on the new task handle
                 task.notify();
             }
+            _ => {
+                // TODO: Assert valid state
```
cramertj: In this branch, two people called register at ~ the same time, so it doesn't matter that we're technically the *second* one to call register if we reach this case?

src/task_impl/atomic_task.rs
```diff
@@ -86,90 +73,96 @@ impl AtomicTask {
...
-                    return;
+                if let Some(task) = task {
+                    task.notify();
                 }
-
-                // update current state variable and try again
-                curr = actual;
-
+            }
+            _ => {
```
cramertj: Nit: can you list out the remaining cases, just to make it clear that all of them should indeed be no-ops?

(REGISTERING, NOTIFYING, REGISTERING | NOTIFYING)

src/task_impl/atomic_task.rs
```diff
@@ -86,90 +73,96 @@ impl AtomicTask {
...
+                            Ok(_) => {
+                                // The atomic exchange was successful, now break
+                                // out of the loop os that task stored in
+                                // `notify` can be notified (if set).
+                                break;
+                            }
+                            Err(actual) => {
+                                // Update `curr` for the next iteration of the
+                                // loop
+                                curr = actual;
```
cramertj: `actual` should be only `REGISTERING | NOTIFYING`, right? Can you update the code to specifically handle this case, rather than matching on any value of `actual`? (and panic or something in other cases)

src/task_impl/atomic_task.rs
```diff
@@ -86,90 +73,96 @@ impl AtomicTask {
...
-                // guarantees), so we can just do nothing.
-            }
-            state => {
-                debug_assert!(state != LOCKED_WRITE, "unexpected state LOCKED_WRITE");
-                debug_assert!(state != LOCKED_WRITE_NOTIFIED, "unexpected state LOCKED_WRITE_NOTIFIED");
-
-                // Currently in a read locked state, this implies that `notify`
-                // is currently being called on the old task handle. So, we call
-                // notify on the new task handle
+            NOTIFYING => {
```
cramertj: Shouldn't it also call `task.notify()` if we read `REGISTERING | NOTIFYING`?

src/task_impl/atomic_task.rs
```diff
@@ -86,90 +73,96 @@ impl AtomicTask {
...
+                    // Start by assuming that the state is `REGISTERING` as this
+                    // is what we jut set it to.
+                    let mut curr = REGISTERING;
+
+                    // If a task has to be notified, it will be set here.
+                    let mut notify = None;
+
+                    loop {
+                        let res = self.state.compare_exchange(
+                            curr, WAITING, AcqRel, Acquire);
```
cramertj: Should this be `Release, Acquire`? On success, we just need the previous writes to `self.task` to be written. On failure, we only need to be able to read others' writes to `self.task`.

cramertj:
Thanks for doing this! This code was in some definite need of TLC. Overall, it would help me be more confident about the code if you explicitly enumerated all of the possible cases and state transitions with every atomic op. Exhaustive matching makes it much easier for me to reason about.

src/task_impl/atomic_task.rs
carllerche: Basically... concurrently calling `register` doesn't really make sense. `AtomicTask` only tries to maintain memory safety, but the `register` is lost.

src/task_impl/atomic_task.rs
carllerche: I'm treating `REGISTERING` being set when calling `register` as a caller bug. If you want, I can add a `panic` in this case?

src/task_impl/atomic_task.rs
carllerche: It has to be `AcqRel` as the success ordering must be stronger than the failure ordering (i.e. a super set).

carllerche:
I addressed the feedback (or replied to your comment). I kept the state assertions as debug asserts, but if you think it would be better to perform the checks at runtime, I can switch it.

src/task_impl/atomic_task.rs
cramertj: TIL! Thanks.

src/task_impl/atomic_task.rs
```diff
@@ -86,39 +73,75 @@ impl AtomicTask {
...
+                                }
+
+                                return;
+                            }
+                            Err(actual) => {
+                                // This branch can only be reached if a
+                                // concurrent thread called `notify`. In this
+                                // case, `actual` **must** be `REGISTERING |
+                                // `NOTIFYING`.
+                                debug_assert_eq!(curr, REGISTERING | NOTIFYING);
```
cramertj: Should this be comparing `actual`, rather than `curr`?

cramertj:
LGTM modulo the one nit about the `assert`.

src/task_impl/atomic_task.rs
carllerche: ugh yeah, I moved that assert up (from after curr was set to actual).

carllerche:
I pushed docs. I'm going to try to do some more extensive testing against Tokio now.

cramertj:
SGTM-- thanks again for doing this. Let me know when you're comfortable with it and I'll merge.

carllerche:
Ok, as far as I can tell, everything works.

carllerche:
Looks like CI is passing!

cramertj:
@carllerche To confirm: you'll follow up w/ a matching change to 0.2?

carllerche:
@cramertj yeah, I can work on that right after I am done w/ https://github.com/tokio-rs/tokio/pull/243

cramertj:
@carllerche Great, thanks!

PR #884, opened 2018-03-20
Close out 0.2 milestone

aturon:
This PR resolves the remaining issues on the 0.2 milestone.

### DIFF
--- a/futures-core/src/task/mod.rs
+++ b/futures-core/src/task/mod.rs
@@ -1,16 +1,11 @@
 //! Task notification.
 
 use core::fmt;
+use executor::Executor;
 
 if_std! {
     use core::marker::PhantomData;
     use never::Never;
-    use executor::Executor;
-    type Exec<'a> = &'a mut Executor;
-}
-
-if_not_std! {
-    type Exec<'a> = ();
 }
 
 #[cfg_attr(feature = "nightly", cfg(target_has_atomic = "ptr"))]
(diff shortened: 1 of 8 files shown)


futures-core/src/task/mod.rs
```diff
@@ -66,21 +82,21 @@ if_std! {
...
+            Context { waker, map, executor: Some(executor) }
         }
 
         /// Get the default executor associated with this task.
         ///
         /// This method is useful primarily if you want to explicitly handle
         /// spawn failures.
         pub fn executor(&mut self) -> &mut Executor {
-            self.executor
+            *self.executor.as_mut().unwrap()
```
cramertj: Seems like we should make this `expect`, since users might actually encounter this in practice.

futures-core/src/task/mod.rs
```diff
@@ -25,10 +20,16 @@ pub use self::atomic_waker::AtomicWaker;
...
     waker: &'a Waker,
     map: &'a mut LocalMap,
-    executor: Exec<'a>,
+    executor: Option<&'a mut Executor>,
 }
 
 impl<'a> Context<'a> {
+    /// Create a new task context without the ability to `spawn`. Useful for
+    /// `no_std` contexts.
+    pub fn without_spawn(map: &'a mut LocalMap, waker: &'a Waker) -> Context<'a> {
```
cramertj: Can we add a note about how use of this method is explicitly discouraged outside of `no_std`, and how use of it with an `std`-using library may result in panics?

aturon:
Commit pushed to address comments.

PR #885, opened 2018-03-20
Clean up a few futures-util issues

aturon:
This PR knocks out a couple of the remaining breaking changes to `futures-util` ahead of the initial 0.2.0 release. Remaining breaking issues are punted.

### DIFF
--- a/futures-util/src/stream/once.rs
+++ b/futures-util/src/stream/once.rs
@@ -19,10 +19,12 @@ pub struct Once<T, E>(Option<Result<T, E>>);
 ///
 /// # fn main() {
 /// let mut stream = stream::once::<(), _>(Err(17));
-/// assert_eq!(Err(17), block_on(stream.collect()));
+/// let collected: Result<Vec<_>, _> = block_on(stream.collect());
+/// assert_eq!(collected, Err(17));
 ///
 /// let mut stream = stream::once::<_, ()>(Ok(92));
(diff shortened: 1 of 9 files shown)


futures-util/src/stream/once.rs
```diff
@@ -18,26 +18,28 @@ pub struct Once<T, E>(Option<Result<T, E>>);
...
+    type Error = F::Error;
 
-    fn poll_next(&mut self, _: &mut task::Context) -> Poll<Option<T>, E> {
-        match self.0.take() {
-            Some(Ok(e)) => Ok(Async::Ready(Some(e))),
-            Some(Err(e)) => Err(e),
-            None => Ok(Async::Ready(None)),
+    fn poll_next(&mut self, cx: &mut task::Context) -> Poll<Option<F::Item>, F::Error> {
+        if let Some(mut f) = self.0.take() {
+            Ok(f.poll(cx)?.map(Some))
```
cramertj: If the future was `Async::Pending`, this will remove it and yield `Ok(Async::Pending)`, and all future calls will result in `Ok(Async::Ready(None))`. In order to fix this, you have to either (a) not `take` unless the value from `poll` is `Ready`, or  (b) put the future back into `self.0`.

cramertj:
LGTM aside from `Once` bug.

cramertj:
@aturon :shipit:

PR #890, opened 2018-03-21
Revise `block_on` to use a global thread pool for spawned tasks 

aturon:
Per discussion with @seanmonstar, revise the `block_on` to set the default executor to be a lazily-initialized global thread pool. This provides a much more usable bridge between the sync and async worlds, since spawned tasks are no longer dropped when the future completes.

Ultimately we'll probably want to make the executor used here configurable, but since this is purely about the sync bridge, I don't think it's very pressing.

This PR also includes an unrelated drive-by refactor of the `task` module into submodules; I thought I was going to work the code there but ended up not -- but I've been wanting to do the refactor anyway.

### DIFF
--- a/futures-executor/src/local_pool.rs
+++ b/futures-executor/src/local_pool.rs
@@ -188,7 +188,9 @@ impl LocalPool {
 }
 
 lazy_static! {
-    static ref GLOBAL_POOL: ThreadPool = ThreadPool::new();
+    static ref GLOBAL_POOL: ThreadPool = ThreadPool::builder()
+        .name_prefix("block_on-")
+        .create() ;
 }
 
 /// Run a future to completion on the current thread.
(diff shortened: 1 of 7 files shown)


cramertj:
:shipit:

futures-executor/src/local_pool.rs
```diff
@@ -186,26 +187,20 @@ impl LocalPool {
     }
 }
 
+lazy_static! {
+    static ref GLOBAL_POOL: ThreadPool = ThreadPool::new();
```
seanmonstar: Perhaps this could add a name prefix to help in case tasks panic?

aturon:
Thanks @seanmonstar! I've pushed fixes.

seanmonstar:
As an experiment, I threw together a PR that did what I mentioned last night: panic if block_on is used for futures that perhaps it shouldn't be: #896

PR #896, opened 2018-03-21
change block_on to not allow background tasks

seanmonstar:
This is an alternative to #890. Instead of using a global thread pool that can't be configured, and making it easy for users to use `block_on` with bigger tasks than they should, this instead makes it so the executor used with `block_on` universally panics when used.

This means it's still a fine tool for cross-thread synchronization:

```rust
let (tx, rx) = oneshot::channel();

thread::spawn(move || {
    let msg = block_on(rx);
    // ...
});
```

However, it is **not recommended** to use this for things like hyper's responses futures, which should instead have a better setup for task execution.

```rust
// this panics when the `ResponseFuture` tries to spawn its
// background Connection task, but with a message that
// problem was trying to use `block_on`, and the user should
// use something else.
let res = block_on(client.get(uri));
```

### DIFF
--- a/futures-executor/src/blocking.rs
+++ b/futures-executor/src/blocking.rs
@@ -1,0 +1,37 @@
+use std::boxed::Box;
+
+use futures_core::Future;
+use futures_core::executor::{Executor, SpawnError};
+use futures_core::never::Never;
+
+use local_pool::LocalPool;
+
+/// Run a future to completion on the current thread.
+///
+/// This function will block the caller until the given future has completed.
+///
+/// Use a [`LocalPool`](super::LocalPool) if you need finer-grained control over
+/// spawned tasks.
+///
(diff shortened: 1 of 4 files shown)


aturon:
The lack of configuration in the existing PR was just to keep the initial API surface small. As I said in that PR, we'll likely add configuration if/when it becomes an issue. And as we talked about at some length last night, I much prefer to make `block_on` a generally usable tool for invoking async code in sync contexts.

seanmonstar:
🤷‍♂️

I just have reservations, and a feeling that using `block_on` in this way could cause problems, or at least be the starting point of harder to understand bugs, and so figured perhaps it better to reduce the amount of places it should be used until it's better understood/has a better solution. It seems easier to enable `block_on` in more places later, than it does to take it away.

PR #899, opened 2018-03-22
Migrate from pin_api crate to std pin feature

Nemo157:
Now that there's a nightly with it released, may as well use it.


withoutboats:
Looks great, thanks so much!

PR #900, opened 2018-03-22
Fix no_std

Nemo157:
Also adds a representative `no_std` target into the travis build matrix to avoid accidental breakage in the future. Hopefully this doesn't slow CI down too much.

### DIFF
--- a/futures-async-runtime/src/lib.rs
+++ b/futures-async-runtime/src/lib.rs
@@ -1,17 +1,22 @@
+#![no_std]
 #![cfg_attr(feature = "nightly", feature(generator_trait))]
 #![cfg_attr(feature = "nightly", feature(on_unimplemented))]
 #![cfg_attr(feature = "nightly", feature(arbitrary_self_types))]
 #![cfg_attr(feature = "nightly", feature(optin_builtin_traits))]
 #![cfg_attr(feature = "nightly", feature(pin))]
 
-macro_rules! if_nightly {
+macro_rules! if_nightly_and_std {
     ($($i:item)*) => ($(
-        #[cfg(feature = "nightly")]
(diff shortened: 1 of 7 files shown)


Nemo157:
Checking the travis run it seems that the time to:

* install `rust-src` component
* install `xargo`
* build for `thumbv6-none-eabi`

is comparable to the time for just building and running the tests, so there's no real slow down assumng there's enough parallelism available.

Nemo157:
Rebased and updated to fix building `futures-async-runtime` as well, will open another PR in the future to get `futures-async-runtime` actually working on core, for now it builds but is empty.

withoutboats:
Exactly what I would've recommended :)

withoutboats:
The CI seems to have hung in a weird way, any idea what to do?

withoutboats:
Maybe amend your commit message and force push to trigger a rebuild?

Nemo157:
Rebased, looks like CI is running now

futures-async-runtime/src/lib.rs
```diff
@@ -6,11 +7,15 @@
 
 macro_rules! if_nightly {
     ($($i:item)*) => ($(
-        #[cfg(feature = "nightly")]
+        #[cfg(all(feature = "nightly", feature = "std"))]
```
stbuehler: This looks wrong; `if_nightly` should not check for `std` - maybe rename to `if_nightly_and_std` if you actually need that?

Nemo157: This is just a temporary change to get `futures` compiling, I've got another branch that will get `futures-async-runtime` actually working with `no_std` almost ready to go. But sure, I can rename the macro in the meantime.

withoutboats:
Looks good, merging :) Thanks for fixing this.

PR #902, opened 2018-03-24
Introduce WithExecutor adaptor

srijs:
Introduces the `WithExecutor` adapter, which assigns an executor to be used when spawning tasks from within the future.

The main driver for me is to enable the use-case of overriding the default thread pool for `block_on`:

```rust
block_on(future.with_executor(pool))
```

By making this a combinator it opens up more use-cases, such as this pattern:

```rust
spawn_with_handle(future).with_executor(pool)
```

### DIFF
--- a/futures-core/src/task/context.rs
+++ b/futures-core/src/task/context.rs
@@ -110,6 +110,19 @@ if_std! {
                 .expect("No default executor found")
                 .spawn(Box::new(f)).unwrap()
         }
+
+        /// Produce a context like the current one, but using the given executor
+        /// instead.
+        ///
+        /// This advanced method is primarily used when building "internal
+        /// schedulers" within a task.
+        pub fn with_executor<'b>(&'b mut self, executor: &'b mut Executor)
+                            -> Context<'b>
+        {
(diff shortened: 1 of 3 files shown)


futures-core/src/task/context.rs
```diff
@@ -110,6 +110,19 @@ if_std! {
                 .expect("No default executor found")
                 .spawn(Box::new(f)).unwrap()
         }
+
+        /// Produce a context like the current one, but using the given executor
+        /// instead.
+        ///
+        /// This advanced method is primarily used when building "internal
+        /// schedulers" within a task.
+        pub fn with_executor<'b>(&'b mut self, executor: Option<&'b mut Executor>)
```
cramertj: IMO this should take the executor directly instead of an `Option`: the executor value should only ever be `None` when using the no-std version.

srijs: Ah, okay! I think I was just a little confused by the fact that the `Context::executor` method is also flagged `if_std!` and returns an `Option`. Will change this!

srijs:
Addressed feedback and rebased!

cramertj:
Thanks!

PR #905, opened 2018-03-25
Merge futures-await into futures-rs

withoutboats:
This adds the three crates of futures-await into the futures repository, under the names `futures-async-runtime`, `futures-macro-async`, `futures-macro-await`. This allows the async/await functionality to be activated in futures by only turning on the `nightly` feature.

This PR was created using `git merge --allow-unrelated-histories`, so it carries with it the entire history of `futures-await`.


withoutboats:
Still need to get async/await tests both running & passing properly

carllerche:
@withoutboats Is there an eventual goal of adding `&pin self` as that would be a lot more ergonomic than importing `Pin` + `self: Pin<Self>`.

futures-async-runtime/src/lib.rs
```diff
@@ -0,0 +1,90 @@
...
+    pub use self::pinned_stream::*;
+
+    pub use futures_core::{Async, Future, Stream};
+    pub use futures_stable::{StableFuture, StableStream};
+
+    pub use std::ops::Generator;
+
+#[rustc_on_unimplemented = "async functions must return a `Result` or \
+                                a typedef of `Result`"]
+    pub trait IsResult {
```
srijs: Out of curiosity, is the plan to move this trait into the standard library at one point? It looks like it could serve as the base trait for `Try`, so that:

```rust
pub trait Try: IsResult {
    fn from_ok(v: Self::Ok) -> Self;
    fn from_err(v: Self::Err) -> Self;
}
```

withoutboats: This is just a hack to make error messages better, its not supposed to be a part of the API for end users.

Nemo157: I was actually recently wondering if this could be extended to `Try` instead of forcing `Result`. Currently that would only additionally allow creating an `impl Future<Item=T, Error=NoneError>` from an `Option<T>` from `std`, but I wonder if there are any interesting implementations of `Try` outside the standard library?

futures-async-runtime/Cargo.toml
```diff
@@ -0,0 +1,27 @@
...
+Runtime for the async/await macros in the `futures` crate.
+"""
+
+[dependencies.futures-core]
+version = "0.2.0-beta"
+path = "../futures-core"
+
+[dependencies.futures-stable]
+version = "0.2.0-beta"
+path = "../futures-stable"
```
Nemo157: These two dependencies should have `default-features = false` to avoid `futures-core/std` and `futures-stable/std` (which then I think recursively activates most of the `std` features except for `futures` itself) always being active when a library depends on `futures = { version = "0.2.0", default-features = false }`.

withoutboats: Could you fix this in a rebase of #900? If not, I'll open a second PR. I'm going to merge this now so that we don't have to block fixing other things in other repos on this PR.

Nemo157: Sure

PR #906, opened 2018-03-25
Relax constraint on `impl Future for SpawnWithHandle`

### LINKED ISSUE #903: impl Future for SpawnWithHandle is too strict to be useful
Hey, on `0.2.0-beta`, you can only use `spawn_with_handle` with Futures returning unit, which isn't very useful for a handle. I think this can be relaxed, but I am not sure.

```rust
impl<F> Future for SpawnWithHandle<F>
    where F: Future<Item = (), Error = Never> + Send + 'static,
          F::Item: Send,
          F::Error: Send,
{
```

Can we make it accept any `Item` and `Error`, as long as they both implement `Send`?

Nemo157:
The constraint that you must have `Future<Item = (), Error = Never>`
appears to have been a mistake, the rest of the code already intends to
support passing any `Send`able `Item` or `Error` back.

Also add doc tests showing passing both items and errors back to ensure
test coverage.

Fixes #903

### DIFF
--- a/futures-executor/src/spawn.rs
+++ b/futures-executor/src/spawn.rs
@@ -51,6 +51,48 @@ pub struct SpawnWithHandle<F>(Option<F>);
 /// onto the default executor. On completion, that future will yield a
 /// [`JoinHandle`](::JoinHandle) that can itself be used as a future
 /// representing the completion of the spawned task.
+///
+/// # Examples
+///
+/// ```
+/// # extern crate futures;
+/// #
+/// use futures::prelude::*;
+/// use futures::future;
+/// use futures::executor::{block_on, spawn_with_handle};
+///
+/// # fn main() {
+/// # fn inner() -> Result<(), Never> {
+/// # Ok({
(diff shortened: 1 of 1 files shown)


futures-executor/src/spawn.rs
```diff
@@ -51,14 +51,58 @@ pub struct SpawnWithHandle<F>(Option<F>);
...
+///
+/// # Examples
+///
+/// ```
+/// # extern crate futures;
+/// # extern crate futures_executor;
+/// #
+/// use futures::prelude::*;
+/// use futures::future;
+/// use futures_executor::{block_on, spawn_with_handle};
```
Nemo157: One question, should this be accessed via `futures::executor` instead of `futures_executor`? What is the policy on using the facade paths vs local paths in examples?

cramertj: I'd use the facade path to e consistent with the other examples.

cramertj:
thanks!

PR #908, opened 2018-03-25
Add FutureExt::inspect_err method

### LINKED ISSUE #898: FutureExt::inspect_err() ?
In continuation of #700 

Inspecting errors in a future could be quite useful. Since we have the same API in `Stream` perhaps we could include one in `Future` as well?

Nemo157:
Fixes #898

### DIFF
--- a/futures-util/src/future/mod.rs
+++ b/futures-util/src/future/mod.rs
@@ -31,6 +31,7 @@ mod or_else;
 mod select;
 mod then;
 mod inspect;
+mod inspect_err;
 mod recover;
 
 // impl details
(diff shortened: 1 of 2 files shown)


futures-util/src/future/mod.rs
```diff
@@ -714,6 +716,35 @@ pub trait FutureExt: Future {
...
+    /// call to `inspect_err`.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// # extern crate futures;
+    /// # extern crate futures_executor;
+    /// use futures::prelude::*;
+    /// use futures::future;
+    /// use futures_executor::block_on;
```
cramertj: As in the other change, I'd use `futures::executor` rather than `futures_executor`.

Nemo157:
Updated

cramertj:
Thanks!

PR #911, opened 2018-03-26
Revert "Replace the Future impl for Option with IntoFuture"

### LINKED ISSUE #892: v0.1.19 has broken backward compatibility
The following code works with v0.1.18 (or earlier versions).

```rust
extern crate futures;

 use futures::Future;

 fn main() {
     let future = futures::lazy(|| {
         println!("Hello, world!");

         let result: Result<(), ()> = Ok(());
         result
     });

     let mut future = Some(future);
     let _ = future.poll();
 }
```

But in v0.1.19, the compiler emits the below error message.

```
error[E0599]: no method named `poll` found for type `std::option::Option<futures::Lazy<[closure@src/main.rs:6:32: 11:6], std::result::Result<(), ()>>>` in the current scope
  --> src/main.rs:13:20
   |
13 |     let _ = future.poll();
   |                    ^^^^
```

### LINKED ISSUE #904: breaking stream::unfold change between 0.1.18 and 0.1.19
It looks like there was a signature change to stream::unfold ( my favorite futures interface! ) in a recent 0.1 release. My crate had an issue reported reported [here](https://github.com/softprops/hubcaps/issues/112) that a consumer wasn't able to build after doing a cargo update. I was able to confirm myself. 

I took a quickly and it looks like a signature change in futures-rs between [0.1.8](https://docs.rs/futures/0.1.18/futures/stream/fn.unfold.html) and [0.1.9](https://docs.rs/futures/0.1.19/futures/stream/fn.unfold.html). Notable there's [no release notes](https://github.com/rust-lang-nursery/futures-rs/blob/8c949e3daac8cb45f83b70ed0e039bde303e061e/CHANGELOG.md) to indicate the changes but for now docs.rs is a fine way to notice the change.

I can deal with it in my crate but I wanted to drop a note on here in case any down stream consumes of futures-rs may run into the same problem .

I know and understand yall are busy working towards a 0.2.* release so I don't expect a remedy. I hope this note is able to help others in the meantime.

cc @matthiasbeyer

seanmonstar:
Closes #892 
Closes #904

### DIFF
--- a/src/future/option.rs
+++ b/src/future/option.rs
@@ -1,33 +1,13 @@
 //! Definition of the `Option` (optional step) combinator
 
-use {Future, IntoFuture, Poll, Async};
-use core::option;
+use {Future, Poll, Async};
 
-/// An optional future.
-///
-/// Created by `Option::into_future`.
-#[derive(Debug, Clone)]
-#[must_use = "futures do nothing unless polled"]
-pub struct Option<F> {
-    inner: option::Option<F>,
-}
+impl<F, T, E> Future for Option<F> where F: Future<Item=T, Error=E> {
+    type Item = Option<T>;
+    type Error = E;
 
-impl<F> IntoFuture for option::Option<F> where F: Future {
-    type Future = Option<F>;
-    type Item = option::Option<F::Item>;
-    type Error = F::Error;
-
-    fn into_future(self) -> Self::Future {
-        Option { inner: self }
-    }
-}
-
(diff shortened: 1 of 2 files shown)


mati865:
Reverted in https://github.com/rust-lang-nursery/futures-rs/commit/c5c976d9ee90311ac517c2b5ffe685081709da7c#diff-80398c5faae3c069e4e6aa2ed11b28c0

PR #913, opened 2018-03-27
Fixed Fanout sink which was always ready

### LINKED ISSUE #912: Fanout sink is always ready
The `Fanout` code appears to have been incorrectly adapted for futures 0.2.0. The `state` field of `Downstream` is initialized to `None` and never changed, causing the `Fanout` sink to be always ready, regardless of the states of the underlying sinks.

I have a fix ready and will create a PR.

stephaneyfx:
Resolves #912.


cramertj:
Thanks!

PR #914, opened 2018-03-27
Revert "Replace the Future impl for Option with IntoFuture"

cramertj:
This reverts commit db7799587562627e03ba011058cb4f02645f76fb.

This was a breaking change which accidentally slipped into the 0.1.19 release.


aturon:
r+ with travis; let's yank 0.1.19

cramertj:
Yup, already yanked :)

PR #916, opened 2018-03-28
Fix pinned boxed async functions

raviqqe:
This change fixes pinned boxed async functions which were unable to compile.

### DIFF
--- a/futures/src/lib.rs
+++ b/futures/src/lib.rs
@@ -268,6 +268,15 @@ pub mod prelude {
     #[cfg(feature = "std")]
     pub use futures_core::executor::Executor;
 
+    #[cfg(feature = "nightly")]
+    pub use futures_stable::{
+        StableFuture,
+        StableStream
+    };
+
+    #[cfg(all(feature = "nightly", feature = "std"))]
+    pub use futures_stable::StableExecutor;
+
     pub use futures_sink::Sink;
 
     #[cfg(feature = "std")]
--- a/futures-macro-async/src/attribute.rs
+++ b/futures-macro-async/src/attribute.rs
@@ -1,0 +1,27 @@
+use self::Attribute::*;
+
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+pub enum Attribute {
+    Pinned,
+    PinnedBoxed,
+    PinnedBoxedSend,
+    Unpinned,
+    UnpinnedBoxed,
+    UnpinnedBoxedSend,
+}
+
+impl Attribute {
+    pub fn boxed(&self) -> bool {
+        match *self {
+            PinnedBoxed | PinnedBoxedSend | UnpinnedBoxed | UnpinnedBoxedSend => true,
+            Pinned | Unpinned => false,
+        }
+    }
+
+    pub fn pinned(&self) -> bool {
+        match *self {
+            Pinned | PinnedBoxed | PinnedBoxedSend => true,
+            Unpinned | UnpinnedBoxed | UnpinnedBoxedSend => false,
+        }
+    }
+}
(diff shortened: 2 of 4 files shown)


raviqqe:
I made the `Attribute` enum to pass information of if futures should be `Send` or not and reduce arguments of the `async_inner()` function. However, I don't know whether it's a correct approach.

ngg:
It's working for me pretty well, can you please merge this?

futures/src/lib.rs
```diff
@@ -264,6 +264,15 @@ pub mod prelude {
     #[cfg(feature = "std")]
     pub use futures_core::executor::Executor;
 
+    #[cfg(feature = "nightly")]
+    pub use futures_stable::{
```
withoutboats: Aren't these already exported through the `stable` submodule?

ngg:
This code doesn't work for me (it works if one of the parameters is `i32` instead of `&i32`)

```rust
#[async(pinned)]
fn bar2(x: &i32, y: &i32) -> Result<i32, i32> {
    Ok(x+y)
}
```

It fails with the following error:
```rust
error[E0226]: only a single explicit lifetime bound is permitted
  --> futures/tests/async_await/pinned.rs:16:45
   |
16 | fn bar2(x: &i32, y: &i32) -> Result<i32, i32> {
   |                                             ^

error[E0623]: lifetime mismatch
  --> futures/tests/async_await/pinned.rs:16:30
   |
16 | fn bar2(x: &i32, y: &i32) -> Result<i32, i32> {
   |                     ----     ^^^^^^^^^^^^^^^^
   |                     |        |
   |                     |        ...but data from `y` is returned here
   |                     this parameter and the return type are declared with different lifetimes...
```

withoutboats:
@ngg's first error is because `Box<Trait + 'a + 'b>` isn't valid syntax. In contrast, `impl Trait + 'a + 'b` is valid syntax, *but it doesn't work*. (This seems pretty bad). The second error is the result of neither working, even if syntactically its allowed.

The expansion of lifetimes needs to be changed to instead of just adding `+ 'a + 'b ...` to the return type, instead adding an additional lifetime and bounding every lifetime by that. So the expansion changes:

```rust
#[async] fn foo(x: &i32, y: &i32) -> Result<i32, !> { ... }

// currently:
fn foo<'_async0, '_async1>(x: &'_async0 i32, y: &'_async1 i32)
    -> impl Future<Item = i32, Error = !> + '_async0 + '_async1 { ... }

// changed to:
fn foo<'_ret, '_async0: '_ret, '_async1: '_ret>(x: &'_async0, y: &'_async1 i32)
    -> impl Future<Item = i32, Error = !> + '_ret { ... }
```

withoutboats:
Its a separate issue from this PR though!

Nemo157:
See https://github.com/rust-lang/rust/issues/49431 for a rust issue about `impl Trait + 'a + 'b`.

ngg:
Yeah I realized it's different but only after writing this here. (I wrongly remembered that it was working with normal `#[async]`, and I tried it only with `#[async(pinned)]`).
Even if https://github.com/rust-lang/rust/issues/49431 will be fixed it won't work with pinned as currently `Box<T + 'a + 'b>` and `PinBox<T + 'a + 'b>` are not permitted either.

ngg:
@raviqqe I reverted it pretty fast when I realized you already did it and that https://github.com/Nemo157/futures-rs/commit/5c32c057390100445617b09bf70108931fc708cb is based on your version. I actually prefer your solution with not creating `gen_pinbox` and `gen_pinbox_local`, but calling `pin()` and `pin_local()` inline, the other differences are quite irrelevant after #918.

withoutboats:
Will merge if tests pass, thanks for this!

withoutboats:
one failure, missing `spawn_pinned` method

withoutboats:
Merged, thanks so much for doing this! :tada:

PR #919, opened 2018-03-29
Remove some feature attributes which got stable in Rust 1.26

raviqqe:
I forgot to remove some other stable features in \#915.


withoutboats:
Thanks!!

PR #921, opened 2018-03-29
Fix broken tokio link

ChristophWurst:
Possibly related to https://github.com/tokio-rs/tokio/issues/13 and https://github.com/tokio-rs/tokio/issues/165, there's a broken link in the Sink documentation.

I'm not entirely sure if that's the correct page to link to, but of the ones available on the tokio site right now, this seemed to be the best choice.


cramertj:
Hm... that doesn't seem like quite the right link to me. Would you mind removing it for now, and we can add it back later when the relevant Async Programming in Rust chapter is added? cc @aturon

cramertj:
Thanks!

PR #925, opened 2018-03-30
Use different await! macro between async and async_move

### LINKED ISSUE #924: Possible to poll moved StableFuture when using #[async_move]
Given this code saved as `futures/examples/foo.rs`:

```rust
#![feature(generators, pin, proc_macro, arbitrary_self_types)]

extern crate futures;

use std::ptr;
use std::mem::Pin;
use futures::{Poll, Async, Future, stable::StableFuture, prelude::{async_move, await}, task};

fn baz() -> impl StableFuture<Item=(), Error=()> {
    #[derive(Debug)]
    struct Baz;

    impl StableFuture for Baz {
        type Item = ();
        type Error = ();

        fn poll(self: Pin<Self>, _cx: &mut task::Context) -> Poll<Self::Item, Self::Error> {
            println!("Poll {:?} at {:?}", *self, &*self as *const Baz);
            Ok(Async::Pending)
        }
    }

    Baz
}

#[async_move]
fn foo() -> Result<(), ()> {
    await!(baz())
}

#[inline(never)]
fn bar<F: Future<Item=(), Error=()>>(mut f: F, cx: &mut task::Context) {
    f.poll(cx).unwrap();
}

fn main() {
    let mut map = task::LocalMap::new();
    let mut waker = NoWake::waker();
    let mut context = task::Context::without_spawn(&mut map, &mut waker);

    let mut future = foo();
    future.poll(&mut context).unwrap();
    bar(future, &mut context);
}


// Not important to the example

struct NoWake;

impl NoWake {
    fn waker() -> task::Waker {
        unsafe { task::Waker::new(ptr::null() as *const NoWake as *const task::UnsafeWake) }
    }
}

unsafe impl task::UnsafeWake for NoWake {
    unsafe fn clone_raw(&self) -> task::Waker { NoWake::waker() }
    unsafe fn drop_raw(&self) {}
    unsafe fn wake(&self) {}
}
```

running it from the `futures` folder with `cargo run --example foo --features nightly` gives:

```
Poll Baz at 0x7ffee1048e98
Poll Baz at 0x7ffee1048e04
```

Which should not be allowed to happen, the pin _must_ point at the same memory location every time.

I believe the error is that the `await!` macro's unsafe pin construction is only valid when it is a part of an `#[async]`  future, when used inside `#[async_move]` there's nothing stopping the generator being moved in between calls to `resume`, causing the pin to point at different memory each time.

CC @withoutboats

Nemo157:
Fixes #924, asserts that we're currently in an immovable generator when polling `StableFuture`s inside `#[async]` functions, and polls via the `Future` trait inside `#[async_move]` functions (so you need to `PinBox` any `StableFuture`s you are `await`ing there).

I'm not certain this is the best solution, but after actually implementing it it seems reasonable to me.

@cramertj I believe you had some reservations about this stack pinning API? Doing a closure based API here is difficult since you can't yield through the closure, I think it should be possible with a couple of layers of extra generators, but I wasn't able to get the lifetimes working out on my first attempt.

r? @withoutboats

### DIFF
--- a/futures-macro-async/src/lib.rs
+++ b/futures-macro-async/src/lib.rs
@@ -191,7 +191,15 @@ if_nightly! {
         });
         syn::token::Semi([block.brace_token.0]).to_tokens(&mut result);
 
+        let await = if pinned {
+            await()
+        } else {
+            await_move()
+        };
+
         let gen_body_inner = quote_cs! {
+            #await
+
             let __e: #output = #result
 
             // Ensure that this closure is a generator, even if it doesn't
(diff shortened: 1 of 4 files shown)


futures-macro-async/src/lib.rs
```diff
@@ -191,7 +191,15 @@ if_nightly! {
...
         syn::token::Semi([block.brace_token.0]).to_tokens(&mut result);
 
+        let await = if pinned {
+            await()
+        } else {
+            await_move()
+        };
+
         let gen_body_inner = quote_cs! {
+            #await
```
cramertj: What happens here if you declare an `#[async]` function inside an `#[async_move]` function (or vice versa)?

futures-macro-async/src/lib.rs
```diff
@@ -631,4 +639,62 @@ if_nightly! {
             |p| AsyncStreamArgs(p.map(|d| d.1.into_iter().collect()).unwrap_or_default())
         ));
     }
+
+    fn await() -> Tokens {
+        quote_cs! {
+            #[allow(unused_macros)]
+            macro_rules! __await {
+                ($e:expr) => ({
+                    let mut future = ::futures::__rt::pinned($e);
```
cramertj: I would just use `new_unchecked` here, and add a fake borrow across the yield point to make sure we're not in a movable generator.

cramertj:
Yes, I have different thoughts about how the stack-pinning API should work, but for now it seems fine to do what you did or just use `new_unchecked`.

futures-macro-async/src/lib.rs
Nemo157: Good question, I assumed macros would follow normal scoping/shadowing rules and this would work. Macros have enough _unique_ behaviour that I'm not certain though, I'll add a test covering this to check.

Nemo157: See the test [`qux2`](https://github.com/rust-lang-nursery/futures-rs/pull/925/files#diff-27836183fc84a021f648e6405d935276R31). I believe this should error if the macro were to leak either from inner -> outer function or outer -> inner function.

cramertj:
This looks great! Thank you.

PR #930, opened 2018-04-01
Add documentation to futures-macro-await macros

Nemo157:


cramertj:
This is awesome! Thanks so much!

PR #932, opened 2018-04-02
fix compilation of #[async(pinned)] and #[async(pinned_send)]

ngg:
On the current master branch, #[async(pinned)] and #[async(pinned_send)] are not working.
It seems to be working for me with this commit, but I'm not a 100% sure if I understand pinning correctly.


ngg:
Closing in favor of #916 (I didn't realize it was already there for a few days)

PR #935, opened 2018-04-03
WIP: no_std support for async macro

Nemo157:
All changes in the first commit should be good to go.

The second commit has a **SUPER UNSAFE** implementation of context passing. It works in my specific use case because I promise to never touch futures in interrupt handlers and run on a single-threaded microcontroller, but definitely needs to be implemented differently to ensure users can't accidently pass the wrong context to different futures.

I have had a brief look at how [`futures` 0.1 handled supporting basically a user overridable thread local implementation](https://docs.rs/futures/0.1.21/futures/task/fn.init.html), putting the onus of safety on the user that called `futures::task::init`. I would really love to avoid having to reintroduce that API (especially having to fully understand that implementation, there appeared to be a lot of implementation details concerned with ensuring any cost introduced by it is optimized out on `std` builds), so any hints on potential alternatives would be greatly appreciated.

If no-one has any potential alternatives, I'll take a better look at the old `futures::task::init` sometime this/next week and see how horrible supporting it will be.


Nemo157:
I never came up with a nice solution to this that doesn't involve changing the definition of generators in the compiler, I'm going to abandon this on 0.2 for now and maybe getting something working for 0.3 (depending on how far away the builtin async support will be).

Issue #918, opened 2018-03-29
Name consistency of async function attributes


raviqqe:
Attributes of `async` and `async_stream` to box them which are currently named `pinned` and `pinned_send` should be the same as `async_move` and `async_stream_move` like `boxed` and `boxed_send` from users' aspects. That is because `async` or `async_stream` functions are pinned anyway whether they are boxed or not.

withoutboats:
The reason for the difference is that they put them in a `PinBox` whereas the move ones put them in a `Box`. But I'm also fine with using the same name for both, I didn't think too hard about this decision.

Nemo157:
I was recently wondering why we have separate `async` and `async_move` functions instead of an `async(move)` attribute argument (although, I think with the `std` nomenclature it should be `async(unpin)` now)? It seems to me most consistent to pass all these modifiers via arguments and try and reduce the set of macros to the minimum possible.

raviqqe:
I mean that it doesn't make sense for users to put `pinned` attribute to `async` or `async_stream` functions as normal `async` or `async_stream` functions which are not in `PinBox` are also pinned. But I'm not sure how other users feel about that because I know their implementation and am not a "pure" user.

withoutboats:
The whole API was just whatever made sense as I was working on it, and it could be a good idea to change it.

We could have three attribute arguments:

* `move`
* `boxed`
* `send`

And the combination of these produces the outcomes we have today.

raviqqe:
Another option is:

- `move`
- `boxed`
- `boxed_send`

as the `send` attribute is always used with `boxed` but cannot without it.

withoutboats:
Its true that `send` without `boxed` wouldn't mean anything right now. I wonder if it has to, though? Couldn't we generate `impl Future + Send` as a way of checking that your future stays `Send`?

raviqqe:
Is there any use case where `impl Future + Send` is necessary? I guess the `Send` annotation is guessed automatically by Rust compiler because it's just like a shorthand for `fn foo<I: Future>() -> I` and `impl Trait` indicates the least requirement unlike trait objects.

withoutboats:
You're right, its not necessary. But if you wanted to get an error if you accidentally made your future `!Send`, you could use this to get that (basically as a check on yourself).

Nemo157:
[...]

PR #936, opened 2018-04-03
Update async_block* macros to support Stable & Unpin 

Nemo157:
I'm not sold on the name, especially since it uses a completely different meaning to the `pinned` argument to `#[async]`. I could do the same as has happened to `#[async]` and change `async_block` to return a `StableFuture` along with a new `async_block_move` macro for `Future`.

My preferred way to do this would be to keep a single `async_block` macro and pass an argument to it as posited in #918 for `#[async(move)]` (along with the change to `StableFuture` by default); I've been trying to come up with a way to nicely pass attributes to the `async_block` macro, but I can't think of anything I'm really attached to. The best I came up with was something like:

```rust
async_block!(move, {
    await!(bar(&x));
});
```

which I believe is unambiguous because `$($ident,)+ { $($tt)+ }` cannot be parsed as an `expr`.

### DIFF
--- a/futures-macro-async/src/lib.rs
+++ b/futures-macro-async/src/lib.rs
@@ -456,6 +456,35 @@ if_nightly! {
         let expr = ExpandAsyncFor.fold_expr(expr);
 
         let mut tokens = quote_cs! {
+            ::futures::__rt::gen_pinned
+        };
+
+        // Use some manual token construction here instead of `quote_cs!` to ensure
+        // that we get the `call_site` span instead of the default span.
+        let span = Span::call_site();
+        syn::token::Paren(span).surround(&mut tokens, |tokens| {
+            syn::token::Static(span).to_tokens(tokens);
+            syn::token::Move(span).to_tokens(tokens);
+            syn::token::OrOr([span, span]).to_tokens(tokens);
+            syn::token::Brace(span).surround(tokens, |tokens| {
+                (quote_cs! {
+                    if false { yield ::futures::__rt::Async::Pending }
+                }).to_tokens(tokens);
+                expr.to_tokens(tokens);
+            });
+        });
+
+        tokens.into()
+    }
+
+    #[proc_macro]
+    pub fn async_block_unpin(input: TokenStream) -> TokenStream {
+        let input = TokenStream::from(TokenTree::Group(Group::new(Delimiter::Brace, input)));
+        let expr = syn::parse(input)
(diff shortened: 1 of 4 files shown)


cramertj:
Perhaps the naming here should be inverted to `async_block` and `movable_async_block`?

cc @withoutboats

awestlake87:
`async_block!` is essentially an `#[async]` closure with empty params. Maybe the answer is to keep the attribute the same and apply it to a `|| { }` expression instead.

Nemo157:
@awestlake87 unfortunately the async/await RFC has [reserved that syntax for async closures](https://github.com/withoutboats/rfcs/blob/async-await/text/0000-async_await.md#async--closures) (which I'm slightly tempted to try implementing at some point). It does [mention](https://github.com/withoutboats/rfcs/blob/async-await/text/0000-async_await.md#initialization-pattern) having an `async_block! { ... }` macro which may become something like `async { ... }` in the future, but presumably relies on the eventual automatic `impl !Unpin for <unnamed self referential Generator>` to distinguish between `Unpin` and immovable futures returned from that macro.

I think changing `async_block!` to be immovable by default is probably the best way to go, if the argument naming in #964 is accepted then I'm leaning toward `async_block_unpin!` for the `Unpin` variant, otherwise I guess `async_block_move!`. I think having a suffix rather than a prefix keeps this more consistent with the `async` macro taking arguments, it's sort of like an argument built into the macro name itself. I'll rebase and update to that now, but feel free to disagree here or there about the naming.

awestlake87:
@Nemo157 I see, I forgot it needed to be called in order to act as a future. The macro does look cleaner, although I'd definitely prefer the first class syntax in that RFC in the longterm. I'm glad there's talk about adding futures to core and stabilizing async/await separately from proc_macro2 and generators.

As for the naming, I agree with @Nemo157 when it comes to changing async_block! to immovable by default because that's how closures act in rust. I think I'm leaning more towards `async_block_move!` than `async_block_unpin!` though, as it's more consistent with rust's concept of `move` for closures. Pin/Unpin is a concept that users will need to learn in order to understand why `yield` points make borrowing unsafe, but I don't think it conveys enough information about how it's treating the environment around the block to someone who's naively using async/await.

withoutboats:
I'm inclined to do just remove the current `async_block` and only support the unpinned version, per my comment on #964.

Nemo157:
Subsumed by #964

PR #939, opened 2018-04-04
Allow creation of CpuPool to be fallible.

djg:
To allow error handling in the face of errors returned by thread creation.

The use of `futures-cpupool` in Gecko leads to content process `panic!` because of use of `unwrap()` in pool creation.

### DIFF
--- a/futures-cpupool/src/lib.rs
+++ b/futures-cpupool/src/lib.rs
@@ -19,7 +19,7 @@
 //! # fn main() {
 //!
 //! // Create a worker thread pool with four threads
-//! let pool = CpuPool::new(4);
+//! let pool = CpuPool::new(4).unwrap();
 //!
 //! // Execute some work on the thread pool, optionally closing over data.
 //! let a = pool.spawn(long_running_future(2));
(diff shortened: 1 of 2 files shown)


cramertj:
This seems fine to me. @aturon @alexcrichton wdyt?

alexcrichton:
👍

Nemo157:
Should this be in `0.1`? Is a breaking release of `futures-cpupool(0.2)` based on `futures(0.1)` planned or should this be done for `futures(0.2)::executor::ThreadPool` instead?

cramertj:
Ah, I missed that this was pointed at the 0.1 branch. Thanks for the heads-up! @djg could you open a new PR targeting master? I'll revert this one.

cramertj:
@djg specifically, I think you want to change this line: https://github.com/rust-lang-nursery/futures-rs/blob/master/futures-executor/src/thread_pool.rs#L259

djg:
@cramertj, I really wanted to modify 0.1 because we're using version 0.1.8 in Gecko and I didn't want to wait to upgrade to version 0.2 before having the ability to handle `CpuPool` creation failure gracefully.

What's the process for making a breaking change to 0.1? Do I need to fork the crate?

cramertj:
@djg Unfortunately, we can't make breaking changes to the 0.1 crate. We could add another method that returns `Result` and you could use that one-- would that work for you?

alexcrichton:
Yeah adding an API like that seems fine by me

PR #947, opened 2018-04-06
Stage 0.2 Release

cramertj:


Nemo157:
Should `futures-stable` be removed from the facade as well (it appears building with the `nightly` feature is broken because it's attempted to be facaded but not actually enabled)?

cramertj:
@Nemo157 Ah, I was just a flag too overzealous when removing the nightly features deps. I'll add back the dependency (building w/ nightly fails CI right now anyways because of breaking changes to the proc macro api).

Nemo157:
Damn, I was sort of hoping that `StableFuture` was not going to be published in 0.2 yet. I'm just about to propose a breaking change to how it relates to `Future` to hopefully make it more ergonomic to use.

cramertj:
@Nemo157 yeah, I think it's fine for us to make breaking changes to something like `StableFuture` (which is only usable on nightly and doesn't currently have much of a point without async/await).

Nemo157:
> doesn't currently have much of a point without async/await

Not entirely no point, I have 4 or 5 manual implementations (2 for supporting DMA and a few basic adaptors that can't be written via async/await).

cramertj:
No, not entirely :)

jimmycuadra:
What was the context of this? I had async/await working using futures master from a couple days ago, but it doesn't work now with the real 0.2.0 release. PRs with no description make me sad. :{

cramertj:
@jimmycuadra An API change to proc macros on nightly broke the async/await macros, so they were removed prior to the release. There will be another release next week with them added back.

PR #950, opened 2018-04-06
Test collecting into JoinAll properly

matklad:
cc https://github.com/rust-lang-nursery/futures-rs/pull/773

### DIFF
--- a/futures/tests/all.rs
+++ b/futures/tests/all.rs
@@ -269,6 +269,14 @@ fn join_all_iter_lifetime() {
 }
 
 #[test]
+fn join_all_from_iter() {
+    assert_done(
+        || vec![f_ok(1), f_ok(2)].into_iter().collect::<JoinAll<_>>(),
+        Ok(vec![1, 2]),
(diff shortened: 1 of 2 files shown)


futures/tests/all.rs
```diff
@@ -268,6 +268,14 @@ fn join_all_iter_lifetime() {
     assert_done(|| sizes(vec![&[1,2,3], &[], &[0]]), Ok(vec![3, 0, 1]));
 }
 
+#[test]
+fn join_all_from_iter() {
+    assert_done(
+        || vec![f_ok(1), f_ok(2)].into_iter().collect::<JoinAll<_>>(),
```
stuhood: Is the turbofish actually necessary in general, or can it be inferred?

matklad: It is unfortunately necessary in general: at the call site, you know that you need some T: Future, but you don’t know whit T specifically are you interested in.

cramertj:
Thanks for fixing this!

Issue #952, opened 2018-04-07
future::shared is unsafe (Send/Sync issue)


dathinab:
Hy,
`futures::future::shared` allows Sending non Send types to other threads
and provides Sync shared access for non Sync types, too.
# Problem

Send/Sync bounds for `future::Shared` are wrong

# Explanation

The `Shared` futures works similar to a Mutex, 
as it makes sure there is only mutably-exclusive
access wrt. to poll-ing the underlying future. 
It should only be `Send`/`Sync` if the underlying future is `Send`.

The created `SharedItem`, `SharedError` are basically `Arc`'s
and implement `Send`/`Sync` correct for themself, the problem
is that they are crated using unsafe code when polling
any clone of `Shared` which completes or was completed,
as such they allow Sync access to a non-sync item/error.

# Mini Prog of Concept

[play](https://play.rust-lang.org/?gist=89b9112cfc5778b409b39fc164464a2f&version=stable)
[gist](https://gist.github.com/89b9112cfc5778b409b39fc164464a2f)

```rust
extern crate futures;

use std::thread;
use std::cell::RefCell;

use futures::executor::block_on;
use futures::{future, FutureExt};

fn main() {
    // we use *const u8 as it is neither Send+Sync, through
    // using just e.g. RefCell<&str> would already show part
    // of the bug
    let cell_fut = future::ok::<_,()>(RefCell::new(1usize as *const u8));
    let shared = cell_fut.shared();
    let shared2 = shared.clone();
    let shared3 = shared.clone();
    let join1 = thread::spawn(move || {
        // wups RefCell is not Sync, but we access it
        // here and in the other thread. Even worse the
        // item in the RefCell is not Send but we get
        // mutable access to it here (which requires Send);
        let cell = block_on(shared2).unwrap(); 
        *cell.borrow_mut() = 2usize as *const u8;
    });
    let join2 = thread::spawn(move || {
        let cell = block_on(shared3).unwrap();
        *cell.borrow_mut() = 4usize as *const u8;
    });
    
    join1.join().unwrap();
    join2.join().unwrap();
    
    let cell = block_on(shared).unwrap();
    println!("raced to get: {:?}", *cell.borrow() as usize);
}
```

# Solution

I will make a pull request soon. The solution is
to only impl Send/Sync if the Future is Send
and if the Item+Error are Send+Sync

(also if there would be a method to
 access the inner future it would have
 to be limited to inner futures which
 are `Sync`)

[...]

PR #953, opened 2018-04-07
Fixes future::Shared's Send/Sync bounds (Issue #952)

dathinab:
Send/Sync where implemented even if Shared
should not have been Send/Sync (Issue #952)

### DIFF
--- a/futures-util/src/future/shared.rs
+++ b/futures-util/src/future/shared.rs
@@ -234,8 +234,17 @@ impl Wake for Notifier {
     }
 }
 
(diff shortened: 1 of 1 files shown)


dathinab:
I took a look at the failed checks, and it seems they are
_unrelated_ to the pull request, basically it seems as if
some nightly parts broke, as there are:

1. errors wrt. proc_macro
2. errors wrt. ui/error message tests

cramertj:
Thanks for the fix!

PR #961, opened 2018-04-12
Add Sink and Stream implementation for VecDeque

NikVolf:
Using `VecDeque` for being both `Sink` and `Stream` in library that uses futures is pretty natural for cases where consuming api might be both synchronous and asynchronous, on the sync side, also for tests


cramertj:
Cool, thanks!

PR #962, opened 2018-04-12
Rename pinned attribute

raviqqe:
The first step to solve the name inconsistency issue of `async` attributes #918.


raviqqe:
May be closed in preference to #964.

PR #963, opened 2018-04-12
Fix some warnings caused by test codes

raviqqe:


raviqqe:
May be closed in preference to #964.

cramertj:
Thanks!

PR #964, opened 2018-04-12
Drop support for non-immovable #[async] macros

### LINKED ISSUE #918: Name consistency of async function attributes

Nemo157:
Fixes #918
Subsumes #936


raviqqe:
Personally, I like your naming of `unpin`. But, can `boxed` be `box` instead as `unpin` sounds like a verb?

withoutboats:
How do you feel about removing the `unpin`/`async_move` feature entirely? It seems like it requires duplicating literally every API - including `await` - and the current version of the RFC for built-in syntax doesn't support this at all.

Nemo157:
I'm :+1: for that, I'm only using immovable futures personally. If the wider futures community is planning on skipping 0.2 for 0.3 (which has been my impression of at least tokio and hyper) I think there's less need to try and provide a compatible macro for experimenting with. I don't want to make that call though, since I'm working with quite a niche usage of futures at the moment. If you/someone makes the call I can update this PR to remove it.

withoutboats:
delete it! :tada:

withoutboats:
LGTM, will merge after the tests run

PR #965, opened 2018-04-13
Add a flatten_sink method to FutureExt

ctiller:
Analogous to flatten_stream, this takes a Future<Sink> and transforms it
into a Sink that waits for the future to be resolved before poll_ready
returns Async::Ready.

### DIFF
--- a/futures-util/src/future/flatten_sink.rs
+++ b/futures-util/src/future/flatten_sink.rs
@@ -1,0 +1,80 @@
+use core::fmt;
+
+use futures_core::{task, Async, Future};
+use futures_sink::Sink;
+
+#[derive(Debug)]
+enum State<F> where F: Future, <F as Future>::Item: Sink {
+    Waiting(F),
+    Ready(F::Item),
+    Closed,
+}
+
+/// Future for the `flatten_sink` combinator, flattening a
+/// future-of-a-sink to get just the result of the final sink as a sink.
+///
+/// This is created by the `Future::flatten_sink` method.
+pub struct FlattenSink<F> where F: Future, <F as Future>::Item: Sink {
+    st: State<F>
+}
+
+impl<F> fmt::Debug for FlattenSink<F>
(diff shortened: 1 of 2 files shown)


futures-util/src/future/flatten_sink.rs
```diff
@@ -0,0 +1,81 @@
+use core::fmt;
+
+use futures_core::{task::Context, Async, Future};
```
cramertj: This syntax (the nested multi-element path inside a group) wasn't around in in 1.20, and we try to stay backcompat for the last several major versions of Rust. Can you update to either have a separate import for `futures_core::task::Context` or import `futures::core{task, Async, Future}` and use `task::Context`?

futures-util/src/future/flatten_sink.rs
```diff
@@ -0,0 +1,81 @@
...
+            State::Ready(ref mut s) => s.poll_flush(cx),
+            // if sink not yet resolved, nothing written ==> everything flushed
+            State::Waiting(_) => Ok(Async::Ready(())),
+            State::Closed => panic!("poll_flush called after eof"),
+        }
+    }
+
+    fn poll_close(&mut self, cx: &mut Context) -> Result<Async<()>, Self::SinkError> {
+        match self.st {
+            State::Ready(ref mut s) => return s.poll_close(cx),
```
cramertj: I think this should be `try_ready!(s.poll_close(cx))`. In the case where it successfully closes, you want to proceed to the next line and set `self.st` to `State::Closed`.

cramertj:
LGTM modulo the two nits

futures-util/src/future/flatten_sink.rs
ctiller: Fixed.

futures-util/src/future/flatten_sink.rs
ctiller: Fixed.

PR #968, opened 2018-04-14
Add `oneshot::Receiver::try_recv`.

c0gent:
Attempts to receive a message outside of context of a task, such as
within a `Drop` impl. It does not schedule a task wakeup or have any
other side effects.

### DIFF
--- a/futures-channel/src/oneshot.rs
+++ b/futures-channel/src/oneshot.rs
@@ -246,8 +246,6 @@ impl<T> Inner<T> {
                     return Ok(Some(data.into()));
                 }
             }
-            // Should there be a different error value or a panic in the case
-            // where `self.data.try_lock() == None`?
             Err(Canceled)
         } else {
             Ok(None)
--- a/futures-channel/src/oneshot.rs
+++ b/futures-channel/src/oneshot.rs
@@ -237,6 +237,23 @@ impl<T> Inner<T> {
         }
     }
 
+    fn try_recv(&self) -> Result<Option<T>, Canceled> {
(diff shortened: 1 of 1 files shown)


futures-channel/src/oneshot.rs
```diff
@@ -237,6 +237,23 @@ impl<T> Inner<T> {
...
+    fn try_recv(&self) -> Result<Option<T>, Canceled> {
+        // If we're complete, either `::close_rx` or `::drop_tx` was called.
+        // We can assume a successful send if data is present.
+        if self.complete.load(SeqCst) {
+            if let Some(mut slot) = self.data.try_lock() {
+                if let Some(data) = slot.take() {
+                    return Ok(Some(data.into()));
+                }
+            }
+            // Should there be a different error value or a panic in the case
```
cramertj: I think this error is fine-- it's the same thing you'd get if you tried calling `recv` after it had already completed successfully.

futures-channel/src/oneshot.rs
```diff
@@ -331,8 +346,8 @@ impl<T> Sender<T> {
     ///
     /// # Return values
     ///
-    /// If `Ok(Ready)` is returnd then the associated `Receiver` has been dropped,
-    /// which means any work required for sending should be cancelled.
+    /// If `Ok(Ready)` is returned then the associated `Receiver` has been
+    /// dropped, which means any work required for sending should be canceled.
```
cramertj: was cancelled changed on purpose? it was correct before

c0gent: Cancelled can be spelled either way and the convention in this lib is apparently to use 'canceled'.

Issue #970, opened 2018-04-15
Async iteration for IntoIter<Item=Future>


HyeonuPark:
(Edit: This issue was "Async iteration and IntoStream" but renamed as described at next comment)

I'm pretty sure we need #[async] for loop at some point!

I think previously futures-await crate used `Stream` for this purpose. But with this approach we can't support iterating over `Iterator<Item=Future>` cases correctly. I tried to implement `Stream` for various `Iterator`s, but it failed as `StreamExt` has tons of methods that conflict with `Iterator` and it's definitely intensional choice.

And this is virtually same reason why normal for loop uses `IntoIter` instead of `Iterator` directly. With `IntoIter`, we can add various methods to `Iterator` without exposing them to types like `Vec`. So I suggest a new trait `IntoStream`, with a single required method that converts `self` to `Stream`.

HyeonuPark:
After some experiment I figured out that `IntoStream` trait as I described before is not possible. It always conflict that `impl<S: Stream> IntoStream for S` with `impl<I: Iterator<Future>> IntoStream for StreamIter<I>`. So the only option for #[async] for loop is `Stream` trait.

But I still think async iteration over `IntoIter<Future>` should be supported near-natively. So my next suggestion is a function that converts `IntoIter<Future>` to `Stream`, just like `future::result`.

Nemo157:
There are two functions for converting an iterator of futures to a stream: [`futures_ordered`](https://docs.rs/futures/0.2.0/futures/stream/fn.futures_ordered.html) and [`futures_unordered`](https://docs.rs/futures/0.2.0/futures/stream/fn.futures_unordered.html). Having distinct functions is necessary because an `Iterator<Future>` is fundamentally different to a `Stream`, a `Stream` has a distinct order in which items will resolve, whereas an `Iterator<Future>` could be something like `[Timeout::from_secs(5), Timeout::from_secs(3), Timeout::from_secs(1)].iter()` and have its items resolve in the opposite order to the iteration order.

PR #971, opened 2018-04-15
Add functions to create a new stream

HyeonuPark:
`stream::from_iter` creates a stream from a sequence of futures. Related issue #970 
`stream::empty` creates an empty stream.
`stream::once` creates a single-element stream from a future.


HyeonuPark:
Closed with #970

PR #972, opened 2018-04-18
Either fixes

cramertj:
`.left` and `.right` overlap w/ the inherent methods on the `Either` type, which makes chaining painful (e.g. you need `Either::right(fut.left())` rather than `fut.left().right()`).

This also adds an either impl for `Sink`.


aturon:
r+

PR #974, opened 2018-04-18
Add for_each_concurrent

cramertj:


futures-util/src/stream/mod.rs
```diff
@@ -599,6 +601,30 @@ pub trait StreamExt: Stream {
...
 
+    /// Runs this stream to completion, executing the provided closure for each
+    /// element on the stream. This is similar to `for_each` but may begin
+    /// processing an element while previous elements are still being processed.
+    ///
+    /// When this stream successfully resolves to an item, the closure will be
+    /// called to produce a future. That future will then be added to
+    /// the set of futures to resolve.
+    ///
+    /// The returned value is a `Future` where the `Item` type is `()` and
```
aturon: The docs don't seem to match, since you're actually returning back the stream.

aturon:
r+ modulo minor nit

futures-util/src/stream/mod.rs
cramertj: oh good catch! that's wrong on for_each, too

cramertj: Fixed.

futures-util/src/stream/for_each_concurrent.rs
```diff
@@ -0,0 +1,77 @@
+use futures_core::{Async, Future, IntoFuture, Poll, Stream};
+use futures_core::task;
+
+use super::futures_unordered::FuturesUnordered;
+
+/// A stream combinator which executes a unit closure over each item on a
+/// stream concurrently.
+///
+/// This structure is returned by the `Stream::for_each` method.
```
tikue: `Stream::for_each` -> `Stream::for_each_concurrent`

tikue:
When would one prefer this over `stream.for_each(|item| tokio::spawn(f(item)))`?

PR #975, opened 2018-04-18
Transition futures-core to 0.3.0-alpha

aturon:
Migrates the futures-core crate to the 0.3 design currently under
RFC consideration. In particular:

- Introduces `Pin` throughout
- Removes the `Error` associated type
- Removes task-local data


futures-core/src/executor.rs
```diff
@@ -1,62 +1,61 @@
...
-#[cfg(not(feature = "std"))]
-pub(crate) trait Executor {}
+/// The result of a failed spawn
+#[derive(Debug)]
+pub struct SpawnError<T> {
+    /// The kind of error
+    pub kind: SpawnErrorKind,
+
+    /// The task for which spawning was attempted
+    pub task: T,
```
cramertj: As I mentioned on IRC, I don't know that we want to commit ourselves to handing back the task, since this won't be possible when we move to unsized rvalues (at least not with some novel new stack-DST-returning ABI).

aturon: So, interesting thing: this does *not* make such a commitment for the unsized-rvalue spawning, which will be a distinct method (called `spawn`) once it's possible.

In that world, `spawn_obj` will remain available though largely unused -- but perhaps one use-case for it is precisely when you need to be able to get your task back, at the cost of needing to pre-allocate it.

In that case the eventual `spawn` method would return a `SpawnErrorKind` rather than `SpawnError`.

wdyt? if that seems worthwhile, I should probably rejigger things to get rid of the type parameter and call it `SpawnObjError`.

futures-core/src/lib.rs
```diff
@@ -20,19 +22,33 @@ macro_rules! if_std {
     )*)
 }
 
-#[macro_use]
+#[macro_export]
+macro_rules! pinned_deref {
```
cramertj: Note that these functions aren't safe in general for arbitrary structures, especially with my proposed extension (which would block you from projecting from a `Pin<ManuallyDrop<T>>` to `Pin<T>`). Other examples include structures which have public fields but which also rely on `Pin` methods as a mechanism to ensure that those fields aren't mutated.

aturon: Fixed.

futures-core/src/future/either.rs
```diff
@@ -1,22 +1,26 @@
...
+        unsafe {
+            match *(Pin::get_mut(&mut self)) {
+                Either::Left(ref mut a) => Pin::new_unchecked(a).poll(cx),
+                Either::Right(ref mut b) => Pin::new_unchecked(b).poll(cx),
+            }
         }
     }
 }
 
+/*
```
cramertj: Did you mean to leave this commented out?

futures-core/src/executor.rs
cramertj: yeah, we could add a different method later on, but it seems like it sets up the expectation that users might want their task back, and the name `SpawnError` suggests that it would be the error for the `spawn` method. I think it's fine to keep it this way with some renaming etc.

cramertj:
r+

futures-core/src/executor.rs
aturon: Done

PR #985, opened 2018-04-24
Channel refactor

### LINKED ISSUE #800: Rethinking the bounded mpsc channel

cramertj:
Refactors channels to use "waker" terminology rather than "parking"

Splits unbounded channels into a separate, simplified underlying type

Calls poll_ready in poll_flush impl for bounded channels in order to
provide backpressure on <Sender as Sink>::send.

Replacement for #984.

Fix #800
Fix rust-lang-nursery/net-wg#11

cc @danburkert, @seanmonstar, @carllerche

### DIFF
--- a/futures-channel/src/mpsc/mod.rs
+++ b/futures-channel/src/mpsc/mod.rs
@@ -5,18 +5,19 @@
 //! [`Sender`](Sender) handles. [`Receiver`](Receiver) implements
 //! [`Stream`](futures_core::Stream) and allows a task to read values out of the
 //! channel. If there is no message to read from the channel, the current task
-//! will be notified when a new value is sent. [`Sender`](Sender) implements the
+//! will be awoken when a new value is sent. [`Sender`](Sender) implements the
 //! `Sink` trait and allows a task to send messages into
 //! the channel. If the channel is at capacity, the send will be rejected and
-//! the task will be notified when additional capacity is available. In other
-//! words, the channel provides backpressure.
+//! the task will be awoken when additional capacity is available. This process
+//! of delaying sends beyond a certain capacity is often referred to as
+//! "backpressure".
 //!
-//! Unbounded channels are also available using the [`unbounded`](unbounded)
-//! constructor.
+//! Unbounded channels (without backpressure) are also available using
+//! the [`unbounded`](unbounded) function.
 //!
 //! # Disconnection
 //!
-//! When all [`Sender`](Sender) handles have been dropped, it is no longer
+//! When all [`Sender`](Sender)s have been dropped, it is no longer
 //! possible to send values into the channel. This is considered the termination
 //! event of the stream. As such, [`Receiver::poll_next`](Receiver::poll_next)
 //! will return `Ok(Ready(None))`.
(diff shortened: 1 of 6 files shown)


futures-channel/tests/mpsc.rs
```diff
@@ -401,7 +401,8 @@ fn stress_close_receiver_iter() {
             Some(r) => assert!(i == r),
             None => {
                 let unwritten = unwritten_rx.recv().expect("unwritten_rx");
-                assert_eq!(unwritten, i);
+                assert!(unwritten <= i + 1);
```
cramertj: This is necessary in order to prevent this test from flaking due to the race condition when writing into a closed channel.

carllerche:
It's a pretty large diff and hard for me to reason about due to how terminology changes are mixed with code and behavior changes.

futures-channel/tests/mpsc.rs
carllerche: Skimming over it looks like the unbounded variant also has the same issue?

cramertj: Yes, they both have this issue.

cramertj:
@carllerche 
> Could you explain the reasoning behind splitting up the unbounded & bounded implementations?

Sure! I thought it was clearer and easier to read with them split up this way, and seemed like it might be more performant since there are fewer queues, atomics, and other checks needed to supply simple receiver wakeups than the comparatively complex sender wakeup semantics.

cramertj:
Before this PR:

```
test bounded_100_tx        ... bench:     290,574 ns/iter (+/- 2,578)
test bounded_1_tx          ... bench:     278,866 ns/iter (+/- 11,897)
test unbounded_100_tx      ... bench:     150,284 ns/iter (+/- 7,048)
test unbounded_1_tx        ... bench:     134,440 ns/iter (+/- 3,558)
test unbounded_uncontended ... bench:      66,844 ns/iter (+/- 5,163)
```

After this PR:

```
test bounded_100_tx        ... bench:     281,201 ns/iter (+/- 1,249)
test bounded_1_tx          ... bench:     278,706 ns/iter (+/- 36,851)
test unbounded_100_tx      ... bench:     105,909 ns/iter (+/- 22,730)
test unbounded_1_tx        ... bench:      98,740 ns/iter (+/- 2,073)
test unbounded_uncontended ... bench:      59,876 ns/iter (+/- 5,405)
```

carllerche:
The main argument against duplicating code is that it introduces more locations for bugs. For example, the close race bug is now duplicated in two code locations.

I would suggest that, if performance is the driver, at the very least benchmarks illustrating the performance difference should be present and probably at least a basic tuning pass of the bounded mpsc should be done (given that none have been done to date).

carllerche:
At the end of the day, you are the ones who have to maintain it, so feel free to do whatever.

futures-channel/src/mpsc/mod.rs
```diff
@@ -940,37 +789,35 @@ impl<T> Stream for Receiver<T> {
...
-                            // A message has been sent while attempting to
-                            // park. Loop again, the next iteration is
-                            // guaranteed to get the message.
-                            continue;
-                        }
-                    }
-                }
-            };
-            // Return the message
+        // Try to read a message off of the message queue.
```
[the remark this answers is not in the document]
cramertj: Done

PR #986, opened 2018-04-25
Bring 0.3 branch up to date with new RFC

aturon:
This PR updates the 0.3 work to match the [new RFC](https://github.com/rust-lang/rfcs/pull/2418).

The approach here keeps the `Future` trait with an `Error` type, but uses the new `Poll` type; an alias `PollResult<T, E> = Poll<Result<T, E>>` is defined.

The combinators for `Future` are added back to this branch and updated with the `Poll` changes, but otherwise are as in 0.2. We'll need follow up with changes to bring them into line, e.g. renaming `map` to `map_ok` etc.

My plan is to follow this same approach to get the crates back into working condition (including tests), and then circle back to harmonize the combinators.


futures-core/src/async_trait/mod.rs
```diff
@@ -0,0 +1,210 @@
...
+use core::marker::Unpin;
+
+use {Poll, PollResult, task};
+
+#[cfg(feature = "either")]
+mod either;
+
+/// An `Async` value represents an asychronous computation.
+///
+/// An `Asyn` is a value that may not have finished computing yet. This kind of
```
cramertj: s/`Asyn`/`Async`

futures-core/src/async_trait/mod.rs
```diff
@@ -0,0 +1,210 @@
...
+/// further progress by `poll`ing again. The wake up is performed using
+/// `cx.waker()`, a handle for waking up the current task.
+///
+/// When using an async value, you generally won't call `poll` directly, but instead
+/// use combinators to build up asynchronous computations. A complete
+/// computation can then be spawned onto an
+/// [executor](../futures_core/executor/trait.Executor.html) as a new, independent
+/// task that will automatically be `poll`ed to completion.
+///
+/// # Combinators
```
cramertj: Here and above you mention combinators on `Async`-- this should link to `AsyncExt` (when it exists).

futures-core/src/async_trait/mod.rs
```diff
@@ -0,0 +1,210 @@
...
+
+    /// The type of failures
+    type Error;
+
+    /// Poll this `AsyncResult` as if it were an `Async`.
+    ///
+    /// This method is a stopgap for a compiler limitation that prevents us from
+    /// directly inheriting from the `Async` trait; in the future it won't be
+    /// needed.
+    fn poll(self: Pin<Self>, cx: &mut task::Context) -> PollResult<Self::Item, Self::Error>;
```
cramertj: This should include `into_async` so that `AsyncResult`s can be `await`ed (well, I suppose it could go on `AsyncResultExt`, but the method should exist somewhere).

futures-core/src/future/mod.rs
```diff
@@ -100,75 +100,83 @@ pub trait Future {
...
+    type Error;
+
+    /// Consumes this object and produces a future.
+    fn into_future(self) -> Self::Future;
 }
 
-/// Create a future that is immediately ready with a value.
-pub fn ready<T>(t: T) -> ReadyFuture<T> {
-    ReadyFuture(Some(t))
+impl<F> IntoFuture for F where F: Future {
```
cramertj: I think you also want an `impl<A> IntoFuture for A where A: AsyncResult + Unpin { ... }`

futures-core/src/future/result.rs
```diff
@@ -0,0 +1,77 @@
+use {Future, IntoFuture, Poll, PollResult};
+use task;
+
+/// A future representing a value that is immediately ready.
+///
+/// Created by the [`result`](::future::result), [`ok`](::future::ok) or
+/// [`err`](::future::err) functions.
+#[derive(Debug, Clone)]
+#[must_use = "futures do nothing unless polled"]
+pub struct FutureResult<T, E> {
```
cramertj: Might as well add an impl of `Async` for `FutureResult`.

futures-core/src/never.rs
```diff
@@ -0,0 +1,39 @@
...
+/// This type is a stable equivalent to the `!` type from `std`.
+#[derive(Debug, Copy, Clone, Eq, PartialEq, Ord, PartialOrd)]
+pub enum Never {}
+
+impl Never {
+    /// Convert the `Never` type into any other type.
+    pub fn never_into<T>(self) -> T {
+        match self {}
+    }
+}
```
cramertj: Also want `Async` for `Never`.

futures-core/src/poll.rs
```diff
@@ -38,8 +38,48 @@ impl<T> Poll<T> {
...
+
 impl<T> From<T> for Poll<T> {
     fn from(t: T) -> Poll<T> {
         Poll::Ready(t)
     }
 }
+
+/// Shorthand for a `Poll<Result<_, _>>` value.
+pub type PollResult<T, E> = Poll<Result<T, E>>;
+
```
cramertj: It seems like we should also have a macro here for propagating `Pending` that works on non-`Result`y `Poll`s.

cramertj:
The general strategy in this PR takes is to copy all of the combinators over to `Future`. Is there a reason for organizing it this way, rather than having single combinator types which implement either `Async` and `Future` based on their type parameters?

aturon:
Closing this out, as I'm taking a different approach now.

PR #994, opened 2018-04-29
Fix async/await support for latest nightly

ngg:
After https://github.com/rust-lang/rust/pull/50120 procedural macros do not receive the parentheses as part of their attributes TokenStream.
Also the proc_macro_non_items feature is necessary to use async_block.


cramertj:
Thanks!

PR #995, opened 2018-05-02
WIP: revamp 0.3 branch

aturon:
This is a WIP PR, just to make it easy to keep folks in the loop as I'm working through ideas on 0.3. **Please don't take anything here too seriously!**


aturon:
@cramertj This is my first stab at a unified `Future` trait that provides ergonomic support for working with `Unpin`.

It turns out that the "default impl" stuff is not, in fact, sufficiently implemented to use. So instead I've written an `unpinned!` macro for wrapping a `Future` or `Stream` impl in.

With the macro, you write a `Future` or `Stream` impl that provides only `poll_mut` or `poll_next_mut`, respectively, and the macro gives you `poll`/`poll_next`, *and* implements `Unpin` for your type.

We need to get to the bottom of the design around `Unpin`, but ideally we can make this macro completely safe to use. You can see some uses in the commit for futures-channel.

aturon:
Note also: went with `TryFuture` for the `Result` variant.

futures-core/src/future/mod.rs
```diff
@@ -110,61 +103,140 @@ pub trait Future {
...
-}
 
-impl<'a, F: ?Sized + Future> Future for &'a mut F {
-    type Output = F::Output;
-
-    fn poll(mut self: Pin<Self>, cx: &mut task::Context) -> Poll<Self::Output> {
-        unsafe { pinned_deref!(self).poll(cx) }
+    /// A convenience for calling `poll` when a future implements `Unpin`.
+    #[cfg(feature = "nightly")]
+    fn poll_mut(&mut self, cx: &mut task::Context) -> Poll<Self::Output> where Self: Unpin {
```
cramertj: :bike: :house:: `poll_unpin`?

futures-core/src/impls.rs
```diff
@@ -0,0 +1,132 @@
+use {task, Future, Stream, Poll, Unpin};
+
+#[cfg(feature = "either")]
+use either::Either;
+
+unsafe impl<T: Unpin> Unpin for Box<T> {}
+unsafe impl<T: Unpin> Unpin for Option<T> {}
+unsafe impl<'a, T: Unpin> Unpin for &'a T {}
+unsafe impl<'a, T: Unpin> Unpin for &'a mut T {}
```
cramertj: Also `unsafe impl<T: Unpin> Unpin for AssertUnwindSafe<T> {}`

futures-core/src/impls_nightly.rs
```diff
@@ -0,0 +1,134 @@
...
+        fn poll(mut self: Pin<Self>, cx: &mut task::Context) -> Poll<Self::Output> {
+            unsafe {
+                let this = PinBox::get_mut(Pin::get_mut(&mut self));
+                let re_pinned = Pin::new_unchecked(this);
+                re_pinned.poll(cx)
+            }
+        }
+    }
+
+    impl<'a, F: Future> Future for ::std::panic::AssertUnwindSafe<F> {
```
cramertj: This is safe because `AssertUnwindSafe` will inherit `Unpin`ned-ness from its field (of type `F`).

futures-core/src/task/context.rs
```diff
@@ -81,7 +81,26 @@ if_std! {
         ///
         /// To handle executor errors, use [executor()](self::Context::executor)
         /// instead.
-        pub fn spawn<F>(&mut self, f: F) where F: Future<Output = ()> + 'static + Send {
+        #[cfg(not(feature = "nightly"))]
+        pub fn spawn<F>(&mut self, f: F)
+            where F: Future<Output = ()> + 'static + Send + ::Unpin
```
cramertj: Why `+ Unpin`?

futures-core/src/task/mod.rs
```diff
@@ -120,10 +121,38 @@ if_std! {
         }
     }
 
+    #[cfg(not(feature = "nightly"))]
+    unsafe impl<F: Future<Output = ()> + Send + 'static + ::Unpin> UnsafePoll for Box<F> {
+        fn into_raw(self) -> *mut () {
+            unsafe {
+                mem::transmute(self)
```
cramertj: `Box::into_raw(self) as *mut ()`

futures-core/src/task/mod.rs
```diff
@@ -120,10 +121,38 @@ if_std! {
...
+        }
+
+        unsafe fn poll(task: *mut (), cx: &mut Context) -> Poll<()> {
+            let ptr: *mut F = mem::transmute(task);
+            let fut: &mut F = &mut *ptr;
+            fut.poll_mut(cx)
+        }
+
+        unsafe fn drop(task: *mut ()) {
+            let ptr: *mut F = mem::transmute(task);
```
cramertj: `drop(Box::from_raw(task as *mut F))`

futures-core/src/task/mod.rs
```diff
@@ -120,10 +121,38 @@ if_std! {
         }
     }
 
+    #[cfg(not(feature = "nightly"))]
+    unsafe impl<F: Future<Output = ()> + Send + 'static + ::Unpin> UnsafePoll for Box<F> {
```
cramertj: Why the `Unpin` bound? You're boxing the value, so you know that it's safe to be polled, and it can't ever be moved out, right?

futures-core/src/task/context.rs
cramertj: Why the `Unpin` bound? See below.

futures-core/src/task/mod.rs
aturon: When not on nightly, you *have* to have `Unpin` because the only method you can call on a future requires it.

cramertj: Right, because the other method doesn't exist. Thanks!

futures-core/src/impls_nightly.rs
```diff
@@ -0,0 +1,134 @@
...
+        };
+        unsafe { *Pin::get_mut(&mut self) = None };
+        Poll::Ready(output)
+    }
+}
+
+if_std! {
+    use std::boxed::{Box, PinBox};
+
+    impl<'a, F: ?Sized + Future> Future for Box<F> {
```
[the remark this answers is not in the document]
pythonesque: >  If you have a Pin<Box<F>>, both the Box and the underlying value cannot be moved.

This is not going to be true for long (even for `!Unpin` types), so any safety arguments based on this are highly dubious.

futures-core/src/impls.rs
pythonesque: `Box<T>` should be unconditionally `Unpin` (according to the current understanding in the `Pin` RFC issue).

cramertj: What? Why? Then we couldn't implement `Future` for `Box<T: Future + !Unpin>`

Issue #1002, opened 2018-05-09
0.2.1 fails to build with nightly 8ff4b4206 2018-05-08


sdroege:
```
error[E0433]: failed to resolve. Could not find `Pin` in `mem`
   --> /home/slomo/.cargo/registry/src/github.com-1ecc6299db9ec823/futures-core-0.2.1/src/future/mod.rs:150:35
    |
150 |             unsafe { ::core::mem::Pin::get_mut(&mut self.as_pin()).poll(cx) }
    |                                   ^^^ Could not find `Pin` in `mem`

error[E0433]: failed to resolve. Could not find `Pin` in `mem`
  --> /home/slomo/.cargo/registry/src/github.com-1ecc6299db9ec823/futures-core-0.2.1/src/stream/mod.rs:92:35
   |
92 |             unsafe { ::core::mem::Pin::get_mut(&mut self.as_pin()).poll_next(cx) }
   |                                   ^^^ Could not find `Pin` in `mem`

error[E0599]: no method named `as_pin` found for type `&mut std::boxed::PinBox<F>` in the current scope
   --> /home/slomo/.cargo/registry/src/github.com-1ecc6299db9ec823/futures-core-0.2.1/src/future/mod.rs:150:58
    |
150 |             unsafe { ::core::mem::Pin::get_mut(&mut self.as_pin()).poll(cx) }
    |                                                          ^^^^^^

error[E0599]: no method named `as_pin` found for type `&mut std::boxed::PinBox<S>` in the current scope
  --> /home/slomo/.cargo/registry/src/github.com-1ecc6299db9ec823/futures-core-0.2.1/src/stream/mod.rs:92:58
   |
92 |             unsafe { ::core::mem::Pin::get_mut(&mut self.as_pin()).poll_next(cx) }
   |                                                          ^^^^^^

error: aborting due to 4 previous errors
```

It would be great if a new release could also be made to crates.io :)

Nemo157:
Needs an `s/Pin/PinMut/` (plus maybe some method name updates) to support https://github.com/rust-lang/rust/pull/50497.

PR #1003, opened 2018-05-09
Update for latest nightly changes to rename Pin to PinMut

sdroege:
Fixes https://github.com/rust-lang-nursery/futures-rs/issues/1002

### DIFF
--- a/futures-async-runtime/src/future.rs
+++ b/futures-async-runtime/src/future.rs
@@ -1,5 +1,5 @@
 use std::marker::Unpin;
-use std::mem::Pin;
+use std::mem::PinMut;
 use std::ops::{Generator, GeneratorState};
 
 use super::{IsResult, Reset, CTX};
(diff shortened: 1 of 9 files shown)


sdroege:
The remaining failure requires adding `#![feature(extern_prelude)]` to the futures and futures-channel crates. But futures-channel does not even have a nightly feature right now. What to do about this?

Nemo157:
It's very annoying that it doesn't say where is trying to access a crate through the prelude. Since it's only happening while building the docs I'm guessing it's related to some link attempting to resolve through an extern crate.

Checking the [last successful build shows](https://travis-ci.org/rust-lang-nursery/futures-rs/jobs/373242465#L522): 

```
warning: [futures_core::Stream] cannot be resolved, ignoring it...
```

which is probably [this line](https://github.com/rust-lang-nursery/futures-rs/blob/62827f1776e833a948cb81582988cb7a70a8b700/futures-channel/src/mpsc/mod.rs#L6).

This should be fixable by changing it to one of ``[`Stream`]``, ``[`Stream`](Stream)`` or ``[`Stream`](::futures_core::Stream)``.

The futures crate appears to have a lot of similar issues, judging by the warnings in the linked build.

I think this is worth opening an issue about on rust-lang/rust, during a doc build it should block the `#![feature(extern_prelude)]` check and just leave it as an unresolved path.

Nemo157:
Oh, there's one more at https://github.com/rust-lang-nursery/futures-rs/blob/62827f1776e833a948cb81582988cb7a70a8b700/futures-channel/src/mpsc/mod.rs#L336, it would be great to check if changing both fixes it for `futures-channel` at least to make sure my rationale above is correct.

sdroege:
Oh yes, changing both fixes it

sdroege:
The build failure in the futures crate was just transitive from futures-channel, we're good with this fortunately.

PR #1010, opened 2018-05-15
0.3

cramertj:


Thomasdezeeuw:
I though `Future` likely wasn't going to require `PinMut` based on https://github.com/rust-lang/rfcs/pull/2418#issuecomment-388939069 and https://github.com/rust-lang/rfcs/pull/2418#issuecomment-389377007. But this change does require it?

PR #1011, opened 2018-05-15
Switch to `TryFuture`

aturon:


futures-core/src/future/mod.rs
```diff
@@ -164,6 +164,34 @@ if_std! {
...
+
+    /// The type of failures yielded by this future
+    type Error;
+
+    /// Poll this `TryFuture` as if it were a `Future`.
+    ///
+    /// This method is a stopgap for a compiler limitation that prevents us from
+    /// directly inheriting from the `Future` trait; in the future it won't be
+    /// needed.
+    fn try_poll(self: PinMut<Self>, cx: &mut task::Context) -> Poll<Result<Self::Item, Self::Error>>;
```
MajorBreakfast: Maybe rename this to `try_future_poll`?

This would indicate that this function is not actually a version of `poll` that returns a`Result` (`try_` prefix suggests this). Also, a more obscure name is probably not a bad idea for a method that is already planned to be deprecated in the not so far future.

PR #1023, opened 2018-05-23
Add future::pinned combinator

cramertj:
This is a preliminary attempt at a massively unsafe manual self-borrowing future to allow working with borrowing `futures` combinators without relying on `async`/`await`. I'm uncomfortable about the number of different crazy hacks happening here-- hopefully we can whittle those down. Ideally this wouldn't be replaced entirely by async/await someday, at which point we can remove this, but it could allow moving stable code over to the new borrowing versions of e.g. `Stream::next` and `AsyncRead::read`.

I'm not even sure that this is possible to express correctly in Rust's surface syntax, since the `Self` type contains a value as well as a live field with a live `&mut` inside of it. `UnsafeCell` isn't enough here, and I don't know what is. Perhaps it's enough to know that we never create an `&mut` reference to `self.data` while `self.fn_or_fut` holds a future containing a reference to `self.data`? I wouldn't have thought so, but I don't know what to do otherwise. cc @Zoxc and @alexcrichton, who know how self-referential generators are implemented-- do they rely on special magic for this? cc also @ralfjung, primarily so they're aware of the crazy pinning shenanigans happening here.

### DIFF
--- a/futures-util/src/future/pinned.rs
+++ b/futures-util/src/future/pinned.rs
@@ -11,9 +11,9 @@ pub trait PinnedFnLt<'a, Data: 'a, Output> {
     fn apply(self, data: PinMut<'a, Data>) -> Self::Future;
 }
 
-pub trait PinnedFn<Data, Output>: for<'a> PinnedFnLt<'a, Data, Output> {}
+pub trait PinnedFn<Data, Output>: for<'a> PinnedFnLt<'a, Data, Output> + 'static {}
 impl<Data, Output, T> PinnedFn<Data, Output> for T
-    where T: for<'a> PinnedFnLt<'a, Data, Output> {}
+    where T: for<'a> PinnedFnLt<'a, Data, Output> + 'static {}
 
 impl<'a, Data, Output, Fut, T> PinnedFnLt<'a, Data, Output> for T
 where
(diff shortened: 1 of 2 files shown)


futures-util/src/future/pinned.rs
```diff
@@ -0,0 +1,116 @@
...
+        (self)(data)
+    }
+}
+
+/// A future which borrows a value for an asynchronous lifetime.
+///
+/// Created by the `borrowed` function.
+#[must_use = "futures do nothing unless polled"]
+#[allow(missing_debug_implementations)]
+pub struct PinnedFut<'any, Data: 'any, Output, F: PinnedFn<Data, Output> + 'any> {
```
cramertj: I can't actually make this work because I get stuck on that `PinnedFut` isn't allowed to outlive `'any` no matter what I do. This seems like I need `#[may_dangle]`, but somehow stronger. I don't even need a lifetime to be part of the type-- I just need a lifetime that does not outlive `Data` (can't always use `'static`) but which will unify with other lifetimes.

cramertj: Even using 'static doesn't work-- then the WF checks for HRTB aren't kicking in for some reason: i get errors that `T: for<'a> PinnedFnLt<'a, Data, Output>` isn't satisfied if I add a `where Self: 'a` bound to `PinnedFnLt`.

cramertj:
Okay, I've got a minimal reproduction of the issue [here](https://play.rust-lang.org/?gist=7f5bd9d39b2e40381ca54a1f8c6c48f3&version=stable&mode=debug). It seems like higher ranked bounds for traits implemented by closures like this don't work.

cramertj:
Opened an issue: https://github.com/rust-lang/rust/issues/51004

futures-util/src/future/pinned.rs
```diff
@@ -0,0 +1,116 @@
...
+///
+/// Created by the `borrowed` function.
+#[must_use = "futures do nothing unless polled"]
+#[allow(missing_debug_implementations)]
+pub struct PinnedFut<Data: 'static, Output, F: PinnedFn<Data, Output>> {
+    fn_or_fut: FnOrFut<Data, Output, F>,
+    // TODO:
+    // marker: Pinned,
+    // Data, which may be borrowed by `fn_or_fut`, must be dropped last
+    data: Data,
```
RalfJung: Maybe this should use `ManuallyDrop` so as to not rely on drop order?

(AFAIK drop order is actually "first field is dropped last", so this may even be wrong right now.)

RalfJung:
I don't think I really understand what is going on here... what is the crazy stuff that's going on here, other than some funny shenanigans with `for<'a> ...` as a super trait bound?

cramertj:
@RalfJung One of the fields is being allowed to hold a `PinMut` reference into another field, and it may mutate that field.

PR #1024, opened 2018-05-25
Clear pending item in TakeWhile

DGolubets:
Pending item is no longer needed after termination flag has been set, so it's better to clear it earlier.

Though it's not strictly a bug, but more a surprising behavior: if pending item is not cleared, item's destructor is called after the stream's destructor. In my case it was causing a problem terminating a stream from RdKafka, cos it relies on the destruction order.


cramertj:

PR #1025, opened 2018-05-25
Update for new nightly

cramertj:


futures-core/src/future/mod.rs
```diff
@@ -139,11 +139,11 @@ impl<'a, F: ?Sized + Future> Future for PinMut<'a, F> {
 if_std! {
     use std::boxed::{Box, PinBox};
 
-    impl<'a, F: ?Sized + Future> Future for Box<F> {
+    impl<'a, F: ?Sized + Future + Unpin> Future for Box<F> {
```
cramertj: Note that I changed the `Box` impls not to project-- that allows us to be more conservative while we shake out whether or not we want `Box` to unconditionally be `Unpin`.

cramertj: (personally I'm still not sold on that choice, but I don't think we should take a hard soundness dependency upon unresolved decisions like this)

futures-util/src/try_future/and_then.rs
```diff
@@ -55,7 +55,7 @@ impl<A, B, F> Future for AndThen<A, B, F>
                 }
             };
 
-            // safe because we're using the `&mut` to do an assignment, not for moving out
+            // safe because we're using the `&mut` to do an setment, not for moving out
```
cramertj: Lol-- will fix

futures-channel/src/mpsc/mod.rs
```diff
@@ -110,8 +110,8 @@ pub struct Sender<T> {
     maybe_parked: bool,
 }
 
-// Safe because we treat the `T` opaquely
-unsafe impl<T> Unpin for Sender<T> {}
+// We never project PInMut<Sender> to `PinMut<T>`
```
MajorBreakfast: `PInMut` -> `PinMut`

futures-channel/src/mpsc/mod.rs
```diff
@@ -137,7 +137,7 @@ pub struct Receiver<T> {
 pub struct UnboundedReceiver<T>(Receiver<T>);
 
 // Safe because we treat the `T` opaquely
```
aturon: Should we update the comment?

futures-core/src/future/mod.rs
aturon: Good catch.

futures-channel/src/mpsc/mod.rs
cramertj: Done.

futures-channel/src/mpsc/mod.rs
cramertj: Done.

Issue #926, opened 2018-03-31
Expose inner Arc from SharedItem & SharedError


vorner:
Hello

These are basically just Arc, but less potent (eg. they have much less methods). Furthermore, some APIs expect to work with Arcs and while it's possible to create one from item from ordinary future (because the recipient gets ownership), it is not the case here.

Would it be possible to provide `into_inner` method for them, to convert them to Arc? Or is there a reason why it shouldn't be done?

I'm willing to send the pull request, but I wanted to ask first if it makes sense.

cramertj:
IMO it'd actually make more sense just to yield `Arc`s directly rather than adding a wrapper type. cc @aturon 

It comes at the cost of some backcompat (if we ever want to change the backing store to something other than an `Arc`) but I can't really imagine what changes we'd want to make.

cramertj:
`shared` now operates on arbitrary `clone()` output types

PR #1027, opened 2018-05-26
Add into_inner on SharedItem and SharedError.

estk:
Attempts to address #926


cramertj:
Thanks! I'm going to leave #926 open since I think in the future we should change this combinator to remove `SharedItem` and `SharedError` in favor of explicit `Arc`s, but this seems like a good starting point!

PR #1028, opened 2018-05-26
Sink

cramertj:
based on https://github.com/rust-lang-nursery/futures-rs/pull/1025

### DIFF
--- a/futures-sink/src/lib.rs
+++ b/futures-sink/src/lib.rs
@@ -7,6 +7,8 @@
 #![deny(missing_docs, missing_debug_implementations)]
 #![doc(html_root_url = "https://docs.rs/futures-sink/0.2.0")]
 
+#![feature(pin, arbitrary_self_types)]
+
 #[cfg(feature = "std")]
 extern crate std;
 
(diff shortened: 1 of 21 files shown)


futures-sink/src/lib.rs
```diff
@@ -172,21 +62,21 @@ pub trait Sink {
...
-    /// This method returns `Async::Ready` once the underlying sink is ready to
+    /// This method returns `Poll::Ready` once the underlying sink is ready to
     /// receive data. If this method returns `Async::Pending`, the current task
     /// is registered to be notified (via `cx.waker()`) when `poll_ready`
     /// should be called again.
     ///
     /// In most cases, if the sink encounters an error, the sink will
     /// permanently be unable to receive items.
-    fn poll_ready(&mut self, cx: &mut task::Context) -> Poll<(), Self::SinkError>;
+    fn poll_ready(self: PinMut<Self>, cx: &mut task::Context) -> Poll<Result<(), Self::SinkError>>;
```
aturon: Rationale on keeping the error baked in here?

cramertj: It doesn't necessarily need to be an error, I suppose-- we could instead return an associated `type Result`. I think that'd always be either `()` or `Result`, in most cases it would be `Result`, and keeping it as `Result` allows for having consistently short-circuiting operations, and avoids questions about what to do with the value when making a future that calls multiple poll functions and receives multiple `Sink::Result`s, like the `Send` combinator. I think I could definitely be convinced that it should change to an associated type, but I don't see too strong of a motivation, and the short-circuiting behavior of several combinators is useful.

cramertj:
> On IRC you said you moved to PinMut for with specifically. I assume that's because it ends up storing a futures, which may not be movable.

`with` + `T: Sink + Stream` interop + consistency with `Future` and `Stream`. I could definitely go a different way on `with`, but I don't think there's a way to work around the `Sink + Stream` interop without requiring `T: Unpin` (which we could do for `Sink`, I suppose). WDYT?

PR #1034, opened 2018-06-07
port futures 0.3 to std futures

tinaun:


tinaun:
this won't compile until a nightly with futures in it is built, but i think its ready to be looked at - theres a whole lot of extension traits which im kinda ambivalent about

tinaun:
i have no idea whats causing those doc errors. a missing path in a link maybe?

futures-core/src/future/either.rs
```diff
@@ -1,24 +1,24 @@
...
-}
+//     fn poll(self: PinMut<Self>, cx: &mut task::Context) -> Poll<A::Output> {
+//         unsafe {
+//             match PinMut::get_mut(self) {
+//                 Either::Left(a) => PinMut::new_unchecked(a).poll(cx),
+//                 Either::Right(b) => PinMut::new_unchecked(b).poll(cx),
+//             }
+//         }
+//     }
+// }
```
MajorBreakfast: Why was this removed?

futures-core/src/poll.rs
```diff
@@ -39,24 +42,23 @@ macro_rules! ready {
...
+/// Temporary trait for inherent methods on `Poll` - should go away before merge
+pub trait PollExt<T> {
+    /// Change the ready value of this `Poll` with the closure provided
+    fn map<U, F>(self, f: F) -> Poll<U>
+        where F: FnOnce(T) -> U;
+    /// Returns whether this is `Poll::Ready`
+    fn is_ready(&self) -> bool;
+    /// Returns whether this is `Poll::Pending`
+    fn is_pending(&self) -> bool {
+        !self.is_ready()
```
MajorBreakfast: Your PR to core has been merged https://github.com/rust-lang/rust/pull/51442 These methods are now not required

futures-core/src/lib.rs
```diff
@@ -63,7 +64,7 @@ macro_rules! unsafe_unpinned {
 }
 
 mod poll;
-pub use poll::Poll;
+pub use poll::{Poll, PollExt, PollResultExt};
```
MajorBreakfast: Also not required

futures-core/src/future/either.rs
tinaun: since either is defined in https://github.com/bluss/either-rs and Future in std it can't exist here because of orphan rules - i might write a pr moving this to the either crate as an unstable feature

futures-core/src/future/mod.rs
```diff
@@ -120,49 +24,52 @@ pub trait Future {
     }
 }
 
-impl<'a, F: ?Sized + Future + Unpin> Future for &'a mut F {
-    type Output = F::Output;
+impl<T: ?Sized> CoreFutureExt for T where T: Future {}
+ 
+// should be in std
```
cramertj: This can be removed now, right?

futures-core/src/task/mod.rs
```diff
@@ -1,129 +1,53 @@
...
+        /// Create a new `TaskObj` by boxing the given future.
+        fn new<F: Future<Output = ()> + Send + 'static>(f: F) -> TaskObj;
     }
-}
 
-impl Drop for TaskObj {
-    fn drop(&mut self) {
-        unsafe {
-            (self.drop)(self.ptr)
+    impl TaskObjExt for TaskObj {
```
cramertj: This implements the trait for `dyn TaskObj`-- did you mean to implement `TaskObjExt` for `T: TaskObj`?

futures-core/src/task/mod.rs
```diff
@@ -1,129 +1,53 @@
...
-    /// Poll the task.
-    ///
-    /// The semantics here are identical to that for futures, but unlike
-    /// futures only an `&mut self` reference is needed here.
-    pub fn poll_task(&mut self, cx: &mut Context) -> Poll<()> {
-        unsafe {
-            (self.poll)(self.ptr, cx)
-        }
+    /// Extension trait for `TaskObj`, adding methods that require allocation.
+    pub trait TaskObjExt {
```
cramertj: What's the purpose of this trait? Since the method doesn't take `self`, it's not actually callable using method syntax, so it might as well be a free function. IMO users should either call `TaskObj::from_poll_task` directly or call the `spawn` method on the `ContextExt` trait (yet to be written).

futures-core/src/task/mod.rs
```diff
@@ -1,129 +1,53 @@
...
     }
 
-    impl TaskObj {
-        /// Create a new `TaskObj` by boxing the given future.
-        pub fn new<F: Future<Output = ()> + Send + 'static>(f: F) -> TaskObj {
-            TaskObj::from_poll_task(Box::new(f))
+    impl<'a> ContextExt for Context<'a> {
+        fn spawn<F>(&mut self, f: F) where F: Future<Output = ()> + 'static + Send {
+            self.executor()
+                .spawn_obj(TaskObj::new(f)).unwrap()
```
cramertj: As I mentioned above, I'd change this to call `from_poll_task` directly.

futures-core/src/task/mod.rs
tinaun: isn't TaskObj a struct that pretends to be a trait object currently?

MajorBreakfast:
About reexports: I'm not sure whether we should offer reexports at all. Normally reexports are used for compatibility. However futures 0.3 can't be compatible to 0.2 in any meaningful way because nearly all signatures change a little bit. Therefore should there be reexports at all? I think we shouldn't offer reexports and instead encourage to either import the prelude or import from core directly.

Terkwood:
How hard is it to preserve the re-export capability? It seems useful in the case where one _isn't _facing a compatibility issue.

MajorBreakfast:
>  It seems useful in the case where one isn't facing a compatibility issue.

@Terkwood Please elaborate

The reason I propose to remove the reexports is for consistency. It seems inconsistent to import `core::mem::PinMut` directly, but not do the same for `core::future::Future`. And to offer reexports for `core::future::Future`, but not for `core::mem::PinMut`.

futures-core/src/future/either.rs
Terkwood: I'd gently suggest removing the commented code, if it is, in fact, going away

futures-core/src/future/mod.rs
```diff
@@ -12,106 +12,10 @@ pub use self::option::FutureOption;
...
-    /// Once a future has completed (returned `Ready` from `poll`),
-    /// then any future calls to `poll` may panic, block forever, or otherwise
-    /// cause bad behavior. The `Future` trait itself provides no guarantees
-    /// about the behavior of `poll` after a future has completed.
-    ///
-    /// Callers who may call `poll` too many times may want to consider using
-    /// the `fuse` adaptor which defines the behavior of `poll`, but comes with
-    /// a little bit of extra cost.
-    fn poll(self: PinMut<Self>, cx: &mut task::Context) -> Poll<Self::Output>;
+pub use core::future::Future;
```
Terkwood: 🎉

MajorBreakfast:
> You're talking about removing reexports from the language entirely, right?

@Terkwood No no ^^' I merely suggest to remove the reexport statements in futures-core. In other words: Only export the stuff that is actually defined in futures-core (e.g. `Stream`)

Terkwood:
Oh, ok😀
Then my earlier comment was misinformed. Thanks for bearing with me! Please
proceed!

MajorBreakfast:
@withoutboats That requires:
- Not removing the definitions of `Future`, `Executor`, etc. from `futures-core`. Instead, they need to be updated so that they're identical to their counterparts in core.
- `PinMut` needs a polyfill/dummy impl

MajorBreakfast:
It was decided to leave the reexports in, because they give us the possibility to maybe do the "stable shim dance" later

aturon:
Merged, thanks much @tinaun!

There are a couple of minor things I will clean up in a follow-up PR.

PR #1035, opened 2018-06-07
Fix a bug in stream::select_all

stjepang:
In the previous implementation, `stream::select_all` completes as soon as any of the streams it is selecting over is exhausted. The code in this PR will continue until all streams are exhausted.


cramertj:
Thanks!

PR #1037, opened 2018-06-12
Executors and partly reenable futures crate

MajorBreakfast:
This needs careful review. I'm not entirely sure what I did ^^' All I can say is that it builds now and some of the tests run.

Notes:
- ~~`futures-core` subcrate does **not** yet reexport from `core`~~
- `executor.spawn_obj(Box::new(future).into())` looks ugly ^^' I think we should introduce an `IntoTaskObj` trait. Unless there's going to be a simpler API layered on top of course.
- The doc tests seem to require `#![feature(arbitrary_self_types)]`. Can this be done? Didn't seem to work for me. Maybe I just did it the wrong way.
- Various places require `Executor` to be `Sized`. Should it become the default, so that we don't need to type it everywhere?
- The Rust compiler is genius. It reports new errors blazing fast and the error messages are extremely helpful
- [x] The `Task` structs should probably be renamed to something like `TaskContainer { task_obj, ... }`.


MajorBreakfast:
Because this PR is so big and requires careful vetting, https://github.com/rust-lang-nursery/futures-rs/pull/1034 should be reviewed an merged first. I will then rebase on top of it once it landed.

MajorBreakfast:
In this recent commit I renamed `Task` to `TaskContainer`.  Previously we had two things called task:
- `TaskObj`: The root future wrapped in a custom trait object
- `Task`: The environment `{ task, exec, wake_handle }` that contains it and other stuff

Calling both things "task" would be too confusing.

Alternative to what I implemented:
- Rename `TaskObj` to something else. It's only a stopgap solutions until dyn can be passed by value after all. One possible good name would be `RootFutureObj`
- Rename `TaskContainer` back to `Task`

I quite like the alternative. What do you think?

MajorBreakfast:
Okay here are some updated thoughts on this:

What about we rename `TaskObj` to `TaskFutureObj`:
- "Task" signifies that it's for a task and should have `Output = ()`
- "Future" signifies that it can be polled. `TaskFutureObj` should therefore also implement `Future` to make this perfectly clear: Currently it just implements a method call `poll_task()` (which I already suggested to rename to `poll`). But we should go the extra mile and make it a real future type. Because it is `Unpin` executors will be able to call `poll_unpin`
- "Obj" signifies that it's a custom trait object
- The trait will be called `UnsafeTaskFuture`

@cramertj 
@aturon

cramertj:
> executor.span_obj(Box::new(future).into()) looks ugly ^^' I think we should introduce an IntoTaskObj trait. Unless there's going to be a simpler API layered on top of course.

The plan was originally to add a method to `Context` that would take a generic future, box it, and spawn it. However, that impl needs std and `Context` is defined in core, so I think we should add it via an extension trait here instead.

cramertj:
> Various places require Executor to be Sized. Should it become the default, so that we don't need to type it everywhere?

No-- the executor inside of `task::Context` is `&mut dyn Executor`.

MajorBreakfast:
I now think we should stick with calling the top level future "task".

MajorBreakfast:
I rebased and updated it

futures-executor/src/local_pool.rs
```diff
@@ -2,21 +2,21 @@ use std::prelude::v1::*;
...
 use futures_util::stream::StreamExt;
 
 use thread::ThreadNotify;
 use enter;
 use ThreadPool;
 
-struct Task {
-    fut: Box<Future<Item = (), Error = Never>>,
-    map: LocalMap,
+struct TaskContainer {
```
cramertj: Can you add a comment to this that describes what it's for? Something like "wrapper for TaskObj that implements `Future` with a unit output" would help-- it's less obvious what its purpose is now that it no longer pairs the task with a LocalMap.

N.B. I think that TaskObj should implement `Future`-- we should add that impl in std. Opened https://github.com/rust-lang/rust/issues/51667

futures-executor/src/local_pool.rs
```diff
@@ -132,35 +132,31 @@ impl LocalPool {
     /// be inert after the call completes, but can continue with further use of
     /// `run` or `run_until`. While the function is running, however, all tasks
     /// in the pool will try to make progress.
-    pub fn run_until<F>(&mut self, mut f: F, exec: &mut Executor) -> Result<F::Item, F::Error>
-        where F: Future
+    pub fn run_until<F, Exec>(&mut self, mut f: F, exec: &mut Exec) -> F::Output
+        where F: Future + Unpin, Exec: Executor + Sized
```
cramertj: This `Unpin` bound isn't needed. Can you add [this macro](https://play.rust-lang.org/?gist=79d54732533d79bf007121bd80a6fbf3&version=nightly&mode=debug) to futures-core and use it instead?

futures-executor/src/local_pool.rs
```diff
@@ -202,14 +197,14 @@ lazy_static! {
 ///
 /// Use a [`LocalPool`](LocalPool) if you need finer-grained control over
 /// spawned tasks.
-pub fn block_on<F: Future>(f: F) -> Result<F::Item, F::Error> {
+pub fn block_on<F: Future + Unpin>(f: F) -> F::Output {
```
cramertj: rm `Unpin`

futures-executor/src/thread_pool.rs
```diff
@@ -96,23 +96,22 @@ impl ThreadPool {
     ///
     /// Note that the function will return when the provided future completes,
     /// even if some of the tasks it spawned are still running.
-    pub fn run<F: Future>(&mut self, f: F) -> Result<F::Item, F::Error> {
+    pub fn run<F: Future + Unpin>(&mut self, f: F) -> F::Output {
```
cramertj: rm Unpin

futures-util/src/sink/mod.rs
```diff
@@ -9,7 +9,7 @@ use super::future::Either;
 use core::mem::PinMut;
 
 mod close;
-mod fanout;
+// mod fanout;
```
cramertj: Why is `fanout` commented out?

futures-util/src/stream/futures_unordered.rs
```diff
@@ -36,8 +39,9 @@ pub struct FuturesUnordered<F> {
     head_all: *const Node<F>,
 }
 
-impl<T: Send> Send for FuturesUnordered<T> {}
-impl<T: Sync> Sync for FuturesUnordered<T> {}
+unsafe impl<T: Send> Send for FuturesUnordered<T> {}
+unsafe impl<T: Sync> Sync for FuturesUnordered<T> {}
+impl<T: Unpin> Unpin for FuturesUnordered<T> {}
```
cramertj: `FuturesUnordered` should be `Unpin` regardless of `T`.

futures-util/src/stream/futures_unordered.rs
```diff
@@ -99,12 +103,12 @@ enum Dequeue<T> {
 }
 
 impl<T> FuturesUnordered<T>
-    where T: Future,
+    where T: Future + Unpin
```
cramertj: Here and elsewhere, `FuturesUnordered` shouldn't require that the future is `Unpin`, since it's allocating it. Instead it should track usage to ensure the future isn't moved.

futures-util/src/stream/futures_unordered.rs
```diff
@@ -533,13 +535,26 @@ impl<'a, T> Clone for NodeToHandle<'a, T> {
...
 #[doc(hidden)]
 impl<'a, T> From<NodeToHandle<'a, T>> for Waker {
     fn from(handle: NodeToHandle<'a, T>) -> Waker {
         unsafe {
-            let ptr = handle.0.clone();
-            let ptr = mem::transmute::<Arc<Node<T>>, *const ArcNode<T>>(ptr);
-            Waker::new(hide_lt(ptr))
+            let ptr: Arc<Node<T>> = handle.0.clone();
+            let ptr: *mut ArcNode<T> = mem::transmute(ptr);
+            let ptr = mem::transmute(ptr as *mut UnsafeWake); // Hide lifetime
```
cramertj: Can you supply manual type arguments to `mem::transmute`? It's hard to tell what it's transmuting between here.

futures-executor/src/local_pool.rs
MajorBreakfast: Already made TaskObj implement Future

I‘ll add the comment

MajorBreakfast:
I've split `FuturesUnordered` into multiple files. Whenever a file is large I spend most of my time scrolling and trying to find stuff instead of getting things done ^^'

futures-executor/src/local_pool.rs
MajorBreakfast: I'm getting rid of TaskContainer entirely. We can use `TaskObj` directly. Can't do that in `ThreadPool`, but here it's simple

futures-util/src/sink/mod.rs
MajorBreakfast: Because the tests fail. There is no more `iter_ok()` for instance. Fixing now...

MajorBreakfast: No this is a rabbit hole... it requires converting `future_util/src/stream/forwards.rs` and probably various other things

MajorBreakfast:
@cramertj To make reviewing easier, look at the "`Future + !Unpin` support for `FuturesUnordered`" commit directly. I split the file in the preceding commit. This PR seems a bit bigger than it really is (it's still big of course)

futures-util/src/sink/mod.rs
cramertj: Can you leave `fanout`, but disable the tests? I'd like to make sure I know what code still needs fixing to work on 0.3 (regardless of tests) and what still needs fixing

futures-util/src/stream/futures_unordered/mod.rs
```diff
@@ -0,0 +1,412 @@
...
+                        return Poll::Ready(None);
+                    } else {
+                        return Poll::Pending;
+                    }
+                }
+                Dequeue::Inconsistent => {
+                    // At this point, it may be worth yielding the thread &
+                    // spinning a few times... but for now, just yield using the
+                    // task system.
+                    cx.waker().wake();
```
cramertj: Nit: you can use `local_waker().wake()` here

futures-util/src/stream/futures_unordered/iter_mut.rs
```diff
@@ -0,0 +1,34 @@
+use std::marker::PhantomData;
+use super::FuturesUnordered;
+use super::node::Node;
+
+#[derive(Debug)]
+/// Mutable iterator over all futures in the unordered set.
```
cramertj: Subtle: this is unsound now because it gives mutable access to the futures allowing them to be `mem::replaced`, although they may have been polled. This should either yield `PinMut<F>` or `iter_mut` should be changed to include an `F: Unpin` bound. Perhaps offer both as `iter_mut` and `iter_pin_mut`?

futures-util/src/stream/futures_unordered/node.rs
```diff
@@ -0,0 +1,145 @@
...
+
+#[doc(hidden)]
+impl<'a, T> From<NodeToHandle<'a, T>> for LocalWaker {
+    fn from(handle: NodeToHandle<'a, T>) -> LocalWaker {
+        unsafe {
+            let ptr: Arc<Node<T>> = handle.0.clone();
+            let ptr = mem::transmute::<Arc<Node<T>>, *mut ArcNode<T>>(ptr);
+            let ptr = ptr as *mut UnsafeWake;
+            // Hide lifetime
+            let ptr = mem::transmute::<*mut UnsafeWake, *mut UnsafeWake>(ptr);
```
cramertj: This looks like a no-op?

futures-util/src/stream/futures_unordered/node.rs
```diff
@@ -0,0 +1,145 @@
...
+}
+
+struct ArcNode<T>(PhantomData<T>);
+
+// We should never touch `T` on any thread other than the one owning
+// `FuturesUnordered`, so this should be a safe operation.
+unsafe impl<T> Send for ArcNode<T> {}
+unsafe impl<T> Sync for ArcNode<T> {}
+
+unsafe impl<T> UnsafeWake for ArcNode<T> {
```
cramertj: This impl looks like it could be written as an impl of the `Wake` trait for `Node<T>`, which would make the `From` impls above trivial.

futures-util/src/stream/futures_unordered/node.rs
```diff
@@ -0,0 +1,145 @@
...
+
+impl<'a, T> Clone for NodeToHandle<'a, T> {
+    fn clone(&self) -> Self {
+        NodeToHandle(self.0)
+    }
+}
+
+#[doc(hidden)]
+impl<'a, T> From<NodeToHandle<'a, T>> for LocalWaker {
+    fn from(handle: NodeToHandle<'a, T>) -> LocalWaker {
```
cramertj: This and the function below are the same modulo the call to either `LocalWaker::new` or `Waker::new`-- can you break out the shared bits into a function?

futures-util/src/stream/futures_unordered/node.rs
MajorBreakfast: It hides the lifetime. Before my PR it was in a function called `hide_lt`. Otherwise we get:
```
error[E0310]: the parameter type `T` may not live long enough
  --> futures-util/src/stream/futures_unordered/node.rs:96:23
   |
91 | impl<'a, T> From<NodeToHandle<'a, T>> for LocalWaker {
   |          - help: consider adding an explicit lifetime bound `T: 'static`...
...
96 |             let ptr = ptr as *mut UnsafeWake;
   |                       ^^^
   |
note: ...so that the type `stream::futures_unordered::node::ArcNode<T>` will meet its required lifetime bounds
```

cramertj: Something else is going on here-- the error you're getting is on line 96, not the usage of the ptr. `*mut UnsafeWake` is one type, and it doesn't contain any lifetimes.

futures-util/src/stream/futures_unordered/node.rs
MajorBreakfast: It can't because the `UnsafeWake` trait is used to hide the lifetime. I've added a bunch of comments in my next commit.

futures-util/src/stream/futures_unordered/node.rs
MajorBreakfast: Done

futures-util/src/stream/futures_unordered/node.rs
cramertj: Gotcha, thanks!

futures-util/src/stream/futures_unordered/mod.rs
MajorBreakfast: Done

futures-util/src/sink/mod.rs
MajorBreakfast: Done. Tests are disabled now.

futures-util/src/stream/futures_unordered/iter_mut.rs
MajorBreakfast: Good catch! Done.

futures-util/src/stream/futures_unordered/iter_pin_mut.rs
```diff
@@ -0,0 +1,36 @@
+use std::marker::PhantomData;
+use std::mem::PinMut;
+
+use super::FuturesUnordered;
+use super::node::Node;
+
+#[derive(Debug)]
+/// Mutable iterator over all futures in the unordered set.
+pub struct IterPinMut<'a, F: 'a> {
```
cramertj: Rather than duplicating the logic between `IterPinMut` and `IterMut`, I'd implement `IterMut` in terms of `PinMut`. The `Unpin` bound will make it safe to go from `PinMut` of the yielded value to `&mut` of the value.

futures-util/src/stream/futures_unordered/mod.rs
```diff
@@ -155,6 +157,15 @@ impl<T> FuturesUnordered<T> {
         }
     }
 
+    /// Returns an iterator that allows modifying each future in the set.
```
cramertj: Can you add an `Unpin` bound to the `iter_mut` method itself? It could be confusing to have `iter_mut` be callable but the result not be usable as an iterator.

futures-util/src/stream/futures_unordered/iter_mut.rs
```diff
@@ -10,7 +10,7 @@ pub struct IterMut<'a, F: 'a> {
     pub(super) _marker: PhantomData<&'a mut FuturesUnordered<F>>
```
cramertj: Can you add an `Unpin` bound to the `IterMut` type itself, too?

MajorBreakfast: Done

futures-util/src/stream/futures_unordered/iter_pin_mut.rs
MajorBreakfast: Done

futures-util/src/stream/futures_unordered/iter_mut.rs
```diff
@@ -1,33 +1,21 @@
...
-            return None;
-        }
-        unsafe {
-            let future = (*(*self.node).future.get()).as_mut().unwrap();
-            let next = *(*self.node).next_all.get();
-            self.node = next;
-            self.len -= 1;
-            Some(future)
-        }
+        PinMut::new(&mut self.0).next().map(|f| unsafe { PinMut::get_mut(f) })
```
cramertj: Why the `PinMut::new` here? You should be able to remove that and have `self.0.next().map(|f| unsafe { PinMut::get_mut(f) })`

MajorBreakfast: Yeah ^^'

cramertj:
@MajorBreakfast Looks great! Thanks for being so patient with all the changes.

MajorBreakfast:
🎉

PR #1040, opened 2018-06-21
Docs: make oneshot more discoverable

matklad:
It took me a non-trivial amount of time to figure out how to complete a future "manually". `oneshot` is super neat, but it is so much different from `Promise.complete` style APIs from other libraries. Let's explain *why* you need one-short to make it hopefully easier to discover?

### DIFF
--- a/futures-channel/src/oneshot.rs
+++ b/futures-channel/src/oneshot.rs
@@ -76,6 +76,9 @@ struct Inner<T> {
 ///
 /// Each half can be separately owned and sent across tasks.
 ///
+/// One-shot channel is used for creating a pending future and completing it manually
+/// later. In other langauges, this functionality is also known under the name of "promise".
+///
 /// # Examples
 ///
 /// ```


matklad:
Example of problem this is trying to prevent: https://users.rust-lang.org/t/async-completion-of-a-future-via-promise/18222/

futures-channel/src/oneshot.rs
```diff
@@ -76,6 +76,9 @@ struct Inner<T> {
 ///
 /// Each half can be separately owned and sent across tasks.
 ///
+/// One-shot channel is used for creating a pending future and completing it manually
```
avkonst: is "manually" better word than "explicitly" in this context?

matklad:
A crazy idea: we could also add an empty `fn promise();` function to the `oneshot` module with a comment that says: "you are looking for oneshot".

avkonst:
Maybe new crate future_promise could be more discoverable? This crate would
simply call oneshot...

MajorBreakfast:
@avkonst To clarify: In JavaScript (and thus in TypeScript because TypeScript is mostly just JavaScript plus types) the difference between a promise and a deferred is that a deferred has the `resolve()` and `reject()` methods publicly accessible while a promise only gives you access to to those inside the executor function `new Promise(function executor (resolve, reject) { ... })`. Thus the analogy between a deferred and oneshot.

> A promise which is completed immediately returns completed future.

I can't figure out what you mean to say ^^'

MajorBreakfast:
The main issue at hand seems to be that `oneshot` isn't discoverable enough. I think the best way to address this issue is to teach about `oneshot` in the asynchronous programming book (which doesn't exist yet, but is planned to exist one day not so far away).

My main issues with this change are:
- JavaScript uses the term "promise" for futures in general, not for this particular use case
- The existing doc text already mentions that the receiver implements `Future` further down

@matklad How should we resolve this?

matklad:
@MajorBreakfast I still maintain that `promise` is a rather popular name for this API in other languages, such as [clojure](https://clojuredocs.org/clojure.core/promise), [Scala](https://docs.scala-lang.org/overviews/core/futures.html#promisehttps://en.wikipedia.org/wiki/Futures_and_promisess) and [C++](https://en.cppreference.com/w/cpp/thread/promise), so I think it is important to include "this is known as promise in some other libraries". Moreover, Wikipedia quotes scala docs for a possible definition of the Promise: https://en.wikipedia.org/wiki/Futures_and_promises.

>The existing doc text already mentions that the receiver implements Future further down

This means that my phrasing could be improved :) What I am trying to call out is that you can use the **sender** as you would use promise in other languages. 


>in the asynchronous programming book 

That is a good idea, but I think it is also important that the API docs themselves are useful for people who are familiar with futures from other languages, who know what they are looking for, and who just don't know the name. 

What about including a section `# Creating a Future` in the docs for `trait Future`, which talks about leaf combinators, like `ok`, `err`, `never`,  and `oneshot`?

MajorBreakfast:
> this is known as promise in some other libraries

Yes, and I think it'd be helpful to mention it, but nevertheless "promise" is not a generally accepted term for this concept because some programming languages don't adhere to it. JavaScript's promises don't have anything like C++'s [std::promise#set_value()](https://en.cppreference.com/w/cpp/thread/promise/set_value) Just mention the programming languages and it's 100% fine. However, I suggest to mention this comparison a bit later after the main description.

> This means that my phrasing could be improved :)

Go for it :) Better phrasing is always good!

> What about including a section # Creating a Future in the docs for trait Future, which talks about leaf combinators, like ok, err, never, and oneshot?

We'll definitely need to improve things in that department. We already mention `lazy` and `ready` on the overview page for `Future`, but I'm certain there's room for improvement. However, the next main focus point will be to port tokio and hyper to futures 0.3 :)

avkonst:
Aleksey, this is great answer!

In general, futures is the strongest scala library. I already mentioned
rust could follow this closely and it would make things right.

MajorBreakfast:
@matklad Just now the former 0.3 branch became the master. You originally created this PR for the 0.2 branch, so I changed it to that. Feel free to change it the new 0.3 based master branch.

futures-channel/src/oneshot.rs
```diff
@@ -76,6 +76,9 @@ struct Inner<T> {
...
+/// later. In other langauges, this functionality is also known under the name of "promise".
```
hcpl: s/langauges/languages/

cramertj:
Closing due to inactivity (it's also still targeting the 0.2 branch)

PR #1041, opened 2018-06-21
Use a Vec for Shared waiters to reduce memory usage.

stuhood:
As discovered in https://github.com/pantsbuild/pants/issues/5990, the minimum allocation of the `waiters` `HashMap` in `future::Shared` is ~3kb. In discussions in the tokio gitter room, @tanriol pointed out that the minimum capacity of a `HashMap` is 32 entries, and the key and value in the waiters map add up to 80 bytes.

Our application creates a DAG of `future::Shared` instances, with an average dependency count per graph node of about 2. For a large usecase in that context, I saw:
```
futures 0.1.21:
  cpu:  5:06.17
  wall: 4m32.984s
  mem:  3495M

this patch:
  cpu:  5:06.19
  wall: 4m27.861s
  mem:  2759M
```
So: effectively no CPU usage difference, but a 21% memory savings.

### DIFF
--- a/src/future/shared.rs
+++ b/src/future/shared.rs
@@ -22,7 +22,7 @@ use std::cell::UnsafeCell;
 use std::sync::{Arc, Mutex};
 use std::sync::atomic::AtomicUsize;
 use std::sync::atomic::Ordering::SeqCst;
-use std::collections::HashMap;
+use std::vec::Vec;
 
 /// A future that is cloneable and can be polled in multiple threads.
 /// Use `Future::shared()` method to convert any future into a `Shared` future.
@@ -54,7 +54,7 @@ struct Inner<F: Future> {
 
 struct Notifier {
     state: AtomicUsize,
-    waiters: Mutex<HashMap<usize, Task>>,
+    waiters: Mutex<Vec<(usize, Task)>>,
 }
 
 const IDLE: usize = 0;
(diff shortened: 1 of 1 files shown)


tanriol:
Not sure whether keeping the vector sorted is worth it, especially for very short vectors.

For long vectors, insertion will normally be almost-at-the-end (waiter IDs are increasing, but a clone created later can be polled and inserted earlier). Deletion depends on the usage pattern -- if most clones are long-lived except for a few, it will be almost-at-the-end too, but not if all the clones have approximately the same max age. So looks like a sorted vector may be better after all, but not sure.

cramertj:
I use a slab (https://github.com/carllerche/slab) for this in my own applications, and I'd suggest doing that here. It's a vec with a built-in free list.

stuhood:
@cramertj : I might not be understanding, but to use `Slab`, we'd need to have somewhere to store the keys, right? Is the idea that we would have a `Vec<usize>` containing the slab keys, and a static `Slab` of the waiters?

stuhood:
> a static `Slab` of the waiters

Although that would introduce a `Mutex` here I think? @cramertj : Is that what you had in mind?

cramertj:
@stuhood `Mutex<Slab<Task>>` would go exactly where the current `Mutex<HashMap<usize, Task>>` is.

stuhood:
@cramertj : Ah, ok.

I thought that you were suggesting that many `Shared` instances would use the same `Slab`, which would likely be very efficient in terms of allocation, but potentially bottlenecked on synchronization.

cramertj:
I've done a refactor of Shared for 0.3 that also includes the change to use `Slab`-- feel free to take a look and backport what you like 0.1: https://github.com/rust-lang-nursery/futures-rs/pull/1047/commits/8632fc0390f90aa96a1324ce4829ad58828d8db8

stuhood:
@cramertj : We ended up switching away from `Shared` for our usecase for other reasons, so we don't need a backport. I'll close this for now.

Thanks!

PR #1042, opened 2018-06-23
Small refactoring for `FutureUnodered` iterators

MajorBreakfast:
- Put both iterators into a single file
- Make `iter_mut()` use `iter_pin_mut()` internally
- Add `Unpin` impl for `IterMut` (because why not)


futures-util/src/stream/futures_unordered/iter.rs
```diff
@@ -12,6 +12,10 @@ pub struct IterPinMut<'a, F: 'a> {
     pub(super) _marker: PhantomData<&'a mut FuturesUnordered<F>>
 }
 
+#[derive(Debug)]
+/// Mutable iterator over all futures in the unordered set.
+pub struct IterMut<'a, F: 'a + Unpin> (pub(super) IterPinMut<'a, F>);
+
 impl<'a, F: Unpin> Unpin for IterPinMut<'a, F> {}
```
cramertj: `IterPinMut` is `Unpin` whether or not `F` is `Unpin`, right?

MajorBreakfast: Yes because it doesn't store the data. I've removed this restriction

MajorBreakfast: Should I remove both `Unpin` impls? `Bilock` and other types that don't use pinning typically don't implement `Unpin`

MajorBreakfast: I've removed them. If you think I should put the back I can put them back in, but I don't think these impls are useful.

cramertj:
LGTM! Thanks for cleaning this up!

PR #1044, opened 2018-06-24
Update some more combinators for 0.3

cramertj:


futures-util/src/io/split.rs
```diff
@@ -21,7 +22,9 @@ fn lock_and_then<T, U, E, F>(lock: &BiLock<T>, cx: &mut task::Context, f: F) ->
     where F: FnOnce(&mut T, &mut task::Context) -> Poll<Result<U, E>>
 {
     match lock.poll_lock(cx) {
-        Poll::Ready(ref mut l) => f(l, cx),
+        // Safety: the value behind the bilock used by `ReadHalf` and `WriteHalf` is never exposed
+        // as a `PinMut` anywhere other than here as a way to get to `&mut`.
```
cramertj: This is pretty sketchy and I can imagine it might not hold up in a types-as-contracts-based model that formally understood `PinMut` as a new typestate, but it seems "fine". The alternatives are (a) take `PinMut` in the `AsyncRead` and `AsyncWrite` methods, (b) put an `Unpin` bound on the `AsyncRead` and `AsyncWrite` types used in a split, (c) make two versions of `BiLock`, one which pins its innards, and one which does not.

futures-util/src/lock.rs
```diff
@@ -55,12 +55,12 @@ impl<T> BiLock<T> {
...
     pub fn new(t: T) -> (BiLock<T>, BiLock<T>) {
-        let inner = Arc::new(Inner {
+        let arc = Arc::new(Inner {
             state: AtomicUsize::new(0),
-            inner: Some(UnsafeCell::new(t)),
+            value: Some(UnsafeCell::new(t)),
         });
 
-        (BiLock { inner: inner.clone() }, BiLock { inner: inner })
+        (BiLock { arc: arc.clone() }, BiLock { arc: arc })
```
MajorBreakfast: `arc: arc` can be just `arc`

futures-util/src/lock.rs
```diff
@@ -135,27 +135,30 @@ impl<T> BiLock<T> {
...
     }
 
     /// Attempts to put the two "halves" of a `BiLock<T>` back together and
     /// recover the original value. Succeeds only if the two `BiLock<T>`s
     /// originated from the same call to `BiLock::new`.
-    pub fn reunite(self, other: Self) -> Result<T, ReuniteError<T>> {
-        if &*self.inner as *const _ == &*other.inner as *const _ {
+    pub fn reunite(self, other: Self) -> Result<T, ReuniteError<T>>
+    where
+        T: Unpin,
```
MajorBreakfast: I think the `Unpin` bound is not required because it takes self by value, so it couldn't have been pinned before.

futures-util/src/stream/forward.rs
```diff
@@ -1,112 +1,94 @@
...
 use stream::{StreamExt, Fuse};
 
+const INVALID_POLL: &str = "polled `Forward` after completion";
+
 /// Future for the `Stream::forward` combinator, which sends a stream of values
-/// to a sink and then waits until the sink has fully flushed those values.
+/// to a sink and then flushes the sink.
+///
+/// Note: this is only usable with `Unpin` sinks, so `Sink`s that aren't `Unpin`
+/// will need to be pinned in order to be used with this combinator.
```
MajorBreakfast: - It would be good if the concrete reason was given. Maybe in parenthesis.
- I think you men "...will need to be ~~pinned~~ boxed in order..."

futures-util/src/lock.rs
cramertj: The bilock couldn't have been pinned, but this PR changes to assume that the innards of a bilock (behind the arc) are always pinned. This is required in order to call `self: PinMut<Self>` methods on the innards of the bilock.

futures-util/src/stream/forward.rs
cramertj: > I think you mean "...will need to be pinned boxed in order..."

I meant that they need to be pinned. This can be done by putting them in a `PInBox`, but it's not the only way.

futures-util/src/lock.rs
MajorBreakfast: Ah, I see

futures-util/src/lock.rs
```diff
@@ -55,12 +55,12 @@ impl<T> BiLock<T> {
     /// that will ever be available to the lock. These can then be sent to separate
     /// tasks to be managed there.
     pub fn new(t: T) -> (BiLock<T>, BiLock<T>) {
-        let inner = Arc::new(Inner {
+        let arc = Arc::new(Inner {
             state: AtomicUsize::new(0),
-            inner: Some(UnsafeCell::new(t)),
+            value: Some(UnsafeCell::new(t)),
```
MajorBreakfast: Could you add a comment: "We now consider the value pinned"

futures-util/src/stream/forward.rs
MajorBreakfast: Ah I get now what you mean. Maybe it's just me, but I think it is really hard to understand what is meant by this sentence. It would be clearer if it was expanded a bit. "e.g. by putting them in `PinBox`es or by pinning them on the stack and giving `PinMut` references to `forward` instead"

cramertj:
@aturon says he doesn't have time to review at the moment, so I'm going to go ahead and merge this for now. @MajorBreakfast Thanks for taking a look-- let me know if anything stands out to you as fishy, and I'll make sure to get it cleared up.

futures-util/src/stream/forward.rs
MajorBreakfast: Ok. Then I suppose you could remove "so `Sink`s that aren't `Unpin` will need to be pinned in order to be used with this combinator". Without this it's much clearer.

My main gripe with that sentence is that it kinda sounds like "Pinning makes a type `Unpin`". But that's not true of course, instead the type that we get by pinning it will be `Unpin`.

cramertj: Pinning results in a value that is `Unpin`, yeah. The naming works out to be sort of funny.

PR #1045, opened 2018-06-24
Local waker reference optimization

MajorBreakfast:
The context API was specifically designed so that it only returns a reference to a local waker. This PR uses this fact and returns a local waker that is created without cloning the `Arc<Node<T>>`. Cloning the `Arc` is not required because the `LocalWaker` reference cannot outlive it anyway.

@cramertj This requires careful review. I'm not so sure whether we should merge it right away.


futures-util/src/stream/futures_unordered/node.rs
```diff
@@ -146,3 +167,24 @@ unsafe impl<T> UnsafeWake for ArcNode<T> {
         Node::wake(node)
     }
 }
+
+unsafe impl<T> UnsafeWake for ArcNodeUnowned<T> {
+    #[inline]
+    unsafe fn clone_raw(&self) -> Waker {
```
cramertj: This needs to actually clone the underlying `Arc`, since the downstream future being polled could call `cx.local_waker().clone()` and then store the value indefinitely.

cramertj:
This is an interesting idea and could possibly be worth it, but I'm inclined to hold off on any more optimizations to this code before we do a cleanup of some of the unsafety and do careful profiling. `FuturesUnordered` as a whole is a large body of unsafe code that is difficult to reason about correctly.

MajorBreakfast:
@cramertj Let's let this simmer a bit. I got the idea from how Alex does it with the two queues `FuturesUnordered` has. He also made it so that the `Arc` is never cloned. I agree that we should hold off and wait until we have a large codebase we can test it on and see if it works as intended.

futures-util/src/stream/futures_unordered/node.rs
MajorBreakfast: Yes. It clones the `Arc` with `Node::waker(node)`.

cramertj: Oh gosh that was not obvious to me at all, but I see it now. Can you leave a comment?

cramertj: And the one created by `Node::waker` here will drop itself because it has been cast to a `*const Arc<Node<T>>`, rather than a `*const ArcNodeUnowned<T>`?

MajorBreakfast: Added the comment

futures-util/src/stream/futures_unordered/node.rs
```diff
@@ -59,32 +73,54 @@ impl<Fut> Node<Fut> {
         }
     }
 
-    // Saftey: The returned `NonNull<Unsafe>` needs to be put into a `Waker`
-    // or `LocalWaker`
-    unsafe fn clone_as_unsafe_wake_without_lifetime(self: &Arc<Node<Fut>>)
-        -> NonNull<dyn UnsafeWake>
-    {
+    /// Returns a local waker for this node without cloning the Arc
+    pub(super) fn local_waker(self: &'a Arc<Node<Fut>>) -> LocalWakerRef<'a> {
```
MajorBreakfast: This is now also safe. I created a wrapper struct with a lifetime that stores the `LocalWaker`. Can't return `&LocalWaker` directly from this function because the `LocalWaker` struct has to be stored somewhere. The wrapper struct implements `Deref`, though. So, the usage is nearly identical.

MajorBreakfast:
I've reworked this code, so that it is based on the current head of the 0.3 branch.

MajorBreakfast:
After my last refactoring a few days ago I feel now pretty confident about the quality of this code and I think we should merge it.

@cramertj Can you review my latest changes?

PR #1046, opened 2018-06-25
 Introduce LocalTaskObj

MajorBreakfast:
- Introduce `LocalTaskObj` and integrate it into `LocalPool`
- Export all the stuff form `core::task` under `futures::task` (some was previously exported under `futures::executor`) This is probably just a start. There're probably still some inconsistencies in the exports.
- This PR makes the tests in `futures-executor/tests/local_pool.rs` work

### DIFF
--- a/futures-core/src/executor.rs
+++ b/futures-core/src/executor.rs
@@ -1,4 +1,0 @@
-//! Executors.
-
-pub use core::task::{Executor, SpawnErrorKind, SpawnObjError};
-
--- a/futures-core/src/lib.rs
+++ b/futures-core/src/lib.rs
@@ -83,5 +83,3 @@ pub use stream::Stream;
 
 pub mod task;
 pub use task::Poll;
-
-pub mod executor;
--- a/futures-core/src/task/mod.rs
+++ b/futures-core/src/task/mod.rs
@@ -2,22 +2,23 @@
 
 use Future;
 
-pub use core::task::{UnsafeWake, Waker, LocalWaker};
+mod poll;
+pub use self::poll::Poll;
+
+pub use core::task::{
+    Context, Executor,
+    Waker, LocalWaker, UnsafeWake,
+    TaskObj, LocalTaskObj, UnsafeTask,
+    SpawnErrorKind, SpawnObjError, SpawnLocalObjError,
+};
 
 #[cfg(feature = "std")]
 pub use std::task::{Wake, local_waker, local_waker_from_nonlocal};
-
-pub use core::task::Context;
-
-mod poll;
-pub use self::poll::Poll;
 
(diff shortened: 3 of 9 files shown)


cramertj:
@MajorBreakfast Pinged you on discord about this. I think, rather than a separate trait, I'd prefer to remove the `Send` bound on the `UnsafeTask` trait, and only introduce it in `TaskObj::new`. I also think I'd prefer for `LocalTaskObj` to live in `core` so that the conversion from `TaskObj` to `LocalTaskObj` doesn't depend on unstable representation details of std.

MajorBreakfast:
I agree. It would then be very similar to how we do it with the `UnsafeWake` trait. The only thing different is that `UnsafeWake` needs bounds for `Sync` and `Send` because of the cloning as `Waker`. For this here however we, as you say, we should add no bounds to the trait, so that `PinBox` with a `Future<Output = ()> + 'static` can implement it

MajorBreakfast:
Do we want to add `SpawnLocalObjError` to libcore as well? It's not absolutely necessary because the fields are public.

cramertj:
@MajorBreakfast Hm, good question! I think either way is fine-- these are all going to be deprecated once unsized-rvalues + arbitrary-self-type-object-safety are fixed, so I don't think it matters particularly much either way.

MajorBreakfast:
We will still need need the distinction for between `LocalSpawnError` and `SpawnError` because of the `Send` difference.

If you let me pick, then I'd put them into core because this way we can use `transmute` for conversion between them. `LocalPools`'s `spawn_obj` calls `spawn_local_obj` internally, so this conversion is on a hot code path.

cramertj:
@MajorBreakfast In the long term, I figured we wouldn't have `LocalSpawnObjError` or `LocalSpawnError`, only the current `SpawnErrorKind`.

MajorBreakfast:
Here is the PR that puts `LocalTaskObj` into libcore https://github.com/rust-lang/rust/pull/51814

`SpawnError` gives us the task back when spawning isn't possible. This makes it possible to spawn it somewhere else. I don't think that it will go away. (And `LocalSpawnError` does the same for non-`Send` tasks)

cramertj:
@MajorBreakfast It will have to go away once spawning goes through unsized rvalues, since there's no way to hand them back.

MajorBreakfast:
That's a bummer.

So, this won't work? I thought the last element of a struct can be unsized?
```rust
pub struct SpawnError<F: Future<Output = ()> + Send> {
    pub kind: SpawnErrorKind,
    pub future: F,
}
```

MajorBreakfast:
The new nightly which now includes `LocalTaskObj` in libcore requires some fixes to the futures crate. Therefore this is now blocked on https://github.com/rust-lang-nursery/futures-rs/pull/1049

cramertj:
Nice! This is great!

PR #1047, opened 2018-06-26
0.3 more

cramertj:
cc @MajorBreakfast (I couldn't add you as a reviewer since you're not currently a member of @rust-lang-nursery, but I'd love it if you could take a look!)

### DIFF
--- a/futures-util/src/future/shared.rs
+++ b/futures-util/src/future/shared.rs
@@ -24,7 +24,7 @@ use slab::Slab;
 use std::fmt;
 use std::cell::UnsafeCell;
 use std::marker::Unpin;
-use std::mem::{self, PinMut};
+use std::mem::PinMut;
 use std::sync::{Arc, Mutex};
 use std::sync::atomic::AtomicUsize;
 use std::sync::atomic::Ordering::SeqCst;
@@ -251,7 +251,7 @@ impl<F> Drop for Shared<F> where F: Future {
     fn drop(&mut self) {
         if self.waker_key != NULL_WAKER_KEY {
             if let Ok(mut wakers) = self.inner.notifier.wakers.lock() {
-                mem::replace(&mut *wakers, Slab::new());
+                wakers.remove(self.waker_key);
             }
         }
     }
--- a/futures-util/src/future/shared.rs
+++ b/futures-util/src/future/shared.rs
@@ -19,48 +19,74 @@
 //! # }
 //! ```
 
-use std::{error, fmt, mem, ops};
+use slab::Slab;
+
+use std::fmt;
 use std::cell::UnsafeCell;
+use std::marker::Unpin;
+use std::mem::PinMut;
 use std::sync::{Arc, Mutex};
 use std::sync::atomic::AtomicUsize;
 use std::sync::atomic::Ordering::SeqCst;
-use std::collections::HashMap;
+use std::task::local_waker_from_nonlocal;
 
-use futures_core::{Future, Poll, Async};
-use futures_core::task::{self, Wake, Waker, LocalMap};
+use futures_core::{Future, Poll};
+use futures_core::task::{self, Wake, Waker};
 
 /// A future that is cloneable and can be polled in multiple threads.
 /// Use `Future::shared()` method to convert any future into a `Shared` future.
 #[must_use = "futures do nothing unless polled"]
 pub struct Shared<F: Future> {
     inner: Arc<Inner<F>>,
-    waiter: usize,
-}
-
-impl<F> fmt::Debug for Shared<F>
-    where F: Future + fmt::Debug,
-          F::Item: fmt::Debug,
-          F::Error: fmt::Debug,
-{
-    fn fmt(&self, fmt: &mut fmt::Formatter) -> fmt::Result {
-        fmt.debug_struct("Shared")
-            .field("inner", &self.inner)
-            .field("waiter", &self.waiter)
-            .finish()
-    }
+    waker_key: usize,
 }
 
 struct Inner<F: Future> {
-    next_clone_id: AtomicUsize,
-    future: UnsafeCell<Option<(F, LocalMap)>>,
-    result: UnsafeCell<Option<Result<SharedItem<F::Item>, SharedError<F::Error>>>>,
+    future_or_output: UnsafeCell<FutureOrOutput<F>>,
     notifier: Arc<Notifier>,
 }
 
 struct Notifier {
     state: AtomicUsize,
-    waiters: Mutex<HashMap<usize, Waker>>,
+    wakers: Mutex<Slab<Option<Waker>>>,
 }
+
+// The future itself is polled behind the `Arc`, so it won't be moved
+// when `Shared` is moved.
+impl<F: Future> Unpin for Shared<F> {}
+
+impl<F: Future> fmt::Debug for Shared<F> {
+    fn fmt(&self, fmt: &mut fmt::Formatter) -> fmt::Result {
+        fmt.debug_struct("Shared")
+            .field("inner", &self.inner)
+            .field("waker_key", &self.waker_key)
+            .finish()
+    }
+}
+
+impl<F: Future> fmt::Debug for Inner<F> {
+    fn fmt(&self, fmt: &mut fmt::Formatter) -> fmt::Result {
+        fmt.debug_struct("Inner")
+            .finish()
+    }
+}
+
+enum FutureOrOutput<F: Future> {
+    Future(F),
+    Output(Arc<F::Output>),
+}
+
+unsafe impl<F> Send for Inner<F>
+where
+    F: Future + Send,
+    F::Output: Send + Sync,
+{}
+
+unsafe impl<F> Sync for Inner<F>
+where
+    F: Future + Send,
+    F::Output: Send + Sync,
+{}
 
 const IDLE: usize = 0;
 const POLLING: usize = 1;
@@ -68,18 +94,18 @@ const REPOLL: usize = 2;
 const COMPLETE: usize = 3;
 const POISONED: usize = 4;
 
+const NULL_WAKER_KEY: usize = usize::max_value();
+
 pub fn new<F: Future>(future: F) -> Shared<F> {
     Shared {
         inner: Arc::new(Inner {
-            next_clone_id: AtomicUsize::new(1),
+            future_or_output: UnsafeCell::new(FutureOrOutput::Future(future)),
             notifier: Arc::new(Notifier {
                 state: AtomicUsize::new(IDLE),
-                waiters: Mutex::new(HashMap::new()),
+                wakers: Mutex::new(Slab::new()),
             }),
-            future: UnsafeCell::new(Some((future, LocalMap::new()))),
-            result: UnsafeCell::new(None),
         }),
-        waiter: 0,
+        waker_key: NULL_WAKER_KEY,
     }
 }
 
@@ -87,7 +113,7 @@ impl<F> Shared<F> where F: Future {
     /// If any clone of this `Shared` has completed execution, returns its result immediately
     /// without blocking. Otherwise, returns None without triggering the work represented by
     /// this `Shared`.
-    pub fn peek(&self) -> Option<Result<SharedItem<F::Item>, SharedError<F::Error>>> {
+    pub fn peek(&self) -> Option<Arc<F::Output>> {
         match self.inner.notifier.state.load(SeqCst) {
             COMPLETE => {
                 Some(unsafe { self.clone_result() })
@@ -97,36 +123,54 @@ impl<F> Shared<F> where F: Future {
         }
     }
 
-    fn set_waiter(&mut self, cx: &mut task::Context) {
-        let mut waiters = self.inner.notifier.waiters.lock().unwrap();
-        waiters.insert(self.waiter, cx.waker().clone());
+    /// Registers the current task to receive a wakeup when `Inner` is awoken.
+    fn set_waker(&mut self, cx: &mut task::Context) {
+        let mut wakers = self.inner.notifier.wakers.lock().unwrap();
+        if self.waker_key == NULL_WAKER_KEY {
+            self.waker_key = wakers.insert(Some(cx.waker().clone()));
+        } else {
+            let waker_slot = &mut wakers[self.waker_key];
+            let needs_replacement = if let Some(old_waker) = waker_slot {
+                // If there's still an unwoken waker in the slot, only wake
+                // if the current one wouldn't wake the same task.
+                !old_waker.will_wake(cx.waker())
+            } else {
+                true
+            };
+            if needs_replacement {
+                *waker_slot = Some(cx.waker().clone());
+            }
+        }
+        debug_assert!(self.waker_key != NULL_WAKER_KEY);
     }
 
-    unsafe fn clone_result(&self) -> Result<SharedItem<F::Item>, SharedError<F::Error>> {
-        match *self.inner.result.get() {
-            Some(Ok(ref item)) => Ok(SharedItem { item: item.item.clone() }),
-            Some(Err(ref e)) => Err(SharedError { error: e.error.clone() }),
-            _ => unreachable!(),
+    /// Safety: callers must first ensure that `self.inner.state`
+    /// is `COMPLETE`
+    unsafe fn clone_result(&self) -> Arc<F::Output> {
+        if let FutureOrOutput::Output(item) = &*self.inner.future_or_output.get() {
+            item.clone()
+        } else {
+            unreachable!()
         }
     }
 
-    fn complete(&self) {
-        unsafe { *self.inner.future.get() = None };
+    /// Safety: polling must have completed and self.inner.result must be
+    /// FutureOrOutput::Output
+    unsafe fn complete(&self) {
         self.inner.notifier.state.store(COMPLETE, SeqCst);
         Wake::wake(&self.inner.notifier);
     }
 }
 
-impl<F> Future for Shared<F>
-    where F: Future
-{
-    type Item = SharedItem<F::Item>;
-    type Error = SharedError<F::Error>;
+impl<F: Future> Future for Shared<F> {
+    type Output = Arc<F::Output>;
 
-    fn poll(&mut self, cx: &mut task::Context) -> Poll<Self::Item, Self::Error> {
-        self.set_waiter(cx);
+    fn poll(mut self: PinMut<Self>, cx: &mut task::Context) -> Poll<Self::Output> {
+        let this = &mut *self;
 
-        match self.inner.notifier.state.compare_and_swap(IDLE, POLLING, SeqCst) {
+        this.set_waker(cx);
+
+        match this.inner.notifier.state.compare_and_swap(IDLE, POLLING, SeqCst) {
             IDLE => {
                 // Lock acquired, fall through
             }
@@ -134,14 +178,17 @@ impl<F> Future for Shared<F>
                 // Another task is currently polling, at this point we just want
                 // to ensure that our task handle is currently registered
 
-                return Ok(Async::Pending);
+                return Poll::Pending
             }
             COMPLETE => {
-                return unsafe { self.clone_result().map(Async::Ready) };
+                return unsafe { Poll::Ready(this.clone_result()) };
             }
             POISONED => panic!("inner future panicked during poll"),
             _ => unreachable!(),
         }
+
+        let waker = local_waker_from_nonlocal(this.inner.notifier.clone());
+        let mut cx = cx.with_waker(&waker);
 
         loop {
             struct Reset<'a>(&'a AtomicUsize);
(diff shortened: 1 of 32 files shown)


futures-util/src/future/shared.rs
```diff
@@ -19,75 +19,101 @@
 //! # }
```
MajorBreakfast: This doc test needs to be updated. Just use `future::ready(6)` and `assert_eq!(6, *block_on(shared1))` (no `unwrap`)

futures-util/src/sink/mod.rs
```diff
@@ -98,12 +97,11 @@ pub trait SinkExt: Sink {
     /// ```
     fn with_flat_map<U, St, F>(self, f: F) -> WithFlatMap<Self, U, St, F>
         where F: FnMut(U) -> St,
-              St: Stream<Item = Self::SinkItem, Error=Self::SinkError>,
+              St: Stream<Item = Result<Self::SinkItem, Self::SinkError>>,
```
MajorBreakfast: Doc test fails

futures-util/src/future/shared.rs
```diff
@@ -97,52 +123,73 @@ impl<F> Shared<F> where F: Future {
...
-        waiters.insert(self.waiter, cx.waker().clone());
+    /// Registers the current task to receive a wakeup when `Inner` is awoken.
+    fn set_waker(&mut self, cx: &mut task::Context) {
+        let mut wakers = self.inner.notifier.wakers.lock().unwrap();
+        if self.waker_key == NULL_WAKER_KEY {
+            self.waker_key = wakers.insert(Some(cx.waker().clone()));
+        } else {
+            let waker_slot = &mut wakers[self.waker_key];
+            let needs_replacement = if let Some(old_waker) = waker_slot {
+                // If there's still an unwoken waker in the slot, only wake
```
MajorBreakfast: "... the slot, only ~wake~ replace"

futures-util/src/future/shared.rs
```diff
@@ -156,153 +203,76 @@ impl<F> Future for Shared<F>
...
+            let _reset = Reset(&this.inner.notifier.state);
 
             // Poll the future
             let res = unsafe {
-                let (ref mut f, ref mut data) = *(*self.inner.future.get()).as_mut().unwrap();
-                let waker = Waker::from(self.inner.notifier.clone());
-                let mut cx = task::Context::new(data, &waker, cx.executor());
-                f.poll(&mut cx)
+                if let FutureOrOutput::Future(future) = &mut *this.inner.future_or_output.get() {
+                    PinMut::new_unchecked(future).poll(&mut cx)
```
MajorBreakfast: I suggest to put the call to `poll` outside of the unsafe block

futures-util/src/future/shared.rs
```diff
@@ -156,153 +203,76 @@ impl<F> Future for Shared<F>
                 }
             }
 
-            let _reset = Reset(&self.inner.notifier.state);
+            let _reset = Reset(&this.inner.notifier.state);
```
MajorBreakfast: Maybe call this `Bomb` like in `FuturesOnordered`

MajorBreakfast: `let _ = ...`

futures-util/src/future/shared.rs
```diff
@@ -156,153 +203,76 @@ impl<F> Future for Shared<F>
...
-                            let prev = self.inner.notifier.state.swap(POLLING, SeqCst);
+                            let prev = this.inner.notifier.state.swap(POLLING, SeqCst);
                             assert_eq!(prev, REPOLL);
                         }
                         _ => unreachable!(),
                     }
-
                 }
-                Ok(Async::Ready(i)) => {
+                Poll::Ready(i) => {
```
MajorBreakfast: `Poll::Ready(output)`

futures-util/src/future/shared.rs
```diff
@@ -156,153 +203,76 @@ impl<F> Future for Shared<F>
...
-
                     break;
                 }
             }
         }
 
-        self.complete();
-        unsafe { self.clone_result().map(Async::Ready) }
+        unsafe {
+            this.complete();
```
MajorBreakfast: Why is the `this.complete()` call here and not after line 234? (After setting `this.inner.future_or_output`)

MajorBreakfast: Ah, took another look and realized that the result is the same. I'd still change it, though. Using `break` seems unintuitive.

futures-util/src/future/shared.rs
```diff
@@ -97,52 +123,73 @@ impl<F> Shared<F> where F: Future {
...
     }
 
-    unsafe fn clone_result(&self) -> Result<SharedItem<F::Item>, SharedError<F::Error>> {
-        match *self.inner.result.get() {
-            Some(Ok(ref item)) => Ok(SharedItem { item: item.item.clone() }),
-            Some(Err(ref e)) => Err(SharedError { error: e.error.clone() }),
-            _ => unreachable!(),
+    /// Safety: callers must first ensure that `self.inner.state`
+    /// is `COMPLETE`
+    unsafe fn clone_result(&self) -> Arc<F::Output> {
```
MajorBreakfast: Should be called `clone_output()`

futures-util/src/future/shared.rs
```diff
@@ -97,52 +123,73 @@ impl<F> Shared<F> where F: Future {
...
+        } else {
+            unreachable!()
         }
     }
 
-    fn complete(&self) {
-        unsafe { *self.inner.future.get() = None };
+    /// Safety: polling must have completed and self.inner.result must be
+    /// FutureOrOutput::Output
+    unsafe fn complete(&self) {
```
MajorBreakfast: This function merely contains two lines of code and is called only in a single place => It should be inlined

futures-util/src/future/shared.rs
```diff
@@ -156,153 +203,76 @@ impl<F> Future for Shared<F>
...
 impl Wake for Notifier {
     fn wake(arc_self: &Arc<Self>) {
         arc_self.state.compare_and_swap(POLLING, REPOLL, SeqCst);
 
-        let waiters = mem::replace(&mut *arc_self.waiters.lock().unwrap(), HashMap::new());
-
-        for (_, waiter) in waiters {
-            waiter.wake();
+        let mut lock = arc_self.wakers.lock().unwrap();
+        for (_key, opt_waker) in &mut *lock {
```
MajorBreakfast:
```rust
let wakers = &mut *arc_self.wakers.lock().unwrap();
    for (_key, opt_waker) in wakers {
```
(similar to `Drop` impl)

futures-util/src/future/shared.rs
```diff
@@ -19,75 +19,101 @@
...
 }
 
-impl<F> fmt::Debug for Shared<F>
-    where F: Future + fmt::Debug,
-          F::Item: fmt::Debug,
-          F::Error: fmt::Debug,
-{
+struct Notifier {
+    state: AtomicUsize,
+    wakers: Mutex<Slab<Option<Waker>>>,
```
MajorBreakfast: I suggest putting the Slab into an option. That way it can be dropped immediately and efficiently after the future completes.

`wakers: Mutex<Option<Slab<Option<Waker>>>>`

futures-util/src/future/shared.rs
```diff
@@ -156,153 +203,76 @@ impl<F> Future for Shared<F>
...
     }
 }
 
 impl<F> Drop for Shared<F> where F: Future {
     fn drop(&mut self) {
-        let mut waiters = self.inner.notifier.waiters.lock().unwrap();
-        waiters.remove(&self.waiter);
+        if self.waker_key != NULL_WAKER_KEY {
+            if let Ok(mut wakers) = self.inner.notifier.wakers.lock() {
+                wakers.remove(self.waker_key);
```
MajorBreakfast: BTW `Slab::remove()` doesn't actually shrink the `Vec` inside the `Slab`. See suggestion above about wrapping the `Slab` in an `Option`,

futures-util/src/sink/with_flat_map.rs
```diff
@@ -52,6 +63,11 @@ where
         &mut self.sink
     }
 
+    /// Get a pinned mutable reference to the inner sink.
+    pub fn get_pin_mut<'a>(self: PinMut<'a, Self>) -> PinMut<'a, S> {
+        unsafe { PinMut::new_unchecked(&mut PinMut::get_mut_unchecked(self).sink) }
```
MajorBreakfast: `self.sink()`

futures-util/src/sink/with_flat_map.rs
```diff
@@ -60,74 +76,89 @@ where
...
+    fn try_empty_stream(self: PinMut<Self>, cx: &mut task::Context) -> Poll<Result<(), S::SinkError>> {
+        let WithFlatMap {
+            sink,
+            f: _,
+            stream,
+            buffer,
+            _phantom,
+        } = unsafe { PinMut::get_mut_unchecked(self) };
+        let mut sink = unsafe { PinMut::new_unchecked(sink) };
+        let mut stream = unsafe { PinMut::new_unchecked(stream) };
```
MajorBreakfast: It'd be nicer if these used the methods. The unsafe code seems unnecessary.

MajorBreakfast: Also, I think temporary variables are not really needed. `self.sink()` can simply be called where it's needed

futures-util/src/sink/with_flat_map.rs
```diff
@@ -60,74 +76,89 @@ where
...
+
+
+        if let Some(x) = buffer.take() {
+            match try_poll!(sink.reborrow().poll_ready(cx)) {
+                Poll::Ready(()) => try_ready!(Poll::Ready(sink.reborrow().start_send(x))),
+                Poll::Pending => {
+                    *buffer = Some(x);
```
MajorBreakfast: This seems weird. Using `if buffer.is_some()` above seems simpler

futures-util/src/sink/with_flat_map.rs
```diff
@@ -60,74 +76,89 @@ where
...
-                    Async::Ready(()) => self.sink.start_send(x)?,
-                    Async::Pending => {
-                        self.stream = Some(stream);
-                        self.buffer = Some(x);
-                        return Ok(Async::Pending);
+        if let Some(mut some_stream) = stream.reborrow().as_pin_mut() {
+            while let Some(x) = ready!(some_stream.reborrow().poll_next(cx)) {
+                let x = try_ready!(Poll::Ready(x));
+                match try_poll!(sink.reborrow().poll_ready(cx)) {
+                    Poll::Ready(()) => try_poll!(Poll::Ready(sink.reborrow().start_send(x))),
```
MajorBreakfast: `try_poll!(Poll::Ready...` is not so nice to read

futures-util/src/sink/with_flat_map.rs
```diff
@@ -60,74 +76,89 @@ where
...
-            while let Some(x) = ready!(stream.poll_next(cx)) {
-                match self.sink.poll_ready(cx)? {
-                    Async::Ready(()) => self.sink.start_send(x)?,
-                    Async::Pending => {
-                        self.stream = Some(stream);
-                        self.buffer = Some(x);
-                        return Ok(Async::Pending);
+        if let Some(mut some_stream) = stream.reborrow().as_pin_mut() {
+            while let Some(x) = ready!(some_stream.reborrow().poll_next(cx)) {
+                let x = try_ready!(Poll::Ready(x));
```
MajorBreakfast: I think the `x` should be stored immediately. Macros that follow might return and then it is lost.

futures-util/src/stream/buffer_unordered.rs
```diff
@@ -66,6 +79,15 @@ impl<S> BufferUnordered<S>
         self.stream.get_mut()
     }
 
+    /// Acquires a pinned mutable reference to the underlying stream that this
+    /// combinator is pulling from.
+    ///
+    /// Note that care must be taken to avoid tampering with the state of the
+    /// stream which may otherwise confuse this combinator.
+    pub fn get_pin_mut<'a>(self: PinMut<'a, Self>) -> PinMut<'a, S> {
+        unsafe { PinMut::new_unchecked(&mut PinMut::get_mut_unchecked(self).stream) }.get_pin_mut()
```
MajorBreakfast: `self.stream().get_pin_mut()`

MajorBreakfast:
It will take a while before I inspected all the code.

One thought though: The macros `try_poll!` `try_ready!` `ready!` are awful to read because they hide return statements which I would rather see.

futures-util/src/stream/buffer_unordered.rs
cramertj: That isn't the same-- the generated `stream()` functions take self by `&mut PInMut` so that they get automatic reborrowing.

MajorBreakfast: Interesting. Well I already suspected that something was up when I started seeing it more than once. I've removed the other GitHub comment about this.

cramertj: Nice! That is better, thanks!

futures-util/src/stream/buffer_unordered.rs
```diff
@@ -76,45 +98,41 @@ impl<S> BufferUnordered<S>
...
+            }
         }
 
-        // Try polling a new future
-        if let Some(val) = ready!(self.queue.poll_next(cx)) {
-            return Ok(Async::Ready(Some(val)));
+        // Attempt to pull the next value from the queue
+        if let Some(val) = ready!(PinMut::new(self.queue()).poll_next(cx)) {
+            return Poll::Ready(Some(val));
         }
```
MajorBreakfast: Without the `ready!` macro it could look like this:
```rust
match PinMut::new(self.queue()).poll_next(cx) {
    x @ Poll::Pending => return x,
    x @ Poll::Ready(Some(_)) => return x,
    Poll::Ready(None) => {},
}
``` 
This is IMO way easier to understand what's going

futures-util/src/stream/buffer_unordered.rs
```diff
@@ -14,18 +15,25 @@ use stream::{Fuse, FuturesUnordered};
...
 pub struct BufferUnordered<S>
-    where S: Stream,
-          S::Item: IntoFuture,
+where
+    S: Stream,
+    S::Item: Future,
 {
     stream: Fuse<S>,
-    queue: FuturesUnordered<<S::Item as IntoFuture>::Future>,
+    queue: FuturesUnordered<S::Item>,
```
MajorBreakfast: Maybe rename this to `in_progress_queue`

futures-util/src/stream/buffered.rs
```diff
@@ -13,18 +14,35 @@ use stream::{Fuse, FuturesOrdered};
...
 /// is created by the `Stream::buffered` method.
 #[must_use = "streams do nothing unless polled"]
-pub struct Buffered<S: Stream> {
+pub struct Buffered<S: Stream>
+where
+    S: Stream,
+    S::Item: Future,
+{
     stream: Fuse<S>,
     queue: FuturesOrdered<S::Item>,
```
MajorBreakfast: Maybe rename this to `in_progress_queue`

futures-util/src/stream/futures_ordered.rs
```diff
@@ -88,11 +90,13 @@ pub struct FuturesOrdered<T>
     where T: Future
 {
     in_progress: FuturesUnordered<OrderWrapper<T>>,
```
MajorBreakfast: Maybe rename this to `in_progress_queue`

futures-util/src/stream/buffered.rs
```diff
@@ -13,18 +14,35 @@ use stream::{Fuse, FuturesOrdered};
...
+    S::Item: Future,
+{}
+
+impl<S: Stream> Buffered<S>
+where
+    S: Stream,
+    S::Item: Future,
+{
+    unsafe_pinned!(stream -> Fuse<S>);
+    unsafe_unpinned!(queue -> FuturesOrdered<S::Item>);
```
MajorBreakfast: Maybe use `unsafe_pinnned` for this. It's `Unpin`, so `DerefMut` works automatically, but a pinned projection let's us avoid the `PinMut::new(self.queue())` further down => Nicer code

MajorBreakfast:
I'm finished with reviewing.

I still suggest to get rid of the `try_poll!` `ready!` and `try_ready!` macros. They make the code a lot harder to understand. I constantly think that I'm missing something. Not to speak of <br>`try_ready!(Poll::Ready(...))` which is just an eyesore. And I think [here](https://github.com/rust-lang-nursery/futures-rs/pull/1047/files#r199229980) is even a bug that would be a lot more obvious to spot if the return statements weren't hidden.

Otherwise, good work! Especially the the usage of `Slab` in `Shared` is quite elegant.

futures-util/src/future/shared.rs
cramertj: Done.

futures-util/src/sink/mod.rs
cramertj: Fixed.

futures-util/src/future/shared.rs
cramertj: It gets messy here because it's prefixed by the call to `new_unchecked`, which uses a temporary `&mut` created by the `&mut *` on `UnsafeCell::get`, so getting the lifetimes right so that we don't create a temporary `&mut` which is alive until the following `get` on line 234 requires an extra set of braces. I'd prefer to leave it as-is :laughing:

futures-util/src/future/shared.rs
cramertj: `let _` doesn't do the same thing as `let _reset`-- `let _ = ` will instantly drop, rather than dropping at the end of the scope. Also, note that this isn't a `Bomb`-- it just resets the state, it doesn't trigger an abort.

futures-util/src/future/shared.rs
cramertj: Done.

futures-util/src/future/shared.rs
cramertj: Done-- I agree, that reads much better, thanks!

futures-util/src/future/shared.rs
cramertj: Done.

futures-util/src/future/shared.rs
cramertj: Done.

futures-util/src/future/shared.rs
cramertj: Done.

futures-util/src/future/shared.rs
cramertj: @MajorBreakfast Yeah, I opened an issue about fixing `Slab::remove` long ago-- see https://github.com/carllerche/slab/issues/38

futures-util/src/future/shared.rs
cramertj: I don't need an option to free it-- I can use `mem::replace` with a new slab.

futures-util/src/sink/with_flat_map.rs
cramertj: `self.sink()` and other methods produce overlapping borrows of `self` (because each individually borrows `self` mutably, so you can't access more than one at a time this way).

futures-util/src/future/shared.rs
MajorBreakfast: Yes definitely! This makes sense

futures-util/src/sink/with_flat_map.rs
cramertj: I don't disagree that it's awkward-- IMO the solution to this is to implement `Try` for `Poll` so that we can get reasonable `?` impls (so that `?` can be used on `Result<T, E>` to convert cleanly into `Poll<Result<T, E>>`) but that doesn't exist yet, so I'd prefer to leave as-is.

futures-util/src/sink/with_flat_map.rs
cramertj: If that errors, then it is intentional that `x` is lost.

futures-util/src/stream/buffer_unordered.rs
cramertj: Done.

futures-util/src/future/shared.rs
MajorBreakfast: I didn't know that o_O I'm sure there has been a long debate about this once ^^'

futures-util/src/future/shared.rs
MajorBreakfast: I don't think that will work. `Slab::remove` will panic if you call it with a key that does not exist.

MajorBreakfast: I mean the call to `Slab::remove` in `drop`

futures-util/src/future/shared.rs
```diff
@@ -156,153 +196,76 @@ impl<F> Future for Shared<F>
...
     }
 }
 
 impl<F> Drop for Shared<F> where F: Future {
     fn drop(&mut self) {
-        let mut waiters = self.inner.notifier.waiters.lock().unwrap();
-        waiters.remove(&self.waiter);
+        if self.waker_key != NULL_WAKER_KEY {
+            if let Ok(mut wakers) = self.inner.notifier.wakers.lock() {
+                mem::replace(&mut *wakers, Slab::new());
```
MajorBreakfast: I think this is a bug. You don't want to remove all the wakers. Just the one for this `Shared`

cramertj: Haha yup-- I was just fixing that. Where did you imagine the `Option` being set to `None`?

cramertj: I think it's probably fine to keep the Slab around-- shrinking it before all of the children have received the value and are ready to drop their `Shared` seems unnecessary.

futures-util/src/sink/with_flat_map.rs
MajorBreakfast: If that's the general rule, then it's okay

futures-util/src/future/shared.rs
```diff
@@ -156,153 +196,76 @@ impl<F> Future for Shared<F>
...
-                        (*self.inner.result.get()) = Some(Err(SharedError { error: Arc::new(e) }));
+                        *this.inner.future_or_output.get() =
+                            FutureOrOutput::Output(output.clone());
                     }
-
-                    break;
+                    // Complete the future
+                    this.inner.notifier.state.store(COMPLETE, SeqCst);
+                    Wake::wake(&this.inner.notifier);
+                    return Poll::Ready(output);
```
MajorBreakfast: Here is where I suggest to destroy the slab

cramertj: It will get immediately reset if done there because `Shared::poll` sets the `waker` before doing anything else (like checking if the value is complete). We could set COMPLETE while holding onto the `notifiers` lock, and check it before amending after taking out the lock in `Shared::poll`-- I'll give that a shot.

MajorBreakfast: set_waker could be a nop if *the option* is `None` :)

cramertj: Done.

futures-util/src/future/shared.rs
```diff
@@ -124,8 +124,14 @@ impl<F> Shared<F> where F: Future {
     }
 
     /// Registers the current task to receive a wakeup when `Inner` is awoken.
-    fn set_waker(&mut self, cx: &mut task::Context) {
+    fn set_waker_if_not_complete(&mut self, cx: &mut task::Context) {
+        // Acquire the lock first before checking COMPLETE to ensure there
+        // isn't a race.
         let mut wakers = self.inner.notifier.wakers.lock().unwrap();
+        if COMPLETE == self.inner.notifier.state.load(SeqCst) {
```
MajorBreakfast: we can save this atomic op if we put the wakers into an option :D hehe ^^'

cramertj: lol sure XD

cramertj: Done.

futures-util/src/future/shared.rs
MajorBreakfast: Bene, bene 👌

PR #1050, opened 2018-06-27
Remove async/await crates

MajorBreakfast:
I don't think we need these anymore


cramertj:
This sounds good to me-- the async/await stuff landed faster than I expected, so I don't think we'll need the macros anymore. Still, it is a bit sad to lose the ability to create streams via generators. cc @withoutboats @aturon Do either of you think that we should keep the macros around (and fix/maintain them) for the sake of stream-generators, or are you okay with killing them?

withoutboats:
If anyone wants to maintain the stream generator stuff against futures 0.3 in a separate repo, big :+1: from me, but I don't think they should be a part of the futures crate.

cramertj:
SGTM-- I'll go ahead and merge this then, and they can get pulled out of the history if someone so desires.

PR #1051, opened 2018-06-28
Add preliminary versions of variadic join and select

cramertj:
This is a super-simplified first pass at `join` and `select` macros for the async/await world. They both work, but `select` especially has lots of room for improvement: futures and streams that are passed to it multiple times (after they have previously completed) will panic since futures can't be polled after completion. One option to address this would be to add `FusedFuture` and `FusedStream` traits similar to `FusedIterator` that provide `is_done` methods for checking whether the future in question has completed (since if all futures/streams are exhausted, we need to visit some kind of "default" case for the `select`). There's also the additional complication that such an `is_done` method would also have to discern between "yielded `None`" and "will yield `None`", since it may be the case that you want to handle the moment a stream is closed, but after that you still want to `select` over it and have it act like a completed future (that is, provide is_done = true and fall through to `default` rather than providing `None`).


futures-util/src/join.rs
```diff
@@ -0,0 +1,91 @@
...
+        ($(
+            $fut.reborrow().take_output().unwrap(),
+        )*)
+    } }
+}
+
+/// `MaybeDone`, a future that may have completed.
+#[doc(hidden)]
+#[derive(Debug)]
+pub enum MaybeDone<A: Future> {
```
MajorBreakfast: We should find a nice place for utility futures like this. `FutureOption`, `ReadyFuture`, `PendingOnce` etc. could all live there.

futures-util/src/select.rs
```diff
@@ -0,0 +1,38 @@
+//! The `select` macro.
+
+#[macro_export]
+macro_rules! select {
+    ($(
+        $name:ident => $body:expr,
+    )*) => { {
+        $(
+            $crate::assert_unpin(&$name);
```
MajorBreakfast: `join!` also moves the value. I think it would be more convenient if the macro works out of the box with `!Unpin` futures by calling the `pin_mut!` macro automatically.

It wouldn't be possible to access the futures after `select()` anymore. But that seems dangerous anyway. If you poll the wrong one, you get a panic.

futures-util/src/lib.rs
```diff
@@ -54,6 +53,85 @@ macro_rules! delegate_sink {
...
+
+    #[doc(hidden)]
+    pub fn assert_unpin<T: Future + Unpin>(_: &T) {}
+
+    /// A macro which returns the result of polling a future once within the
+    /// current `async` context.
+    ///
+    /// This macro is only usable inside of `async` functions, closures, and blocks.
+    #[macro_export]
+    macro_rules! poll {
```
MajorBreakfast: Maybe move these out of `lib.rs`

It would be more ideal if `lib.rs` had no definitions in it.

MajorBreakfast:
These already look really convenient!

- There probably should be some clear distinction between macros intended for use in async functions and other macros. Macros that internally use `await` should be marked somehow
- For developers coming from JavaScript it would be extremely useful if `join` was called `all`. I'm just sayin' 🙂

futures-util/src/select.rs
cramertj: > It wouldn't be possible to access the futures after select() anymore. But that seems dangerous anyway. If you poll the wrong one again, you get a panic.

I think that we strongly want to support this use case-- see my comment in the PR description for how this could work. Every `select` I've ever written aside from timeouts was inside of a loop.

futures-util/src/lib.rs
cramertj: What would you think about moving them all to `src/futures-util/await/*` to signify that they're macros that interact with async/await?

cramertj:
@MajorBreakfast `select!` needs to do the `await!` internally (rather than providing a future as a result) in order to allow the compiler to be smarter about control-flow (e.g. allowing `return` to work normally, enabling the initialized-ness checks to understand the different conditions under which a break can occur, etc.). Since `select!` doessn't return a future, I thought it would be confusing for `join!` to return a future-- the two seem like somewhat natural doubles, and it seems surprising that they would work differently in that respect.

Additionally, users of the 2018 edition can choose the other behavior by wrapping the `join!` or `select!` in an `async { ... }` block.

futures-util/src/select.rs
cramertj: I also considered the possibility of having `select_loop!` with the loop baked in, but I think that approach would be too restrictive.

futures-util/src/select.rs
```diff
@@ -0,0 +1,38 @@
...
+        )*
+        loop {
```
cramertj: I'm going to attempt to change this loop to use `MaybeDone` like `join!` does so that `$body` can go outside the loop, making `break` work like you'd expect.

futures-util/src/lib.rs
```diff
@@ -54,6 +53,85 @@ macro_rules! delegate_sink {
...
+    /// This macro is only usable inside of `async` functions, closures, and blocks.
+    #[macro_export]
+    macro_rules! pending {
+        () => {
+            await!($crate::pending_once())
+        }
+    }
+
+    #[doc(hidden)]
+    pub fn pending_once() -> impl Future<Output = ()> {
```
MajorBreakfast: I've discovered that the tests of the `futures` crate use something similar called `DelayFuture`. Only difference is that it lets you specify the output.

MajorBreakfast:
> Since select! doessn't return a future, I thought it would be confusing for join! to return a future

I agree.

However maybe rename the macro to `await_join!` and `await_select!`? I'm sceptic that hiding the `await` is a good idea.

futures-util/src/select.rs
cramertj: Done.

futures-util/src/lib.rs
cramertj: Done.

futures-util/src/join.rs
```diff
@@ -0,0 +1,91 @@
+//! The `join` macro.
```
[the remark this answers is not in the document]
cramertj: Moved into `src/await/...`

futures-util/src/await/select.rs
```diff
@@ -0,0 +1,49 @@
...
+        if false {
+            unreachable!()
+        }
+        $(
+            else if let Some($name) = $name.take_output() {
+                $body
+            }
+        )*
+        else {
+            unreachable!()
```
cramertj: Note: this is actually reachable by writing `select!{}`-- perhaps worth a special panic message?

futures-util/src/await/join.rs
```diff
@@ -0,0 +1,38 @@
...
+macro_rules! join {
+    ($($fut:ident),*) => { {
+        $(
+            let mut $fut = $crate::future::maybe_done($fut);
+            pin_mut!($fut);
+        )*
+        loop {
+            let mut all_done = true;
+            $(
+                if let ::core::task::Poll::Pending = poll!($fut.reborrow()) {
```
cramertj: Currently this and all other macros in the `futures` crate assume that `::core` is available-- we should make all the macro-containing `futures` crates re-export `core`, or at least `core::{task, future}` and `PInMut` so that the macros can use `$crate::core_reexport::...` cc @MajorBreakfast @aturon

futures-util/src/await/join.rs
```diff
@@ -0,0 +1,38 @@
+//! The `join` macro.
+
+#[macro_export]
+macro_rules! join {
+    ($($fut:ident),*) => { {
```
cramertj: It would be nice to allow `join!` to take exprs rather than specifically idents, but I don't know of a way to do that while still being able to give unique names to each of the inputs.

futures-util/src/await/join.rs
MajorBreakfast: Yes this makes sense. Without this we just get `failed to resolve. Maybe a missing extern crate core;?` errors.

cramertj: Yeah-- not sure what happens in the new edition.

MajorBreakfast: Does `#[doc(hidden)]` work for modules?

futures-util/src/await/select.rs
```diff
@@ -0,0 +1,52 @@
...
+            $(
+                if let ::core::task::Poll::Ready(()) = poll!($name.reborrow()) {
+                    break;
+                }
+            )*
+            pending!();
+        }
+        if false {
+            unreachable!()
+        }
```
MajorBreakfast: `if false { ... }`?

MajorBreakfast: Ah never mind

futures-util/src/await/join.rs
cramertj: Yeah, presumably it would be a `#[doc(hidden)]` reexport of `core`.

futures-util/src/await/mod.rs
```diff
@@ -0,0 +1,81 @@
...
+
+#[doc(hidden)]
+pub fn assert_unpin<T: Future + Unpin>(_: &T) {}
+
+/// A macro which returns the result of polling a future once within the
+/// current `async` context.
+///
+/// This macro is only usable inside of `async` functions, closures, and blocks.
+#[macro_export]
+macro_rules! poll {
```
MajorBreakfast: Maybe put the `poll` and `pending` in their own files

futures-util/src/await/join.rs
withoutboats: One option I think is to use a head/rest style unrolling series of macro invocations, allowing you to exploit hygiene to use the same variable name in the same scope. A trivial example of what I mean:

```rust
macro_rules! m {
    ($head:expr, $($rest:expr),*) => {{
        let x = $head;
        m!($($rest),*);
        println!("{}", x);
    }};
    ($last:expr) => {
        let x = $last;
        println!("{}", x);
    };
}
```

withoutboats: Actually, that way definitely doesn't work.. What if instead you wrote a macro which expanded to an array of pin_mut'd futures; could you avoid naming them directly that way?

cramertj: @withoutboats That was actually the first way I tried, but because the `Future` trait isn't object-safe right now (because of arbitrary_self_types) it doesn't work :(

PR #1053, opened 2018-06-28
Fix compilation when "nightly" feature is enabled

ngg:
This commit just removes the references to the already removed sub-crates.


cramertj:
Thanks!

PR #1054, opened 2018-06-28
Rust 2018

MajorBreakfast:
Since we're on nightly already, we might as well use Rust 2018. This gives Rust 2018 some usage and it costs us nothing because futures wont stabilize before Rust 2018.

I especially like `dyn Trait`. It makes it much clearer where trait objects are used.

This PR includes https://github.com/rust-lang-nursery/futures-rs/pull/1053

### DIFF
--- a/futures-channel/src/mpsc/mod.rs
+++ b/futures-channel/src/mpsc/mod.rs
@@ -88,7 +88,7 @@ use std::usize;
 use futures_core::task::{self, Waker};
 use futures_core::{Poll, Stream};
 
-use mpsc::queue::{Queue, PopResult};
+use crate::mpsc::queue::{Queue, PopResult};
 
 mod queue;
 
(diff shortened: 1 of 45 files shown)


MajorBreakfast:
Hmm. "costs us nothing" isn't exactly right. If we want to do "the stable dance", then we need to keep the old module system and only use dyn trait

futures-channel/src/mpsc/mod.rs
```diff
@@ -88,7 +88,7 @@ use std::usize;
 use futures_core::task::{self, Waker};
 use futures_core::{Poll, Stream};
 
-use mpsc::queue::{Queue, PopResult};
+use crate::mpsc::queue::{Queue, PopResult};
```
cramertj: Is this what `rustfix` generates? I think `use self::queue::{....};` makes more sense than walking down from the crate root.

MajorBreakfast: Yes. This is what was generated. However, the line it replaced seems to have used a crate relative path as well.

MajorBreakfast:
I'm closing this now. I think we want to do this conversion at some point, but not quite yet

PR #1056, opened 2018-06-29
0.2: Fix compilation when "nightly" feature is enabled

ngg:
Some of the crates are not compiling with the latest nightly compiler when "nightly" feature is enabled (because of https://github.com/rust-lang/rust/pull/51730).
Please release a new version of the "futures-core-preview", "futures-stable-preview" and "futures-async-runtime-preview" after merging this.

### DIFF
--- a/futures/tests/async_await/pinned.rs
+++ b/futures/tests/async_await/pinned.rs
@@ -1,6 +1,7 @@
 use futures::stable::block_on_stable;
 use futures::executor::{block_on, ThreadPool};
 use futures::prelude::*;
+use futures::prelude::await;
 
 #[async]
 fn foo() -> Result<i32, i32> {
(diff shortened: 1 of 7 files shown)


futures/tests/async_await/pinned.rs
```diff
@@ -1,6 +1,7 @@
 use futures::stable::block_on_stable;
 use futures::executor::{block_on, ThreadPool};
 use futures::prelude::*;
+use futures::prelude::await;
```
cramertj: @ngg Is `await` not included automatically by the `*` import above?

ngg: Without this the compiler gives an error "error[E0659]: `await` is ambiguous", "note: `await` could refer to the name imported here" (referring to the prelude), "note: `await` is also a builtin macro", "note: consider adding an explicit import of `await` to disambiguate". Explicitly importing it solves this.

cramertj: Ah, gotcha. Thanks for explaining!

cramertj:
@ngg I've published the updates to these three crates. Thanks for fixing this!

PR #1057, opened 2018-06-30
0.3: Fix compilation with no_std

ngg:
await! won't work with no_std, but at least the crates compile.

### DIFF
--- a/futures-core/src/lib.rs
+++ b/futures-core/src/lib.rs
@@ -73,7 +73,7 @@ macro_rules! pin_mut {
         // Shadow the original binding so that it can't be directly accessed
         // ever again.
         #[allow(unused_mut)]
-        let mut $x = unsafe { ::std::mem::PinMut::new_unchecked(&mut $x) };
+        let mut $x = unsafe { ::core::mem::PinMut::new_unchecked(&mut $x) };
     )* }
 }
 
--- a/futures-core/src/lib.rs
+++ b/futures-core/src/lib.rs
@@ -18,6 +18,11 @@ extern crate std;
 #[cfg(feature = "either")]
 extern crate either;
 
+#[doc(hidden)]
+pub mod core_reexport {
+    pub use core::*;
+}
+
 macro_rules! if_std {
     ($($i:item)*) => ($(
         #[cfg(feature = "std")]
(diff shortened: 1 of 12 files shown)


futures-core/src/lib.rs
```diff
@@ -73,7 +73,7 @@ macro_rules! pin_mut {
         // Shadow the original binding so that it can't be directly accessed
         // ever again.
         #[allow(unused_mut)]
-        let mut $x = unsafe { ::std::mem::PinMut::new_unchecked(&mut $x) };
+        let mut $x = unsafe { ::core::mem::PinMut::new_unchecked(&mut $x) };
```
MajorBreakfast: Could you implement it via a reexport of libcore and see if the strategy @cramertj proposes [here](https://github.com/rust-lang-nursery/futures-rs/pull/1051#discussion_r198942744) works?

ngg: Yes, it seems to work.

futures-core/src/task/mod.rs
```diff
@@ -1,5 +1,6 @@
 //! Task notification.
 
+#[cfg(feature = "std")]
```
MajorBreakfast: Maybe move this into `if_std! { ... }`

futures-executor/src/lib.rs
```diff
@@ -13,9 +13,13 @@
 #[macro_use]
 extern crate std;
 
+#[cfg(feature = "std")]
 #[macro_use]
 extern crate futures_core;
 
+#[cfg(not(feature = "std"))]
+extern crate futures_core;
```
MajorBreakfast: Same here

futures-util/src/lib.rs
```diff
@@ -31,9 +31,13 @@ macro_rules! if_std {
     )*)
 }
 
+#[cfg(feature = "std")]
```
MajorBreakfast: And here :)

MajorBreakfast:
I made a few notes. Thanks for fixing this! I also consider good `no_std` support important.

BTW `await` currently only works with `std` because it *currently* relies on thread local storage to access the `Context`.

MajorBreakfast:
Could you apply the reexport thing also to `join!` and `select!`? Then it's applied to everything 🙂

https://github.com/rust-lang-nursery/futures-rs/tree/0.3/futures-util/src/await

MajorBreakfast:
I must have missed that somehow. Good work!

BTW I‘m not authorited to merge PRs. We‘ll need to wait for @cramertj to do it. Likely on Monday

cramertj:
This looks great, thanks so much!

PR #1063, opened 2018-07-03
More tests

MajorBreakfast:


futures-channel/tests/mpsc.rs
```diff
@@ -410,7 +410,7 @@ async fn stress_poll_ready_sender(mut sender: mpsc::Sender<u32>, count: u32) {
     for i in (1..=count).rev() {
         // FIXME: should assert `is_ok()`, but cannot because assigning the result
         // of this expression to a variable causes an ICE. See rust-lang/rust/issues/51995.
-        await!(sender.send(i));
+        await!(sender.send(i)).unwrap();
```
cramertj: huh-- interesting that this works.

MajorBreakfast: It returns a `Result` which throws a warning if we don't handle it

cramertj: @MajorBreakfast Yes-- I meant that it's interesting that `unwrap()` doesn't cause the same invalid MIR error that `.is_ok()` does.

cramertj: (see the linked issue)

MajorBreakfast: Haha I didn‘t even read that xD All I saw was a pesky warning

PR #1065, opened 2018-07-03
Implement a lot of clippy suggestions

MajorBreakfast:


futures-util/src/future/join.rs
```diff
@@ -57,7 +59,7 @@ macro_rules! generate {
                 )*
 
                 if all_done {
-                    Poll::Ready(($(self.$Fut().take_output().unwrap_or_else(|| unreachable!())),*))
```
MajorBreakfast: This looks strange, but at the same time it feels like I'm missing something

futures-util/src/stream/unfold.rs
```diff
@@ -90,21 +90,19 @@ impl<T, F, Fut, It> Stream for Unfold<T, F, Fut>
...
-                *self.state() = Some(next_state);
-                return Poll::Ready(Some(item))
-            } else {
-                return Poll::Ready(None)
-            }
+        if let Some((item, next_state)) = step {
+            *self.state() = Some(next_state);
+            return Poll::Ready(Some(item))
+        } else {
+            return Poll::Ready(None)
```
MajorBreakfast: Here clippy said the loop was useless.

futures-util/src/future/join.rs
cramertj: These aren't quite the same-- `unwrap` only works for `Debug` types, while `unwrap_or_else(|| unreachable!())` works for all types.

cramertj: Edit: ignore last-- that used to be true when this was a `Result`, but now that it is an `Option`, unwrapping is fine.

futures-util/src/stream/concat.rs
```diff
@@ -46,17 +46,21 @@ impl<S: Stream + Unpin> Unpin for Concat<S> {}
...
+    fn poll(
+        mut self: PinMut<Self>, cx: &mut task::Context
+    ) -> Poll<Self::Output> {
         loop {
             match self.stream().poll_next(cx) {
                 Poll::Pending => return Poll::Pending,
                 Poll::Ready(None) => {
-                    return Poll::Ready(self.accum().take().unwrap_or_else(|| Default::default()))
+                    return Poll::Ready(
+                        self.accum().take().unwrap_or_else(Default::default)
```
cramertj: Even better: swap out for `unwrap_or_default`: https://doc.rust-lang.org/std/option/enum.Option.html#method.unwrap_or_default

futures-util/src/stream/unfold.rs
cramertj: Nice! Looks like an artifact of how this used to be written: the `self.state`-setting case would wrap around: https://github.com/rust-lang-nursery/futures-rs/blob/master/futures-util/src/stream/unfold.rs#L89

cramertj:
Cool! Thanks for these fixes.

futures-util/src/stream/concat.rs
MajorBreakfast: cool. Now it will fit into one line

MajorBreakfast: I'll simply include this in the next PR

PR #1067, opened 2018-07-03
More tests 3

MajorBreakfast:


futures-util/src/sink/map_err.rs
```diff
@@ -59,19 +59,19 @@ impl<S, F, E> Sink for SinkMapErr<S, F>
...
     }
 
     fn poll_flush(mut self: PinMut<Self>, cx: &mut task::Context) -> Poll<Result<(), Self::SinkError>> {
-        self.sink().poll_flush(cx).map_err(self.expect_f())
+        self.sink().poll_flush(cx).map_err(|e| self.expect_f()(e))
     }
 
     fn poll_close(mut self: PinMut<Self>, cx: &mut task::Context) -> Poll<Result<(), Self::SinkError>> {
-        self.sink().poll_close(cx).map_err(self.expect_f())
+        self.sink().poll_close(cx).map_err(|e| self.expect_f()(e))
```
MajorBreakfast: Why did this fail?

cramertj: Ah-- this wasn't the same. `expect_f` is only called in the case of an error in this version. With the clippy-suggested change, `expect_f` is called regardless of whether an error occurs or not.

MajorBreakfast: Ah I see. It‘s not a method created by the unsafe_unpinned macro. Good that we have unit tests :)

cramertj: :)

PR #1069, opened 2018-07-04
More tests 4

MajorBreakfast:


futures-core/src/lib.rs
```diff
@@ -30,57 +30,8 @@ macro_rules! if_std {
     )*)
 }
 
-#[macro_export]
-macro_rules! pinned_deref {
```
MajorBreakfast: I removed this macro. It was never used anywhere

futures-core/src/lib.rs
```diff
@@ -30,57 +30,8 @@ macro_rules! if_std {
...
-    ($e:expr) => (
-        $crate::core_reexport::mem::PinMut::new_unchecked(
-            &mut **$crate::core_reexport::mem::PinMut::get_mut_unchecked($e.reborrow())
-        )
-    )
-}
-
-#[macro_export]
-macro_rules! pinned_field {
```
MajorBreakfast: I removed this macro because it's an anti pattern. Better alternatives are:
- Usage of `PinMut::map_unchecked`
- `unsafe_pinned!` macro

futures-core/src/lib.rs
```diff
@@ -30,57 +30,8 @@ macro_rules! if_std {
...
-        // Move the value to ensure that it is owned
-        let mut $x = $x;
-        // Shadow the original binding so that it can't be directly accessed
-        // ever again.
-        #[allow(unused_mut)]
-        let mut $x = unsafe { $crate::core_reexport::mem::PinMut::new_unchecked(&mut $x) };
-    )* }
-}
+#[macro_use]
+mod macros;
```
MajorBreakfast: I've put them in a macro module This is cleaner, but ultimately when macro imports are implemented I'd like to:
- Put the poll macros under `futures::task`
- Put the pin macros... not sure where. We need to decide that once macros can be imported via `use` statements

futures-core/src/macros/pin.rs
```diff
@@ -0,0 +1,43 @@
+#[macro_export]
+macro_rules! unsafe_pinned {
+    ($f:tt -> $t:ty) => (
+        fn $f<'a>(
+            self: &'a mut $crate::core_reexport::mem::PinMut<Self>
+        ) -> $crate::core_reexport::mem::PinMut<'a, $t> {
+            unsafe {
+                $crate::core_reexport::mem::PinMut::map_unchecked(
+                    self.reborrow(), |x| &mut x.$f
+                )
```
MajorBreakfast: I've changed this to `map_unchecked`. It works the same.

Also, here and in the following macros the `$crate::core_reexport::` prefixes were partly missing.

futures-util/src/future/flatten.rs
```diff
@@ -43,6 +53,6 @@ impl<A> Future for Flatten<A>
     type Output = <A::Output as Future>::Output;
 
     fn poll(mut self: PinMut<Self>, cx: &mut task::Context) -> Poll<Self::Output> {
-        unsafe { pinned_field!(self, state) }.poll(cx, |a, ()| a)
+        self.state().poll(cx, |a, ()| a)
```
MajorBreakfast: I removed all occurences of the `pinned_field!` macro

futures/src/macros/mod.rs
```diff
@@ -0,0 +1,7 @@
+// Macros redefined here because macro re-exports are unstable.
```
MajorBreakfast: That's the reason why all macros are duplicated

futures-util/src/sink/send_all.rs
```diff
@@ -65,7 +65,7 @@ where
 impl<'a, T, U> Future for SendAll<'a, T, U>
 where
     T: Sink + Unpin + 'a + ?Sized,
-    U: Stream<Item = Result<T::SinkItem, T::SinkError>> + Unpin + 'a + ?Sized,
+    U: Stream<Item = T::SinkItem> + Unpin + 'a + ?Sized,
```
MajorBreakfast: Signature was wrong

futures/tests/recurse.rs
```diff
@@ -0,0 +1,25 @@
...
+        } else {
+            FutureObj::new(Box::new(future::ready((n - 1, x + n)).then(do_it)))
+        }
+    }
+
+    let (tx, rx) = mpsc::channel();
+    ::std::thread::spawn(|| {
+        block_on(do_it((1_000, 0)).map(move |x| tx.send(x).unwrap()))
+    });
+    assert_eq!(500_500, rx.recv().unwrap());
```
MajorBreakfast: I made it a bit fancier that it was before. Now it adds the numbers `0 + 1 + ... + 1000 = 500,500`

futures/tests_disabled/unfold.rs
```diff
@@ -1,52 +0,0 @@
...
-
-    sassert_empty(&mut stream);
-    sassert_next(&mut stream, 4);
-
-    // no more items
-    sassert_done(&mut stream);
-}
-
-#[test]
-fn unfold_err1() {
```
MajorBreakfast: Streams are no longer error-aware, therefore this test doesn't apply anymore.

MajorBreakfast:
@aturon This can be reviewed

futures-core/src/lib.rs
MajorBreakfast: Correction: Macro imports via use are implemented. They‘re just feature gated. We could use them.

futures-util/src/try_future/try_chain.rs
```diff
@@ -0,0 +1,66 @@
+use core::mem::PinMut;
+use futures_core::TryFuture;
+use futures_core::task::{Context, Poll};
+
+#[must_use = "futures do nothing unless polled"]
+#[derive(Debug)]
+crate enum TryChain<Fut1, Fut2, Data> {
```
MajorBreakfast: TryChain is a new mechanism so that `and_then` and `or_else` can share a single implementation.

futures-util/src/try_future/try_chain.rs
```diff
@@ -0,0 +1,66 @@
...
+                TryChain::Second(fut2) => {
+                    // Poll the second future
+                    return unsafe { PinMut::new_unchecked(fut2) }.try_poll(cx)
+                }
+                TryChain::Empty => {
+                    panic!("future must not be polled after it returned `Poll::Ready`");
+                }
+            };
+
+            *this = TryChain::Empty; // Drop fut1
```
MajorBreakfast: The eager_drop tests want this: They expect that the old futures is dropped before the closure runs.

futures-util/src/try_future/mod.rs
```diff
@@ -100,11 +112,11 @@ pub trait TryFutureExt: TryFuture {
...
     /// assert_eq!(block_on(new_future), Err(1));
     /// ```
-    fn map_ok<U, F>(self, f: F) -> MapOk<Self, F>
-        where F: FnOnce(Self::Item) -> U,
+    fn map_ok<T, F>(self, op: F) -> MapOk<Self, F>
+        where F: FnOnce(Self::Item) -> T,
               Self: Sized,
     {
-        map_ok::new(self, f)
+        MapOk::new(self, op)
```
MajorBreakfast: I made these `new` functions methods

futures-util/src/try_future/map_err.rs
```diff
@@ -1,39 +1,43 @@
...
 ///
 /// This is created by the `Future::map_err` method.
 #[derive(Debug)]
 #[must_use = "futures do nothing unless polled"]
-pub struct MapErr<A, F> {
-    future: A,
-    f: Option<F>,
+pub struct MapErr<Fut, F> {
+    future: Fut,
+    op: Option<F>,
```
MajorBreakfast: `op` is how the standard library's `Result` methods call this

futures-util/src/try_future/into_future.rs
```diff
@@ -0,0 +1,28 @@
+use core::mem::PinMut;
+use futures_core::{Future, TryFuture};
+use futures_core::task::{Context, Poll};
+
+/// Converts a `TryFuture` into a normal `Future`
+#[derive(Debug)]
+#[must_use = "futures do nothing unless polled"]
+pub struct IntoFuture<Fut> {
```
MajorBreakfast: This is a new invention. I think that could be useful. I originally planned to use it for something, but my plan changed.

futures-util/src/try_future/unwrap_or_else.rs
```diff
@@ -0,0 +1,44 @@
...
+use core::mem::PinMut;
+use futures_core::{Future, TryFuture, Poll};
+use futures_core::task;
+
+/// Future for the `unwrap_or_else` combinator. It unwraps the result, returning
+/// the content of the `Ok` as `Output` or if the value is an `Err` then it
+/// calls `op` with its value.
+#[derive(Debug)]
+#[must_use = "futures do nothing unless polled"]
+pub struct UnwrapOrElse<Fut, F> {
```
MajorBreakfast: This was previously called `Recover`. The made sense for futures 0.1 and 0.2, but now it works *exactly* like `Result::unwrap_or_else`

futures-util/src/try_future/or_else.rs
```diff
@@ -1,65 +1,45 @@
...
-                        }
-                    }
-                }
-                State::Second(ref mut fut2) => {
-                    // Safe to create a new `PinMut` because `fut2` will never move
-                    // before it's dropped; once we're in `Chain::Second` we stay
-                    // there forever.
-                    return unsafe { PinMut::new_unchecked(fut2) }.try_poll(cx)
-                }
-            };
```
MajorBreakfast: All this code now lives in `TryChain`

futures-util/src/future/chain.rs
```diff
@@ -1,54 +1,52 @@
...
-use futures_core::task;
+use futures_core::Future;
+use futures_core::task::{Context, Poll};
 
 #[must_use = "futures do nothing unless polled"]
 #[derive(Debug)]
-pub enum Chain<Fut1, Fut2, Data> {
-    First(Fut1, Option<Data>),
+crate enum Chain<Fut1, Fut2, Data> {
+    First(Option<Fut1>, Option<Data>),
```
seanmonstar: I can't find when this option is used, is it still needed?

futures-util/src/async_await/poll.rs
```diff
@@ -0,0 +1,32 @@
+use futures_core::{task, Future, Poll};
+use core::marker::Unpin;
+use core::mem::PinMut;
+
+/// A macro which returns the result of polling a future once within the
+/// current `async` context.
+///
+/// This macro is only usable inside of `async` functions, closures, and blocks.
```
seanmonstar: I might be missing some context, how or why is this macro to be used?

futures-util/src/future/map.rs
```diff
@@ -9,40 +8,37 @@ use futures_core::task;
...
-}
+impl<Fut, F> Map<Fut, F> {
+    unsafe_pinned!(future -> Fut);
+    unsafe_unpinned!(op -> Option<F>);
 
-impl<A: Future, F> Map<A, F> {
-    unsafe_pinned!(future -> A);
-    unsafe_unpinned!(f -> Option<F>);
+    /// Creates a new Map.
+    pub fn new(future: Fut, op: F) -> Map<Fut, F> {
```
seanmonstar: The previous `new` functions where actually not exported publicly. This change makes the constructors public and thus exposes what types are used internally. If moved from module functions to inherent functions, they should be `pub(super)`.

futures-util/src/try_future/unwrap_or_else.rs
seanmonstar: As another data point, Scala futures call this operation `recover`. That may be where the name came from here. I think I prefer `recover` to `unwrap_or_else`. Especially since it's not actually unwrapping...

futures-util/src/future/chain.rs
MajorBreakfast: `Option<Data>` is for line 38: It is used to safely take out the data from the enum variant:
```rust
Poll::Ready(output) => (output, data.take().unwrap()),
```

`Option<Fut1>` is unnecessary. Good catch! I played with this code for quite a while and I ended up introducing  `Chain::Empty` to drop the future before the callback runs because that seemed more efficient. But it seems I messed it up. I got it right in `Chain`'s equivalent `TryChain`, though :)

I've removed it. Thanks!

futures-util/src/async_await/poll.rs
MajorBreakfast: These helper macros are used in @cramertj's `join!` and `select!`:
- [join.rs line 13](https://github.com/MajorBreakfast/futures-rs/blob/0ce2116a0b50355386760ba0df8df8cf11491f4c/futures-util/src/async_await/join.rs#L13)
- [select.rs line 18](https://github.com/MajorBreakfast/futures-rs/blob/0ce2116a0b50355386760ba0df8df8cf11491f4c/futures-util/src/async_await/select.rs#L18)

I extracted `poll!` and `pending!` out of `mod.rs` because `mod.rs` felt a bit unwieldy.

MajorBreakfast:
@seanmonstar Thanks for reviewing! Keep the comments coming :)

futures-util/src/try_future/unwrap_or_else.rs
MajorBreakfast: Its behavior changed in futures 0.3. It now transforms a `Future<Output = Result<T, E>>` into a `Future<Output = T>`. According to git blame this seems to have been defined by @aturon approximately 3 months ago. In think we should align with libstd terminology because we've got an exact match here. Rust users will appreciate it. It is also fitting to change the name because it works now differently than in 0.2.

MajorBreakfast: I took a look at Scala's `recover`. Scala futures seem to have a success and error state like we had in futures 0.2 or like in JavaScript.

BTW the only thing that bugs me is that for `TryFuture` the `and_then` and `or_else` methods, which `std::result::Result` both also has, are asynchronous, the others `map_ok` (called `map` on `Result`), `map_err` and now `unwrap_or_else` have synchronous callbacks. In futures 0.2 this was partly sidestepped by implementing `IntoFuture` on `Result`. That can't be done in 0.3. It would work for these `TryFuture` functions, but it wouldn't work on `FutureExt::then` because the return value of the callback can be anything.

futures/tests/basic_combinators.rs
```diff
@@ -0,0 +1,106 @@
...
+        })
+        .map_err(move |x: i32| {
+            tx9.send(x).unwrap(); // Send 9
+            tx9.send(10).unwrap(); // Send 10
+            11
+        })
+        .map(move |x: Result<i32, i32>| {
+            tx10.send(x.err().unwrap()).unwrap(); // Send 11
+            tx10.send(12).unwrap(); // Send 12
+        });
```
MajorBreakfast: The tests for these combinators were previously different. Rather than going through all of the tests and making sure that they cover all possibilities, I came up with this approach that tests each of the 4 combinators first with an `Ok` as input and then with an `Err`.

futures-channel/src/oneshot.rs
```diff
@@ -92,18 +92,18 @@ struct Inner<T> {
 /// use futures::*;
 ///
 /// fn main() {
-///     let (p, c) = oneshot::channel::<i32>();
+///     let (tx, rx) = oneshot::channel::<i32>();
```
MajorBreakfast: Use standard naming

MajorBreakfast:
Added tests for basic combinators and `oneshot::channel()`

MajorBreakfast:
Woohoo. Green CI 🎉

And I didn't cheat. To the contrary: The docs now actually build and the CI runs the tests again

futures/src/lib.rs
```diff
@@ -23,126 +25,25 @@
 
 #![no_std]
 
-#![deny(missing_docs, missing_debug_implementations, warnings)]
+#![deny(missing_docs, missing_debug_implementations)]
```
MajorBreakfast: Rustdoc for reexports seems broken. The crates all build fine individually (e.g. `cargo doc -p futures-core`). But in when reexported in the futures crate we get a bunch of warnings

futures-channel/src/mpsc/mod.rs
```diff
@@ -1,38 +1,42 @@
...
 //! such, it is usually desirable to perform a "clean" shutdown. To do this, the
 //! receiver will first call `close`, which will prevent any further messages to
 //! be sent into the channel. Then, the receiver consumes the channel to
 //! completion, at which point the receiver can be dropped.
+//!
+//! [`Sender`]: struct.Sender.html
+//! [`Receiver`]: struct.Receiver.html
+//! [`Stream`]: ../../futures_core/stream/trait.Stream.html
+//! [`Receiver::poll_next`]:
+//!     ../../futures_core/stream/trait.Stream.html#tymethod.poll_next
```
MajorBreakfast: I added these links like so. I wasn't able to make it build otherwise. Seems to be a bug in all `//!` comments

futures/src/lib.rs
```diff
@@ -23,126 +25,18 @@
...
-extern crate futures_util;
-
-#[doc(hidden)]
-pub use futures_core::core_reexport;
-pub use futures_core::future::Future;
-pub use futures_util::future::FutureExt;
-pub use futures_core::stream::Stream;
-pub use futures_util::stream::StreamExt;
-pub use futures_sink::Sink;
-pub use futures_util::sink::SinkExt;
```
MajorBreakfast: I removed these top level exports because using them is an anti-pattern. It's much easier to either import from the right location like `futures::task::Poll` or to import the prelude `futures::prelude::*`. Anything in-between doesn't make sense.

seanmonstar: I disagree. I don't see any anti-pattern. I frequently have `use futures::{Future, Stream}` in modules. Making them harder to import doesn't improve anything.

Glob imports and preludes, however, definitely *are* an anti-pattern.

Nemo157: I agree with @seanmonstar, having to begin every file with

```rust
use futures::{core::{future::Future, stream::Stream}, utils::{future::FutureExt, stream::StreamExt}};
```

Instead of

```rust
use futures::{Future, Stream, FutureExt, StreamExt};
```

Would be very annoying.

MajorBreakfast: @Nemo157 @seanmonstar Thanks for your comments!

My reaction is that all this top-level stuff accumulates. There is also:
```
Context, ContextExt, CoreFutureExt, FutureObj, LocalFutureObj
```
These are just the most common imports, but there are obviously more. (`FutureObj` exists because `dyn Trait` is currently impossible for traits that use arbitrary self types)  My intention behind removing these top-level imports is because it makes the docs better. Top level exports are a bad thing because they destroy the structure. If everything is exported at the top level anyway, there is no real reason to have submodules at all.  How can we decide which things are considered important enough to be exported at the top level? How will it grow in the future? How would it look like if the standard library did it? I think that it doesn't scale and we shouldn't do it.
The hierarchy I'm proposing here is the same as in libcore: libcore also has differentiates between the submodules `future` and `task`. And a `stream` submodule will probably exist in the future.

@Nemo157 `FutureExt` is defined in the utils crate, but there's there's no such module

This is how I imagine it:
```rust
use futures::future::{Future, FutureExt};
use futures::stream::{Stream, StreamExt};
use futures::task::{Task, Context};
```
I think that actually looks pretty nice. That said, the approach I'm proposing here is just a suggestion. This PR has become larger large, so I explictly highlighted this change to make it absolutely obvious. I certainly don't want to decide that for everyone. If you all really dislike it, we can revert to the old approach. I'm confident, though, that the approach of clear separation is better.

futures/src/lib.rs
```diff
@@ -140,21 +140,34 @@ pub mod future {
...
+        // For FutureExt:
+        CatchUnwind, Shared
+
+        // ToDo: JoinAll, SelectAll, SelectOk, join_all, select_all, select_ok
+    };
+
+    pub use futures_util::try_future::{
+        TryFutureExt,
+        AndThen, ErrInto, FlattenSink, IntoFuture, MapErr, MapOk, OrElse,
+        UnwrapOrElse,
```
MajorBreakfast: I've put the stuff for `TryFuture` into the `future` module

MajorBreakfast: Currently the `TryFuture` trait does not have `Future` as super trait. But, this is planned to change  as soon as it becomes technically possible. Therefore, merging the two makes sense to me. What do you think?

MajorBreakfast:
@cramertj In case you're wondering: I moved the `futures-util/benches/bilock.rs` to `futures-util/benches_disabled/bilock.rs` in order to disable it. It hasn't been converted to 0.3 yet.

futures/src/lib.rs
nayato: I think idea of top-level re-exports serves both convenience for common case (easy to reach things you need without clutter) and discoverability. It's hard to argue that Future, Stream and Sink (and their extension traits since 0.2) are core concepts to futures crates. In my view, public submodules still make sense for things that are more specialized, e.g. deeper dive into streaming functionality for writing your own combinators. Being able to import things through `futures::{Future, FutureExt, Stream, StreamExt}` and `futures::stream::{Stream, StreamExt, FilterMap, Empty}` is both useful in my opinion if only in different contexts.

MajorBreakfast: Clean:
screen shot 2018-07-09 at 17 32 27

Cluttered:
screen shot 2018-07-09 at 17 38 22

seanmonstar: I don't see any problem with the traits list there at the bottom. *If* it were a big deal, they can be `doc(hidden)` from the main page, but I don't think it's an issue.

I definitely wouldn't make actually *using* these traits harder just for the docs change.

Nemo157: One thing to note is that `futures` 0.2 didn't export `Context` to the top-level. As far as I am aware the standard was to use `task::Context` everywhere. That doesn't reduce it much though 😦

I wonder if trait aliases could help with all those extension traits. In `futures` have something like

```rust
trait Future = ::core::future::Future + CoreFutureExt + FutureExt;
```

(although currently the RFC doesn't specify that trait aliases bring methods into scope, [I left a comment on the tracking issue to find out](https://github.com/rust-lang/rust/issues/41517#issuecomment-403550419)).

MajorBreakfast: @seanmonstar My main concern is the impression a new user gets when looking at the docs for the first time. I've added a commit that adds the reexports as `doc(hidden)`. As long as there is no big pile of unsorted stuff visible to users, it's fine I guess.

@Nemo157 I think it would be confusing if more than one thing was called a `Future`. Would `impl Future for MyType` even work for a trait alias?

futures-util/src/async_await/poll.rs
cramertj: @seanmonstar `poll` is used to poll a future once inside of an `async` function, block, or closure. Internally, it works by `await`ing a future which polls once and yields the result.

futures-util/src/future/map.rs
```diff
@@ -1,48 +1,44 @@
...
-}
+impl<Fut, F> Map<Fut, F> {
+    unsafe_pinned!(future -> Fut);
+    unsafe_unpinned!(op -> Option<F>);
 
-impl<A: Future, F> Map<A, F> {
-    unsafe_pinned!(future -> A);
-    unsafe_unpinned!(f -> Option<F>);
+    /// Creates a new Map.
+    pub(super) fn new(future: Fut, op: F) -> Map<Fut, F> {
```
cramertj: Why change these functions to `pub(super)` methods in some places and not others? I'm fine with it either way, but we should be consistent.

futures-util/src/try_future/unwrap_or_else.rs
cramertj: > As another data point, Scala futures call this operation recover. That may be where the name came from here. I think I prefer recover to unwrap_or_else. Especially since it's not actually unwrapping...

@seanmonstar Yeah, that is where I got the name `recover` from. I don't have a strong opinion here-- I personally think `recover` is clearer, but the symmetry of `unwrap_or_else` with `Result` is nice.

MajorBreakfast:
🎉

Issue #993, opened 2018-04-29
Data for 0.1.21 is missing, tag 1.20 still present


stbuehler:
Hi,

there is a release for 0.1.21, but no tag. But there is https://github.com/cramertj/futures-rs/commits/futures-0.21

Also there is now a tag 0.1.20 (as requested in #920), but doesn't point to the same commit as 1.20 (but probably the same tree), and 1.20 still exists; you may want to think about removing the old tag (`git push origin :refs/tags/1.20` should do the trick)

It seems https://github.com/cramertj/futures-rs/commits/futures-0.21 is not fast-forwardable onto the 0.1 branch, but merging should work afaict. Could the 0.1 branch please be kept in a maintainable state?

berkus:
Yes, folks, please fix the tagging situation, it's currently somewhat misleading.

dekellum:
@cramertj it sounds like futures release 0.1.21 on crates.io is your work? Could you merge this here? Tag and CHANGELOG?

dekellum:
I arrived here looking for the changelog (or lacking that, tag and commits) for 0.1.21, but I supposed if I can upgrade off of 0.1.21 to 0.1.22 and have those things, then that would be sufficient, thanks.

dekellum:
On the other hand, if it's going to take weeks to release 0.1.22, what's the problem with merging 0.1.21 here? Key people out of town (bus factor++)?

cramertj:
Sorry for the delay responding @dekellum-- yeah, I was out of town this last week. I can push up 0.1.22 now (opened https://github.com/rust-lang-nursery/futures-rs/pull/1072). As for why we couldn't just merge in 0.1.21, there's not a clear commit to tag with 0.1.21 as the 0.1.21 commit that was released doesn't have the same history as the futures-rs 0.1 branch, so we'd have to edit the history of the 0.1 branch in order to match. This wouldn't be the end of the world, but since a 0.1.22 release had been requested for other reasons, I figured I'd go ahead and push up 0.1.22.

cramertj:
0.1.22 has been published.

dekellum:
Thanks and I see a basic tag (without changes listed) is in place.  No CHANGELOG entry for 0.1.22 in the expected place(s), though?

* https://github.com/rust-lang-nursery/futures-rs/blob/0.1/CHANGELOG.md
* https://github.com/rust-lang-nursery/futures-rs/blob/master/CHANGELOG.md

[...]

PR #1073, opened 2018-07-09
Note that CHANGELOG.md (0.1 branch) is no longer maintained

dekellum:
Per discussion with @cramertj  in #993.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+**Note**: This CHANGELOG is no longer maintained for newer 0.1.x releases.
+See instead the github release tags and individual git commits.
+
+-----
+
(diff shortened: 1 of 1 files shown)


carllerche:
I don't see where in the linked issue it says that 0.1 is unmaintained. Given that Tokio and a large chunk of the ecosystem is on 0.1 and will remain so for the foreseeable future, I would think it wise to continue maintaining 0.1.

dekellum:
Ooops, I might have chose my words more carefully in the git commit and this PR, but if you look at the diff, I'm only indicating that the *CHANGELOG.md file* is not being maintained (it hasn't since 1.1.17)!  

Apologies for any alarming FUD!

cramertj:
Thanks!

PR #1074, opened 2018-07-09
Note that CHANGELOG.md (master) is no longer maintained

dekellum:
Per discussion with @cramertj in #993.


dekellum:
As this is only a small addition of a note in CHANGELOG.md, the CI failure is unrelated.

cramertj:
Thanks!

PR #1077, opened 2018-07-10
Associated `new` functions

MajorBreakfast:
- Turn `new` functions into associated functions
- 80char idents
- Enabled test for `fanout`
- Consistently call `Stream` type param `St`
- Consistently call `Sink` type param `Si`
- Consistently call `Future` type param `Fut`


MajorBreakfast:
@cramertj It's probably better if you land this PR first instead of the pinning PR

I didn't add the associated `new` function to all combinators yet because there are so many. This PR applies it to the `Future`, i/o and `Sink` combinators.

cramertj:
This looks great! Thanks!

PR #1078, opened 2018-07-10
Core cleanup

MajorBreakfast:
- `AtomicWaker` and `ContextExt` are now in `futures_util::task` and `ContextExt` is now actually reexported in `futures`
- Macros are now in `futures_util::macros`
- `ReadyFuture` is now called `Ready` an in `futures_util::future`
- `FutureOption` is now called `OptionFuture` an in `futures_util::future` with a code example
- `futures_util` top-level exports are now also hidden in the docs


cramertj:
Works for me!

Issue #942, opened 2018-04-05
AtomicWaker doc example depends on its own explicitly undocumented behavior


adlerd:
In `0.2.0-beta` the docs state "**AtomicWaker does not provide any memory ordering guarantees**, as such the user should use caution and use other synchronization primitives to guard the result of the underlying computation." But consider the documentation example (expanded to include a second thread for concreteness, and with the actual Future trait dropped for simplicity):

```rust
struct Flag {
    waker: AtomicWaker,
    set: AtomicBool,
}

impl Flag {
    fn thread1_poll(&mut self, cx: &mut task::Context) -> Poll<(), Never> {
        self.waker.register(cx.waker());

        if self.set.load(SeqCst) {
            Ok(Ready(()))
        } else {
            Ok(Pending)
        }
    }
    fn thread2(&mut self) {
        self.set.store(true, SeqCst);
        self.waker.wake();
    }
}
```

In order for this to work we need either `load` to return `true` or someone to actually wake thread1 (or both) every time these two threads come through. This happens in one of two ways:

1. `wake` observes `register`[0]
2. `load` observes `store`.

This means that every time `wake` does not observe the store in `register`, `store` must *happen before* `load`. That requires that a failed `wake` *inter-thread happens before* `register`, since there's no other synchronization here.[1] But you can't normally get a *inter-thread happens before* relation from a read to a write without another way to synchronize. In fact, the only way I know is to guarantee a total order on all reads and writes of that location. The takeaway point is that `AtomicWaker` explicitly declines to guarantee that total order.

Now, with the current implementation the example works because there is in fact a total order. The two ways to get that are 1. critical sections and 2. doing everything entirely inside SeqCst atomics, and the current implementation uses critical sections. But if, for instance, the implementation of `register`/`wake` switched to using an `AtomicPtr` with any memory ordering less than `SeqCst`, then the example would break. It would be possible for `wake` to read null *and* `load` to read `false`, thus causing `thread1_poll` to return `Async::Pending` without anyone being responsible for waking it up (assuming `thread2()` will never be called again). Before reading the implementation, a user has no way of knowing that there was a total order on `register`/`wake`; therefore, they could not have written the above code safely.

If I'm right about a total ordering being the only way this example works, then the documentation should just document that `AtomicWaker` guarantees it (and `set.load` can use `Relaxed`, for that matter). But if not then at least the example should be updated not to rely on the explicitly unguaranteed behavior.

[...]

PR #1079, opened 2018-07-10
Add abortable combinator

cramertj:
Retargeting of #696
cc @carllerche

### DIFF
--- a/futures-util/src/future/abortable.rs
+++ b/futures-util/src/future/abortable.rs
@@ -1,0 +1,169 @@
+use futures_core::future::Future;
+use futures_core::task::{self, Poll};
+use crate::task::AtomicWaker;
+use std::marker::Unpin;
+use std::mem::PinMut;
+use std::sync::Arc;
+use std::sync::atomic::{AtomicBool, Ordering};
+
+/// A future which can be remotely short-circuited using an `AbortHandle`.
+#[derive(Debug, Clone)]
+#[must_use = "futures do nothing unless polled"]
+pub struct Abortable<Fut> {
+    future: Fut,
+    inner: Arc<AbortInner>,
+}
+
+impl<Fut: Unpin> Unpin for Abortable<Fut> {}
+
+impl<Fut> Abortable<Fut> where Fut: Future {
+    unsafe_pinned!(future: Fut);
+
+    /// Creates a new `Abortable` future using an existing `AbortRegistration`.
+    /// `AbortRegistration`s can be acquired through `AbortHandle::new`.
+    ///
+    /// When `abort` is called on the handle tied to `reg` or if `abort` has
+    /// already been called, the future will complete immediately without making
+    /// any further progress.
+    ///
+    /// Example:
+    ///
+    /// ```rust
+    /// # extern crate futures;
+    /// use futures::prelude::*;
+    /// use futures::future::{ready, Abortable, AbortHandle, Aborted};
+    /// use futures::executor::block_on;
+    ///
+    /// # fn main() {
+    /// let (abort_handle, abort_registration) = AbortHandle::new_pair();
+    /// let future = Abortable::new(ready(2), abort_registration);
+    /// abort_handle.abort();
+    /// assert_eq!(block_on(future), Err(Aborted));
+    /// # }
+    /// ```
+    pub fn new(future: Fut, reg: AbortRegistration) -> Self {
+        Abortable {
+            future: future,
+            inner: reg.inner,
+        }
+    }
+}
+
+/// A registration handle for a `Abortable` future.
+/// Values of this type can be acquired from `AbortHandle::new` and are used
+/// in calls to `Abortable::new`.
+#[derive(Debug)]
+pub struct AbortRegistration {
+    inner: Arc<AbortInner>,
+}
+
+/// A handle to a `Abortable` future.
+#[derive(Debug, Clone)]
+pub struct AbortHandle {
+    inner: Arc<AbortInner>,
+}
+
+impl AbortHandle {
+    /// Creates an (`AbortHandle`, `AbortRegistration`) pair which can be used
+    /// to abort a running future.
+    ///
+    /// This function is usually paired with a call to `Abortable::new`.
+    ///
+    /// Example:
+    ///
+    /// ```rust
+    /// # extern crate futures;
+    /// use futures::prelude::*;
+    /// use futures::future::{ready, Abortable, AbortHandle, Aborted};
+    /// use futures::executor::block_on;
+    ///
+    /// # fn main() {
+    /// let (abort_handle, abort_registration) = AbortHandle::new_pair();
+    /// let future = Abortable::new(ready(2), abort_registration);
+    /// abort_handle.abort();
+    /// assert_eq!(block_on(future), Err(Aborted));
+    /// # }
+    pub fn new_pair() -> (Self, AbortRegistration) {
+        let inner = Arc::new(AbortInner {
+            waker: AtomicWaker::new(),
+            cancel: AtomicBool::new(false),
+        });
+
+        (
+            AbortHandle {
+                inner: inner.clone(),
+            },
+            AbortRegistration {
+                inner: inner,
+            },
+        )
+    }
+}
+
+// Inner type storing the waker to awaken and a bool indicating that it
+// should be cancelled.
+#[derive(Debug)]
+struct AbortInner {
+    waker: AtomicWaker,
+    cancel: AtomicBool,
+}
+
+/// Creates a new `Abortable` future and a `AbortHandle` which can be used to stop it.
+///
+/// This function is a convenient (but less flexible) alternative to calling
+/// `AbortHandle::new` and `Abortable::new` manually.
+pub fn abortable<Fut>(future: Fut) -> (Abortable<Fut>, AbortHandle)
+    where Fut: Future
+{
+    let (handle, reg) = AbortHandle::new_pair();
+    (
+        Abortable::new(future, reg),
+        handle,
+    )
+}
+
+/// Indicator that the `Abortable` future was aborted.
+#[derive(Copy, Clone, Debug, Eq, PartialEq)]
+pub struct Aborted;
+
+impl<Fut> Future for Abortable<Fut> where Fut: Future {
+    type Output = Result<Fut::Output, Aborted>;
+
+    fn poll(mut self: PinMut<Self>, cx: &mut task::Context) -> Poll<Self::Output> {
+        // Check if the future has been aborted
+        if self.inner.cancel.load(Ordering::Relaxed) {
+            return Poll::Ready(Err(Aborted))
+        }
+
+        // attempt to complete the future
+        if let Poll::Ready(x) = self.future().poll(cx) {
+            return Poll::Ready(Ok(x))
+        }
+
+        // Register to receive a wakeup if the future is aborted in the... future
+        self.inner.waker.register(cx.waker());
+
+        // Check to see if the future was aborted between the first check and
+        // registration.
+        // Checking with `Relaxed` is sufficient because `register` introduces an
+        // `AcqRel` barrier.
+        if self.inner.cancel.load(Ordering::Relaxed) {
+            return Poll::Ready(Err(Aborted))
+        }
+
+        Poll::Pending
+    }
+}
+
+impl AbortHandle {
+    /// Abort the `Abortable` future associated with this handle.
+    ///
+    /// Notifies the Abortable future associated with this handle that it
+    /// should abort. Note that if the future is currently being polled on
+    /// another thread, it will not immediately stop running. Instead, it will
+    /// continue to run until its poll method returns.
+    pub fn abort(&self) {
+        self.inner.cancel.store(true, Ordering::Relaxed);
+        self.inner.waker.wake();
+    }
+}
--- a/futures/tests/abortable.rs
+++ b/futures/tests/abortable.rs
@@ -1,0 +1,46 @@
+#![feature(pin, arbitrary_self_types, futures_api)]
+
+#[macro_use] extern crate futures;
+
+use futures::FutureExt;
+use futures::channel::oneshot;
+use futures::executor::block_on;
+use futures::future::{abortable, Aborted};
+use futures::task::Poll;
+
+mod support;
+use self::support::with_counter_waker_context;
+
+#[test]
+fn abortable_works() {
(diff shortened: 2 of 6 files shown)


futures-util/src/future/abortable.rs
```diff
@@ -0,0 +1,131 @@
...
+    inner: Arc<AbortInner>,
+}
+
+impl AbortHandle {
+    /// Creates a new `AbortHandle` which can be used to abort a running future.
+    /// This function also returns a `AbortRegistration` which is used to
+    /// pair the handle with a `Abortable` future.
+    ///
+    /// This function is usually paired with a call to `Abortable::new`.
+    pub fn new() -> (Self, AbortRegistration) {
```
MajorBreakfast: Probably should be private

futures-util/src/future/abortable.rs
```diff
@@ -0,0 +1,131 @@
...
+impl<T: Unpin> Unpin for Abortable<T> {}
+
+impl<T> Abortable<T> where T: Future {
+    /// Creates a new `Abortable` future using an existing `AbortRegistration`.
+    /// `AbortRegistration`s can be acquired through `AbortHandle::new`.
+    ///
+    /// When `abort` is called on the handle tied to `reg` or if `abort` has
+    /// already been called, the future will complete immediately without making
+    /// any further progress.
+    pub fn new(future: T, reg: AbortRegistration) -> Self {
```
MajorBreakfast: Probably should be private

futures-util/src/future/abortable.rs
```diff
@@ -0,0 +1,131 @@
...
+struct AbortInner {
+    waker: AtomicWaker,
+    cancel: AtomicBool,
+}
+
+/// Creates a new `Abortable` future and a `AbortHandle` which can be used to stop it.
+///
+/// This function is a convenient (but less flexible) alternative to calling
+/// `AbortHandle::new` and `Abortable::new` manually.
+pub fn abortable<T>(future: T) -> (Abortable<T>, AbortHandle)
```
MajorBreakfast: Probably should be a method on `FutureExt`

The signature is similar to `StreamExt::split`:
```rust
fn split(self) -> (SplitSink<Self>, SplitStream<Self>) // method
```

futures-util/src/future/abortable.rs
```diff
@@ -0,0 +1,131 @@
...
+        Abortable::new(future, reg),
+        handle,
+    )
+}
+
+/// Indicator that the `Abortable` future was aborted.
+#[derive(Copy, Clone, Debug, Eq, PartialEq)]
+pub struct Aborted;
+
+impl<T> Future for Abortable<T> where T: Future {
```
MajorBreakfast: Nit: It'd be better if you either:
- listed first all defs and then all impls
- or kept def and impls for each type together

futures-util/src/future/abortable.rs
```diff
@@ -0,0 +1,131 @@
...
+    /// already been called, the future will complete immediately without making
+    /// any further progress.
+    pub fn new(future: T, reg: AbortRegistration) -> Self {
+        Abortable {
+            future: future,
+            inner: reg.inner,
+        }
+    }
+
+    unsafe_pinned!(future: T);
```
MajorBreakfast: I think we elsewhere always list pinned projections at the top

futures-util/src/future/abortable.rs
```diff
@@ -0,0 +1,131 @@
...
+        if self.inner.cancel.load(Ordering::Relaxed) {
+            return Poll::Ready(Err(Aborted))
+        }
+
+        Poll::Pending
+    }
+}
+
+impl AbortHandle {
+    /// Abort the `Abortable` future associated with this handle.
```
MajorBreakfast: The docs should mention that `abort()` will abort the future ASAP, but does not block until it has actually been done.

futures-util/src/future/abortable.rs
```diff
@@ -0,0 +1,131 @@
...
+        }
+
+        // Register to receive a wakeup if the future is aborted in the... future
+        self.inner.waker.register(cx.waker());
+
+        // Check to see if the future was aborted between the first check and
+        // registration
+        if self.inner.cancel.load(Ordering::Relaxed) {
+            return Poll::Ready(Err(Aborted))
+        }
```
MajorBreakfast: I don't think this will work. I think `Ordering::Relaxed` makes this check unreliable.

My suggestion for `poll`:
```rust
// Register to receive a wakeup if the future is aborted in the... future
self.inner.waker.register(cx.waker());

// Check if the future has been aborted
if self.inner.cancel.load(Ordering::Aquire) {
    return Poll::Ready(Err(Aborted))
}

// Attempt to complete the future
match self.future().poll(cx) {
    Poll::Ready(output) => Poll::Ready(Ok(output)),
    Poll::Pending => Poll::Pending,
}
```

And `abort`:
```
self.inner.cancel.store(true, Ordering::Release);
self.inner.waker.wake();
```

I might be wrong of course, because atomic ops are not exactly my forte

futures-util/src/future/abortable.rs
```diff
@@ -0,0 +1,131 @@
...
+impl AbortHandle {
+    /// Creates a new `AbortHandle` which can be used to abort a running future.
+    /// This function also returns a `AbortRegistration` which is used to
+    /// pair the handle with a `Abortable` future.
+    ///
+    /// This function is usually paired with a call to `Abortable::new`.
+    pub fn new() -> (Self, AbortRegistration) {
+        let inner = Arc::new(AbortInner {
+            waker: AtomicWaker::new(),
+            cancel: AtomicBool::new(false),
```
MajorBreakfast: Maybe call this field `aborted`

futures-util/src/future/abortable.rs
MajorBreakfast: Ah no, you intend to make in to possible to create an `AbortHandle` without creating an `Abortable` future. I'm not against adding this functionality, but it seems a bit over-engineered. Is there an actual use case for this?

*Anywho*, there should probably be a unit test and a quick code example in the docs if this functionality should exist.

futures/tests/abortable.rs
```diff
@@ -0,0 +1,39 @@
...
+    assert_eq!(Err(Aborted), block_on(abortable_rx));
+}
+
+#[test]
+fn abortable_awakens() {
+    let (_tx, a_rx) = oneshot::channel::<()>();
+    let (abortable_rx, abort_handle) = abortable(a_rx);
+
+    thread::spawn(move || {
+        thread::sleep(Duration::from_millis(100));
```
MajorBreakfast: Tests should not use `sleep`. Here is some (untested) code that should do the same:

```rust
let (tx1, rx1) = mpsc::channel::<()>();
let (tx2, rx2) = mpsc::channel::<()>();

let (fut, abort_handle) = future::poll_fn(move |cx| {
    tx1.send(()).unwrap();
    Poll::Pending
}).abortable()

fut.map_err(err => tx2.send(err).unwrap())
    .run_in_background();

assert_eq!((), rx1.recv().unwrap()); // Future inside `Abortable` runs

assert_eq!(Err(mpsc::TryRecvError::Empty), rx2.try_recv()); // Not aborted yet
abort_handle.abort(); // `abort` awakens the `Abortable` future
assert_eq!(Aborted, rx2.recv().unwrap()) // `Abortable` future has completed
```

futures-util/src/future/abortable.rs
```diff
@@ -0,0 +1,131 @@
...
+            return Poll::Ready(Err(Aborted))
+        }
+
+        // attempt to complete the future
+        if let Poll::Ready(x) = self.future().poll(cx) {
+            return Poll::Ready(Ok(x))
+        }
+
+        // Register to receive a wakeup if the future is aborted in the... future
+        self.inner.waker.register(cx.waker());
```
Nemo157: Does `AtomicWaker` support registering multiple wakers across different tasks? (I would assume not otherwise there would need to be some explicit de-registration?)

The fact that `AbortRegistration` is clonable means you can do something like

```rust
    let (_a_tx, a_rx) = oneshot::channel::<()>();
    let (_b_tx, b_rx) = oneshot::channel::<()>();
    let (abort_handle, reg) = AbortHandle::new();
    let a_rx = Abortable::new(a_rx, reg.clone());
    let b_rx = Abortable::new(b_rx, reg);
```

but it seems like that won't wake both when aborted.

MajorBreakfast: The `AtomicWaker::register` docs say:
> This function is safe to call concurrently, but this is generally a bad idea. Concurrent calls to `register` will attempt to register different tasks to be notified. One of the callers will win and have its task set, but there is no guarantee as to which caller will succeed.

Well spotted! I agree the scenario in your code sample won't work as intended. `abort` will always wake just one task of the two tasks, but it should wake both.

futures-util/src/future/abortable.rs
```diff
@@ -0,0 +1,131 @@
+use futures_core::future::Future;
+use futures_core::task::{self, AtomicWaker, Poll};
```
MajorBreakfast:
```
use crate::task::AtomicWaker;
```

futures-util/src/future/abortable.rs
MajorBreakfast: This `new` function doesn't follow the usual `fn new() -> Self` convention

Making it a free function would be a good alternative. (Assuming `AbortRegistration` stays)

futures-util/src/future/abortable.rs
MajorBreakfast: @cramertj I've found your old comment where you explain why you want this. I now fully agree that we want this functionality.

https://github.com/rust-lang-nursery/futures-rs/pull/696#issuecomment-356692685

futures-util/src/future/abortable.rs
MajorBreakfast: Additionally we could also add the method `future.abortable_via(abort_reg)`

futures-util/src/future/abortable.rs
cramertj: Yeah, that changed after I wrote this. Will rebase.

futures-util/src/future/abortable.rs
```diff
@@ -0,0 +1,131 @@
...
+use std::marker::Unpin;
+use std::mem::PinMut;
+use std::sync::Arc;
+use std::sync::atomic::{AtomicBool, Ordering};
+
+/// A future which can be remotely short-circuited using an `AbortHandle`.
+#[derive(Debug, Clone)]
+#[must_use = "futures do nothing unless polled"]
+pub struct Abortable<T> {
+    future: T,
```
MajorBreakfast: Instead of `T`, consider using `Fut` as type param. I'm currently introducing this convention in my PRs. I've found it to be really helpful when reading the code later.
`F` function, `Fut` future, `St` stream, `Si` sink, `T` output, `R` `AsyncRead`, `W` `AsyncWrite`, `E` error, `S` sink + stream

futures-util/src/future/abortable.rs
cramertj: Personal preference but I think this makes more sense as a free function-- you wouldn't chain it like normal combinators since it doesn't return a `Future`.

futures-util/src/future/abortable.rs
cramertj: Aborting it isn't an operation that requires a delay, so I don't think it makes sense to refer to it as blocking/non-blocking.

futures-util/src/future/abortable.rs
cramertj: The call to `register` introduces an `AcqRel` memory barrier.

futures-util/src/future/abortable.rs
cramertj: Done.

futures-util/src/future/abortable.rs
cramertj: Done.

futures-util/src/future/abortable.rs
cramertj: Done.

futures-util/src/future/abortable.rs
cramertj: Renamed to `new_pair`.

futures-util/src/future/abortable.rs
cramertj: Done.

futures/tests/abortable.rs
cramertj: Replaced this with a test using an AtomicBool counter as a waker.

futures-util/src/future/abortable.rs
MajorBreakfast: - The API docs for `AtomicWaker::wake` should mention this
- A comment about this that explains why `Ordering::Relaxed` is enough would be good
- I still think that one check of `self.inner.cancel` per `poll` call should suffice

cramertj: > The API docs for AtomicWaker::wake should mention this

There's [an open issue about documenting this](https://github.com/rust-lang-nursery/futures-rs/issues/942), and it has yet to be decided exactly what guarantees we want to offer.

>  A comment about this that explains why Ordering::Relaxed is enough would be good

Done

> I still think that one check of self.inner.cancel per poll call should suffice

It does not-- either we need to register to receive a wakeup unconditionally (prior to checking the value of `cancel` or polling the future) or we need to check both before and after.

futures-util/src/future/abortable.rs
MajorBreakfast: Here's what I'd like the docs to say:
Notifies the `Abortable` future associated with this handle that it should abort. Note that if the future is currently being polled on another thread, it will not immediately stop running, instead it will continue to run until its `poll` method returns.

cramertj: Done.

MajorBreakfast:
> either we need to register to receive a wakeup unconditionally (prior to checking the value of cancel or polling the future) or we need to check both before and after.

@cramertj The comment is now hidden, so I write it here: It makes sense to register the waker unconditionally. Here's why: The most common code path is that the future returns `Poll::Pending`. The current implementation will register the waker in this case, but it also checks the bool twice. It makes more sense to always register the waker and check the bool only once.

cramertj:
@MajorBreakfast Registering the waker requires several SeqCst accesses and an AcqRel access, while checking the boolean (Release) is basically free.

MajorBreakfast:
Ah I see. Well that reasoning is sound

PR #1080, opened 2018-07-11
Recommend 0.1 in README on master

mbrubeck:
For the benefit of potential users who view the README via the default GitHub view.


cramertj:
Works for me!

PR #1081, opened 2018-07-11
`clone_as_unsafe_wake_without_lifetime` is safe

MajorBreakfast:
It seems that I incorrectly marked this method as `unsafe`. Calling it is not actually unsafe. It just leaks memory. Although memory leaks are undesirable, they are safe in nature.

### DIFF
--- a/futures-util/src/stream/futures_unordered/node.rs
+++ b/futures-util/src/stream/futures_unordered/node.rs
@@ -59,24 +59,29 @@ impl<Fut> Node<Fut> {
         }
     }
 
-    // Saftey: The returned `NonNull<Unsafe>` needs to be put into a `Waker`
-    // or `LocalWaker`
-    unsafe fn clone_as_unsafe_wake_without_lifetime(self: &Arc<Node<Fut>>)
-        -> NonNull<dyn UnsafeWake>
-    {
+    // Clones the `Arc` and casts it to `dyn UnsafeWake`. This method is safe,
+    // but the returned `NonNull<dyn UnsafeWake>` must be placed in a waker,
+    // otherwise there will be a memory leak.
+    fn clone_as_unsafe_wake_without_lifetime(
+        self: &Arc<Node<Fut>>,
+    ) -> NonNull<dyn UnsafeWake> {
         let clone = self.clone();
 
         // Safety: This is save because an `Arc` is a struct which contains
         // a single field that is a pointer.
-        let ptr = mem::transmute::<Arc<Node<Fut>>,
-                                   NonNull<ArcNode<Fut>>>(clone);
-
+        let ptr = unsafe {
+            mem::transmute::<Arc<Node<Fut>>,
+                             NonNull<ArcNode<Fut>>>(clone)
+        };
         let ptr = ptr as NonNull<dyn UnsafeWake>;
 
         // Hide lifetime of `Fut`
-        // Safety: This is safe because `UnsafeWake` is guaranteed not to
+        // Safety: This is safe because `UnsafeWake` is guaranteed to not
         // touch `Fut`
-        mem::transmute::<NonNull<dyn UnsafeWake>, NonNull<dyn UnsafeWake>>(ptr)
+        unsafe {
+            mem::transmute::<NonNull<dyn UnsafeWake>,
+                             NonNull<dyn UnsafeWake>>(ptr)
+        }
     }
 
     pub(super) fn local_waker(self: &Arc<Node<Fut>>) -> LocalWaker {


carllerche:
How do you figure? Losing the lifetime information is very unsafe?

MajorBreakfast:
@carllerche The methods implemented in the `dyn UnsafeWake` impl guarantees to not touch the type `T` for which the lifetime is for. Its `Drop` impl would even panic if by the time it runs the `T` value was still around (just a development safeguard, this cannot happen). Full credit for this algorithm goes to alexcrichton. In my previous PRs in which I adapted this code for 0.3, I inspected it quite thoroughly. If you're interested, there are a lot of comments in the code which explain why everything holds in great detail. In this case the wakers can live als long as they want, there is no danger.

MajorBreakfast:
The lifetime hiding already exists in futures 0.1:
https://github.com/rust-lang-nursery/futures-rs/blob/8965117895dff48de18e70918fd375a041a93877/src/stream/futures_unordered.rs#L596-L598

However, the `hide_lt` method really has to be marked as `unsafe` because it takes a pointer as argument and the user must ensure that the pointer is valid.

`clone_as_unsafe_wake_without_lifetime` on the other hand is not unsafe because there are no extra conditions that must hold to make calling it safe.

carllerche:
I'm well aware of the history and impl details of `FuturesUnordered`. The function is unsafe and callers must take caution. Your argument simply states that `FuturesUnordered` itself is not unsafe.

cramertj:
I guess technically you could argue that this is sort of vacuously safe because it returns a raw pointer (`NonNull`) whose only invariant is that it is nonzero, which this definitely is.

MajorBreakfast:
@carllerche But that's my point: The caller does not have to have to take caution. Aside from a memory leak there is nothing that can go wrong. The unsafety stays completely inside the function. Therefore it should not be marked as unsafe. Only functions that require certain extra conditions to be met when they are called should be marked as `unsafe`. This is not the case here.

MajorBreakfast:
@carllerche The lifetime hiding is of course terribly unsafe. That's why an unsafe block is required inside the function. The unsafe block guarantees the safety and the function as a whole can therefore be considered safe because there are no external requirements to be met by the caller.

carllerche:
In what way is the unsafety completely hidden inside the function? By virtue of hiding the lifetime, you are able to use the type past the duration of the of its existence?

> The unsafe block guarantees the safety of the lifetime hiding.

How does an unsafe block guarantee the safety? I am reading your argument as "the `fn` doesn't need to be `unsafe` because the caller is bug free).

> The caller does not have to take caution

The caller has to ensure that the type is not used past its lifetime.

The only argument that makes sense to me is that `NonZero` is an unsafe pointer, thus you would need to wrap all usage of it in an `unsafe` block. But, afaik unsafe pointers still track type lifetimes (which is why hide_lt is needed), which are lost by the call to this fn.

MajorBreakfast:
> In what way is the unsafety completely hidden inside the function? By virtue of hiding the lifetime, you are able to use the type past the duration of the of its existence?

@carllerche 
- The returned `NonNull<dyn UnsafeWake>` makes it impossible to access the future `T` inside which has the lifetime. There is no way to get to it. The trait has just three methods `clone_raw`, `drop_raw` and `wake`. The implementations for all three methods promise to not access or drop the future `T`. The comment above the lifetime hiding unsafe block explains this: "Hide lifetime of `T` Safety: This is safe because `UnsafeWake` is guaranteed to not touch `T`"
- The retuned `NonNull<dyn UnsafeWake>` may outlive the future because `Node<T>` is meant to outlive the future. It is `FuturesUnordered`'s responsibility to drop the future. `Node<T>::drop` would panic if the future was still around by the time it runs:

https://github.com/rust-lang-nursery/futures-rs/blob/5f33992bd2c8de1bff8206903c0748dcd2536fa2/futures-util/src/stream/futures_unordered/node.rs#L101-L103

> How does an unsafe block guarantee the safety? I am reading your argument as "the fn doesn't need to be unsafe because the caller is bug free).

The caller can't do anything wrong that could cause a bug.

> The only argument that makes sense to me is that NonZero is an unsafe pointer, thus you would need to wrap all usage of it in an unsafe block.

That would be cheating. Also, I wouldn't buy it :)

MajorBreakfast:
@seanmonstar Indeed. https://github.com/rust-lang-nursery/futures-rs/pull/1045 would fix this. We were cautious and didn't merge it right away.

MajorBreakfast:
@seanmonstar I think 0.1 does:
https://github.com/rust-lang-nursery/futures-rs/blob/8965117895dff48de18e70918fd375a041a93877/src/stream/futures_unordered.rs#L361-L364

https://github.com/rust-lang-nursery/futures-rs/blob/8965117895dff48de18e70918fd375a041a93877/src/task_impl/mod.rs#L467-L472

There is a `notify.clone()`

MajorBreakfast:
@seanmonstar It seems that in 0.1 `clone()` is triggered when `current()` is called:
https://github.com/rust-lang-nursery/futures-rs/blob/8965117895dff48de18e70918fd375a041a93877/src/task_impl/mod.rs#L89-L91

Ah well...

MajorBreakfast:
Closing in favor of https://github.com/rust-lang-nursery/futures-rs/pull/1045

PR #1083, opened 2018-07-12
fix poll function in ReadExact

laizy:


cramertj:
@laizy Good catch! Would you mind adding a test for this as well?

cramertj:
Additionally, could you also add a 0.3 backport? No worries if that's a hassle-- I'll do it otherwise.

laizy:
@cramertj ，have added a test case. But I do not know how to  add 0.3 backport.  😅

MajorBreakfast:
@laizy Yeah "0.3 backport" surprised me as well when I was reading this ^^'

PR #1085, opened 2018-07-12
shared: fix unsound Send and Sync impl for Shared

seanmonstar:
The `Future::shared` combinator wrongly allowed both `!Send` futures, and futures with `!Send + !Sync` items or errors, to be sent across threads. Here's an example that compiles with 0.1.21, allowing the sending and cloning of an `Rc` across threads: https://gist.github.com/rust-play/1b25c790a21d34e653a8cbf3e528ae08

### DIFF
--- a/src/future/shared.rs
+++ b/src/future/shared.rs
@@ -235,8 +235,20 @@ impl Notify for Notifier {
     }
 }
 
-unsafe impl<F: Future> Sync for Inner<F> {}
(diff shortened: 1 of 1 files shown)


cramertj:
Thanks so much for this! I'll push a new release right away.

MajorBreakfast:
I just checked: On the 0.3 branch this is correctly implemented

PR #1093, opened 2018-07-15
Allow ownership of the output of Shared to be reclaimed

### LINKED ISSUE #1091: Provide a way to reclaim ownership of the result of `shared`
The [shared](https://docs.rs/futures/0.1.23/futures/future/trait.Future.html#method.shared) combinator mentions that there is currently no way to reclaim ownership of the output returned from the future.

* Remove the `Arc` wrapping that is currently done and require that `Output: Clone`.

This avoids the reclamation issue entirely as there is no "inner value" to reclaim. As a bonus, cheap to clone values such as integers which don't need `Arc` will be faster.

The downside of this is that one needs to manually wrap the output of the inner future with `Arc` for expensive to clone values `future.map(Arc::new).map_err(Arc::new).shared()`. Due to the way the future works, the output will always need to be cloned as well, even if the future only outputs once.

(Assuming this can be implemented this is the approach I would prefer as it is more general).

* Use `Arc` but use/expose [try_unwrap](https://doc.rust-lang.org/std/sync/struct.Arc.html#method.try_unwrap) to reclaim ownership.

This preserves the current API, making it easy to work with expensive to Clone items and avoids the cost of always needing to clone the output once.

There is a subtle problem here however, if one where to use `map/map_err` to reclaim the inner value, one wouldn't actually be able to do that (without a clone) as the `Shared` future will still be alive and have an extra reference to the output, preventing `try_unwrap` from working (this is related to the always clone issue in the previous point).

```rust
fut.shared().map(|t| {
    t.try_unwrap()// Reference count is at least 2
})
```

I believe this can be fixed with a small change in the semantics of calling `poll` (which may or may not be desirable), but if we state that `Shared` will be invalidate if there is only a single owner to the instance at the time that `poll` is called. If that is the case, it should be possible to move the output out of the future and mark the `Shared` instance as already polled. This should be fine if we assume that (just like any other future) one does not call poll again on a future that has returned, however it can cause some confusing errors if someone polls and then clones the `Shared` instance to poll again it later.

This should fix both of the clone issues in either approach but is probably more important for the first approach.

Marwes:
This implements the ideas in #1091. Namely that `Shared` now returns the value as is, without wrapping in `Arc` (first commit) and that the `Shared` instance now gets invalidated if, when it is polled, it is the last instance left which allows `Shared` to avoid cloning the output if only one instance of it remains.

I'd like to document this behavior but I have so far struggled to come up with succinct wording. So instead I figured I'd post a PR to see if it is desirable in the first place 

Closes #1091

### DIFF
--- a/futures-util/src/future/shared.rs
+++ b/futures-util/src/future/shared.rs
@@ -50,7 +50,7 @@ impl<Fut: Future> fmt::Debug for Inner<Fut> {
 
 enum FutureOrOutput<Fut: Future> {
     Future(Fut),
-    Output(Arc<Fut::Output>),
+    Output(Fut::Output),
 }
 
 unsafe impl<Fut> Send for Inner<Fut>
@@ -88,15 +88,17 @@ impl<Fut: Future> Shared<Fut> {
     }
 }
 
-impl<Fut> Shared<Fut> where Fut: Future {
+impl<Fut> Shared<Fut>
+where
+    Fut: Future,
+    Fut::Output: Clone,
+{
     /// If any clone of this `Shared` has completed execution, returns its result immediately
     /// without blocking. Otherwise, returns None without triggering the work represented by
     /// this `Shared`.
-    pub fn peek(&self) -> Option<Arc<Fut::Output>> {
+    pub fn peek(&self) -> Option<Fut::Output> {
         match self.inner.notifier.state.load(SeqCst) {
-            COMPLETE => {
-                Some(unsafe { self.clone_output() })
-            }
+            COMPLETE => Some(unsafe { self.clone_output() }),
             POISONED => panic!("inner future panicked during poll"),
             _ => None,
         }
@@ -133,7 +135,7 @@ impl<Fut> Shared<Fut> where Fut: Future {
 
     /// Safety: callers must first ensure that `self.inner.state`
     /// is `COMPLETE`
-    unsafe fn clone_output(&self) -> Arc<Fut::Output> {
+    unsafe fn clone_output(&self) -> Fut::Output {
         if let FutureOrOutput::Output(item) = &*self.inner.future_or_output.get() {
             item.clone()
         } else {
@@ -142,8 +144,11 @@ impl<Fut> Shared<Fut> where Fut: Future {
     }
 }
 
-impl<Fut: Future> Future for Shared<Fut> {
-    type Output = Arc<Fut::Output>;
+impl<Fut: Future> Future for Shared<Fut>
+where
+    Fut::Output: Clone,
+{
+    type Output = Fut::Output;
 
     fn poll(mut self: PinMut<Self>, cx: &mut task::Context) -> Poll<Self::Output> {
         let this = &mut *self;
@@ -210,7 +215,6 @@ impl<Fut: Future> Future for Shared<Fut> {
                     }
                 }
                 Poll::Ready(output) => {
-                    let output = Arc::new(output);
                     unsafe {
                         *this.inner.future_or_output.get() =
                             FutureOrOutput::Output(output.clone());
@@ -232,7 +236,10 @@ impl<Fut: Future> Future for Shared<Fut> {
     }
 }
 
-impl<Fut> Clone for Shared<Fut> where Fut: Future {
+impl<Fut> Clone for Shared<Fut>
+where
+    Fut: Future,
+{
     fn clone(&self) -> Self {
         Shared {
             inner: self.inner.clone(),
@@ -241,7 +248,10 @@ impl<Fut> Clone for Shared<Fut> where Fut: Future {
     }
 }
 
-impl<Fut> Drop for Shared<Fut> where Fut: Future {
+impl<Fut> Drop for Shared<Fut>
+where
+    Fut: Future,
+{
     fn drop(&mut self) {
         if self.waker_key != NULL_WAKER_KEY {
             if let Ok(mut wakers) = self.inner.notifier.wakers.lock() {
--- a/futures-util/src/future/mod.rs
+++ b/futures-util/src/future/mod.rs
@@ -578,10 +578,6 @@ pub trait FutureExt: Future {
     /// The shared() method provides a method to convert any future into a
     /// cloneable future. It enables a future to be polled by multiple threads.
     ///
-    /// The returned `Shared` future resolves with `Arc<Self::Output>`,
-    /// which implements `Deref` to allow shared access to the underlying
-    /// result. Ownership of the underlying value cannot currently be reclaimed.
-    ///
     /// This method is only available when the `std` feature of this
     /// library is activated, and it is activated by default.
     ///
@@ -596,8 +592,8 @@ pub trait FutureExt: Future {
     /// let shared1 = future.shared();
     /// let shared2 = shared1.clone();
     ///
-    /// assert_eq!(6, *await!(shared1));
-    /// assert_eq!(6, *await!(shared2));
+    /// assert_eq!(6, await!(shared1));
+    /// assert_eq!(6, await!(shared2));
     /// # });
     /// ```
     ///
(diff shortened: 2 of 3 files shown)


futures-util/src/future/shared.rs
```diff
@@ -229,26 +246,32 @@ impl<Fut: Future> Future for Shared<Fut> {
...
-                    this.inner.notifier.state.store(COMPLETE, SeqCst);
-                    let wakers = &mut lock.take().unwrap();
-                    for (_key, opt_waker) in wakers {
-                        if let Some(waker) = opt_waker.take() {
-                            waker.wake();
-                        }
-                    }
-                    return Poll::Ready(output);
-                }
+        if let Some(_) = Arc::get_mut(&mut this.inner) {
```
Marwes: Had to move all this code out here since `this.inner` was borrowed by `Reset`. Does the code after this check also need to be covered by `Reset`?

futures-util/src/future/shared.rs
```diff
@@ -70,7 +70,8 @@ impl<Fut: Future> fmt::Debug for Inner<Fut> {
 
 enum FutureOrOutput<Fut: Future> {
     Future(Fut),
-    Output(Arc<Fut::Output>),
+    Output(Fut::Output),
+    Done,
 }
 
 unsafe impl<Fut> Send for Inner<Fut>
```
MajorBreakfast: @cramertj Do we actually need the `where Fut::Output: Sync`?

futures-util/src/future/shared.rs
```diff
@@ -181,7 +198,7 @@ impl<Fut: Future> Future for Shared<Fut> {
                 return Poll::Pending
             }
             COMPLETE => {
-                return unsafe { Poll::Ready(this.clone_output()) };
+                return unsafe { Poll::Ready(this.take_output()) };
```
MajorBreakfast: That's too drastic: *Every* instance of `Shared` needs produce the same `Output`. `Shared`'s main purpose is enable sharing of an eventual value that hasn't been produced yet.

futures-util/src/future/shared.rs
```diff
@@ -153,17 +154,33 @@ impl<Fut> Shared<Fut> where Fut: Future {
...
+                FutureOrOutput::Done
+            );
+            if let FutureOrOutput::Output(item) = output {
+                return item;
+            }
+        }
+        self.clone_output()
+    }
+
+    unsafe fn clone_output(&self) -> Fut::Output {
```
MajorBreakfast: Marked as unsafe without precise comment why it is unsafe

futures-util/src/future/shared.rs
```diff
@@ -153,17 +154,33 @@ impl<Fut> Shared<Fut> where Fut: Future {
...
+
```
MajorBreakfast: Unnecessary newline

futures-util/src/future/shared.rs
```diff
@@ -108,11 +109,11 @@ impl<Fut: Future> Shared<Fut> {
     }
 }
 
-impl<Fut> Shared<Fut> where Fut: Future {
+impl<Fut> Shared<Fut> where Fut: Future, Fut::Output: Clone {
```
MajorBreakfast: each where clauses should be on a new line + trailing comma after each

futures-util/src/future/shared.rs
```diff
@@ -153,17 +154,33 @@ impl<Fut> Shared<Fut> where Fut: Future {
 
     /// Safety: callers must first ensure that `self.inner.state`
     /// is `COMPLETE`
-    unsafe fn clone_output(&self) -> Arc<Fut::Output> {
-        if let FutureOrOutput::Output(item) = &*self.inner.future_or_output.get() {
-            item.clone()
-        } else {
-            unreachable!()
+    unsafe fn take_output(&mut self) -> Fut::Output {
+        if let Some(inner) = Arc::get_mut(&mut self.inner) {
```
MajorBreakfast: To make this effective `Arc<Inner>` would need to be put into an Option so that after the future returns `Poll::Ready`, it destroys gives up its strong reference to Inner. Otherwise I think we'll rarely hit this branch.

MajorBreakfast:
AFAIK @cramertj is out this week and we need his opinion on this. It will likely take a while before this can be merged. In the meantime, I left some comments

futures-util/src/future/shared.rs
```diff
@@ -153,17 +154,33 @@ impl<Fut> Shared<Fut> where Fut: Future {
...
+            let output = mem::replace(
+                &mut *inner.future_or_output.get(),
```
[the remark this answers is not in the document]
Marwes: Rewrote it to only use safe code.

futures-util/src/future/shared.rs
Marwes: Unsure what you mean about `Option` wrapping but the point about rarely being hit is true. If one makes a 100 clones of a `Shared` instance then this branch won't get hit in 99 of them.

In that case this may be a pessimisation as the branch could be more expensive than cloning the value one extra time (101 times instead of 100).

Marwes:
I am in no rush to see this merged (cloning the inner value works just fine for me, even on 0.1).

futures-util/src/future/shared.rs
MajorBreakfast: `get_mut` only returns `Some(_)` if the `Arc` is the last one that points to the value. What I mean is that it will not hit it in 100 of 100 cases if the other 99 `Shared` instances holding references to `Inner` weren't dropped before the 100th is polled.

futures-util/src/future/shared.rs
```diff
@@ -153,17 +158,35 @@ impl<Fut> Shared<Fut> where Fut: Future {
 
     /// Safety: callers must first ensure that `self.inner.state`
     /// is `COMPLETE`
-    unsafe fn clone_output(&self) -> Arc<Fut::Output> {
-        if let FutureOrOutput::Output(item) = &*self.inner.future_or_output.get() {
-            item.clone()
-        } else {
-            unreachable!()
+    unsafe fn try_take_output(&mut self) -> Fut::Output {
```
MajorBreakfast: `take_or_clone_output()` would be better. 

(The convention is that functions prefixed with `try_` should return something the implements the `Try` trait - often `Result`)

futures-util/src/future/shared.rs
MajorBreakfast: I suppose I didn't edit my comment above quickly enough ^^'

futures-util/src/future/shared.rs
seanmonstar: It is required if the output is going to accessed from an `Arc` (since that is `Arc<T: Send + Sync>`), but if `Arc` is removed from this type, then thread safety of the `Clone` is then left up to the user.

MajorBreakfast: `Arc` only implements both `Send` and `Sync` only for `T: Send + Sync`. So, how much can we reduce these bounds? Maybe `Fut::Output: Send` for `Send`, but do we require something for `Sync`? My thinking is that we don't actually have shared references, so no restrictions for `Fut::Ouptut` are required to make it `Sync`.

MajorBreakfast:
> To clarify: the idea here is that, in order to get the current behavior when using this type, you'd write something like my_future.map(Arc::new).shared() rather than my_future.shared()?

@cramertj Yes, That's correct. I just noticed one caveat though: The new combinator will be slower for storing an `Arc` because it puts the `Arc` inside its `Arc`. So, we have two `Arc`s.

Marwes:
> The new combinator will be slower for storing an Arc because it puts the Arc inside its Arc. So, we have two Arcs

That is the case before this PR as well?

MajorBreakfast:
@Marwes You're right

MajorBreakfast:
I think `peek` panics with `"Shared future polled again after completion"` if it is called after `Arc::get_mut`  in `try_take_output` was called.

futures-util/src/future/shared.rs
cramertj: Yes-- can you move it all into a block containing the `Reset`?

Marwes:
> I think peek panics with "Shared future polled again after completion" if it is called after Arc::get_mut in try_take_output was called.

Yeah, that falls under the same caveat about polling the same instance of `Shared` again after it has been resolved. So we need to decide whether to accept https://github.com/rust-lang-nursery/futures-rs/pull/1093/commits/0041116f4d3633a48c6af6706d11ac9bee62dbf6 or not.

Marwes:
Rebased and removed the invalidation behaviour since I think the `Arc` removal can be added on its own and should be less controversial.

cramertj:
Thanks @marwes, this looks great! Would you mind opening a follow-up PR for the invalidation? I still think it's a good idea, but I want to talk through how it affects things like 'peek'.

Issue #1090, opened 2018-07-14
Associated types of TryFuture and TryStream


MajorBreakfast:
Currently we have:
```rust
pub trait Future { type Output; ... }
pub trait TryFuture { type Item; type Error; ... }
```

```rust
pub trait Stream { type Item; ... }
pub trait TryStream { type TryItem; type TryError; ... }
```

- `TryStream`'s success associated type cannot be called `Item` because `Stream` already uses that name for its associated type. That's why we end up with the not so great names `TryItem` and `TryError`
- `TryFuture` and `TryStream` have inconsistently named associated types

Here's what I propose:
```rust
pub trait Future { type Output; ... }
pub trait TryFuture { type Ok; type Err; ... }
```

```rust
pub trait Stream { type Item; ... }
pub trait TryStream { type Ok; type Err; ... }
```

That way the naming would be consistent with the variants of the `Result` enum!

```rust
Future<Output = Result<T, E>>
TryFuture<Ok = T, Err = E>
```

```rust
Stream<Item = Result<T, E>>
TryStream<Ok = T, Err = E>
```

MajorBreakfast:
Maybe `Error` instead of `Err` to align with the [`Try` trait](https://doc.rust-lang.org/std/ops/trait.Try.html)

MajorBreakfast:
Implemented by #1094

PR #1094, opened 2018-07-16
Associated types of TryStream and TryFuture

MajorBreakfast:
Implements https://github.com/rust-lang-nursery/futures-rs/issues/1090

To minimize churn I recommend merging this before the release of the alpha. I think users will appreciate the similarity to the `Result` type and `Try` trait. Consistency with the standard library is great.

### DIFF
--- a/futures-core/src/future.rs
+++ b/futures-core/src/future.rs
@@ -9,7 +9,7 @@ pub use core::future::{Future, FutureObj, LocalFutureObj, UnsafeFutureObj};
 /// a variety of adapters tailored to such futures.
 pub trait TryFuture {
     /// The type of successful values yielded by this future
-    type Item;
+    type Ok;
 
     /// The type of failures yielded by this future
     type Error;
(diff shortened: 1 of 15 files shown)


seanmonstar:
I still don't love the `Try_` traits, but in the context of this change, it does seem like `Ok` and `Error` is better than `TryItem` and `TryError`.

MajorBreakfast:
> it does seem like Ok and Error is better than TryItem and TryError

@seanmonstar 👍

> I still don't love the Try_ traits

@seanmonstar The main alternative names are `ResultFuture` and `ResultStream`. I like both options equally. `Result` makes sense because that's what the output type is and `Try` makes sense because we're talking about eventual values which we "try" to produce.

seanmonstar:
Oh, if they have to be a separate trait, `TryFuture` is far better than `ResultFuture`. I'm just bemoaning the general situation of moving `Error` out of `Future`; I realize that ship has sailed for 0.3.

PR #1095, opened 2018-07-17
Distribute license files with all subcrates

MajorBreakfast:


Nemo157:
The symlinks should be to `../LICENSE-MIT` rather than `LICENSE-MIT`

MajorBreakfast:
@Nemo157 I know ^^' I'm not good at creating symlinks. Now I managed it, though :P (I used your PR https://github.com/rust-lang-nursery/futures-rs/pull/1026 as template, so I knew how it *should* look like)

BTW I don't like that this was resolved in tokio by not dual-licensing anymore. Seems very un-Rust-ish...

seanmonstar:
FWIW, the legal implications of dual licensing are unclear. For instance, the protection that apache gives the repo over shared patents: it seems anyone could just claim to license with MIT, and then sue you.

(Note, I didn't make the decision) It seems very fair to prioritize legal clarity over "feeling Rustish".

withoutboats:
The dual licensing scheme is intended to protect against patent claims from *contributors*, who are providing their contributions under both licenses, not against *users*, who can choose (because of the requirement on contributors) to use the code under either license.

Nemo157:
It's also of note that tokio went ahead and included the MIT license text in its subcrates in the end anyway.

seanmonstar:
Yes, tokio stuck to a single license, MIT, instead of dual. The reasoning is that apache alone is too strict, so may as well clear things up and offer only the "weaker" one.

I'm not going to otherwise argue why dual licensing may be really cool. There's no point arguing that it offers more protection, as it's not proven in any court case. My point is that it's fair for an author to realize it's confusing, and prioritize clarity instead of just following example without asking questions.

seanmonstar:
Oh, I wasn't trying to say there shouldn't be a dual license here. I was simply saying that calling out tokio here seemed out of place.

MajorBreakfast:
@Nemo157 Yeah, but only MIT, not Apache 2.0. So, it's now, like mio, MIT only.

@seanmonstar Every other crate uses dual licensing. Also calling out tokio is fine because it is closely related to the futures crate and the decision to only license under the MIT license is very recent (March 2018)

Nemo157:
Luckily including MIT licensed code into a project imposes strictly less restrictions on the project owner, so if it's ok for a project to depend on an Apache-2.0 project it's guaranteed they can also depend on MIT. (And by dual licensing as `MIT OR Apache-2.0` the inverse also works in this case as users of `tokio` that for some reason can't have an Apache-2.0 licensed dependency can choose to use `futures` under MIT).

seanmonstar:
Every other crate does **not** use that scheme. Official rust-lang crates do, and many others do, but not all. Why is it useful to becry tokio for picking a clearer license option? Does it inhibit you or others?

MajorBreakfast:
> Does it inhibit you or others

@seanmonstar Maybe? See @withoutboats's comment. To me it feels like a step backwards. That's my personal opinion.

Also, the real reason I mentioned tokio is because I browsed its codebase, saw the "Relicense Tokio exclusively under the MIT license." commit and wondered how licensing is handled in the futures crate. Then, I opened this PR because I saw that @Nemo157's PR only applied the licenses to the current master branch. And here we are

withoutboats:
I've marked the discussion of tokio's licensing off topic. Let's not continue that conversation on this PR. The tokio project is free to choose how to license their code and unless they are not complying with the license of the futures project (and they are complying), its not on topic in this repository.

MajorBreakfast:
I agree with you both. Nevertheless I mildly dislike the use of moderation tools. It would have been courteous to ask first.

Issue #1096, opened 2018-07-18
0.2 fails to compile with latest nightly


sdroege:
See e.g. `cargo +nightly test --features nightly` with latest git master (and the 0.2.2/0.2.3 versions on crates.io). This is `rustc 1.29.0-nightly (4f3c7a472 2018-07-17)`

```
error[E0432]: unresolved import `futures::prelude::await`
 --> futures/tests/async_await/pinned.rs:4:5
  |
4 | use futures::prelude::await;
  |     ^^^^^^^^^^^^^^^^^^^^^^^ no `await` in `prelude`

[...]

error[E0658]: attribute procedural macros are experimental (see issue #38356)
 --> futures/tests/async_await/elisions.rs:6:1
  |
2 | use futures::prelude::*;
  |     ------------------- procedural macro imported here
...
6 | #[async]
  | ^^^^^^^^
  |
  = help: add #![feature(proc_macro)] to the crate attributes to enable

[...]

error[E0659]: `await` is ambiguous
  --> futures/tests/async_await/smoke.rs:71:5
   |
71 |     await!(foo())
   |     ^^^^^
   |
note: `await` could refer to the name imported here
  --> futures/tests/async_await/smoke.rs:14:5
   |
14 | use futures::prelude::*;
   |     ^^^^^^^^^^^^^^^^^^^
   = note: `await` is also a builtin macro
   = note: consider adding an explicit import of `await` to disambiguate
```

Nemo157:
```
error[E0432]: unresolved import `futures::prelude::await`
 --> futures/tests/async_await/pinned.rs:4:5
  |
4 | use futures::prelude::await;
  |     ^^^^^^^^^^^^^^^^^^^^^^^ no `await` in `prelude`
```

This requires `#![feature(use_extern_macros)]`, that seems to have been split off `proc_macro` since `use_extern_macros` is stable in Edition 2018 while `proc_macro` is still unstable. 

```
error[E0658]: attribute procedural macros are experimental (see issue #38356)
 --> futures/tests/async_await/elisions.rs:6:1
  |
2 | use futures::prelude::*;
  |     ------------------- procedural macro imported here
...
6 | #[async]
  | ^^^^^^^^
  |
  = help: add #![feature(proc_macro)] to the crate attributes to enable
```

Probably just needs `feature(proc_macro)` in the test file, I guess the fact that it's a procedural macro is no longer hidden behind the re-export.

```
[...]

PR #1098, opened 2018-07-19
Fix test build and doctests with latest nightly

sdroege:
Fixes https://github.com/rust-lang-nursery/futures-rs/issues/1096

----

This does not fix `futures/testcrate` as that one seems to have quite a few more problems

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,9 +27,8 @@ matrix:
         - cargo test --manifest-path futures/Cargo.toml --features nightly --test async_await_tests
     - rust: nightly
       script:
-        - rustup component add rust-src
-        - cargo install xargo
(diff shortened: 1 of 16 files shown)


sdroege:
It would probably make sense to add the nightly tests to travis. I can add that if you agree

MajorBreakfast:
Feel free to add the nightly tests to the CI. I agree that having them in there is a good idea.

The CI fails because of some changes to `cfg(target_has_atomic)`. Fixing this is super easy, though. More about it here: https://github.com/rust-lang-nursery/futures-rs/pull/1089

sdroege:
Ok, this is all running now. The only thing left is the test-crate, that seems to need some more attention

MajorBreakfast:
And it's green. Wonderful! Thanks :)

PR #1102, opened 2018-07-19
Fix crate name in import

Nemo157:
If you don't override the lib name then Cargo will automatically convert the `-` in the package name to an `_` for the lib.


MajorBreakfast:
The `[lib.name]` continues to be "futures". So, this might change again :)

PR #1108, opened 2018-07-21
futures-util/io examples

Nemo157:
Also fixes `WriteAll`, it was broken in the same way as `ReadExact` in #1083

PR #1109, opened 2018-07-21
Implement AsyncWrite for the generic Cursor<T: AsMut<[u8]>>

Nemo157:
This introduces an unfortunate point of difference between `futures::io::AsyncWrite` and `std::io::Write`, but I think the increased ergonomics around writing to statically sized in memory buffers (presumably just for test purposes) is useful.

`impl<T: AsRef<[u8]>> Read for Cursor<T>` was added in rust-lang/rust#27197, I'm not sure why `impl<T: AsMut<[u8]>> Write for Cursor<T>` wasn't added at the same time; I would propose doing this change in `std` and just piggybacking off it here, but the breakage is almost certainly not worth it by this point.

(This PR is based off #1108 in order to have some "real" code changes to evaluate the ergonomics on)

### DIFF
--- a/futures-io/src/lib.rs
+++ b/futures-io/src/lib.rs
@@ -23,9 +23,9 @@ macro_rules! if_std {
 if_std! {
     use futures_core::task::{self, Poll};
     use std::boxed::Box;
+    use std::cmp;
     use std::io as StdIo;
     use std::ptr;
-    use std::vec::Vec;
 
     // Re-export IoVec for convenience
     pub use iovec::IoVec;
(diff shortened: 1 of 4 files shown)


futures-io/src/lib.rs
```diff
@@ -354,16 +354,29 @@ if_std! {
         }
     }
 
-    impl<'a> AsyncWrite for StdIo::Cursor<&'a mut [u8]> {
-        delegate_async_write_to_stdio!();
-    }
+    impl<T: AsMut<[u8]>> AsyncWrite for StdIo::Cursor<T> {
+        fn poll_write(&mut self, _: &mut task::Context, buf: &[u8])
+            -> Poll<Result<usize>>
```
MajorBreakfast: Style :)

https://github.com/rust-lang-nursery/fmt-rfcs/blob/master/guide/items.md#function-definitions

MajorBreakfast: I see that the whole file is full of this unusual style. Ignore this then. At some point I'll just apply rustfmt to the whole project and fix such things

Nemo157: `rustfmt` just removes some blank lines from this file, what do you see wrong here?

Nemo157: Ah, the actual RFC doesn't allow for just splitting the return onto its own line when the args fit. I'd prefer to go with what `rustfmt` does for ease of application.

MajorBreakfast:
```rust
fn poll_write(&mut self, _: &mut task::Context, buf: &[u8])
    -> Poll<Result<usize>>
{
```

should be
```rust
fn poll_write(
    &mut self,
    _: &mut task::Context,
    buf: &[u8],
) -> Poll<Result<usize>> {
```
according to the style guide.

Nemo157: Oh, the `if_std!`macro call messes `rustfmt` up since it doesn't know that the internals should be normal Rust code.

Nemo157: I'll fix it up anyway and rebase once #1108 is merged so this will have a decent diff.

MajorBreakfast:
I would like that the others look at this as well. I'm not familiar enough with the i/o stuff.

cramertj:
This looks fine to me!

PR #1111, opened 2018-07-21
Only deny warnings in CI to improve dev experience

Nemo157:
Having the build break because of missing docs when developing locally can get really annoying. I'm also trying to fixup clippy noted issues which trigger the current `deny(warnings)`. Instead it seems better to allow developers to temporarily have warnings while developing, but deny them in CI so that PRs are gated on the code being warning free.

I also tidied the CI config a little bit, a bit of extra whitespace around the jobs, split the benchmarks up to multiple separate commands and used a multiline string for the `thumbv6m` build (yaml will merge multiline indented strings into a single line).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,25 +5,49 @@ sudo: false
 
 matrix:
   include:
-    - os: osx
-    - os: linux
+    - name: cargo test
+      os: osx
+
+    - name: cargo test
+      os: linux
+
+    - name: cargo rustc -- -D warnings
+      rust: nightly
+      script:
+        - cargo rustc --manifest-path futures/Cargo.toml -- -Dwarnings
+        - cargo rustc --manifest-path futures-core/Cargo.toml -- -Dwarnings
+        - cargo rustc --manifest-path futures-channel/Cargo.toml -- -Dwarnings
+        - cargo rustc --manifest-path futures-executor/Cargo.toml -- -Dwarnings
+        - cargo rustc --manifest-path futures-io/Cargo.toml -- -Dwarnings
+        - cargo rustc --manifest-path futures-sink/Cargo.toml -- -Dwarnings
+        - cargo rustc --manifest-path futures-util/Cargo.toml -- -Dwarnings
+
     - name: cargo bench
       rust: nightly
-      script: cargo bench --all && cd futures-util && cargo bench --features=bench
+      script:
+        - cargo bench --all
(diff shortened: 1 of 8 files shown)


.travis.yml
```diff
@@ -5,25 +5,49 @@ sudo: false
...
+    - name: cargo rustc -- -D warnings
+      rust: nightly
+      script:
+        - cargo rustc --manifest-path futures/Cargo.toml -- -Dwarnings
+        - cargo rustc --manifest-path futures-core/Cargo.toml -- -Dwarnings
+        - cargo rustc --manifest-path futures-channel/Cargo.toml -- -Dwarnings
+        - cargo rustc --manifest-path futures-executor/Cargo.toml -- -Dwarnings
+        - cargo rustc --manifest-path futures-io/Cargo.toml -- -Dwarnings
+        - cargo rustc --manifest-path futures-sink/Cargo.toml -- -Dwarnings
+        - cargo rustc --manifest-path futures-util/Cargo.toml -- -Dwarnings
```
MajorBreakfast: I really like the idea! I did some digging and I've discovered [this great post](https://users.rust-lang.org/t/persistent-warnings-from-cargo-build/15693/2?u=majorbreakfast) by cuviper with an even better approach to do it 😁:

`RUSTFLAGS=-Dwarnings cargo test`

Benefit: No extra CI job and its a super clean oneliner 😋

Nemo157: I also spent a while trying out tools like [`cargo-workspace`](https://crates.io/crates/cargo-workspace) or ([`rq`](https://crates.io/crates/record-query) + a small amount of inline bash) to avoid having to list out all the crates. These both had issues, so I didn't think using them was worth it over the mass list of crates.

MajorBreakfast: Okay I'm sold on the extra CI job. How about: `RUSTFLAGS=-Dwarnings cargo check`? I don't like listing all the individual packages. Breaking at the first error is fine because the warnings are all listed in the other jobs.

***

Using clippy has some problems because there are some suggestions that we can't implement, e.g.:

Calls `take_f` even if it shouldn't (the fn has a side effect):
```
  --> futures-util/src/sink/map_err.rs:64:44
   |
64 |         self.sink().poll_ready(cx).map_err(|e| self.take_f()(e))
```

And the suggestion in the join macro can't be implemented:
```
warning: `if _ { .. } else { .. }` is an expression
  --> futures-util/src/future/join.rs:54:17
```

PR #1112, opened 2018-07-21
100% Clippy

Nemo157:
I think merging all these changes is only worth it if we start gating CI on Clippy, otherwise the buggy allows should be dropped (just means dropping the last 2 commits). ~~With this and #1111 the deny warnings job could be changed to `cargo clippy -- -D warnings` and hit all builtin + Clippy warnings on all the crates at once.~~ (EDIT: Done since #1111 is merged)

I think the `needless_lifetimes` one could be fixed in Clippy relatively easily, presumably it just doesn't support `arbitrary_self_types` properly. The `redundant_closures` one seems like it won't be able to get a fix soon, it would need Clippy to somehow detect side-effect producing code.

I'm opening this mostly for discussion on whether gating CI on Clippy is of interest to the team. If it is then I might take a look at doing the upstream fixes to reduce the bad parts of this.

### DIFF
--- a/futures-executor/src/local_pool.rs
+++ b/futures-executor/src/local_pool.rs
@@ -160,9 +160,9 @@ impl LocalPool {
                 let mut main_cx = Context::new(local_waker, exec);
 
                 // if our main task is done, so are we
-                match future.reborrow().poll(&mut main_cx) {
-                    Poll::Ready(output) => return Poll::Ready(output),
-                    _ => {}
+                let result = future.reborrow().poll(&mut main_cx);
+                if let Poll::Ready(output) = result {
+                    return Poll::Ready(output);
                 }
             }
 
@@ -202,6 +202,12 @@ impl LocalPool {
                 _ => {}
             }
         }
+    }
+}
+
+impl Default for LocalPool {
+    fn default() -> Self {
+        Self::new()
     }
 }
 
--- a/futures-util/src/io/mod.rs
+++ b/futures-util/src/io/mod.rs
@@ -250,12 +250,12 @@ pub trait AsyncWriteExt: AsyncWrite {
     /// assert_eq!(output, [1, 2, 3, 4, 0]);
     /// # Ok::<(), Box<std::error::Error>>(()) }).unwrap();
     /// ```
-    fn flush<'a>(&'a mut self) -> Flush<'a, Self> {
+    fn flush(&mut self) -> Flush<'_, Self> {
         Flush::new(self)
     }
 
     /// Creates a future which will entirely close this `AsyncWrite`.
-    fn close<'a>(&'a mut self) -> Close<'a, Self> {
+    fn close(&mut self) -> Close<'_, Self> {
         Close::new(self)
     }
 
@@ -281,7 +281,7 @@ pub trait AsyncWriteExt: AsyncWrite {
     /// assert_eq!(writer.into_inner(), [1, 2, 3, 4, 0]);
     /// # Ok::<(), Box<std::error::Error>>(()) }).unwrap();
     /// ```
-    fn write_all<'a>(&'a mut self, buf: &'a [u8]) -> WriteAll<'a, Self> {
+    fn write_all(&'a mut self, buf: &'a [u8]) -> WriteAll<'a, Self> {
         WriteAll::new(self, buf)
     }
 }
--- a/futures-util/src/sink/mod.rs
+++ b/futures-util/src/sink/mod.rs
@@ -161,7 +161,7 @@ pub trait SinkExt: Sink {
     }
 
     /// Close the sink.
-    fn close<'a>(&'a mut self) -> Close<'a, Self>
+    fn close(&mut self) -> Close<'_, Self>
         where Self: Unpin,
     {
         Close::new(self)
@@ -183,7 +183,7 @@ pub trait SinkExt: Sink {
     ///
     /// This adapter is intended to be used when you want to stop sending to the sink
     /// until all current requests are processed.
-    fn flush<'a>(&'a mut self) -> Flush<'a, Self>
+    fn flush(&mut self) -> Flush<'_, Self>
         where Self: Unpin,
     {
         Flush::new(self)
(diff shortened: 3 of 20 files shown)


futures-executor/src/local_pool.rs
```diff
@@ -205,6 +205,12 @@ impl LocalPool {
     }
 }
 
+impl Default for LocalPool {
```
Nemo157: I could see arguments for not implementing `Default` for `LocalPool`, if so I think having an explicit `allow` for it is good documentation that it has been considered and decided against.

MajorBreakfast: I can't think of any such arguments ^^' Could you list some?

futures-util/src/sink/with.rs
```diff
@@ -58,7 +58,8 @@ enum State<Fut, T> {
 }
 
 impl<Fut, T> State<Fut, T> {
-    fn as_pin_mut<'a>(
+    #[allow(needless_lifetimes, wrong_self_convention)] // https://github.com/rust-lang/rfcs/pull/2362#issuecomment-406804183
+    fn as_pin_mut(
         self: PinMut<'a, Self>
```
MajorBreakfast: missing trailing comma

MajorBreakfast:
I'm in favor of merging this. We can always remove it again from the CI if it proves to be a hassle. However, I'm optimistic that we'll benefit from having it in there.

futures-util/src/io/mod.rs
```diff
@@ -288,7 +288,7 @@ pub trait AsyncWriteExt: AsyncWrite {
     /// assert_eq!(output, [1, 2, 3, 4, 0]);
     /// # Ok::<(), Box<std::error::Error>>(()) }).unwrap();
     /// ```
-    fn write_all<'a>(&'a mut self, buf: &'a [u8]) -> WriteAll<'a, Self> {
+    fn write_all(&'a mut self, buf: &'a [u8]) -> WriteAll<'a, Self> {
```
cramertj: Was it intentional to use in-band-lifetimes here?

futures-util/src/sink/mod.rs
```diff
@@ -212,7 +212,7 @@ pub trait SinkExt: Sink {
     /// Doing `sink.send_all(stream)` is roughly equivalent to
     /// `stream.forward(sink)`. The returned future will exhaust all items from
     /// `stream` and send them to `self`.
-    fn send_all<'a, St>(
+    fn send_all<St>(
```
cramertj: Ditto WRT in-band lifetimes

futures-util/src/sink/with.rs
```diff
@@ -58,7 +58,8 @@ enum State<Fut, T> {
 }
 
 impl<Fut, T> State<Fut, T> {
-    fn as_pin_mut<'a>(
+    #[allow(needless_lifetimes, wrong_self_convention)] // https://github.com/rust-lang/rfcs/pull/2362#issuecomment-406804183
+    fn as_pin_mut(
```
cramertj: And again

futures-util/src/sink/with_flat_map.rs
```diff
@@ -60,7 +60,8 @@ where
     }
 
     /// Get a pinned mutable reference to the inner sink.
-    pub fn get_pin_mut<'a>(self: PinMut<'a, Self>) -> PinMut<'a, Si> {
+    #[allow(needless_lifetimes)] // https://github.com/rust-lang/rfcs/pull/2362#issuecomment-406804183
+    pub fn get_pin_mut(self: PinMut<'a, Self>) -> PinMut<'a, Si> {
```
cramertj: and again

futures-util/src/stream/buffer_unordered.rs
```diff
@@ -84,7 +84,8 @@ where
     ///
     /// Note that care must be taken to avoid tampering with the state of the
     /// stream which may otherwise confuse this combinator.
-    pub fn get_pin_mut<'a>(self: PinMut<'a, Self>) -> PinMut<'a, St> {
+    #[allow(needless_lifetimes)] // https://github.com/rust-lang/rfcs/pull/2362#issuecomment-406804183
+    pub fn get_pin_mut(self: PinMut<'a, Self>) -> PinMut<'a, St> {
```
cramertj: etc...

futures-util/src/io/mod.rs
Nemo157: Yep, as far as I’m aware they’re still on track to be stable in 2018 Edition. If you’ve got some reason not to use them I can revert these changes.

futures-executor/src/local_pool.rs
Nemo157: Mostly that a `LocalPool` isn’t a data structure, so being able to create a default in a generic context isn’t that useful. IMO it’s still worth having the impl anyway though.

futures-util/src/io/mod.rs
MajorBreakfast: @cramertj My local waker opt PR you just merged also [had an in-band lifetime](https://github.com/rust-lang-nursery/futures-rs/blob/476fc37b05c9387d8d8604e03f3a11aea1ec0f3f/futures-util/src/stream/futures_unordered/node.rs#L101) in it 😇

MajorBreakfast:
@Nemo157 It seems that my waker reference optimization PR triggered a clippy warning that broke the CI

Nemo157:
Oh, interesting, I didn't realise that Travis tested the merge commit. I was confused how those changes made it into this build. Rebased and fixed the new  warning.

MajorBreakfast:
@Nemo157 I didn't know that as well. Interesting!

cramertj:
@MajorBreakfast Yup, seems fine to me! I just wanted to check that it was an intentional decision.

MajorBreakfast:
I merge this with the intention to remove the `#[allow(needless_lifetimes)` annotations once `'_` lifetimes are supported in arbitrary self types. In the meantime these annotations cane be happily in there 🙂

PR #1119, opened 2018-07-25
`Compat` implementation

tinaun:
based on @MajorBreakfast's approach here https://github.com/rust-lang-nursery/net-wg/issues/26#issuecomment-407577152

issues: 
* executors are blocked on https://github.com/rust-lang/rust/pull/52674
* more tests / examples
* ~where should 0.1 -> 0.3 combinators live?~
* renaming blocked on https://github.com/rust-lang/cargo/pull/5794
* name of NeverError type

### DIFF
--- a/futures-util/Cargo.toml
+++ b/futures-util/Cargo.toml
@@ -19,6 +19,7 @@ name = "futures_util"
 [features]
 std = ["futures-core-preview/std", "futures-io-preview/std", "futures-sink-preview/std", "either/use_std", "slab"]
 default = ["std", "futures-core-preview/either", "futures-sink-preview/either"]
+compat = ["std", "futures"]
 bench = []
 nightly = []
 
@@ -29,6 +30,7 @@ futures-io-preview = { path = "../futures-io", version = "0.3.0-alpha.2", defaul
 futures-sink-preview = { path = "../futures-sink", version = "0.3.0-alpha.2", default-features = false}
 either = { version = "1.4", default-features = false }
 slab = { version = "0.4", optional = true }
+futures = { version = "0.1", optional = true }
 
 [dev-dependencies]
 futures-preview = { path = "../futures", version = "0.3.0-alpha.2" }
--- a/futures-util/src/compat/executor.rs
+++ b/futures-util/src/compat/executor.rs
@@ -1,0 +1,47 @@
+
+pub trait ExecCompat: Executor01<
+        Compat<FutureObj<'static, ()>, BoxedExecutor>
+    > + Clone + Send + 'static
+{
+    fn compat(self) -> ExecutorCompat<Self> 
+        where Self: Sized;
+}
+
+impl<E> ExecCompat for E
+where E: Executor01<
+        Compat<FutureObj<'static, ()>, BoxedExecutor>
+      >,
+      E: Clone + Send + 'static
+{
+    fn compat(self) -> ExecutorCompat<Self> {
+        ExecutorCompat {
+            exec: self,
+        }
+    }
+}
+
+#[derive(Clone)]
+pub struct ExecutorCompat<E> {
+    exec: E
+}
+
+impl<E> Executor03 for ExecutorCompat<E> 
+    where E: Executor01<
+        Compat<FutureObj<'static, ()>, Box<Executor03>>
+    >,
+    E: Clone + Send + 'static,
+{
+    fn spawn_obj(&mut self, obj: FutureObj<'static, ()>) -> Result<(), task::SpawnObjError> {
+        
+        self.exec.execute(obj.compat(Box::new(self.clone())))
+                 .map_err(|exec_err| {
+                     use futures_core::task::{SpawnObjError, SpawnErrorKind};
+                     
+                     let fut = exec_err.into_future().compat().map(|_| ());
+                     SpawnObjError {
+                         kind: SpawnErrorKind::shutdown(),
+                         task: Box::new(fut).into(),   
+                     }
+                 })
+    }
+}
--- a/futures-util/src/compat/executor.rs
+++ b/futures-util/src/compat/executor.rs
@@ -23,7 +23,7 @@ pub type ExecutorFuture01 = Compat<NeverError<FutureObj<'static, ()>>, BoxedExec
 pub trait Executor01CompatExt: Executor01<ExecutorFuture01> 
     + Clone + Send + 'static
 {
-    /// Creates an exector compatable with futures 0.3.
+    /// Creates an `Executor` compatable with futures 0.3.
     fn compat(self) -> CompatExecutor<Self> 
         where Self: Sized;
 }
@@ -39,7 +39,7 @@ where E: Executor01<ExecutorFuture01>,
     }
 }
 
-/// Converts `futures 0.1` Executors into `futures 0.3` Executors
+/// Converts a futures 0.1 `Executor` into a futures 0.3 `Executor`.
 #[derive(Clone)]
 pub struct CompatExecutor<E> {
     exec: E
--- a/futures-util/src/compat/executor.rs
+++ b/futures-util/src/compat/executor.rs
@@ -1,47 +1,70 @@
 
-pub trait ExecCompat: Executor01<
-        Compat<FutureObj<'static, ()>, BoxedExecutor>
-    > + Clone + Send + 'static
+use futures::future::Executor as Executor01;
+
+use futures_core::task::Executor as Executor03;
+use futures_core::task as task03;
+use futures_core::future::FutureObj;
+
+use super::Compat;
+use crate::{TryFutureExt, FutureExt, future::NeverError};
+
+pub struct BoxedExecutor(Box<dyn Executor03 + Send>);
+
+impl Executor03 for BoxedExecutor {
+    fn spawn_obj(&mut self, future: FutureObj<'static, ()>) -> Result<(), task03::SpawnObjError> {
+        (&mut *self.0).spawn_obj(future)
+    }
+}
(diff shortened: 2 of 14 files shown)


futures-util/src/future/never_error.rs
```diff
@@ -0,0 +1,39 @@
...
+        NeverError { future }
+    }
+}
+
+impl<Fut: Unpin> Unpin for NeverError<Fut> {}
+
+impl<Fut, T> Future for NeverError<Fut>
+    where Fut: Future<Output = T>,
+{
+    type Output = Result<T, ()>;
```
MajorBreakfast: I didn't know that [`Never`](https://github.com/rust-lang-nursery/futures-rs/blob/0.2/futures-core/src/never.rs) was introduced only in 0.2 when I proposed the name `NeverError`. I think we should come up with another name for this type. Maybe `UnitError`?

That said, I continue to believe that this combinator is useful. `.map(|x| Ok::<_, ()>(x))` would be damn annoying to type.

futures-util/src/future/never_error.rs
```diff
@@ -0,0 +1,39 @@
...
+
+    fn poll(mut self: PinMut<Self>, cx: &mut task::Context) -> Poll<Result<T, ()>> {
+        match self.future().poll(cx) {
```
MajorBreakfast: `self.future().poll(cx).map(|x| Ok(x))` is shorter :)

futures-util/src/future/mod.rs
```diff
@@ -83,6 +86,8 @@ if_std! {
 
     mod shared;
     pub use self::shared::Shared;
+
+    use std::boxed::PinBox;
```
MajorBreakfast: Best placed right underneath `if_std! {`. We usually have the imports on top

futures-util/src/compat/executor.rs
```diff
@@ -0,0 +1,47 @@
+
+pub trait ExecCompat: Executor01<
+        Compat<FutureObj<'static, ()>, BoxedExecutor>
+    > + Clone + Send + 'static
```
MajorBreakfast:
```rust
pub trait ExecCompat: Executor01<Fut> + Clone + Send + 'static
where Fut: Compat<FutureObj<'static, ()>, BoxedExecutor>
{
```
The multiline `<...>` reminds me of C++ xD

futures-util/src/compat/executor.rs
```diff
@@ -0,0 +1,47 @@
...
+{
+    fn compat(self) -> ExecutorCompat<Self>
```
MajorBreakfast: To be consistent that should be in `ExecutorExt` in `futures_util/task/executor.rs` (doesn't exist yet, but `ContextExt` does, so `ExecutorExt` should as well)

futures-util/src/compat/executor.rs
MajorBreakfast: Can you rename the trait to `Executor01CompatExt`?

futures-util/src/compat/executor.rs
```diff
@@ -0,0 +1,47 @@
...
+{
+    fn compat(self) -> ExecutorCompat<Self> {
+        ExecutorCompat {
+            exec: self,
+        }
+    }
+}
+
+#[derive(Clone)]
+pub struct ExecutorCompat<E> {
```
MajorBreakfast: Rename to `CompatExecutor`

futures-util/src/compat/executor.rs
```diff
@@ -0,0 +1,47 @@
...
+    exec: E
+}
+
+impl<E> Executor03 for ExecutorCompat<E> 
+    where E: Executor01<
+        Compat<FutureObj<'static, ()>, Box<Executor03>>
+    >,
```
MajorBreakfast: Fits on single line (or if you prefer to split it use a type param)

futures-util/Cargo.toml
```diff
@@ -29,6 +30,7 @@ futures-io-preview = { path = "../futures-io", version = "0.3.0-alpha.1", defaul
 futures-sink-preview = { path = "../futures-sink", version = "0.3.0-alpha.1", default-features = false}
 either = { version = "1.4", default-features = false }
 slab = { version = "0.4", optional = true }
+futures = { version = "0.1", optional = true }
```
Nemo157: I’m surprised this plus the dev-dependency below don’t cause issues for the tests (both have a lib name of `futures`). Again here’s a place where having crate renaming for the edition would be super useful to make this `futures01` or something.

futures-util/src/compat/executor.rs
```diff
@@ -0,0 +1,47 @@
...
+    E: Clone + Send + 'static,
+{
+    fn spawn_obj(&mut self, obj: FutureObj<'static, ()>) -> Result<(), task::SpawnObjError> {
```
MajorBreakfast: Long line

```rust
fn spawn_obj(
    &mut self,
    obj: FutureObj<'static, ()>,
) -> Result<(), task::SpawnObjError> {
```

(Note: Trailing comma)

MajorBreakfast:
@tinaun Brilliant work!

ATM it doesn't compile because the `mod executor` statement was missing. I'll comment more later once you've fixed that. (Doesn't make sense to fix it myself if you do it anyway)

- Add to all `01` imports the suffix `01` and no suffix for `03`. Or, alternatively (if you prefer) add suffixes to everything. Currently it's a bit mixed.
- Try if `extern crate futures as futures01` in `lib.rs` can be used for referring to the futures 0.1 crate as `futures01` in use statements

Nemo157:
@tinaun are you planning on working on adaptors for `Stream`, `Sink`, or the io traits as well? If not I can build off this for some of those.

futures-util/src/future/never_error.rs
MajorBreakfast: @Nemo157 Yeah ^^'
@tinaun I just noticed something funny: *You* added the `Poll::map` method to libcore yourself 😁

MajorBreakfast:
> where should 0.1 -> 0.3 combinators live?

I think your approach to pack everything related to compatibility into `futures_util::compat` is the right solution.

futures-util/Cargo.toml
MajorBreakfast: We should convene the council of elders

One temporary solution *could* be to fully rename our libraries to `futures-preview`. Long term though, we'll need crate renaming because both crates will share a name in the end.

tinaun: could we get away with moving `Compat` into its own `futures-compat` sub crate?

Nemo157: I think that'll work for now, since it doesn't need to import the 0.3 `futures` facade it won't run into the name conflict. Once renaming is working then it could be merged back into `futures-util`.

MajorBreakfast: @tinaun We might. It only happens in the doc tests because `futures-util` has `futures-preview` as dev dependency. So the normal dep `futures@0.1` and dev dep collide.

I think it is problematic that one can't depend on `futures` 0.3 and 0.1 in the same project at the same time

Nemo157: Alright, testing with a locally built `cargo` including https://github.com/rust-lang/cargo/pull/5794 using `futures01 = { version = "0.1", optional = true, package = "futures" }` (and adjusting all the `use`s) appears to run the tests fine.

So, if you're willing to wait for a new `cargo` to be released to nightly that's one option.

MajorBreakfast: @Nemo157 How long do you think will this take? If it's just a day or two, then we should wait. @tinaun requires some time to tweak this PR anyway

Nemo157: I'm not sure what the Cargo release cycle is like, my current nightly toolchain has `rustc` from 2018-07-23 but `cargo` from 2018-07-17, so I don't think it's always the latest master.

tinaun: this pr requires a new nightly anyway for the Executor stuff

MajorBreakfast: @tinaun Wouldn't the newtype approach be more lightweight?

Nemo157:
Since I need them for a project I want to use as a test case for this I've got `Stream` and (uni-directional) `Sink` compat implementations ready. Once the executor stuff is sorted I'll hopefully have a small ported project to show. (Although, it's really not the nicest codebase to start with).

tinaun:
besides the NeverError bikeshed and the cargo renamings this should be about done

Nemo157:
IMO the cargo renaming can wait. This compiles and works currently, we just can’t run the tests with the feature active until renaming is supported.

MajorBreakfast:
I'm going to merge this now. We still need need to document it (I think `cargo doc` ignores it ATM) and add some further polish. However that can be done in follow-up PRs :)

Big thanks to @tinaun !

Nemo157:
Guess that means I should get onto doing the rebase of my changes I'd planned 😄

futures-util/src/compat/executor.rs
```diff
@@ -0,0 +1,70 @@
...
+        let fut = future.never_error().compat(BoxedExecutor(Box::new(self.clone())));
+
+        self.exec.execute(fut)
+            .map_err(|exec_err| {
+                use futures_core::task::{SpawnObjError, SpawnErrorKind};
+                     
+                let fut = exec_err.into_future().into_inner().unwrap_or_else(|_| ());
+                SpawnObjError {
+                    kind: SpawnErrorKind::shutdown(),
+                    task: Box::new(fut).into(),
```
Nemo157: This field is now `future`.

PR #1122, opened 2018-07-25
`unwrap` combinator

MajorBreakfast:


Nemo157:
What about `expect` as well?

MajorBreakfast:
@Nemo157 Good idea. I also want to go through all the documentation texts. I glanced "can conform to any specific `Error` type, including `Never`." in the `unwrap_or_else` combinators description ^^'

Nemo157:
Oh yeah, there's a whole lot of `unwrap_*` combinators, I forgot about those as well. Is it worth replicating most of the `Result` API when users could just `.map(|a| a.unwrap_or_else(|| foo))`?

MajorBreakfast:
@Nemo157 I investigated this and I think I won't add `expect`. `Result::expect` takes a `&str`. That means a faithful impl would only be useful inside an async fn. One might as well `await!(_).expect("...")` then. IMO `.map_err(|_| panic!("..."))` is nice enough.

And I won't be the one who adds `unwrap_or` 'cause that method is a footgun

cramertj:
I'm not sure I'm sold on the usefulness of this. Between `unwrap` being generally considered an antipattern and `await!(fut.unwap())` being the same as `await!(fut).unwrap()` it seems like there wouldn't be a lot of utility here. Is there a specific use-case you had in mind?

MajorBreakfast:
@cramertj Yes. https://github.com/rust-lang-nursery/net-wg/issues/26#issuecomment-407577152

cramertj:
Ick... That feels like something that maybe deserves a special-case, especially since in the case of 0.1 "spawned" futures returning an error of type `()` would actually *succeed*, not fail: https://docs.rs/futures/0.1.23/src/futures/task_impl/std/mod.rs.html#453

MajorBreakfast:
Okay let's not add it then.

`.map(|x| x.unwrap())` is a good alternative

PR #1123, opened 2018-07-25
Macros

cramertj:


futures/src/lib.rs
```diff
@@ -50,8 +50,15 @@
...
-mod macros;
+// Macro reexports
+pub use futures_util::{
+    // Async-await
+    join, select, pending, poll,
+    // Pinning
+    pin_mut, unsafe_pinned, unsafe_unpinned,
+    // Error/readiness propagation
+    try_ready, try_poll, ready,
+};
```
MajorBreakfast: I'd prefer if we namespaced these macros:
- Poll macros belong in `task`
- async/await macros in `async_await`
- and pin macros in `pin_macros`

Top level reexports are fine, but I'd prefer if they were doc hidden

futures/tests/async_await_macros.rs
```diff
@@ -0,0 +1,48 @@
...
+}
+
+#[test]
+fn select() {
+    let (tx1, mut rx1) = oneshot::channel::<i32>();
+    let (_tx2, mut rx2) = oneshot::channel::<i32>();
+    tx1.send(1).unwrap();
+    block_on(async {
+        select! {
+            rx1 => assert_eq!(Ok(1), rx1),
```
MajorBreakfast: It'd be good if the test could check that this arm is actually executed

futures/src/lib.rs
cramertj: We can't namespace them inside of the crates in which they are defined (futures-util), so I thought it better to be consistent and keep them at top-level here. Do you disagree?

futures/tests/async_await_macros.rs
cramertj: Done.

futures/src/lib.rs
Nemo157: I would say `join` and `select` should be in `future` since that's what they operate on, e.g. there could be a `stream::select!` macro for selecting the next ready item from multiple streams.

cramertj: Right, but they *can't* be in `futures-util::future`, they can only be in `future::future`. Do you think it's better to be consistent, or to have them in the inner module?

MajorBreakfast: I'd prefer to namespace them. Since they're reexported we have the luxury to do this. It's great for documentation because a tutorial can say something like "the `async_await` module contains tools for use inside async functions"

cramertj: I feel like that's sort of misleading, since the goal is that most people will interact with *the whole crate* from nearly exclusively the context of async functions / blocks / closures.

Nemo157: Add a private helper crate `futures-util-macros` so you can re-export them in `futures-util`? (not really a serious suggestion, but I wonder how often that hack is going to be used until macros 2.0 gets here)

Since I'm still holding out hope for public/private dependencies to be a thing at some point I lean on consistencies side to reduce the overhead to migrate away from the facade.

cramertj: > Add a private helper crate futures-util-macros so you can re-export them in futures-util?

Yeah, I considered that, but then we can't use things defined in futures-util itself from within the macros. Perhaps we could move those dependencies into the helper crate, but then we start getting all mixed up

cramertj:
I'm going to go ahead and merge this since either way I think we''ll want a top-level reexport (even if doc-hidden), and I'd prefer for now to keep the API surface of `futures` symmetric with that of `futures-util` and the other child crates.

MajorBreakfast:
It's a tricky situation. I agree that merging this is fine. After all this PR didn't actually change how things work. Ultimately when macros 2.0 land we should do namespacing and have doc hidden top level reexports for compatibility.

PR #1129, opened 2018-07-26
Refactor `TryFuture` docs

MajorBreakfast:


futures-util/src/try_future/and_then.rs
```diff
@@ -3,10 +3,7 @@ use core::mem::PinMut;
 use futures_core::future::{Future, TryFuture};
 use futures_core::task::{self, Poll};
 
-/// Future for the `and_then` combinator, chaining a computation onto the end of
-/// another future which completes successfully.
-///
-/// This is created by the `Future::and_then` method.
+/// Future for the [`and_then`](super::TryFutureExt::and_then) combinator.
```
MajorBreakfast: I replaced all these with just a link to the detailed documentation of the combinator method. This avoids duplication and is better for the user because the method's documentation far better.

futures-util/src/try_future/mod.rs
```diff
@@ -51,19 +51,42 @@ pub use self::unwrap_or_else::UnwrapOrElse;
...
+    /// use futures::sink::Sink;
+    /// # use futures::channel::mpsc::{self, SendError};
+    /// # type T = i32;
+    /// # type E = SendError;
+    ///
+    /// fn make_sink_async() -> impl Future<Output = Result<
+    ///     impl Sink<SinkItem = T, SinkError = E>,
+    ///     E,
+    /// >> { // ... }
+    /// # let (tx, _rx) = mpsc::unbounded::<i32>();
```
Nemo157: Just realised we’re missing an equivalent of [`io::Sink`](https://doc.rust-lang.org/nightly/std/io/struct.Sink.html) for `Sink`s (unfortunate naming collision though). I might submit a PR for one tomorrow.

cramertj: Please don't call it `Sink` or `Sync` XD

futures-util/src/try_future/mod.rs
```diff
@@ -193,39 +220,43 @@ pub trait TryFutureExt: TryFuture {
...
-    ///
-    /// This function can be used to chain two futures together and ensure that
-    /// the final future isn't resolved until both have finished. The closure
-    /// provided is yielded the successful result of this future and returns
-    /// another value which can be converted into a future.
-    ///
-    /// Note that because `Result` implements the `IntoFuture` trait this method
-    /// can also be useful for chaining fallible and serial computations onto
-    /// the end of one future.
+    /// Chains on a computation returning another future with this future's
```
cramertj: This phrasing seems confusing to me-- I actually like previous description, and I have an easier time understanding the behavior of the method and the resulting future.

futures-util/src/try_future/mod.rs
```diff
@@ -193,39 +220,43 @@ pub trait TryFutureExt: TryFuture {
...
+    /// success value as input.
     ///
-    /// If this future is dropped, panics, or completes with an error then the
-    /// provided closure `f` is never called.
+    /// The provided closure `f` will only be called if this future is resolved
```
cramertj: This bit is much better though, thanks :)

futures-util/src/try_future/mod.rs
```diff
@@ -409,23 +444,26 @@ pub trait TryFutureExt: TryFuture {
     }
 */
 
-    /// Handle errors generated by this future by converting them into
-    /// `Self::Item`.
+    /// Unwraps this future's ouput, producing a future with this future's
+    /// [`Ok`](TryFuture::Ok) type as
```
cramertj: "as [its] output type"?

futures-util/src/try_future/mod.rs
```diff
@@ -409,23 +444,26 @@ pub trait TryFutureExt: TryFuture {
...
+    /// [`Output`](std::future::Future::Output) type.
     ///
-    /// Because it can never produce an error, the returned `UnwrapOrElse` future can
-    /// conform to any specific `Error` type, including `Never`.
+    /// If this futures is resolved successfully, the returned future will
```
cramertj: s/futures/future

cramertj: Maybe "the original future" rather than "this future"? It's not totally clear which future is "this future" to me

cramertj:
LGTM with comments fixed. These doctests are awesome! Thank you so much!

futures-util/src/try_future/mod.rs
MajorBreakfast: Changed it to:

> Executes another future after this one resolves successfully. The success value is passed to a closure to create this subsequent future.

That's close to the original phrasing but it also mentions the closure which is IMO crucial.

Issue #1128, opened 2018-07-26
select! and join! macros need documentation


MajorBreakfast:

MajorBreakfast:
Indeed. Big thanks to @tinaun for providing the PR!

PR #1130, opened 2018-07-27
added macro doctests

tinaun:
part of #1128

### DIFF
--- a/futures-util/src/async_await/join.rs
+++ b/futures-util/src/async_await/join.rs
@@ -1,5 +1,26 @@
 //! The `join` macro.
 
+/// Polls multiple futures simultaneously, returning a tuple
+/// of all results once complete. 
+/// 
+/// While `join!(a, b)` is similar to `(await!(a), await!(b))`,
+/// `join!` polls both futures concurrently and therefore is more efficent.
+/// 
+/// This macro is only usable inside of async functions, closures, and blocks.
+/// 
+/// # Examples
+/// 
+/// ```
+/// #![feature(pin, async_await, await_macro, futures_api)]
+/// # futures::executor::block_on(async {
+/// use futures::{join, future};
+///
+/// let a = future::ready(1);
+/// let b = future::ready(2);
+/// 
+/// assert_eq!(join!(a, b), (1, 2));
+/// # });
+/// ```
 #[macro_export]
 macro_rules! join {
     ($($fut:ident),*) => { {
--- a/futures-util/src/async_await/pending.rs
+++ b/futures-util/src/async_await/pending.rs
@@ -3,11 +3,13 @@ use futures_core::future::Future;
 use futures_core::task::{self, Poll};
 
 /// A macro which yields to the event loop once.
-/// This is similar to returning `Poll::Pending` from a `Future::poll` implementation.
-/// If `pending!` is used, the current task should be scheduled to receive a wakeup
-/// when it is ready to make progress.
+/// 
+/// This is equivalent to returning [`Poll::Pending`](futures_core::task::Poll) 
+/// from a [`Future::poll`](futures_core::future::Future::poll) implementation. 
+/// Similarly, when using this macro, it must be ensured that [`wake`](std::task::Waker::wake) 
+/// is called somewhere when further progress can be made.
 ///
-/// This macro is only usable inside of `async` functions, closures, and blocks.
+/// This macro is only usable inside of async functions, closures, and blocks.
 #[macro_export]
 macro_rules! pending {
     () => {
--- a/futures-util/src/async_await/select.rs
+++ b/futures-util/src/async_await/select.rs
@@ -1,5 +1,30 @@
 //! The `select` macro.
 
+/// Polls multiple futures simultaneously, executing the branch for the future 
+/// that finishes first.
+/// 
+/// `select!` can select over futures with different output types, but each 
+/// branch has to have the same return type. Inside each branch, the respective 
+/// future's output is available via a variable with the same name as the future.
+/// 
+/// This macro is only usable inside of async functions, closures, and blocks.
+/// 
+/// # Examples
+/// 
+/// ```
+/// #![feature(pin, async_await, await_macro, futures_api)]
+/// # futures::executor::block_on(async {
+/// use futures::{select, future};
+/// let mut a = future::ready(4);
+/// let mut b: future::Empty<()> = future::empty();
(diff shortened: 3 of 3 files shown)


futures-util/src/async_await/join.rs
```diff
@@ -1,5 +1,25 @@
-//! The `join` macro.
 
+/// The `join` macro.
```
MajorBreakfast: This text was a summary for the entire module. It's not required because the module isn't exported publicly. Now, you moved into the description for `join!`. Usually descriptions don't start with a headline, but with a summary sentence.

futures-util/src/async_await/join.rs
```diff
@@ -1,5 +1,25 @@
...
+/// 
+/// This macro polls multiple futures simultaneously, returning a tuple
+/// of all results once complete. While `join!(a, b)` is similar to `(await!(a), await!(b))`,
+/// unlike the latter `join!` can make progress on multiple pending futures at once.
+/// 
+/// This macro is only usable inside of `async` functions, closures, and blocks.
+/// 
+/// # Example
```
MajorBreakfast: We always use plural here

futures-util/src/async_await/join.rs
```diff
@@ -1,5 +1,25 @@
-//! The `join` macro.
 
+/// The `join` macro.
+/// 
+/// This macro polls multiple futures simultaneously, returning a tuple
+/// of all results once complete. While `join!(a, b)` is similar to `(await!(a), await!(b))`,
+/// unlike the latter `join!` can make progress on multiple pending futures at once.
```
MajorBreakfast: - "~~This macro~~ Polls multiple futures..." Rustdoc shows that it's a macro right above the text. Also, it always looks a bit odd if we have multiple descriptions next to each other and every one of them starts with "This marcro/method/function ..."
- Make a new paragraph after the first sentence. The first sentence is a great summary sentence

futures-util/src/async_await/join.rs
```diff
@@ -1,5 +1,25 @@
...
+/// 
+/// This macro is only usable inside of `async` functions, closures, and blocks.
```
MajorBreakfast: Highlight "async" only when you mean the keyword. (Same with `Future`, we usually write just "future"). I once submitted a PR to the async functions RFC that highlighted all the keywords. But it was rejected because it looked cluttered (and it did!) ^^'

futures-util/src/async_await/pending.rs
```diff
@@ -3,6 +3,7 @@ use futures_core::future::Future;
 use futures_core::task::{self, Poll};
 
 /// A macro which yields to the event loop once.
+///
```
MajorBreakfast: Replace the last paragraph with ``"This is equivalent to returning [`Poll::Pending`] from a [`Future::poll`] implementation. Similarly, when using this macro, it must be ensured that [`wake`](std::task::Waker::wake) is called somewhere when further progress can be made."``

(I rephrased it a bit an added links)

futures-util/src/async_await/select.rs
```diff
@@ -1,5 +1,30 @@
-//! The `select` macro.
-
+/// The `select` macro.
+/// 
+/// This macro polls multiple futures simultaneously, executing the branch that finishes first.
+/// The output of the future shadows the name of the future within each branch.
+/// 
+/// This macro is only usable inside of `async` functions, closures, and blocks.
+/// 
+/// # Example
```
MajorBreakfast: Same points as above: Heading, async, plural

futures-util/src/async_await/select.rs
```diff
@@ -1,5 +1,30 @@
-//! The `select` macro.
-
+/// The `select` macro.
+/// 
+/// This macro polls multiple futures simultaneously, executing the branch that finishes first.
```
MajorBreakfast: "executing the branch [for the future] that finishes first."

futures-util/src/async_await/select.rs
```diff
@@ -1,5 +1,30 @@
...
+/// The output of the future shadows the name of the future within each branch.
```
MajorBreakfast: This one's tricky... maybe: "Inside each branch, the respective future's output is available via a variable with the same name as the future."

MajorBreakfast:
Great! I left some comments

tinaun:
i can't think of a summary sentence for select! that im happy with

futures-util/src/async_await/join.rs
```diff
@@ -1,5 +1,26 @@
 //! The `join` macro.
 
+/// Polls multiple futures simultaneously, returning a tuple
+/// of all results once complete. 
+/// 
+/// While `join!(a, b)` is similar to `(await!(a), await!(b))`,
+/// unlike the latter `join!` can make progress on multiple pending futures at once.
```
MajorBreakfast: That sentence sounds a bit weird. What do you think about this as replacement?:

``The difference between `join!(a, b)` and `(await!(a), await!(b))` is that `join!` polls both futures concurrently and therefore is more efficient.``

futures-util/src/async_await/select.rs
```diff
@@ -1,5 +1,33 @@
...
+/// let mut a = future::ready(4);
+/// let mut b: future::Empty<()> = future::empty();
+/// 
+/// let res = select! {
+///     a => {
+///         a + 1
+///     },
+///     b => {
+///         0
+///     },
```
MajorBreakfast: I'd make these one-liners

MajorBreakfast:
> i can't think of a summary sentence for select! that im happy with

@tinaun I actually like the current version. If you want to change it later, you can always provide another PR 🙂

BTW, currently the CI breaks

futures-util/src/async_await/select.rs
```diff
@@ -1,5 +1,30 @@
 //! The `select` macro.
 
+/// Polls multiple futures simultaneously, executing the branch for the future 
+/// that finishes first.
+/// 
+/// `select!` can select over futures with different output types, but each 
+/// branch has to resolve to the same type. Inside each branch, the respective
```
MajorBreakfast: > each branch has to resolve to the same type.

- The word resolve is a good choice. "resolve" is usually only used for future. "return" would be a better choice
- Is this true if the output is not used?

tinaun: if the output is not used, the return type of the branch is `()`, isn't it?

PR #1131, opened 2018-07-27
Add sink::drain for discarding items

Nemo157:
Name’s up for bikeshed, some I considered:

* BlackHole
* Null
* Void
* Discard
* Dispose
* Drop

### DIFF
--- a/futures-util/src/sink/drain.rs
+++ b/futures-util/src/sink/drain.rs
@@ -1,0 +1,65 @@
+use core::marker::PhantomData;
+use core::mem::PinMut;
+use futures_core::task::{self, Poll};
+use futures_sink::Sink;
+
+/// A sink that will discard all items given to it.
+///
+/// See the [`drain`] function for more details.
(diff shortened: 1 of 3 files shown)


futures-util/src/sink/drain.rs
```diff
@@ -0,0 +1,65 @@
...
+/// await!(drain.send(5))?;
+/// # Ok::<(), ()>(()) }).unwrap();
+/// ```
+pub fn drain<T>() -> Drain<T> {
+    Drain { marker: PhantomData }
+}
+
+impl<T> Sink for Drain<T> {
+    type SinkItem = T;
+    type SinkError = ();
```
Nemo157: Do we have a good generic void error currently? I would use `!` here, but that’s still a ways from stable. `()` is especially not good since it doesn’t implement `Error`.

cramertj:
LGTM aside from error type-- I'm fine w/ `Drain` as the name.

PR #1137, opened 2018-07-29
Prevent shadowing of future name in select!

cramertj:
This isn't perfect, but I think it's definitely an improvement.

### DIFF
--- a/futures-util/src/async_await/select.rs
+++ b/futures-util/src/async_await/select.rs
@@ -10,13 +10,23 @@ macro_rules! select {
     )*) => { {
         $(
             $crate::async_await::assert_unpin(&$name);
-            let mut $name = $crate::future::maybe_done(&mut $name);
-            let mut $name = $crate::core_reexport::mem::PinMut::new(&mut $name);
         )*
+        #[allow(bad_style)]
+        struct __priv<$($name,)*> {
+            $(
+                $name: $name,
+            )*
+        }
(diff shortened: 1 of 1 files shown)


MajorBreakfast:
A PR from you on the weekend? 😁

So, what's happening here is that inside each branch the respective output still shadows the future, but the bindings of the other futures are now visible?

futures-util/src/async_await/select.rs
```diff
@@ -10,13 +10,23 @@ macro_rules! select {
     )*) => { {
         $(
             $crate::async_await::assert_unpin(&$name);
-            let mut $name = $crate::future::maybe_done(&mut $name);
-            let mut $name = $crate::core_reexport::mem::PinMut::new(&mut $name);
         )*
+        #[allow(bad_style)]
+        struct __priv<$($name,)*> {
```
MajorBreakfast: The name should be capitalized

cramertj: ...why?

cramertj:
>  So, what's happening here is that inside each branch the respective output still shadows the future, but the bindings of the other futures are now visible?

Yup!

futures-util/src/async_await/select.rs
MajorBreakfast: because all structs have a name that starts with an upper case letter by convention. `__Priv` would suggest that it's a type, not a binding

PR #1138, opened 2018-07-29
 Consistent naming for "task"

MajorBreakfast:
- Rename `task` to `future` where appropriate
- Rename `TaskContainer` to `Task` in `thread_pool.rs`
- Rename `Node` to `Task` in `FuturesUnordered` because it fits our revised task definition like a glove
- Tweak various documentation and code comments

If this PR is reviewed before we get the new nightly, we can merge it as is. Otherwise, there are two lines of code that need to be modified.


futures-util/src/stream/futures_unordered/ready_to_run_queue.rs
```diff
@@ -6,39 +6,39 @@ use std::sync::atomic::AtomicPtr;
...
+    Data(*const Task<Fut>),
     Empty,
     Inconsistent,
 }
 
 pub(super) struct ReadyToRunQueue<Fut> {
-    // The task using `FuturesUnordered`.
-    pub(super) parent: AtomicWaker,
+    // The waker of the task using `FuturesUnordered`.
+    pub(super) waker: AtomicWaker,
```
MajorBreakfast: I've renamed this to `waker` instead of `parent`. I think that makes it clearer. I named it `parent_waker` for a while, but I think `waker` is clear enough because the queue has only a single one.

futures-util/src/stream/futures_unordered/mod.rs
```diff
@@ -20,32 +20,34 @@ mod abort;
...
 
-/// A set of `Future`s which may complete in any order.
+/// A set of futures which may complete in any order.
 ///
 /// This structure is optimized to manage a large number of futures.
-/// Futures managed by `FuturesUnordered` will only be polled when they
-/// generate notifications. This reduces the required amount of work needed to
-/// poll large numbers of futures.
+/// Futures managed by [`FuturesUnordered`] will only be polled when they
+/// generate wake-up notifications. This reduces the required amount of work
```
cramertj: nit: wakeup is one word

MajorBreakfast:
My preference is to merge it because it's not really blocked. We just get conflicts if we let it sit

cramertj:
Yeah, I'm just already working off of a new version of rustc locally (not a nightly) so it'd be nice to be able to update, but I understand that we probably shouldn't have a not-working-on-nightly master ;)

PR #1147, opened 2018-07-30
Fix warnings in new nightly

### LINKED ISSUE #1143: New nightly shows warnings that will become hard errors in the future
Nightly from 2018-07-29

```
warning: variable does not need to be mutable
   --> futures-io/src/lib.rs:365:21
    |
365 |                 let mut out = self.get_mut().as_mut();
    |                     ----^^^
    |                     |
    |                     help: remove this `mut`
    |
    = note: `-D unused-mut` implied by `-D warnings`
    = warning: This error has been downgraded to a warning for backwards compatibility with previous releases.
            It represents potential unsoundness in your code.
            This warning will become a hard error in the future.

    Checking futures-util-preview v0.3.0-alpha.1 (file:///Users/josef/Documents/GitHub/futures-rs/futures-util)
warning: variable does not need to be mutable
   --> futures-util/src/sink/with.rs:124:28
    |
124 |             State::Process(mut fut) => Some(try_ready!(fut.poll(cx))),
    |                            ----^^^
    |                            |
    |                            help: remove this `mut`
    |
    = note: `-D unused-mut` implied by `-D warnings`
    = warning: This error has been downgraded to a warning for backwards compatibility with previous releases.
            It represents potential unsoundness in your code.
            This warning will become a hard error in the future.

warning[E0499]: cannot borrow `promoted` as mutable more than once at a time
  --> futures-util/src/io/read_exact.rs:43:61
   |
43 |                 let (_, rest) = mem::replace(&mut this.buf, &mut []).split_at_mut(n);
   |                                                             ^^^^^^^ mutable borrow starts here in previous iteration of loop
   |
note: borrowed value must be valid for the lifetime 'a as defined on the impl at 35:6...
  --> futures-util/src/io/read_exact.rs:35:6
   |
35 | impl<'a, R: AsyncRead + ?Sized> Future for ReadExact<'a, R> {
   |      ^^
   = warning: This error has been downgraded to a warning for backwards compatibility with previous releases.
           It represents potential unsoundness in your code.
           This warning will become a hard error in the future.
```

Nemo157:
Part of #1143.

### DIFF
--- a/futures-io/src/lib.rs
+++ b/futures-io/src/lib.rs
@@ -362,7 +362,7 @@ if_std! {
         ) -> Poll<Result<usize>> {
             let position = self.position();
             let result = {
-                let mut out = self.get_mut().as_mut();
+                let out = self.get_mut().as_mut();
                 let pos = cmp::min(out.len() as u64, position) as usize;
                 StdIo::Write::write(&mut &mut out[pos..], buf)
             };
--- a/futures-util/src/sink/with.rs
+++ b/futures-util/src/sink/with.rs
@@ -121,7 +121,7 @@ impl<Si, U, Fut, F, E> With<Si, U, Fut, F>
     ) -> Poll<Result<(), E>> {
         let buffered = match self.state().as_pin_mut() {
             State::Empty => return Poll::Ready(Ok(())),
-            State::Process(mut fut) => Some(try_ready!(fut.poll(cx))),
+            State::Process(fut) => Some(try_ready!(fut.poll(cx))),
             State::Buffered(_) => None,
         };
         if let Some(buffered) = buffered {


cramertj:
I don't think we should do this-- as I pointed out in https://github.com/rust-lang/rust/issues/52671, I think it should be fine for an *empty* type (like []) to be promoted and have mutable references to it shared.

Nemo157:
Dropped the `ReadExact` change, now this is just removing some unnecessary muts.

futures-io/src/lib.rs
```diff
@@ -362,7 +362,7 @@ if_std! {
         ) -> Poll<Result<usize>> {
             let position = self.position();
             let result = {
-                let mut out = self.get_mut().as_mut();
+                let out = self.get_mut().as_mut();
```
cramertj: Can you explain why this `mut` is unused? It looks used to me (there's an `&mut out` below). Is it just because it's an `&mut` already and some coercion is firing?

cramertj: Got it, thanks.

PR #1149, opened 2018-07-30
Add job to publish docs on tags

Nemo157:
r? @MajorBreakfast 

This reuses the existing `GH_TOKEN` variable (at least, I assume that's what the encrypted environment variable is from looking back at the history) which I believe is from @alexcrichton's account. We may want to change that (especially if that doesn't give it access to push to https://github.com/rust-lang-nursery/futures-doc)?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,11 @@ rust:
 sudo: false
 cache: cargo
 
+stages:
+  - name: test
+  - name: release
+    if: tag IS present
+
 matrix:
   include:
     - name: cargo test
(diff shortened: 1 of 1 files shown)


MajorBreakfast:
I don't like that there's a "secret" key publicly inside our `.travis.yml`. I'll have to check how travis really recommends doing this.

Nemo157:
It’s encrypted using a key that is [only accessible inside the build](https://docs.travis-ci.com/user/encryption-keys/). They also [automatically scrub it from the log if leaked](https://docs.travis-ci.com/user/best-practices-security/#steps-travis-ci-takes-to-secure-your-data).  This is standard practice for giving Travis.CI access to release github pages. Additionally using `.netrc` for authentication should avoid all the potential leaks from `git` (but if it doesn’t somehow then the automatic scrubbing will get it).

Nemo157:
Updated to enable `--all-features` to get the compatibility docs in.

Nemo157:
(also add `--no-deps` to test build to reduce build time marginally)

cramertj:
Ping @MajorBreakfast @Nemo157 is there something blocking this or can we land? (after rebase)

.travis.yml
```diff
@@ -79,11 +84,32 @@ matrix:
...
-        - cargo doc
+        - cargo doc --all --no-deps --all-features
+
+    - name: publish docs
+      stage: release
+      rust: nightly
+      before_script:
+        - echo "machine github.com login $GH_TOKEN password x-oauth-basic" >> ~/.netrc
+        - chmod 0600 ~/.netrc
+        - git clone https://github.com/rust-lang-nursery/futures-doc
```
MajorBreakfast: I renamed this to `futures-api-docs`

MajorBreakfast:
Unfortunately I don't have the time to work on this this weekend.

@Nemo157 In case you have the time, I've made you admin on the futures-api-docs repo, that way you can try the script and force push if you destroy something :D

.travis.yml
Nemo157: Updated url in both places

Nemo157:
I've rechecked the script for the new location and I'm pretty confident on this. Easiest way to test is to push a temporary tag pointing at the latest commit (35927b7), it should push a new commit to futures-api-docs that can then be force-pushed back to what it was or reverted. I don't have access to push tags here though.

Nemo157:
That should have fixed the doc build.

Nemo157:
Rebased on top of #1264 since they touch the same travis jobs and this depends on the build succeeding.

PR #1153, opened 2018-07-31
Add TryStreamExt::try_filter_map

gbonik:
A result-aware version of filter_map would be very useful

### DIFF
--- a/futures-util/src/try_stream/mod.rs
+++ b/futures-util/src/try_stream/mod.rs
@@ -240,6 +240,8 @@ pub trait TryStreamExt: TryStream {
     ///
     /// # Examples
     /// ```
+    /// #![feature(async_await, await_macro)]
+    /// # futures::executor::block_on(async {
     /// use futures::executor::block_on;
     /// use futures::prelude::*;
     ///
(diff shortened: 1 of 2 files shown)


futures-util/src/try_stream/mod.rs
```diff
@@ -219,4 +222,41 @@ pub trait TryStreamExt: TryStream {
...
+    ///
+    /// All errors are passed through without filtering in this combinator.
+    ///
+    /// Note that this function consumes the stream passed into it and returns a
+    /// wrapped version of it, similar to the existing `filter_map` methods in
+    /// the standard library.
+    ///
+    /// # Examples
+    /// ```
+    /// use futures::executor::block_on;
```
cramertj: Nit: we've been migrating the docs examples over to use `async`/`await` syntax, so rather than using two `block_on`s here, switch them to `await!`s and wrap the whole thing in a `block_on(async { .. })` as in [this example](https://github.com/rust-lang-nursery/futures-rs/blob/master/futures-util/src/try_stream/mod.rs#L40).

futures-util/src/try_stream/try_filter_map.rs
```diff
@@ -0,0 +1,88 @@
+use core::marker::Unpin;
+use core::mem::PinMut;
+use futures_core::future::{TryFuture};
+use futures_core::stream::{Stream, TryStream};
+use futures_core::task::{self, Poll};
+
+/// A combinator that attempts to filter the results of a stream
+/// and simultaneously map them to a different type.
+///
+/// This structure is returned by the `Stream::try_filter_map` method.
```
cramertj: Nit: can you make this include a doc link? like [this one](https://github.com/rust-lang-nursery/futures-rs/blob/master/futures-util/src/try_stream/err_into.rs#L6)

cramertj:
This looks great! Let me know when you've had a chance to fix the example and docs nits and I'll merge.

PR #1154, opened 2018-07-31
Fix stream termination in buffered & buffer_unordered

gbonik:
Also add an example for buffer_unordered

### DIFF
--- a/futures-util/src/stream/mod.rs
+++ b/futures-util/src/stream/mod.rs
@@ -740,6 +740,30 @@ pub trait StreamExt: Stream {
     ///
     /// This method is only available when the `std` feature of this
     /// library is activated, and it is activated by default.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// #![feature(async_await, await_macro)]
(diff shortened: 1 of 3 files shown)


futures-util/src/stream/mod.rs
```diff
@@ -740,6 +740,30 @@ pub trait StreamExt: Stream {
     ///
     /// This method is only available when the `std` feature of this
     /// library is activated, and it is activated by default.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// #![feature(async_await, await_macro)]
+    /// # futures::executor::block_on(async {
+    /// use futures::prelude::*;
```
MajorBreakfast: We're removing the prelude from the doc examples. That makes it clearer which traits are involved:

```rust
```

MajorBreakfast:
I've added one comment. Otherwise this looks good. Thanks!

futures-util/src/stream/mod.rs
gbonik: Done

PR #1155, opened 2018-08-01
Try join

cramertj:
cc @gbonik

### DIFF
--- a/futures-util/src/async_await/join.rs
+++ b/futures-util/src/async_await/join.rs
@@ -31,7 +31,7 @@ macro_rules! join {
         loop {
             let mut all_done = true;
             $(
-                if let $crate::core_reexport::task::Poll::Pending = $crate::poll!($fut.reborrow()) {
+                if $crate::poll!($fut.reborrow()).is_pending() {
                     all_done = false;
                 }
             )*
@@ -84,7 +84,7 @@ macro_rules! try_join {
         let res: $crate::core_reexport::result::Result<(), _> = loop {
             let mut all_done = true;
             $(
-                if let $crate::core_reexport::task::Poll::Pending = $crate::poll!($fut.reborrow()) {
+                if $crate::poll!($fut.reborrow()).is_pending() {
                     all_done = false;
                 } else if $fut.reborrow().output_mut().unwrap().is_err() {
                     // `.err().unwrap()` rather than `.unwrap_err()` so that we don't introduce
--- a/futures-util/src/async_await/join.rs
+++ b/futures-util/src/async_await/join.rs
@@ -1,15 +1,15 @@
 //! The `join` macro.
 
 /// Polls multiple futures simultaneously, returning a tuple
-/// of all results once complete. 
-/// 
+/// of all results once complete.
+///
 /// While `join!(a, b)` is similar to `(await!(a), await!(b))`,
 /// `join!` polls both futures concurrently and therefore is more efficent.
-/// 
+///
 /// This macro is only usable inside of async functions, closures, and blocks.
-/// 
+///
 /// # Examples
-/// 
+///
 /// ```
 /// #![feature(pin, async_await, await_macro, futures_api)]
 /// # futures::executor::block_on(async {
@@ -17,7 +17,7 @@
 ///
 /// let a = future::ready(1);
 /// let b = future::ready(2);
-/// 
+///
 /// assert_eq!(join!(a, b), (1, 2));
 /// # });
 /// ```
@@ -45,5 +45,66 @@ macro_rules! join {
         ($(
             $fut.reborrow().take_output().unwrap(),
         )*)
+    } }
+}
+
+/// Attempts to poll multiple futures simultaneously, returning a `Result`
+/// of a tuple of all of the results once complete.
+///
+/// `try_join!` is similar to `join!`, but completes immediately if any of
+/// the futures return an error.
+///
+/// This macro is only usable inside of async functions, closures, and blocks.
+///
+/// # Examples
+///
+/// ```
+/// #![feature(pin, async_await, await_macro, futures_api)]
+/// # futures::executor::block_on(async {
+/// use futures::{try_join, future};
+///
+/// let a = future::ready(Ok::<i32, i32>(1));
+/// let b = future::ready(Err::<u64, i32>(2));
+///
+/// assert_eq!(try_join!(a, b), Err(2));
+///
+/// let a = future::ready(Ok::<i32, i32>(1));
+/// let b = future::ready(Ok::<u64, i32>(2));
+///
+/// assert_eq!(try_join!(a, b), Ok((1, 2)));
+/// # });
+/// ```
+#[macro_export]
+macro_rules! try_join {
+    ($($fut:ident),*) => { {
+        $(
+            let mut $fut = $crate::future::maybe_done($fut);
+            $crate::pin_mut!($fut);
+        )*
+        let res: $crate::core_reexport::result::Result<(), _> = loop {
+            let mut all_done = true;
+            $(
+                if let $crate::core_reexport::task::Poll::Pending = $crate::poll!($fut.reborrow()) {
+                    all_done = false;
+                } else if $fut.reborrow().output_mut().unwrap().is_err() {
+                    // `.err().unwrap()` rather than `.unwrap_err()` so that we don't introduce
+                    // a `T: Debug` bound.
+                    break $crate::core_reexport::result::Result::Err(
+                        $fut.reborrow().take_output().unwrap().err().unwrap()
+                    );
+                }
+            )*
+            if all_done {
+                break $crate::core_reexport::result::Result::Ok(());
+            } else {
+                $crate::pending!();
+            }
+        };
+
+        res.map(|()| ($(
+            // `.ok().unwrap()` rather than `.unwrap()` so that we don't introduce
+            // an `E: Debug` bound.
+            $fut.reborrow().take_output().unwrap().ok().unwrap(),
+        )*))
     } }
 }
--- a/futures-util/src/future/maybe_done.rs
+++ b/futures-util/src/future/maybe_done.rs
@@ -44,8 +44,24 @@ pub fn maybe_done<Fut: Future>(future: Fut) -> MaybeDone<Fut> {
 }
 
 impl<Fut: Future> MaybeDone<Fut> {
+    /// Returns an optional mutable reference to the output of the future.
+    /// The output of this method will be `Some` if and only if the inner
+    /// future has been completed and [`take_output`](MaybeDone::take_output)
+    /// has not yet been called.
+    #[inline]
+    pub fn output_mut<'a>(self: PinMut<'a, Self>) -> Option<&'a mut Fut::Output> {
+        unsafe {
+            let this = PinMut::get_mut_unchecked(self);
+            match this {
+                MaybeDone::Done(res) => Some(res),
+                _ => None,
+            }
+        }
+    }
+
     /// Attempt to take the output of a `MaybeDone` without driving it
     /// towards completion.
+    #[inline]
     pub fn take_output(self: PinMut<Self>) -> Option<Fut::Output> {
         unsafe {
             let this = PinMut::get_mut_unchecked(self);
--- a/futures-util/src/try_future/mod.rs
+++ b/futures-util/src/try_future/mod.rs
@@ -32,6 +32,9 @@ pub use self::err_into::ErrInto;
 mod flatten_sink;
 pub use self::flatten_sink::FlattenSink;
 
+mod try_join;
+pub use self::try_join::{TryJoin, TryJoin3, TryJoin4, TryJoin5};
+
 mod into_future;
 pub use self::into_future::IntoFuture;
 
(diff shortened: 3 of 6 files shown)


futures-util/src/async_await/join.rs
```diff
@@ -47,3 +47,64 @@ macro_rules! join {
...
+///
+/// let a = future::ready(Ok::<i32, i32>(1));
+/// let b = future::ready(Err::<u64, i32>(2));
+///
+/// assert_eq!(try_join!(a, b), Err(2));
+///
+/// let a = future::ready(Ok::<i32, i32>(1));
+/// let b = future::ready(Ok::<u64, i32>(2));
+///
+/// assert_eq!(try_join!(a, b), Ok((1, 2)));
```
MajorBreakfast: Maybe make it two examples [like we have for `map_ok`](https://github.com/rust-lang-nursery/futures-rs/blob/43bfbbe450220db6710706cadd5975b268bb4415/futures-util/src/try_future/mod.rs#L114-L137). Also I suggest to show the successful one first

futures-util/src/async_await/join.rs
```diff
@@ -47,3 +47,64 @@ macro_rules! join {
         )*)
     } }
 }
+
+/// Attempts to poll multiple futures simultaneously, returning a `Result`
+/// of a tuple of all of the results once complete.
```
MajorBreakfast: ``Polls multiple futures simultaneously, resolves to a [`Result`] containing a tuple of all success values.``

(The tuple does not contain results)

futures-util/src/async_await/join.rs
```diff
@@ -47,3 +47,64 @@ macro_rules! join {
...
+///
+/// `try_join!` is similar to `join!`, but completes immediately if any of
```
MajorBreakfast: The should be links

futures-util/src/try_future/try_join.rs
```diff
@@ -0,0 +1,131 @@
...
+                    $($Fut: maybe_done(TryFutureExt::into_future($Fut))),*
+                }
+            }
+        }
+
+        impl<Fut1: TryFuture, $($Fut: TryFuture),*> $Join<Fut1, $($Fut),*> {
+            unsafe_pinned!(Fut1: MaybeDone<IntoFuture<Fut1>>);
+            $(
+                unsafe_pinned!($Fut: MaybeDone<IntoFuture<$Fut>>);
+            )*
```
MajorBreakfast: These `unsafe_pinned!` macro calls could be in the impl above. That's how we usually do it.

futures-util/src/try_future/try_join.rs
```diff
@@ -0,0 +1,131 @@
...
+        ($Join:ident, <Fut1, $($Fut:ident),*>),
+    )*) => ($(
+        $(#[$doc])*
+        #[must_use = "futures do nothing unless polled"]
+        pub struct $Join<Fut1: TryFuture, $($Fut: TryFuture),*> {
+            Fut1: MaybeDone<IntoFuture<Fut1>>,
+            $($Fut: MaybeDone<IntoFuture<$Fut>>,)*
+        }
+
+        impl<Fut1, $($Fut),*> fmt::Debug for $Join<Fut1, $($Fut),*>
```
MajorBreakfast: - It'd be nicer if the `Debug` impl was after the inherent impl

- Should there be an `Unpin` impl?

futures-util/src/try_future/mod.rs
```diff
@@ -501,4 +504,136 @@ pub trait TryFutureExt: TryFuture {
...
+    /// ```
+    fn try_join<Fut2>(self, other: Fut2) -> TryJoin<Self, Fut2>
+    where
+        Fut2: TryFuture<Error = Self::Error>,
+        Self: Sized,
+    {
+        TryJoin::new(self, other)
+    }
+
+    /// Same as `try_join`, but with more futures.
```
MajorBreakfast: This should be a link

futures-util/src/try_future/mod.rs
```diff
@@ -501,4 +504,136 @@ pub trait TryFutureExt: TryFuture {
...
         IntoFuture::new(self)
     }
+
+    /// Joins the result of two futures, waiting for them both to complete or
+    /// for one to produce an error.
+    ///
+    /// This function will return a new future which awaits both this and the
+    /// `other` future to complete. If successful, the returned future will
+    /// finish with a tuple of both results. If unsuccesful, it will complete
+    /// with an error.
```
MajorBreakfast: will complete with *first* error

futures-util/src/try_future/mod.rs
```diff
@@ -501,4 +504,136 @@ pub trait TryFutureExt: TryFuture {
...
+    ///
+    /// Note that this function consumes the receiving future and returns a
+    /// wrapped version of it.
```
MajorBreakfast: Note that this method consumes the futures it is called on and returns a wrapped version of them.

futures-util/src/try_future/mod.rs
```diff
@@ -501,4 +504,136 @@ pub trait TryFutureExt: TryFuture {
...
+    /// #![feature(async_await, await_macro, futures_api)]
+    /// # futures::executor::block_on(async {
+    /// use futures::future::{self, TryFutureExt};
+    ///
+    /// let a = future::ready(Ok::<i32, i32>(1));
+    /// let b = future::ready(Err::<i32, i32>(2));
+    /// let pair = a.try_join(b);
+    ///
+    /// assert_eq!(await!(pair), Err(2));
+    /// # });
```
MajorBreakfast: Missing error example

futures-util/src/future/maybe_done.rs
```diff
@@ -44,8 +44,24 @@ pub fn maybe_done<Fut: Future>(future: Fut) -> MaybeDone<Fut> {
 }
 
 impl<Fut: Future> MaybeDone<Fut> {
+    /// Returns an optional mutable reference to the output of the future.
+    /// The output of this method will be `Some` if and only if the inner
+    /// future has been completed and [`take_output`](MaybeDone::take_output)
+    /// has not yet been called.
+    #[inline]
+    pub fn output_mut<'a>(self: PinMut<'a, Self>) -> Option<&'a mut Fut::Output> {
```
MajorBreakfast: Needs `#[allow(needless_lifetimes)] // https://github.com/rust-lang/rust/issues/52675` to make Clippy happy.

And maybe use an in-band lifetime

MajorBreakfast:
@cramertj 

I left a few comments.

BTW, what do you think about making the `join` methods free functions? It always seems so asymmetric when we use them. (Not in this PR of course)

futures-util/src/future/maybe_done.rs
Nemo157: (I’m sure I saw a post about potentially dropping in-band lifetimes by @aturon in the last few days, but can’t find it now)

futures-util/src/try_future/try_join.rs
Nemo157: Is there some reason `derive(Debug)` doesn't work for this struct, it looks like the same implementation and should be equivalent bounds?

cramertj:
> BTW, what do you think about making the join methods free functions? It always seems so asymmetric when we use them. (Not in this PR of course)

I'd be fine with that. I feel similarly. I do worry that some of the free functions are less discoverable than the methods, but I think this one is common enough that it won't be an issue.

futures-util/src/try_future/try_join.rs
cramertj: Done.

futures-util/src/try_future/try_join.rs
cramertj: Done, also updated future/join.rs

I don't think we need an `Unpin` impl-- it should be the same as the inferred one from being an auto trait.

futures-util/src/try_future/mod.rs
cramertj: Done.

futures-util/src/try_future/mod.rs
cramertj: The existing wording here is the one used elsewhere in the lib. IMO we should just remove this bit altogether, as I think it's obvious from the `self` parameter, but whatever we do should be consistent across the library.

futures-util/src/try_future/mod.rs
cramertj: I'm not sure what you're asking for here-- the docs show an example of an error occurring.

futures-util/src/try_future/try_join.rs
cramertj: These are not the same as the `derive` impls-- deriving fails as it doesn't know to put a bound on the associated `Ok` and `Error` types.

futures-util/src/try_future/mod.rs
MajorBreakfast: Ah yes, you're right. It doesn't show the tuple case. I'd like two examples one with the success value tuple and one with the error

futures-util/src/try_future/try_join.rs
MajorBreakfast: `Unpin` is still an auto trait? I thought that changed o_O

futures-util/src/try_future/mod.rs
cramertj: Done.

futures-util/src/try_future/mod.rs
cramertj: Done.

futures-util/src/try_future/mod.rs
MajorBreakfast: - The wording here needs to be in plural
- It's consistent with the other TryFuture combinators. I changed that a while back and still need to update some locations. I don't like the wording "receiving future"

MajorBreakfast: About removing it: I wouldn't be opposed. But, I suggest to keep it

futures-util/src/async_await/join.rs
```diff
@@ -47,3 +47,77 @@ macro_rules! join {
         )*)
     } }
 }
+
+/// Polls multiple futures simultaneously, returning a [`Result`] containing
+/// either a tuple of the successful outputs or an error.
```
MajorBreakfast: "return" is IMO the wrong word because the result is inside a future

futures-util/src/future/maybe_done.rs
```diff
@@ -44,8 +44,25 @@ pub fn maybe_done<Fut: Future>(future: Fut) -> MaybeDone<Fut> {
 }
 
 impl<Fut: Future> MaybeDone<Fut> {
+    /// Returns an optional mutable reference to the output of the future.
```
MajorBreakfast: ``Returns an [`Option`] containing a mutable reference...``

This currently sounds like "optionally mutable"

futures-util/src/future/maybe_done.rs
```diff
@@ -44,8 +44,25 @@ pub fn maybe_done<Fut: Future>(future: Fut) -> MaybeDone<Fut> {
...
+    /// The output of this method will be `Some` if and only if the inner
```
MajorBreakfast: ``[`Some`]``

futures-util/src/try_future/mod.rs
```diff
@@ -501,4 +504,155 @@ pub trait TryFutureExt: TryFuture {
...
+    /// #![feature(async_await, await_macro, futures_api)]
+    /// # futures::executor::block_on(async {
+    /// use futures::future::{self, TryFutureExt};
+    ///
+    /// let a = future::ready(Ok::<i32, i32>(1));
+    /// let b = future::ready(Ok::<i32, i32>(2));
+    /// let c = future::ready(Err::<i32, i32>(3));
+    /// let tuple = a.try_join3(b, c);
+    ///
+    /// assert_eq!(await!(tuple), Err(3));
```
MajorBreakfast: I think it's better to show the success case like for `Join4` and `Join5`.

futures-util/src/try_future/mod.rs
```diff
@@ -518,6 +518,25 @@ pub trait TryFutureExt: TryFuture {
     ///
     /// # Examples
     ///
+    /// When used on multiple futures that reutrn `Ok`, `try_join` will return
```
MajorBreakfast: ``[`Ok`]``

futures-util/src/try_future/try_join.rs
cramertj: No.

futures-util/src/async_await/join.rs
cramertj: Done.

futures-util/src/future/maybe_done.rs
cramertj: Done.

futures-util/src/future/maybe_done.rs
cramertj: Done.

futures-util/src/try_future/mod.rs
cramertj: Done.

futures-util/src/try_future/mod.rs
cramertj: Done.

futures-util/src/try_future/try_join.rs
MajorBreakfast: No as in "it isn't an auto trait anymore" or "it didn't change, it still is"?

PR #1157, opened 2018-08-01
Add an IntoStream wrapper around TryStream

gbonik:
Similar idea to IntoFuture.
Will use this in the implementation of try_buffer_unordered to wrap a TryStream in a Fuse

### DIFF
--- a/futures-util/src/try_stream/into_stream.rs
+++ b/futures-util/src/try_stream/into_stream.rs
@@ -1,0 +1,48 @@
+use core::mem::PinMut;
+use futures_core::stream::{Stream, TryStream};
+use futures_core::task::{self, Poll};
+
+/// Stream for the [`into_stream`](super::TryStreamExt::into_stream) combinator.
+#[derive(Debug)]
+#[must_use = "streams do nothing unless polled"]
+pub struct IntoStream<St> {
+    stream: St,
+}
+
+impl<St> IntoStream<St> {
+    unsafe_pinned!(stream: St);
+
+    #[inline]
+    pub(super) fn new(stream: St) -> Self {
+        IntoStream { stream }
+    }
+
+    /// Acquires a reference to the underlying stream that this combinator is
+    /// pulling from.
+    pub fn get_ref(&self) -> &St {
+        &self.stream
+    }
+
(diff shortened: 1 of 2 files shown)


cramertj:
Looks great! r=me once travis says okay.

futures-util/src/try_stream/into_stream.rs
```diff
@@ -0,0 +1,48 @@
+use core::mem::PinMut;
+use futures_core::stream::{Stream, TryStream};
+use futures_core::task::{self, Poll};
+
+/// Stream for the [`into_stream`](super::TryStreamExt::into_stream) combinator.
+#[derive(Debug)]
+#[must_use = "streams do nothing unless polled"]
+pub struct IntoStream<St> {
+    stream: St,
+}
```
MajorBreakfast: `impl<St: Unpin> Unpin for IntoStream<St> {}`

futures-util/src/try_stream/mod.rs
```diff
@@ -262,4 +292,5 @@ pub trait TryStreamExt: TryStream {
     {
         TryFilterMap::new(self, f)
     }
+
```
MajorBreakfast: We usually don't have padding in impl blocks.

futures-util/src/try_stream/into_stream.rs
MajorBreakfast: I just checked. The Unpin impl is missing in `IntoFuture`  as well.

@cramertj Do we need an `Unpin` impl here?

futures-util/src/try_stream/mod.rs
gbonik: Added this by accident, done

futures-util/src/try_stream/into_stream.rs
gbonik: Yeah I didn't add it since IntoFuture doesn't have one. Not sure what the practical implications are of either way.

cramertj: These Unpin impls aren't necessary since Unpin is an auto trait. You only need one in places where you want a more generous impl than the automatic one.

PR #1159, opened 2018-08-02
Smaller await

cramertj:
`async` bodies are huge right now, largely due to https://github.com/rust-lang/rust/issues/52924, but this change makes them quite a bit smaller. I've added tests so that we know when the sizes go up or down in response to changes-- these will be somewhat volatile in response to MIR changes, especially any work on https://github.com/rust-lang/rust/issues/52924, but I think it's good for us to watch and be aware of any changes with respect to the size of the generated structures.

### DIFF
--- a/futures-util/src/async_await/join.rs
+++ b/futures-util/src/async_await/join.rs
@@ -31,7 +31,7 @@ macro_rules! join {
         await!($crate::future::poll_fn(move |cx| {
             let mut all_done = true;
             $(
-                if $crate::future::Future::poll($fut.reborrow(), cx).is_pending() {
+                if $crate::core_reexport::future::Future::poll($fut.reborrow(), cx).is_pending() {
                     all_done = false;
                 }
             )*
@@ -96,7 +96,7 @@ macro_rules! try_join {
         let res: $crate::core_reexport::result::Result<_, _> = await!($crate::future::poll_fn(move |cx| {
             let mut all_done = true;
             $(
-                if $crate::future::Future::poll($fut.reborrow(), cx).is_pending() {
+                if $crate::core_reexport::future::Future::poll($fut.reborrow(), cx).is_pending() {
                     all_done = false;
                 } else if $fut.reborrow().output_mut().unwrap().is_err() {
                     // `.err().unwrap()` rather than `.unwrap_err()` so that we don't introduce
@@ -111,10 +111,11 @@ macro_rules! try_join {
             if all_done {
                 $crate::core_reexport::task::Poll::Ready(
                     $crate::core_reexport::result::Result::Ok(($(
-                    // `.ok().unwrap()` rather than `.unwrap()` so that we don't introduce
-                    // an `E: Debug` bound.
-                    $fut.reborrow().take_output().unwrap().ok().unwrap(),
-                )*)))
+                        // `.ok().unwrap()` rather than `.unwrap()` so that we don't introduce
+                        // an `E: Debug` bound.
+                        $fut.reborrow().take_output().unwrap().ok().unwrap(),
+                    )*))
+                )
             } else {
                 $crate::core_reexport::task::Poll::Pending
             }
--- a/futures-util/src/async_await/join.rs
+++ b/futures-util/src/async_await/join.rs
@@ -28,19 +28,19 @@ macro_rules! join {
             let mut $fut = $crate::future::maybe_done($fut);
             $crate::pin_mut!($fut);
         )*
-        loop {
+        await!($crate::future::poll_fn(|cx| {
             let mut all_done = true;
             $(
-                if $crate::poll!($fut.reborrow()).is_pending() {
+                if $crate::future::Future::poll($fut.reborrow(), cx).is_pending() {
                     all_done = false;
                 }
             )*
             if all_done {
-                break;
+                $crate::core_reexport::task::Poll::Ready(())
             } else {
-                $crate::pending!();
+                $crate::core_reexport::task::Poll::Pending
             }
-        }
+        }));
 
         ($(
             $fut.reborrow().take_output().unwrap(),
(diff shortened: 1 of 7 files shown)


futures-util/src/async_await/join.rs
```diff
@@ -28,23 +28,21 @@ macro_rules! join {
             let mut $fut = $crate::future::maybe_done($fut);
             $crate::pin_mut!($fut);
         )*
-        loop {
+        await!($crate::future::poll_fn(move |cx| {
             let mut all_done = true;
             $(
-                if $crate::poll!($fut.reborrow()).is_pending() {
+                if $crate::future::Future::poll($fut.reborrow(), cx).is_pending() {
```
MajorBreakfast: It's safe to use the method call syntax because we know that it's a `MaybeDone` future.

futures-util/src/async_await/join.rs
```diff
@@ -94,30 +92,34 @@ macro_rules! try_join {
             let mut $fut = $crate::future::maybe_done($fut);
             $crate::pin_mut!($fut);
         )*
-        let res: $crate::core_reexport::result::Result<(), _> = loop {
+
+        let res: $crate::core_reexport::result::Result<_, _> = await!($crate::future::poll_fn(move |cx| {
             let mut all_done = true;
             $(
-                if $crate::poll!($fut.reborrow()).is_pending() {
+                if $crate::future::Future::poll($fut.reborrow(), cx).is_pending() {
```
MajorBreakfast: Here as well

futures-util/src/async_await/join.rs
```diff
@@ -94,30 +92,34 @@ macro_rules! try_join {
...
                 }
             )*
             if all_done {
-                break $crate::core_reexport::result::Result::Ok(());
+                $crate::core_reexport::task::Poll::Ready(
+                    $crate::core_reexport::result::Result::Ok(($(
+                    // `.ok().unwrap()` rather than `.unwrap()` so that we don't introduce
+                    // an `E: Debug` bound.
+                    $fut.reborrow().take_output().unwrap().ok().unwrap(),
+                )*)))
```
MajorBreakfast: Inconsistent indentation

futures-util/src/async_await/join.rs
```diff
@@ -94,30 +92,34 @@ macro_rules! try_join {
...
             }
-        };
+        }));
 
-        res.map(|()| ($(
-            // `.ok().unwrap()` rather than `.unwrap()` so that we don't introduce
-            // an `E: Debug` bound.
-            $fut.reborrow().take_output().unwrap().ok().unwrap(),
-        )*))
+        res
```
MajorBreakfast: It's basically:
```rust
let res = /* ... */;
res
```
You could return the value from the expression above directly

futures-util/src/async_await/join.rs
MajorBreakfast: I know this. It's fine here because this type is under our control because we define it.

futures-util/src/future/mod.rs
```diff
@@ -5,7 +5,7 @@
 
 use core::marker::Unpin;
 use core::mem::PinMut;
-use futures_core::future::Future;
+pub use futures_core::future::Future;
```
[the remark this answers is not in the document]
MajorBreakfast: We shouldn't go overboard with reexports. We should try to only use reexports in our facade and in `futures_core`.

futures-util/src/async_await/select.rs
```diff
@@ -47,16 +47,18 @@ macro_rules! select {
...
         }
 
-        let __priv_res = loop {
+        let __priv_res = await!($crate::future::poll_fn(|cx| {
             $(
-                let poll_res = $crate::poll!($crate::core_reexport::mem::PinMut::new(
-                        &mut $name));
-                if let $crate::core_reexport::task::Poll::Ready(x) = poll_res {
-                    break __PrivResult::$name(x);
+                match $crate::future::Future::poll(
```
MajorBreakfast: `$crate::core_reexport::future::Future` would also work. It's IMO no different than `Poll`

futures-util/src/async_await/select.rs
```diff
@@ -47,16 +47,18 @@ macro_rules! select {
...
+                    $crate::core_reexport::mem::PinMut::new(&mut $name), cx)
+                {
+                    $crate::core_reexport::task::Poll::Ready(x) =>
+                        return $crate::core_reexport::task::Poll::Ready(__PrivResult::$name(x)),
+                    $crate::core_reexport::task::Poll::Pending => {},
```
MajorBreakfast: Since this branch is empty, you could use `if poll.is_ready() {}` instead

futures-util/src/async_await/join.rs
Nemo157: @MajorBreakfast the problem is the user _could_ introduce a new trait that has a `poll` method and implement that for `MaybeDone`. That’s highly unlikely, but the only way to guarantee no method resolution errors is to always use expanded function calls.

MajorBreakfast: Ah, I see. And it's okay for `reborrow` because it's an inherent method and those have precedence (I just [checked](http://play.rust-lang.org/?gist=c0e79343117a383e92092046f6e5ad9b&version=stable&mode=debug&edition=2015) that because I was curious)

futures-util/src/async_await/join.rs
cramertj: It's intentionally done this way in order to provide the `: Result<_, _>` annotation .

futures-util/src/future/mod.rs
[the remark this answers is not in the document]
cramertj: Done.

futures-util/src/async_await/select.rs
cramertj: Done.

futures-util/src/async_await/select.rs
cramertj: I intentionally changed it because I think it's easier to read as a `match` since all of the paths included are so long.

futures-util/src/async_await/join.rs
cramertj: Done.

futures-util/src/async_await/join.rs
cramertj: Yup, as others said we can't assume that the `Future` trait is in-scope or that other methods called `poll` haven't been provided via other traits. `.reborrow()` is fine because it's an inherent method.

futures-util/src/async_await/select.rs
MajorBreakfast: I'd have introduced a `poll` variable. But `match` is also fine

PR #1160, opened 2018-08-02
Prevent lifetime name collisions

MajorBreakfast:


futures-util/src/try_stream/try_next.rs
```diff
@@ -15,8 +15,8 @@ pub struct TryNext<'a, St: 'a> {
 
 impl<'a, St: TryStream + Unpin> Unpin for TryNext<'a, St> {}
 
-impl<'b, St: TryStream + Unpin> TryNext<'b, St> {
-    pub(super) fn new(stream: &'b mut St) -> Self {
+impl<'a, St: TryStream + Unpin> TryNext<'a, St> {
+    pub(super) fn new(stream: &'a mut St) -> Self {
```
MajorBreakfast: Oddly enough this did even clash because the pin macros aren't used in this combinator. However, it would clash if they were used (I tried).

PR #1162, opened 2018-08-02
Extend compatibility to Stream and Sink

Nemo157:
Sink is unidirectional because `Sink01::start_send` was split into `Sink03::poll_ready` + `Sink03::start_send` which makes tracking the context across the two method calls painful (might be possible with a custom `Compat` struct for this case, but I didn't investigate much).

### DIFF
--- a/futures-util/src/compat/compat.rs
+++ b/futures-util/src/compat/compat.rs
@@ -1,20 +1,20 @@
-/// Converts a futures 0.3 `TryFuture` into a futures 0.1 `Future`
-/// and vice versa.
+/// Converts a futures 0.3 `TryFuture`, `TryStream` or `Sink` into a futures 0.1
+/// `Future` and vice versa.
 #[derive(Debug)]
 #[must_use = "futures do nothing unless polled"]
-pub struct Compat<Fut, Ex> {
-    crate future: Fut,
+pub struct Compat<T, Ex> {
+    crate inner: T,
     crate executor: Option<Ex>,
 }
 
-impl<Fut, Ex> Compat<Fut, Ex> {
-    /// Returns the inner future.
-    pub fn into_inner(self) -> Fut {
-        self.future
+impl<T, Ex> Compat<T, Ex> {
+    /// Returns the inner item.
+    pub fn into_inner(self) -> T {
+        self.inner
     }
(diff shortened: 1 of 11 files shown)


futures-util/src/sink/mod.rs
```diff
@@ -248,4 +254,14 @@ pub trait SinkExt: Sink {
     {
         Either::Right(self)
     }
+
+    /// Wraps a [`Sink`] into a sink compatible with libraries using
+    /// futures 0.1 `Sink`. Requires the `compat` feature to enable.
```
MajorBreakfast: to be enabled

MajorBreakfast:
Some lines are longer than 80 chars

futures-util/src/compat/compat.rs
```diff
@@ -1,20 +1,20 @@
-/// Converts a futures 0.3 `TryFuture` into a futures 0.1 `Future`
-/// and vice versa.
+/// Converts a futures 0.3 `TryFuture`, `TryStream` or `Sink` into a futures 0.1
+/// `Future` and vice versa.
```
MajorBreakfast: These should be links

MajorBreakfast: Ah never mind. Let's do the docs stuff in another PR

Nemo157: Yeah, I started doing it then decided to just leave it for another pass.

Nemo157:
Oh, are we using an 80 character limit? I just applied `rustfmt` cause there was a lot of copy-paste + renaming and I couldn't be bothered formatting manually. I can add a `.rustfmt.toml` limit it in that case.

MajorBreakfast:
>  I can add a .rustfmt.toml limit it in that case.

@Nemo157 That'd be great!

cramertj:
RE rustfmt: be aware that for `async` blocks and functions you'll need a recent version that includes https://github.com/rust-lang-nursery/rustfmt/pull/2874

Nemo157:
I was only applying it to a couple of files so luckily didn’t hit that, did discover https://github.com/rust-lang-nursery/rustfmt/issues/2887 on my next change though 😢

PR #1163, opened 2018-08-02
Remove unnecessary locals in join! macros

cramertj:


MajorBreakfast:
Doesn't that remove support for `!Unpin` futures?

MajorBreakfast:
Ah no I see. You poll a reference.

cramertj:
@MajorBreakfast lol, good catch-- meant to use `new_unchecked`. This is safe because we're moving the future into a local that isn't accessible by the end user.

MajorBreakfast:
Hehe ^^' I shouldn't have posted the second comment.

Maybe also add a comment to the `let` "Move future into local"

cramertj:
I've added some tests to ensure that we continue to allow `!Unpin` futures.

PR #1164, opened 2018-08-02
Tokio compatibility

Nemo157:
I have tested the example in an external project, I'm not sure what to do about having it actually tested since you can't have optional dev-dependencies and I really don't want to bring in all of `tokio` normally.


futures-util/src/compat/tokio.rs
```diff
@@ -0,0 +1,71 @@
...
+};
+use tokio_executor::{DefaultExecutor, Executor as _};
+
+/// An executor that delegates to `tokio`'s [`DefaultExecutor`], will panic if
+/// used in the context of a task that is not running on `tokio`'s executor.
+///
+/// # Examples
+///
+/// ```ignore
+/// #![feature(async_await, await_macro, underscore_imports, pin)]
```
MajorBreakfast: Is the `underscore_imports` feature required?

futures-util/src/compat/tokio.rs
```diff
@@ -0,0 +1,71 @@
...
+///     future::TryFutureExt as _,
+///     executor::spawn,
+///     compat::TokioDefaultExecutor,
+/// };
+///
+/// let (sender, receiver) = oneshot::channel::<i32>();
+///
+/// let handle = thread::spawn(move || {
+///     futures::executor::block_on(async move {
+///         await!(receiver).unwrap()
```
MajorBreakfast: Why not `block_on(receiver).unwrap()` at the bottom? Does it have to be in another thread?

futures-util/src/compat/tokio.rs
```diff
@@ -0,0 +1,71 @@
...
+/// #![feature(async_await, await_macro, underscore_imports, pin)]
+///
+/// use std::thread;
+/// use std::boxed::PinBox;
+///
+/// use futures::{
+///     channel::oneshot,
+///     future::FutureExt as _,
+///     future::TryFutureExt as _,
+///     executor::spawn,
```
MajorBreakfast: spawn is now macro. See my [blog post PR](https://github.com/rust-lang-nursery/futures-rs/pull/1165) for a quick intro

futures-util/src/compat/tokio.rs
```diff
@@ -0,0 +1,71 @@
...
+///
+/// # Examples
+///
+/// ```ignore
+/// #![feature(async_await, await_macro, underscore_imports, pin)]
+///
+/// use std::thread;
+/// use std::boxed::PinBox;
+///
+/// use futures::{
```
MajorBreakfast: This style is inconsistent with our other doc examples

futures-util/src/compat/tokio.rs
Nemo157: Nope, I just try and use minor unstable features to give them a good testing. I’ll remove it.

futures-util/src/compat/tokio.rs
Nemo157: Changed

futures-util/src/compat/tokio.rs
```diff
@@ -0,0 +1,74 @@
...
+
+impl Executor for TokioDefaultExecutor {
+    fn spawn_obj(
+        &mut self,
+        task: FutureObj<'static, ()>,
+    ) -> Result<(), SpawnObjError> {
+        let fut = Box::new(task.unit_error().compat(*self));
+        DefaultExecutor::current().spawn(fut).map_err(|err| {
+            panic!(
+                "tokio failed to spawn and doesn't return the future: {:?}",
```
Nemo157: Here's another argument for not returning the future when spawning fails, I can't see any way to do anything other than panic here.

Thomasdezeeuw:
Shouldn't this be part of the tokio project, rather then futures project?

Nemo157:
That's definitely a good question. I'm mostly concerned about getting something up and running ASAP so that we can start testing futures 0.3 in a wider context.

Maybe we can add this here for now, with a note that it will likely not make it to the 0.3 release, and look into adding a compatibility feature into tokio instead for the release?

Thomasdezeeuw:
Ok, looks good to me.

futures-util/src/compat/tokio.rs
```diff
@@ -0,0 +1,79 @@
...
+/// deprecated before or soon after the initial 0.3 release and moved to a
+/// feature in `tokio` instead.
+///
+/// # Examples
+///
+/// ```ignore
+/// #![feature(async_await, await_macro, futures_api, pin, use_extern_macros)]
+///
+/// use std::thread;
+/// use std::boxed::PinBox;
```
MajorBreakfast: Can you put these two after `use futures::spawn`? (single import block)

futures-util/src/compat/tokio.rs
MajorBreakfast: I think that's fine

MajorBreakfast: We can always change it if find a better solution 🙂

futures-util/src/compat/tokio.rs
MajorBreakfast: Ah never mind. I'll change it myself 🙂

MajorBreakfast:
I agree that we should get things going ASAP and then move the stuff around until we're happy. Maybe we'll end up putting it into its own crate. Who knows :)

Nemo157:
(What is up with the CI jobs hanging so often?)

PR #1165, opened 2018-08-02
Alpha 3 blog post

MajorBreakfast:
Do not merge.

[rendered](https://github.com/MajorBreakfast/futures-rs/blob/blog-alpha-3/_posts/2018-01-01-futures-0.3.0-alpha.3.md)

Should we make this a standalone blog post or include it in the alpha 3 announcement? (Let's not rush the release of alpha 3, though)

### DIFF
--- a/_posts/2018-01-01-futures-0.3.0-alpha.3.md
+++ b/_posts/2018-01-01-futures-0.3.0-alpha.3.md
@@ -25,13 +25,13 @@ The `std::task::Executor` trait is now called `std::task::Spawn`. The purpose of
 As a consequence of this renaming, some functions on `Context` were adjusted:
 ```rust
 impl<'a> Context<'a> {
-    pub fn new(local_waker: &'a LocalWaker, spawner: &'a mut dyn Spawn) -> Context<'a> { /* --snip-- */ }
+    pub fn new(local_waker: &'a LocalWaker, spawner: &'a mut dyn Spawn) -> Context<'a> { ... }
 
-    pub fn spawner(&mut self) -> &mut dyn Spawn { /* --snip-- */ }
+    pub fn spawner(&mut self) -> &mut dyn Spawn { ... }
 
-    pub fn with_spawner<'b, Sp: Spawn>(&'b mut self, spawner: &'b mut Sp) -> Context<'b> { /* --snip-- */ }
+    pub fn with_spawner<'b, Sp: Spawn>(&'b mut self, spawner: &'b mut Sp) -> Context<'b> { ... }
 
-    // --snip--
+    // ...
 }
 ```
 
@@ -113,7 +113,7 @@ The pin macros now live in their own crate, the `pin-utils` crate. The macros ar
 The crate includes the macros for (un)pin projections `unsafe_pinned!` and `unsafe_unpinned!` and the macro for stack pinning `pin_mut!`:
 
 ```rust
-let foo = Foo { /* ... */ };
+let foo = Foo { ... };
 pin_mut!(foo); // PinMut<Foo>
 ```
 
--- a/_posts/2018-01-01-futures-0.3.0-alpha.3.md
+++ b/_posts/2018-01-01-futures-0.3.0-alpha.3.md
@@ -14,13 +14,13 @@ $ rustup update
 
 ## Compatibility layer
 
-A compatibility layer between 0.3 an 0.1 was developped. It is now possible to convert an 0.3 future into an 0.1 future and vice versa. Similar conversions for streams and sinks are also supported. Additionally it is now possible to run 0.3 futures and async functions on Tokio's executor. We have a dedicated blog post coming up that explains this in more detail.
+A compatibility layer between 0.3 an 0.1 was developed. It is now possible to convert an 0.3 future into an 0.1 future and vice versa. Similar conversions for streams and sinks are also supported. Additionally, it is now possible to run 0.3 futures and async functions on Tokio's executor. We have a dedicated blog post coming up that explains this in more detail.
 
 ## Spawning
 
 ### `Spawn` trait
 
-The `std::task::Executor` trait is now called `std::task::Spawn`. The purpose of the trait remains unchanged - it is to provide a mechanism for spawning tasks. We noticed that in many implementations the responsibilty of spawning tasks is handled by a type that merely sends the task to an associated executor, not by the task executor type itself. Furthermore, it became clear that the former trait name "`Executor`" didn't properly capture the essence of what it means to be an executor. Fundamentally, a task executor is a type that manages the execution of tasks. There exists no fixed set of methods that a trait could define. Instead, each task executor type offers different methods that make sense for its design goals. This is why we chose to rename the trait to `Spawn` to reflect that it merely defines a method for spawning tasks and makes no specifications about how to run them.
+The `std::task::Executor` trait is now called `std::task::Spawn`. We chose to rename the trait to `Spawn` because the functionality exposed by this trait is not the ability to execute, but the ability to spawn.
 
 As a consequence of this renaming, some functions on `Context` were adjusted:
 ```rust
@@ -48,7 +48,7 @@ use futures::spawn;
 spawn!(future).unwrap();
 ```
 
-`spawn_with_handle!` lets you spawn futures with any `Output: Send` type and also lets you await the produced output value:
+`spawn_with_handle!` lets you spawn futures with any `Output: Send` type and returns a future that resolves to the output of the spawned future:
 
 ```rust
 use futures::spawn_with_handle;
@@ -62,10 +62,10 @@ let output = await!(join_handle);
 Should spawning fail, a `SpawnError` error is returned which contains information about why it failed.
 
 So what futures can be spawned? There are some requirements: The future needs to be `Send` and `'static` and the output needs to be `Send`. Here's why:
-- The `'static` lifetime means that the future cannot contain any non-static (exterior) references. This is required so that the spawned task can be run independently to the task that spawns it.
-- The `Send` bounds mean that the future and its output can be safely sent to other operating system threads. This is required so that multithreaded executors can run the spawned task on any operating thread they choose.
+- The `'static` lifetime means that the future cannot contain any non-static references. This is required so that the spawned task can be run independently to the task that spawns it.
+- The `Send` bounds mean that the future and its output can be sent to other threads. This is required so that multithreaded executors can run the spawned task on any thread they choose.
 
-Note: You can execute non-`Send`, non-`'static` futures concurrently via the `join!` macro.
+Note: You can still execute non-`Send`, non-`'static` futures concurrently via the `join!` macro.
 
 ### Spawn methods
 
--- a/_posts/2018-01-01-futures-0.3.0-alpha.3.md
+++ b/_posts/2018-01-01-futures-0.3.0-alpha.3.md
@@ -18,7 +18,7 @@ This release introduces the `spawn!` and `spawn_with_handle!` macros for use in 
 spawn!(future).unwrap();
 ```
 
-`spawn_with_handle!` lets you spawn futures with any `Output` type and also lets you await the output:
+`spawn_with_handle!` lets you spawn futures with any `Output: Send` type and also lets you await the produced output value:
 
 ```rust
 let join_handle = spawn_with_handle!(future).unwrap();
--- a/_posts/2018-01-01-futures-0.3.0-alpha.3.md
+++ b/_posts/2018-01-01-futures-0.3.0-alpha.3.md
@@ -29,7 +29,7 @@ impl<'a> Context<'a> {
 
     pub fn spawner(&mut self) -> &mut dyn Spawn { /* --snip-- */ }
 
-    pub fn with_spawner<'b, Sp: Spawn>(&'b mut self, spawner: &'b mut Sp) -> Context<'b>
+    pub fn with_spawner<'b, Sp: Spawn>(&'b mut self, spawner: &'b mut Sp) -> Context<'b> { /* --snip-- */ }
 
     // --snip--
 }
--- a/_posts/2018-01-01-futures-0.3.0-alpha.3.md
+++ b/_posts/2018-01-01-futures-0.3.0-alpha.3.md
@@ -7,7 +7,7 @@ date:   2018-01-01
 categories: blog
 ---
 
-This release introduces the easy to use `spawn!` and `spawn_with_handle!` macros for use in async functions, closures and blocks. Both let you spawn a task that polls the given future to completion. The task is spawned onto the current context's executor.
+This release introduces the `spawn!` and `spawn_with_handle!` macros for use in async functions, closures and blocks. Both macros let you spawn a task that polls the given future to completion. The task is spawned onto the current context's executor.
 
 `spawn!` lets you spawn futures with `Output = ()`:
 
@@ -68,4 +68,4 @@ This method can't be implemented just yet due to compiler limitations:
 - Arbitrary self types are currently not object-safe: `Future::poll` uses `PinMut` as arbitrary self type. This unfortunately makes it impossible to create trait objects for `Future`. Therefore, `Box<dyn Future>` and `&dyn Future` are *currently* not supported.
 - Unsized rvalues are not yet available: Trait objects (`dyn Trait`) are `!Sized` because the size of a trait object depends on the size of the type inside it and can therfore only be known at runtime. Currently we often use `Box<dyn Trait>` which stores the trait object on the heap to circumvent that. The unsized rvalues feature will make it possible to store `!Sized` types on the stack.
 
-To circumvent these limitations, we currently have `FutureObj` and `Executor::spawn_obj`. These are, however, just a temporary solution until we can introduce the planned `spawn` method. The planned `spawn` method will make it possible to have one heap allocation less because the future is kept on the stack for as long as possible.
+To circumvent these limitations, we currently have `FutureObj` and `Executor::spawn_obj`. These are, however, just a temporary solution until we can introduce the planned `spawn` method. The planned `spawn` method will make it possible to have one fewer heap allocation because the future is kept on the stack for as long as possible.
--- a/_posts/2018-01-01-futures-0.3.0-alpha.3.md
+++ b/_posts/2018-01-01-futures-0.3.0-alpha.3.md
@@ -7,9 +7,14 @@ date:   2018-01-01
 categories: blog
 ---
 
-## `pin-utils` crate
+`0.3.0-alpha.3` requires a recent nightly (2018-08-xx or newer):
+```sh
+$ rustup update
+```
 
 ## Compatibility layer
+
+A compatibility layer between 0.3 an 0.1 was developped. It is now possible to convert an 0.3 future into an 0.1 future and vice versa. Similar conversions for streams and sinks are also supported. Additionally it is now possible to run 0.3 futures and async functions on Tokio's executor. We have a dedicated blog post coming up that explains this in more detail.
 
 ## Spawning
 
@@ -77,7 +82,7 @@ Inside `poll`, you can use these methods by accessing the context's spawner via 
 
 ```rust
 fn poll(self: PinMut<Self>, cx: &mut task::Context) -> Poll<T> {
-    cx.spawner().spawn(future);
+    cx.spawner().spawn(future).unwrap();
 }
 ```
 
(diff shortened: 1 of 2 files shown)


_posts/2018-01-01-futures-0.3.0-alpha.3.md
```diff
@@ -0,0 +1,71 @@
...
+#![feature(async_await, await_macro, futures_api)]
+#[macro_use] extern crate futures;
+
+spawn!(future).unwrap();
+```
+
+`spawn_with_handle!` lets you spawn futures with any `Output` type and also lets you await the output:
+
+```rust
+let join_handle = spawn_with_handle!(future).unwrap();
```
tinaun: i don't know if this is necessary in this post, but i would love to have documentation explaining what makes calling `spawn_with_handle!` and awaiting on the handle different than awaiting the future directly, with usecases explaining when to use one or the other.

_posts/2018-01-01-futures-0.3.0-alpha.3.md
```diff
@@ -0,0 +1,71 @@
...
+    &mut self,
+    future: dyn Future<Output = ()> + Send + 'static,
+) -> Result<(), SpawnError>
+```
+
+This method can't be implemented just yet due to compiler limitations:
+- Arbitrary self types are currently not object-safe: `Future::poll` uses `PinMut` as arbitrary self type. This unfortunately makes it impossible to create trait objects for `Future`. Therefore, `Box<dyn Future>` and `&dyn Future` are *currently* not supported.
+- Unsized rvalues are not yet available: Trait objects (`dyn Trait`) are `!Sized` because the size of a trait object depends on the size of the type inside it and can therfore only be known at runtime. Currently we often use `Box<dyn Trait>` which stores the trait object on the heap to circumvent that. The unsized rvalues feature will make it possible to store `!Sized` types on the stack.
+
+To circumvent these limitations, we currently have `FutureObj` and `Executor::spawn_obj`. These are, however, just a temporary solution until we can introduce the planned `spawn` method. The planned `spawn` method will make it possible to have one heap allocation less because the future is kept on the stack for as long as possible.
```
tinaun: i'd prefer "one fewer heap allocation".

_posts/2018-01-01-futures-0.3.0-alpha.3.md
```diff
@@ -0,0 +1,71 @@
+---
+layout: post
+title:  "Futures 0.3.0-alpha.3"
+author: "Josef Brandl"
+author_github: "MajorBreakfast"
+date:   2018-01-01
+categories: blog
+---
+
+This release introduces the easy to use `spawn!` and `spawn_with_handle!` macros for use in async functions, closures and blocks. Both let you spawn a task that polls the given future to completion. The task is spawned onto the current context's executor.
```
tinaun: get rid of "easy to use"

tinaun:
i think we should have one blog post, with both the Compat stuff and the spawn! macros. this by itself feels a little light

_posts/2018-01-01-futures-0.3.0-alpha.3.md
MajorBreakfast: `spawn_with_handle!` uses a oneshot channel to send the value back. `spawn!` doesn't do that. Hence, it's a bit cheaper.

_posts/2018-01-01-futures-0.3.0-alpha.3.md
MajorBreakfast: Done

_posts/2018-01-01-futures-0.3.0-alpha.3.md
MajorBreakfast: Done

MajorBreakfast:
Alright, then I have to get multi author posts working ^^'

_posts/2018-01-01-futures-0.3.0-alpha.3.md
```diff
@@ -0,0 +1,73 @@
...
+`spawn!` lets you spawn futures with `Output = ()`:
+
+```rust
+#![feature(async_await, await_macro, futures_api)]
+#[macro_use] extern crate futures;
+
+spawn!(future).unwrap();
+```
+
+`spawn_with_handle!` lets you spawn futures with any `Output` type and also lets you await the output:
```
Nemo157: Maybe mention the `Output: Send` requirement?

_posts/2018-01-01-futures-0.3.0-alpha.3.md
```diff
@@ -0,0 +1,73 @@
...
+
+```rust
+let join_handle = spawn_with_handle!(future).unwrap();
+let output = await!(join_handle);
+```
+
+`spawn!` is slightly cheaper than `spawn_with_handle!`.
```
Nemo157: I think this sentence seems a bit throwaway without any explanation, it could be something really simple like

> `spawn!` is slightly cheaper than `spawn_with_handle!` since it doesn't need to provide a way to get the output.

_posts/2018-01-01-futures-0.3.0-alpha.3.md
Nemo157: I think @tinaun was referring to why would you choose one or the other of the following:

```rust
let value = await!(spawn_with_handle!(future).unwrap());
let value = await!(future);
```

I think the answer is that you would always choose the second, `spawn_with_handle!` only really makes sense to use if you are then going to do something concurrently, e.g.

```rust
let first_thing = spawn_with_handle!(future).unwrap();
let second_thing = await!(other_future);
let first_thing = await!(first_thing);
```

I agree that it would be good to include this in the post somehow.

_posts/2018-01-01-futures-0.3.0-alpha.3.md
Nemo157: Oh, just noticed that this is below, not sure if it needs repeating here then, but I missed it during my first reading.

_posts/2018-01-01-futures-0.3.0-alpha.3.md
```diff
@@ -0,0 +1,73 @@
...
+```rust
+fn spawn(
+    &mut self,
+    future: dyn Future<Output = ()> + Send + 'static,
+) -> Result<(), SpawnError>
+```
+
+This method can't be implemented just yet due to compiler limitations:
+- Arbitrary self types are currently not object-safe: `Future::poll` uses `PinMut` as arbitrary self type. This unfortunately makes it impossible to create trait objects for `Future`. Therefore, `Box<dyn Future>` and `&dyn Future` are *currently* not supported.
+- Unsized rvalues are not yet available: Trait objects (`dyn Trait`) are `!Sized` because the size of a trait object depends on the size of the type inside it and can therefore only be known at runtime. Currently we often use `Box<dyn Trait>` which stores the trait object on the heap to circumvent that. The unsized rvalues feature will make it possible to store `!Sized` types on the stack.
```
Nemo157: I don't feel that "the type inside it" is really correct in "the size of a trait object depends on the size of the type inside it", trait objects aren't really wrappers. Maybe something like "the size of a trait object depends on the size of the concrete type behind it"

_posts/2018-01-01-futures-0.3.0-alpha.3.md
```diff
@@ -0,0 +1,73 @@
...
+
+To circumvent these limitations, we currently have `FutureObj` and `Executor::spawn_obj`. These are, however, just a temporary solution until we can introduce the planned `spawn` method. The planned `spawn` method will make it possible to have one fewer heap allocation because the future is kept on the stack for as long as possible.
```
Nemo157: Although I would hope it most cases it never touches the stack and is directly written to the heap allocation. If place expressions were anywhere close to a thing I'd recommend looking at ensuring this were forward compatible with them, but they're so far off at this point there's not much point, maybe it'll be possible to retrofit onto `Executor` if they ever appear.

_posts/2018-01-01-futures-0.3.0-alpha.3.md
MajorBreakfast: Good idea. Mentioning it twice doesn't hurt.

Ich changed it to:
`spawn_with_handle!` lets you spawn futures with any `Output: Send` type and also lets you await the produced output value

_posts/2018-01-01-futures-0.3.0-alpha.3.md
MajorBreakfast: They're not? Aren't `dyn Trait` basically the following?
```rust
pub struct TraitObject {
    vtable: *mut (),
    data: Type,
}
```
If yes, then my formulation should work. But, yours works as well, so I'm going to change it.

_posts/2018-01-01-futures-0.3.0-alpha.3.md
MajorBreakfast: Sounds good.

_posts/2018-01-01-futures-0.3.0-alpha.3.md
MajorBreakfast: I left this untouched. GitHub just hides it because updated the other thing near it.

I'm happy to change this if a better explanation can be found.

_posts/2018-01-01-futures-0.3.0-alpha.3.md
Nemo157: Except in cases other than unsized rvalues the vtable is split from the data out into the pointer type, e.g. `Box<dyn Trait>` is similar to

```
struct Box<dyn Trait> {
    vtable: *mut (),
    data: *mut Type,
}
```

so a "trait object" isn't a real construction in the type system, rather more like an idea that is implemented differently depending on the wrapper type. Once there are unsized rvalues I assume they'll have a construction similar to your example, but that's specifically the unsized rvalue case, not the "trait object" itself. One alternative name I've seen for unsized rvalues is "bare trait object", but that seems to have been co-opted by the lint for renaming `Trait -> dyn Trait`, so I'm not sure if there's a nice way to refer specifically to "unboxed trait objects" now.

MajorBreakfast:
I've updated the blog post. We can now either include @tinaun's upcoming compatibility layer blog post in this post or we can publish it as a standalone post.

@Nemo157 @cramertj please review the newly added sections

_posts/2018-01-01-futures-0.3.0-alpha.3.md
```diff
@@ -0,0 +1,121 @@
...
+    pub fn new(local_waker: &'a LocalWaker, spawner: &'a mut dyn Spawn) -> Context<'a> { ... }
+
+    pub fn spawner(&mut self) -> &mut dyn Spawn { ... }
+
+    pub fn with_spawner<'b, Sp: Spawn>(&'b mut self, spawner: &'b mut Sp) -> Context<'b>
+
+    ...
+}
+```
+### Spawn macros
```
Nemo157: Should have an empty line above a heading (currently breaks github's syntax highlighting, but may work depending on the parser, nicer with the space when reading the source anyway)

_posts/2018-01-01-futures-0.3.0-alpha.3.md
```diff
@@ -0,0 +1,121 @@
...
+
+## Spawning
+
+### `Spawn` trait
+
+The `std::task::Executor` trait is now called `Spawn`. The purpose of the trait remains unchanged - it is to provide a mechanism for spawning tasks. We noticed that in many implementations the responsibilty of spawning tasks is handled by a type that merely sends the task to an associated executor, not by the task executor type itself. Furthermore, it became clear that the former trait name "`Executor`" didn't properly capture the essence of what it means to be an executor. Fundamentally, a task executor is a type that manages the execution of tasks. There exists no fixed set of methods that a trait could define. Instead, each task executor type offers different methods that make sense for its design goals. This is why we chose to rename the trait to `Spawn` to reflect that it merely defines a method for spawning tasks and makes no specifications about how to run them.
```
Nemo157: Maybe use

    The `std::task::Executor` trait was renamed to `std::task::Spawn`.

to highlight that it remains in the same module.

_posts/2018-01-01-futures-0.3.0-alpha.3.md
```diff
@@ -0,0 +1,121 @@
...
+```
+### Spawn macros
+
+This release of futures-rs introduces the `spawn!` and `spawn_with_handle!` macros for use in async functions, closures and blocks. Both macros let you spawn a task that polls the given future to completion. The task is spawned via the current context's spawner.
+
+`spawn!` lets you spawn futures with `Output = ()`:
+
+```rust
+#![feature(async_await, await_macro, futures_api)]
+#[macro_use] extern crate futures;
```
Nemo157: I thought we killed off `#[macro_use]`? 😉

MajorBreakfast: Yeah we did. With 🔥

(I still need to adjust the pin-utils doc examples)

_posts/2018-01-01-futures-0.3.0-alpha.3.md
```diff
@@ -0,0 +1,124 @@
...
+---
+
+`0.3.0-alpha.3` requires a recent nightly (2018-08-xx or newer):
+```sh
+$ rustup update
+```
+
+## Compatibility layer
+
+A compatibility layer between 0.3 an 0.1 was developped. It is now possible to convert an 0.3 future into an 0.1 future and vice versa. Similar conversions for streams and sinks are also supported. Additionally it is now possible to run 0.3 futures and async functions on Tokio's executor. We have a dedicated blog post coming up that explains this in more detail.
```
cramertj: s/developped/developed

cramertj: Additionally[,]

_posts/2018-01-01-futures-0.3.0-alpha.3.md
```diff
@@ -0,0 +1,124 @@
...
+
+## Spawning
+
+### `Spawn` trait
+
+The `std::task::Executor` trait is now called `std::task::Spawn`. The purpose of the trait remains unchanged - it is to provide a mechanism for spawning tasks. We noticed that in many implementations the responsibilty of spawning tasks is handled by a type that merely sends the task to an associated executor, not by the task executor type itself. Furthermore, it became clear that the former trait name "`Executor`" didn't properly capture the essence of what it means to be an executor. Fundamentally, a task executor is a type that manages the execution of tasks. There exists no fixed set of methods that a trait could define. Instead, each task executor type offers different methods that make sense for its design goals. This is why we chose to rename the trait to `Spawn` to reflect that it merely defines a method for spawning tasks and makes no specifications about how to run them.
```
cramertj: s/it is to provide/it provides

cramertj: I'd remove "There exists no fixed set of methods that a trait could define." We could come up with such a trait, but it doesn't exist. In any case, this isn't what the `Executor` trait was. The functionality exposed by the `Executor` trait was not the ability to `execute`, but the ability to `spawn`.

_posts/2018-01-01-futures-0.3.0-alpha.3.md
```diff
@@ -0,0 +1,124 @@
...
+`spawn!` lets you spawn futures with `Output = ()`:
+
+```rust
+#![feature(async_await, await_macro, futures_api)]
+use futures::spawn;
+
+spawn!(future).unwrap();
+```
+
+`spawn_with_handle!` lets you spawn futures with any `Output: Send` type and also lets you await the produced output value:
```
cramertj: s/lets you await the produced output value/returns a future of the result of the spawned task

_posts/2018-01-01-futures-0.3.0-alpha.3.md
```diff
@@ -0,0 +1,124 @@
...
+let join_handle = spawn_with_handle!(future).unwrap();
+let output = await!(join_handle);
+```
+
+`spawn!` is slightly cheaper than `spawn_with_handle!` since it doesn't need to provide a way to get the output.
+
+Should spawning fail, a `SpawnError` error is returned which contains information about why it failed.
+
+So what futures can be spawned? There are some requirements: The future needs to be `Send` and `'static` and the output needs to be `Send`. Here's why:
+- The `'static` lifetime means that the future cannot contain any non-static (exterior) references. This is required so that the spawned task can be run independently to the task that spawns it.
```
cramertj: I'd remove "exterior" here-- I don't think it adds anything. Rust users know what `'static` is.

_posts/2018-01-01-futures-0.3.0-alpha.3.md
```diff
@@ -0,0 +1,124 @@
...
+- The `Send` bounds mean that the future and its output can be safely sent to other operating system threads. This is required so that multithreaded executors can run the spawned task on any operating thread they choose.
```
cramertj: s/that the future is and its output can be safely sent to other operating system threads/the future and its output can be sent to other threads

cramertj: s/on any operating thread/on any thread

_posts/2018-01-01-futures-0.3.0-alpha.3.md
```diff
@@ -0,0 +1,124 @@
...
+
+Note: You can execute non-`Send`, non-`'static` futures concurrently via the `join!` macro.
```
cramertj: can -still- execute

(clarify this was existing behavior)

_posts/2018-01-01-futures-0.3.0-alpha.3.md
```diff
@@ -0,0 +1,124 @@
...
+) -> Result<(), SpawnError>
+```
+
+This method can't be implemented just yet due to compiler limitations:
+- Arbitrary self types are currently not object-safe: `Future::poll` uses `PinMut` as arbitrary self type. This unfortunately makes it impossible to create trait objects for `Future`. Therefore, `Box<dyn Future>` and `&dyn Future` are *currently* not supported.
+- Unsized rvalues are not yet available: Trait objects (`dyn Trait`) are `!Sized` because the size of a trait object depends on the size of the concrete type behind it and can therefore only be known at runtime. Currently we often use `Box<dyn Trait>` which stores the trait object on the heap to circumvent that. The unsized rvalues feature will make it possible to store `!Sized` types on the stack.
+
+To circumvent these limitations, we currently have `FutureObj` and `Spawn::spawn_obj`. These are, however, just a temporary solution until we can introduce the planned `spawn` method. The planned `spawn` method will make it possible to have one fewer heap allocation because the future is kept on the stack for as long as possible.
+
+## `pin-utils` crate
```
cramertj: The other sections had ###, but this one has ### --was the different header levels intentional?

MajorBreakfast: Yes. The previous sections were subsections of "spawning". This is on the top level again, the same as "compatibility layer" and "spawning".

_posts/2018-01-01-futures-0.3.0-alpha.3.md
MajorBreakfast: Done.

What I wanted to say is that it could contain interior references and still be `'static`. The future doesn't at that stage, though, because it isn't pinned yet. So you're right the distinction isn't helpful.

_posts/2018-01-01-futures-0.3.0-alpha.3.md
MajorBreakfast: I've shortened this paragraph

_posts/2018-01-01-futures-0.3.0-alpha.3.md
MajorBreakfast: Change it to

> `spawn_with_handle!` lets you spawn futures with any `Output: Send` type and returns a future that resolves to the output of the spawned future:

_posts/2018-01-01-futures-0.3.0-alpha.3.md
MajorBreakfast: Done

_posts/2018-01-01-futures-0.3.0-alpha.3.md
MajorBreakfast: Done

_posts/2018-01-01-futures-0.3.0-alpha.3.md
MajorBreakfast: Done

MajorBreakfast: Done

_posts/2018-01-01-futures-0.3.0-alpha.3.md
MajorBreakfast: Done

_posts/2018-01-01-futures-0.3.0-alpha.3.md
cramertj: Gotcha, thanks for clarifying.

Issue #1171, opened 2018-08-03
Eager Complete Select


dbcfd:
Select should provide an eager complete functionality for situations where you join to another stream which may never produce a value, such as a stream of signals to interrupt the data stream.

Alternatively, could introduce another stream operator eager_select and its combinator EagerSelect.

dbcfd:
Some kind of real world examples I can think of

- Stream of signals with a stream of data, stream of data completes first, stream of signals may never complete
- Stream of events, selecting a periodic timestamp event stream, interrupting either stream should cause the stream to complete

Both of these are usually desired with a system like kafka, where you consume from an infinite stream, but occasionally interrupt the kafka stream. You'd like the drive the stream to completion so that offsets can be committed based on what was pulled off of kafka (attempted exactly once rather than at least once).

The alternative solution is to interrupt every provider of streams being selected (assuming that functionality is available), but then your ability to finish is a result of when things are interrupted and how they drain.

PR #1172, opened 2018-08-03
Add Eager Select

dbcfd:
Add EagerSelect combinator, and eager_select stream operation. Will perform the same functionality as select, until one of the streams completes, at which point the combinator will complete.

Addresses https://github.com/rust-lang-nursery/futures-rs/issues/1171

### DIFF
--- a/futures-util/src/stream/mod.rs
+++ b/futures-util/src/stream/mod.rs
@@ -71,6 +71,9 @@ pub use self::poll_fn::{poll_fn, PollFn};
 mod select;
 pub use self::select::Select;
 
+mod eager_select;
+pub use self::eager_select::EagerSelect;
+
 mod skip;
 pub use self::skip::Skip;
 
@@ -873,6 +876,21 @@ pub trait StreamExt: Stream {
               Self: Sized,
     {
         Select::new(self, other)
+    }
+
+    /// Creates a stream that selects the next element from either this stream
+    /// or the provided one, whichever is ready first.
+    ///
+    /// This combinator will attempt to pull items from both streams. Each
+    /// stream will be polled in a round-robin fashion, and whenever a stream is
+    /// ready to yield an item that item is yielded.
+    ///
+    /// This will complete once one of the streams completes
+    fn eager_select<St>(self, other: St) -> EagerSelect<Self, St>
+        where St: Stream<Item = Self::Item>,
+              Self: Sized,
+    {
+        EagerSelect::new(self, other)
     }
 
     /// A future that completes after the given stream has been fully processed
--- a/futures/tests_disabled/all.rs
+++ b/futures/tests_disabled/all.rs
@@ -377,6 +377,106 @@ fn select2() {
 }
 
 #[test]
+fn eager_select() {
+    assert_done(|| f_ok(2).select(empty()).then(unselect), Ok(2));
+    assert_done(|| empty().select(f_ok(2)).then(unselect), Ok(2));
+    assert_done(|| f_err(2).select(empty()).then(unselect), Err(2));
+    assert_done(|| empty().select(f_err(2)).then(unselect), Err(2));
+
+    assert_done(|| {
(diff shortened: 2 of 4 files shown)


dbcfd:
Not ready for merge. Tests need to be updated, need feedback on issue whether this is the correct approach.

cramertj:
I'm not opposed to this, but it seems pretty niche to me. Would you mind sharing a bit about the usecase you have in mind for this? It seems like this is only useful if you have two streams where you (a) know that at least one of them will complete and (b) you don't care about receiving more elements after one of the streams completes.

dbcfd:
Will add these to the issue as well.

Some kind of real world examples I can think of

- Stream of signals with a stream of data, stream of data completes first, stream of signals may never complete
- Stream of events, selecting a periodic timestamp event stream, interrupting either stream should cause the stream to complete

Both of these are usually desired with a system like kafka, where you consume from an infinite stream, but occasionally interrupt the kafka stream. You'd like the drive the stream to completion so that offsets can be committed based on what was pulled off of kafka (attempted exactly once rather than at least once).

The alternative solution is to interrupt every provider of streams being selected (assuming that functionality is available), but then your ability to finish is a result of when things are interrupted and how they drain.

futures/tests_disabled/all.rs
```diff
@@ -376,6 +376,106 @@ fn select2() {
     }
 }
 
+#[test]
```
MajorBreakfast: `tests_disabled` contains tests that we haven't ported to 0.3 yet. Our goal is to look at each test and migrate it to 0.3 (or discard if it doesn't apply anymore). (BTW PRs are appreciated 😁)


Adding new tests into this folder isn't a good idea 🙂

futures-util/src/stream/mod.rs
```diff
@@ -71,6 +71,9 @@ pub use self::poll_fn::{poll_fn, PollFn};
 mod select;
 pub use self::select::Select;
 
+mod eager_select;
+pub use self::eager_select::EagerSelect;
```
MajorBreakfast: alphabetic order

MajorBreakfast:
I'm not quite sold on the name. Eager means in this case that the stream wants to finish eagerly. It could however also mean that it it is eager to return everything.

futures/tests_disabled/all.rs
dbcfd: Yep, in the process of attempting to move them out.

dbcfd:
Not entirely sold on the name either.

Another approach is to make select take a flag indicating if it should complete when one stream completes. Alternatively, that could default to false, and could use an attribution style method after select

`futA.select(futB).complete_on_first_stream()`

cramertj:
Q: is it necessary that the early-exit be a property of the stream, or is this something that could be handled by the code handling the stream? e.g.:

```rust
async {
    loop {
        let mut stream1_next = stream1.next();
        let mut stream2_next = stream2.next();
        select! {
            stream1_next => { match stream1_next { Some(..) => ..., None => break, } },
            stream2_next => { match stream2_next { Some(..) => ..., None => break, } },
        }
    }
}
```

dbcfd:
Not necessary to be a property of the stream, this would be purely for ergonomics since 

`
stream1.select_eager_complete(stream2).do_more_stuff()
`

is easier to read and maintain. Although this does bring up the idea of whether there should be similar behavior between select, merge, zip, or other intertwine operations when a stream completes. Was actually a gotcha for me that the select didn't complete when one stream finished.

dbcfd:
Is this functionality worthwhile to merge in, and should I update it and finish out tests?

cramertj:
I appreciate the effort to upstream custom combinators like this, and I understand wanting them as part of the core library. However, this still feels too niche to me, too easy to replicate with a manual `select!` statement, and too hard to come up with a good name for. I'd totally be open to reconsidering this in the future if you can come up with a more general solution or if we can build it on top of the existing `select` function in some way.

PR #1173, opened 2018-08-03
Add support for creating a LocalWaker from a borrowed Wake type

cramertj:


seanmonstar:
Couldn't `UnsafeWake` just be implement for `&'a T where T: UnsafeWake`, and not need the newtype and extra functions?

cramertj:
@seanmonstar That could work, but we don't currently expose the std type that implements `UnsafeWake` and is really the innards of an `Arc`, so there'd be no way to use the function without a custom implementation of `UnsafeWake`. I'm interested in ideas for making this better-- it's sort of an awkward thing at the moment that `UnsafeWake` implementations are the types *pointed to* by the pointer, whereas it seems like most of the implementations are for types that are actually pointers themselves, so we wind up with funky things like `ReferencedArc` and [`ArcWrapped`](https://doc.rust-lang.org/nightly/src/alloc/task.rs.html#51).

seanmonstar:
If that blanket impl also included `where T: ?Sized`, then `&UnsafeWake` basically works, right?

cramertj:
@seanmonstar I'm not sure how that solves the issue-- users going through the `Wake` trait still wouldn't be able to use it, right?

futures-util/src/task/local_waker_ref.rs
```diff
@@ -0,0 +1,140 @@
...
+// Pointers to this type below are really pointers to `Arc<W>`
+struct ReferencedArc<W> {
+    _marker: PhantomData<W>,
+}
+
+unsafe impl<W: Wake + 'static> UnsafeWake for ReferencedArc<W> {
+    #[inline]
+    unsafe fn clone_raw(&self) -> Waker {
+        let me = self as *const ReferencedArc<W> as *const Arc<W>;
+        Waker::from(Arc::clone(&*me))
```
MajorBreakfast: Why not method call syntax?

```rust
Arc::clone(&*me).into()
```

cramertj: No particular reason.

MajorBreakfast:
I like this PR! I think we can use this in `thread_pool.rs`

futures-util/src/task/local_waker_ref.rs
MajorBreakfast: Hehe. Method call syntax would be nice then 😁

cramertj:
@seanmonstar I've opened https://github.com/rust-lang-nursery/futures-rs/issues/1174 to discuss the issues around `UnsafeWake` vs. `Wake`.

futures-util/src/task/local_waker_ref.rs
cramertj: Done.

futures-util/src/task/local_waker_ref.rs
```diff
@@ -0,0 +1,140 @@
...
+        Arc::clone(&*me).into()
+    }
+
+    #[inline]
+    unsafe fn drop_raw(&self) {}
+
+    #[inline]
+    unsafe fn wake(&self) {
+        let me = self as *const ReferencedArc<W> as *const Arc<W>;
+        W::wake(&*me)
```
MajorBreakfast: here

futures-util/src/task/local_waker_ref.rs
```diff
@@ -0,0 +1,140 @@
...
+    }
+
+    #[inline]
+    unsafe fn wake_local(&self) {
+        let me = self as *const ReferencedArc<W> as *const Arc<W>;
+        W::wake_local(&*me)
```
MajorBreakfast: and here

MajorBreakfast: and on the other type

MajorBreakfast: basically like in `futures_unordered/task.rs`

https://github.com/rust-lang-nursery/futures-rs/blob/c8e7446b8d31c3ae0fad6f5d98a830db770fc86a/futures-util/src/stream/futures_unordered/task.rs#L207-L212

futures-util/src/task/local_waker_ref.rs
cramertj: Hm, I guess this is arguable, but I actually like using UFCS here and below because (1) I think it makes it clearer which implementation of the `Wake` trait is being delegated to and (2) it makes it clear that an unsafe re-referencing of a raw pointer is occuring. In any case, it really doesn't matter that much, so I guess I'm fine changing it?

MajorBreakfast: I'd use the method call syntax because it's easier to read (IMO). It's however fine by me if you leave it like that. If you decide to keep it, consider changing the ones in `futures_unordered/task.rs` as well then for consistency.

Issue #1169, opened 2018-08-03
Provide a crate with utilities for testing futures based code


Nemo157:
Related to https://github.com/rust-lang-nursery/wg-net/issues/38

I think it's worth `futures-rs` itself providing a crate with basic test utilities. Over time I would expect more complex utilities to be provided by crates out of tree, but having something available around the time of the 0.3 release for at least basic testing would be very useful for the ecosystem.

Ideas suggested:

 - [x] more introspect-able `Context`/`Waker`/`Spawn` implementations
 - [ ] in-memory IO primitives that can be induced to return `Poll::Pending` sometimes
 - [ ] [`#[async_test]` proc-macro](https://github.com/rust-lang-nursery/futures-rs/issues/1169#issuecomment-416301993)
 - (feel free to throw more ideas out for a first cut)

Nemo157:
I just threw together a start on a crate to provide something more useful for discussion, it trys to implement something around the second bullet above: https://github.com/Nemo157/futures-rs/tree/testing-utilities

cramertj:
I think everything in [`futures/tests/support`](https://github.com/rust-lang-nursery/futures-rs/tree/master/futures/tests/support) is really handy and should be exposed to the public somehow, especially the [waker counter context](https://github.com/rust-lang-nursery/futures-rs/blob/master/futures/tests/support/counter_waker_context.rs), which is super handy for testing readiness notifications from futures primitives.

Nemo157:
> especially the waker counter context

Heh, that’s basically exactly what I implemented on that branch 😄, oh well, it was a good chance for me to finally look at how the 0.3 waker system fits together.

MajorBreakfast:
I agree that we should aim to provide first party testing utilities. Testing is important and every respectable library needs to do it. We can avoid a lot of work duplication throughout the ecosystem if we provide common utilities. And the fewer project specific utilities a project uses, the easier it is of course for others to get started and contribute. Furthermore, we can write a guide that explains how it's done.

Nemo157:
Cool, leads are on board 😁, I’ll take a look at what’s in the futures test support currently and try and pull out a first pass on a crate this weekend.

MajorBreakfast:
Do we want to put this into `futures_util::test` or into a `futures-test` crate?

Nemo157:
IMO separate crate that’s not in the facade to be used as a dev-dependency

MajorBreakfast:
[...]

PR #1176, opened 2018-08-04
Initial testing utilities

Nemo157:
I'm not a massive fan of how the generics worked out here, especially around `with_context` needing two `_, _` at the end... any suggestion on a more ergonomic API?

CC #1169


MajorBreakfast:
I'd prefer the name `futures-test` because the built-in `test` crate (contains benchmarking stuff) also uses that name.

futures-test/src/task/context.rs
```diff
@@ -0,0 +1,146 @@
+use crate::task::{executor, wake};
+use futures_core::{
+    future::Future,
+    task::{local_waker, Context, Executor, Poll, Wake},
```
MajorBreakfast: - For consistency with the rest of our codebase the non-nested style would be better
- `Context` should not be imported directly. Instead it should always be referred to as `task::Context`. This decision was made in the 0.2 days. I wasn't around back then, but I think it's because the word "context" is so nondescript.

MajorBreakfast:
I think the symlinked license files are missing

MajorBreakfast:
> especially around with_context needing two _, _ at the end... any suggestion on a more ergonomic API?

Maybe `impl Trait` can help

futures-test/src/task/context.rs
```diff
@@ -0,0 +1,146 @@
...
+/// ```
+/// #![feature(async_await, await_macro, futures_api, pin)]
+///
+/// use futures::{future::Future, task::Poll};
+/// use futures_test::task::{self, wake, executor};
+/// use std::boxed::PinBox;
+///
+/// let mut future = PinBox::new(async {
+///     5
+/// });
```
MajorBreakfast: I suggest
```rust
let mut future = async { 5 };
pin_mut!(future);
```
I think we should prefer the `pin_mut!` macro in examples because users should reach for it first to avoid the heap allocation

futures-test/src/task/context.rs
Nemo157: I think in this case it's better to use a bare `Context`, this is explicitly a helper for managing that exact type. I think I'll change what's used in the docs though.

Nemo157:
> Maybe `impl Trait` can help

Not for this case, since you can't use turbofish when calling the function to fill in the 2 required parameters.

futures-test/src/task/context.rs
Nemo157: In terms of non-nested style, what about

```rust
use crate::task::{executor, wake};
use std::{marker::Unpin, mem::PinMut, sync::Arc};
```

I assume that the first is definitely ok, but what about the second? What's the actual limit on nesting that you want to enforce?

Nemo157:
😭 features and dev-dependencies seem impossible to work with (I knew there were going to be issues with using `futures-test` in a `no_std` crate, but I can't even get it working properly for `futures` to use it).

futures-test/src/task/context.rs
MajorBreakfast: I'd prefer to expand the second one as well. I'd like to align our code style with the code style of the standard library and compiler. I browsed the code just now and it seems that they consistently expand such things. Also, our existing code does so as well:

```rust
use core::marker::Unpin;
use core::mem::PinMut;
```
can be found many many times in futures-rs 🙂

Nemo157: Alright, so nesting the final segment in a path is ok, but no more than that.

Nemo157:
And that's the entirety of `futures/test/support` moved over, I think that's probably a good initial cut.

futures-test/src/assert.rs
```diff
@@ -2,16 +2,38 @@ use futures_core::stream::Stream;
...
 /// Assert that the next poll to the provided stream will return
 /// [`Poll::Pending`][futures_core::task::Poll::Pending].
+///
+/// # Examples
+///
+/// ```
+/// #![feature(async_await, futures_api, pin)]
+///
+/// #[macro_use]
+/// extern crate futures_test;
```
MajorBreakfast: In other doc example we write this as:
```rust
#![feature(async_await, futures_api, pin)]
#[macro_use] extern crate futures_test;
```

futures-test/src/assert.rs
```diff
@@ -2,16 +2,38 @@ use futures_core::stream::Stream;
...
+///
+/// #[macro_use]
+/// extern crate futures;
```
MajorBreakfast: I think this isn't required. We rely on the Rust 2018 module system and there's also no macro from the futures crate in use

futures-test/src/assert.rs
```diff
@@ -2,16 +2,38 @@ use futures_core::stream::Stream;
...
+///
+/// # fn main() {
```
MajorBreakfast: Main function doesn't need to be explicit

futures-test/src/assert.rs
Nemo157: `pin_mut!` is coming from `futures` still, this can change to `pin_utils` or whatever once that's here.

futures-test/src/assert.rs
```diff
@@ -2,16 +2,38 @@ use futures_core::stream::Stream;
...
+/// use futures::stream;
+/// use futures_test::future::FutureTestExt;
+///
+/// let mut stream = stream::once((async { 5 }).delay());
+/// pin_mut!(stream);
+///
+/// assert_stream_pending!(stream);
```
MajorBreakfast: It'd be good if the examples showed that multiple consecutive calls to these assert function are possible.

futures-test/src/assert.rs
MajorBreakfast: ah right ^^'

futures-test/src/assert.rs
Nemo157: It does when using `#[macro_use]`

MajorBreakfast: 😏
https://github.com/rust-lang-nursery/futures-rs/blob/86c40ff7e0d5c28e37a14edd3732e25efdfb9805/futures-util/src/macros/pin.rs#L88-L94

Nemo157: Ah, but only when `#[macro_use]` is on the same line as `extern crate`...

MajorBreakfast:
I'm not so sure about the `TestContext` API:
- `TestContext` is not actually a `Context` although its name suggests it
- Its name will often be shorted to "context" which leads to two things being called "context"
- It is not `no_std` compatible
- `TestContext::poll` works the other way around

Here is suggestion for an alternative API:
```
let cx = Context::new(&NoopWaker, &NoopExecutor);
let cx = noop_context(); // Short hand

let cx = Context::new(&PanicWaker, &PanicExecutor);
let cx = panic_context(); // Short hand

let wake_counter = WakeCounter::new();
let cx = noop_context().with_waker(wake_counter.local_waker());

let spawn_recorder = SpawnRecorder::new();
let cx = noop_context().with_executor(spawn_recorder.executor())
```

Everything except `SpawnRecorder` is `no_std` compatible. `WakeCounter` just needs atomics.

futures-test/src/lib.rs
```diff
@@ -0,0 +1,33 @@
...
+#[doc(hidden)]
+pub use std as std_reexport;
+
+#[doc(hidden)]
+pub extern crate futures_core as futures_core_reexport;
+
+#[macro_use]
+extern crate futures_util;
+
+#[macro_export]
```
MajorBreakfast: `#[macro_export]` above modules is not required

Nemo157:
The closest I can work out how to do is:

```rust
let wake = wake::panic();
let mut executor = executor::Panic;
let cx = Context::new(&wake, &mut executor);
```

the waker is a `LocalWaker`, which is non-`Sync` so can't be a `static`, so must be created from a function call. Then because the context takes in references the waker and executor both have to be stored in local variables to get a non-temporary lifetime. For `SpawnRecorder` specifically it would also become much more complicated since it would need to start using thread safe internal mutability.

That's the main reason I wanted a type like `TestContext` to handle the creation and lifetimes for the waker and executor, with the goal of compressing those three lines into one.

I don't think being `no_std` compatible is very worthwhile (at least I have not seen any way to actually utilise it in test helpers), it's also basically impossible because of limitations in feature handling with dev-dependencies in Cargo.

futures-test/src/assert.rs
MajorBreakfast: Oh ^^' interesting

Nemo157:
> I continue to believe that we shouldn't introduce `poll` and `poll_unpin` methods on the `TestContext`. They are unintuitive because they work the wrong way around and they can't be used with streams and sinks.

My issue was that using `with` per call is a lot of semantic overhead that makes the poll hard to read

```rust
assert_eq!(Poll::Pending, test_cx.with(|cx| abortable_rx.poll_unpin(cx)));
```

I think that would be helped a lot with the `context` method + NLL you suggest, while it's more code it's simpler to read code than what you get with `with` (although, I'm not sure if NLL adds any extra promotion that would allow this to work without `test_cx.context()` being saved to a local variable)

```rust
assert_eq!(Poll::Pending, abortable_rx.poll_unpin(&mut test_cx.context()));
```

> The code examples should refrain from shortening variables called `test_context` to `context`. `test_cx` is fine though.

:+1:

> ```rust
> with_noop_context(|cx| { ... }); // waker and executor are noops
> with_no_spawn_context(|cx| { ... }); // waker is a noop, executor panics on spawn attempts
> let cx = test_cx.context(); // For use with NLL
> ```

Yep, I realized that a `context` method should work last night. Didn't have time to fully test it. I'll add these soon.

---

I'm wondering if there's some lifetime finagling that would make it possible to have

```rust
let cx = noop_context();
let cx = panic_context();
let cx = no_spawn_context();
```

Since those wakers/executors don't actually do anything having multiple mutable references to a static executor should be safe (but I'm going to have to investigate how to make it sound...), `LocalWaker` on the other hand I'm wondering if there's a way to leverage thread local storage to have static instances.

MajorBreakfast:
@Nemo157 If you could make `noop_context()` and friends work via thread local storage, that would IMO be truly awesome. It could make this possible:

```rust
future.poll(noop_context());
```

futures-test/src/assert.rs
```diff
@@ -30,16 +32,15 @@ macro_rules! assert_stream_pending {
...
-                stream, cx,
-            ) {
-                $crate::futures_core_reexport::task::Poll::Ready(_) => {
-                    panic!("assertion failed: stream is not pending");
-                }
-                $crate::futures_core_reexport::task::Poll::Pending => {}
+        let cx = &mut $crate::task::no_spawn_context();
+        match $crate::futures_core_reexport::stream::Stream::poll_next(
+            stream, cx,
+        ) {
```
MajorBreakfast: I recommend using
```rust
if poll.is_ready() {}
```

futures-test/src/assert.rs
```diff
@@ -69,19 +72,18 @@ macro_rules! assert_stream_next {
...
+        let cx = &mut $crate::task::no_spawn_context();
+        match $crate::futures_core_reexport::stream::Stream::poll_next(stream, cx) {
+            $crate::futures_core_reexport::task::Poll::Ready(Some(x)) => {
+                assert_eq!(x, $item);
             }
-        })
+            $crate::futures_core_reexport::task::Poll::Ready(None) => {
+                panic!("assertion failed: expected stream to provide item but stream is at its end");
+            }
+            $crate::futures_core_reexport::task::Poll::Pending => {
```
MajorBreakfast: It might be nicer to combine these two cases with `|`

futures-test/src/task.rs
```diff
@@ -1,10 +1,19 @@
 //! Task related utilities.
+//!
+//! For the majority of usecases you can use the functions exported below
```
MajorBreakfast: *In* the majority of *use cases*

futures-test/src/task/executor/noop.rs
```diff
@@ -0,0 +1,56 @@
...
+///
+/// ```
+/// #![feature(async_await, futures_api)]
+///
+/// use futures::task::ExecutorExt;
+/// use futures_test::task::{panic_context, executor};
+///
+/// let mut cx = panic_context();
+/// let mut executor = executor::Noop::new();
+/// let cx = &mut cx.with_executor(&mut executor);
```
Nemo157: Two issues with `Context`'s current APIs here:

The first (that's repeated in most of the examples) is that `with_executor` and `with_waker` take `&mut self`, that's useful for being able to create a sub-context with a different executor inside a `poll` function, but it is what makes the first of these three lines necessary. If `with_executor` instead took `self` by value we should be able to merge the first and third lines into

```rust
let cx = &mut panic_context().with_executor(&mut executor);
```

Second, `with_executor` requires `E: Sized` so we can't use the unsized `Noop::executor_mut()`, if we removed this bound then we can merge the second and third lines into

```rust
let cx = &mut cx.with_executor(executor::Noop::executor_mut());
```

futures-test/src/task.rs
MajorBreakfast: and "functions" is mentioned twice in this sentence

futures-test/src/task.rs
```diff
@@ -1,10 +1,19 @@
...
+//! functions to create a [`Context`][futures_core::task::Context] appropriate
+//! to use in your tests.
+//!
+//! For more complex test cases you can take a `Context` from one of these
+//! functions then use the
```
MajorBreakfast: *and * then

futures-test/src/assert.rs
```diff
@@ -0,0 +1,126 @@
...
+
+/// Assert that the next poll to the provided stream will return
+/// [`Poll::Pending`][futures_core::task::Poll::Pending].
+///
+/// # Examples
+///
+/// ```
+/// #![feature(async_await, futures_api, pin)]
+///
+/// use futures::stream;
```
MajorBreakfast: Our code examples usually have no free line between `#![feature]` and `use` statements

futures-test/src/task/executor.rs
```diff
@@ -0,0 +1,8 @@
+//! Implementations of [`Executor`][futures_core::task::Executor] with various
```
MajorBreakfast: Maybe rename this to `executor/mod.rs` instead.

futures-test/src/task/executor.rs
```diff
@@ -0,0 +1,8 @@
...
+//! behaviour for test purposes.
+
+mod noop;
+mod panic;
+mod record;
+
+pub use self::{noop::Noop, panic::Panic, record::Record};
```
MajorBreakfast: I prefer the

```rust
mod noop;
use self::noop::Noop;

mod panic;
use self::panic::Panic;

// --snip--
```

style

futures-test/src/task/executor.rs
Nemo157: Yeah, I've been getting quite annoyed at the new file scheme breaking tab completion, so will switch this back.

MajorBreakfast: They changed it because the old system can lead to loads of `mod.rs` file tabs in an editor. However, that's IMO fine if these files are just lightweight files that just reexport the right things from their child modules. So, I continue to like `mod.rs` files 🙂

MajorBreakfast:
@Nemo157 I changed various things. Can you have a look?

PR #1178, opened 2018-08-05
Toykio blog post

polachok:


Nemo157:
Looks great :smile:

MajorBreakfast:
@polachok

Again, big thanks for writing this post!

I think we should release it shortly after alpha 3 this week, so that it can be compatible with the latest changes. Would that be alright?

polachok:
Sure

2018-08-06-toykio.md
```diff
@@ -0,0 +1,195 @@
+---
+layout: post
+title:  "Toykio"
+author: "Alexander Polakov"
+author_github: "polachok"
+date:   2018-08-01
+categories: blog
+---
+
+In this blog post I'd like to present [toykio](https://github.com/polachok/toykio), a simple futures executor intended for learning about how executors with an event loop work. Toykio only provides a very minimal feature set: An event loop and TCP streams and listners. However, it turns out that due to the fact that futures are composable, this is enough to build complex clients and servers.
```
cramertj: s/listners/listeners/

2018-08-06-toykio.md
```diff
@@ -0,0 +1,195 @@
...
+
+In the following, I'd like to give you a quick overview of toykio's components.
+
+## `AsyncTcpStream`
+
+Tokio defines the `AsyncTcpStream` type, a newtype wrapper around `TcpStream` from the standard library. Just like in std, the `connect` function opens a connection, and sets the socket to non-blocking mode. This means that the `read()` and `write()` methods will return immediately, either with data or an error. If there's not enough data (for reads) or buffer space (for writes) avaialble, a special kind of error - `WouldBlock` - is returned. We'll discuss how to handle it in the next section.
```
cramertj: s/avaialble/available

2018-08-06-toykio.md
```diff
@@ -0,0 +1,195 @@
...
+                REACTOR.with(|reactor| reactor.add_read_interest(fd, waker.clone()));
+
+                Poll::Pending
+            }
+            Err(err) => panic!("error {:?}", err),
+        }
+    }
+}
+```
+It tries to perform a read on the underlying `TcpStream` and if the read succeeds, the slice will be filled with data. If however the read fails with a `WouldBlock` error, then it registers its interest so that it gets notified when the data becomes available. More about this in the next section.
```
cramertj: s/"registers its interest so that it gets notified"/"registers the current task's waker so that it gets awoken"

cramertj:
@polachok Thanks so much for the great post! Is there a reason that toykio uses `select` rather than `epoll` or a cross-platform abstraction like `mio`? It seems like that would be a more realistic, simpler, and safer approach. The current explanation using `select` makes it seem like there's a lot of redundant work that needs to be done every iteration through the event loop.

polachok:
@cramertj 
1. epoll() is not cross-platform, select works across all unixen
2. many people know select from their OS/Unix class
3. Mio hides the complexity, you have to explain mio then. My core idea was that you can build a tokio-core-like reactor from OS primitives, there's no magic in it (indeed the first (unpublished) version of the article was titled "reconstructing tokio from first principles")

cramertj:
@polachok Great, thanks for explaining! It might be good to add a quick note about the expensive-seeming parts of the process (primarily the "iterate through all entries each time around the event loop") that wouldn't be present in a real application. Totally up to you, of course, but I'd like to avoid giving people the false impression that we have multiple `O(n)` lookups as part of handling every event.

polachok:
@cramertj I added a note about select

cramertj:
@polachok great, thanks!

MajorBreakfast:
And publish! 🙂

PR #1179, opened 2018-08-06
Remove code highlighting for project name

MajorBreakfast:


Nemo157:
Heh, new nightly improved on @cramertj's optimizations? Maybe those should be `<` instead of `==`?

MajorBreakfast:
@Nemo157 Yeah these probably should be `<=` signs. I'll create a PR for that

PR #1180, opened 2018-08-06
Update size assertions

MajorBreakfast:
@cramertj Should this be changed to `<=` to detect only regressions?


cramertj:
RE detecting only regressions: no, I think it's worth the breakage to know exactly when the size is changing, but it does require more frequent maintenance. I wouldn't be strongly opposed to changing it if you feel like the maintenance burden is too high.

cramertj:
SGTM!

PR #1183, opened 2018-08-06
Adjust docs on select

dbcfd:
Related to https://github.com/rust-lang-nursery/futures-rs/pull/1172, adding to select docs behavior when one stream ends.

Docs for zip specify this behavior, so no changes needed there.

### DIFF
--- a/futures-util/src/stream/mod.rs
+++ b/futures-util/src/stream/mod.rs
@@ -868,6 +868,9 @@ pub trait StreamExt: Stream {
     /// This combinator will attempt to pull items from both streams. Each
     /// stream will be polled in a round-robin fashion, and whenever a stream is
     /// ready to yield an item that item is yielded.
+    ///
+    /// When either this stream or the provided one ends, the remaining stream will be polled to
+    /// completion or error. The stream that ended will no longer be polled.
     fn select<St>(self, other: St) -> Select<Self, St>
         where St: Stream<Item = Self::Item>,
               Self: Sized,
--- a/futures-util/src/stream/mod.rs
+++ b/futures-util/src/stream/mod.rs
@@ -870,7 +870,7 @@ pub trait StreamExt: Stream {
     /// ready to yield an item that item is yielded.
     ///
     /// When either this stream or the provided one ends, the remaining stream will be polled to
-    /// completion or error. The stream that ended will no longer be polled.
+    /// completion. The stream that ended will no longer be polled.
     fn select<St>(self, other: St) -> Select<Self, St>
         where St: Stream<Item = Self::Item>,
               Self: Sized,
--- a/futures-util/src/stream/select.rs
+++ b/futures-util/src/stream/select.rs
@@ -8,6 +8,9 @@ use futures_core::task::{self, Poll};
 ///
 /// The merged stream produces items from either of the underlying streams as
 /// they become available, and the streams are polled in a round-robin fashion.
+///
+/// When either this stream or the provided one ends, the remaining stream will be polled to
+/// completion or error. The stream that ended will no longer be polled.
 #[derive(Debug)]
 #[must_use = "streams do nothing unless polled"]
 pub struct Select<St1, St2> {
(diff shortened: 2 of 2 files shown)


futures-util/src/stream/mod.rs
```diff
@@ -868,6 +868,9 @@ pub trait StreamExt: Stream {
     /// This combinator will attempt to pull items from both streams. Each
     /// stream will be polled in a round-robin fashion, and whenever a stream is
     /// ready to yield an item that item is yielded.
+    ///
+    /// When either this stream or the provided one ends, the remaining stream will be polled to
+    /// completion or error. The stream that ended will no longer be polled.
```
cramertj: Streams won't necessarily stop being polled on error, especially since this isn't a method on `TryStream`, which has special-casing for error handling. I would leave out the "or error" bit.

cramertj:
LGTM aside from "or error" nit

futures-util/src/stream/mod.rs
dbcfd: Too used to futures 0.1. Will remove the or error part.

cramertj: Thanks!  :smile:

futures-util/src/stream/mod.rs
```diff
@@ -868,6 +868,9 @@ pub trait StreamExt: Stream {
     /// This combinator will attempt to pull items from both streams. Each
     /// stream will be polled in a round-robin fashion, and whenever a stream is
     /// ready to yield an item that item is yielded.
+    ///
+    /// When either this stream or the provided one ends, the remaining stream will be polled to
+    /// completion. The stream that ended will no longer be polled.
```
MajorBreakfast: I'd phrase it like this:

> After the first input stream completes, the remaining one will be polled exclusively. The returned stream completes when both input stream have completed.

(Note: 80 char limit)

Feel free to tweak it further if you want ;)

dbcfd: `
/// After the consumed stream or the argument stream ends, the remaining 
    /// stream will be polled exclusively. The stream returned from select will
    /// complete when both the consumed stream and the argument stream have 
    /// completed.`

First input stream reads a little weird to me, possibly indicating the consumed stream.

Not sure what the correct terminology is for extension/enrichment traits like this, where one argument is self which is moved/consumed by the function, and the other argument is provided.

Maybe actually tackle the whole documentation with something like `
Select is a combinator which consumes this stream and the stream provided as an argument, producing a pair of input streams. These input streams are polled in round-robin fashion, and whenever a stream is ready to yield an item, that item is yielded. 

When an input stream completes, the remaining input stream from the pair is polled exclusively. The stream returned by the Select combinator will complete when both input streams from the pair have completed.`

MajorBreakfast: How about: 
> After one of the two input stream completes, the remaining one will be polled exclusively. The returned stream completes when both input stream have completed.

I don't want to emphasize too much that one stream is the `self` argument and the other the following argument. It doesn't really matter in this context. Also, it's very like that we soon make `join` and `select` free functions. `select(stream1, stream2)` is much more intuitive `stream1.select(stream2)`

As for the other thing you mention, various future combinators mentions this:

> Note that this function consumes this future and returns a wrapped version of it.

This can be adapted for this combinator. Of course use developer can glean the same by looking at the function signature.

dbcfd:
```
This combinator will attempt to pull items from both streams. Each
stream will be polled in a round-robin fashion, and whenever a stream is
ready to yield an item that item is yielded.

After one of the two input stream completes, the remaining one will be 
polled exclusively. The returned stream completes when both input 
streams have completed.

Note that this function consumes this stream and returns a wrapped version of it.```

That work?

dbcfd:
Replaced with https://github.com/rust-lang-nursery/futures-rs/pull/1192

MajorBreakfast:
@dbcfd you might know this already, but just in case you don't: Instead of creating the new `select-docs-2` branch, you could have force pushed to your existing `select-docs` branch to replace its content. PRs on GitHub automatically update whenever changes are made to their underlying branch.

Anyhow, thanks for your contribution! (Also, don't hesitate to provide further PRs if you want to change other stuff in the docs 😉)

PR #1184, opened 2018-08-06
Use default max age for crates.io badge

MajorBreakfast:


MajorBreakfast:
Now the badge finally shows alpha.2 🙂 (it still showed alpha.1 to me)

PR #1185, opened 2018-08-07
pin-utils crate & remove glob imports

MajorBreakfast:
- Add `pin-utils` crate
- Get rid of all `extern crate` statements
- Always use `use` statements to import macros
- Expand most glob imports. In particular remove all remaining `use futures::prelude::*` statements from code examples


futures-util/src/future/shared.rs
```diff
@@ -1,23 +1,3 @@
...
-//! use futures::executor::block_on;
-//!
-//! # fn main() {
-//! let future = future::ready(6);
-//! let shared1 = future.shared();
-//! let shared2 = shared1.clone();
-//! assert_eq!(6, *block_on(shared1));
-//! assert_eq!(6, *block_on(shared2));
-//! # }
-//! ```
```
MajorBreakfast: I removed this module description because module is private and we don't have descriptions for the other modules.

futures-channel/src/lib.rs
```diff
@@ -12,11 +12,6 @@
 
 #![doc(html_root_url = "https://rust-lang-nursery.github.io/futures-doc/0.3.0-alpha.2/futures_channel")]
 
-#[cfg(feature = "std")]
```
cramertj: Oh, I didn't think about that. We'll have to take care not to accidentally use `std` symbols in our `no_std` build, because there won't be anything preventing them in CI (since `std` is still reachable via a path).

MajorBreakfast: I'm pretty sure that this will catch them:

https://github.com/rust-lang-nursery/futures-rs/blob/3fdcfc1c42fd507b3e04c6ed2e99dd7e5cf1b56e/.travis.yml#L50

cramertj: Yeah, that'd get it. I opened https://github.com/rust-lang/rust/issues/53166 to discuss this.

cramertj:
Thanks, this looks great! :shipit:

PR #1190, opened 2018-08-10
Build crates with minimal versions in CI

MajorBreakfast:
Follow-up PR to https://github.com/rust-lang-nursery/futures-rs/pull/1189

The CI should break if we're relying on features not available in the minimal versions we specify.

cc @cramertj

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,7 +29,7 @@ matrix:
       rust: nightly
       script:
         - cargo bench --all
-        - cargo bench -p futures-util-preview --features=bench
+        - cargo bench --manifest-path futures-util/Cargo.toml --features=bench
 
     - name: cargo build --no-default-features
       rust: nightly
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,6 +12,12 @@ matrix:
     - name: cargo test
       os: linux
 
+    - name: cargo test (with minimal versions)
+      os: linux
+      script:
+        - cargo update -Zminimal-versions
+        - cargo test
+
     - name: cargo clippy
       rust: nightly
       install:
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -47,6 +47,18 @@ matrix:
...
+      rust: nightly
+      script:
+        - cargo update -Zminimal-versions
+        - cargo build --manifest-path futures/Cargo.toml --all-features
+        - cargo build --manifest-path futures-core/Cargo.toml --all-features
+        - cargo build --manifest-path futures-channel/Cargo.toml --all-features
+        - cargo build --manifest-path futures-executor/Cargo.toml --all-features
+        - cargo build --manifest-path futures-io/Cargo.toml --all-features
+        - cargo build --manifest-path futures-sink/Cargo.toml --all-features
+        - cargo build --manifest-path futures-util/Cargo.toml --all-features
```
MajorBreakfast: Could we use `cargo build --all-features` or `cargo build -p futures-... --all-features` instead?

`cargo build --no-default-features` seems not to exist at least

MajorBreakfast: `cargo build --all-features` might also not work. It exists but I'm not sure whether it does what we want

Nemo157: Nope, features and workspaces are pretty much broken at the moment. There was an attempt at fixing them a while ago, but I’m not sure if that’s still available even as an unstable flag. I’ll try and find the link to the issue in a sec.

Nemo157: https://github.com/rust-lang/cargo/issues/5364

MajorBreakfast: I tested `-p` instead of `--manifest-path` and it doesn't work

cramertj:
Nice!

PR #1192, opened 2018-08-10
Adjust Docs on Select

dbcfd:
Adjust documentation on StreamExt select combinator to indicate behavior when
one of the stream ends early.

Squashed, updated version of https://github.com/rust-lang-nursery/futures-rs/pull/1183 after comments.

@MajorBreakfast @cramertj


cramertj:
Thanks!

PR #1194, opened 2018-08-11
Add for_each_concurrent and try_for_each_concurrent

cramertj:


cavedweller:
docs seem clear to me. Excited for this!

futures-util/src/stream/for_each_concurrent.rs
```diff
@@ -0,0 +1,106 @@
...
+    unsafe_pinned!(stream: Option<St>);
+    unsafe_unpinned!(f: F);
+    unsafe_unpinned!(futures: FuturesUnordered<Fut>);
+    unsafe_unpinned!(limit: Option<NonZeroUsize>);
+
+    pub(super) fn new(stream: St, limit: Option<usize>, f: F) -> ForEachConcurrent<St, Fut, F> {
+        ForEachConcurrent {
+            stream: Some(stream),
+            // Note: `limit` = 0 gets ignored.
+            limit: limit.and_then(NonZeroUsize::new),
```
Nemo157: I'm guessing there's no ergonomic way to push this up to the public API? What do you think of panicking here on `limit == 0` and calling it developer error instead?

cramertj: I suppose we could panic. I have no strong preference here. I think this is unlikely to be an issue in practice, so either way is fine.

futures-util/src/stream/for_each_concurrent.rs
```diff
@@ -0,0 +1,106 @@
+use core::marker::Unpin;
+use core::mem::PinMut;
+use core::num::NonZeroUsize;
+use futures_core::future::Future;
+use futures_core::stream::Stream;
+use futures_core::task::{self, Poll};
+use pin_utils::{unsafe_pinned, unsafe_unpinned};
+use crate::stream::{FuturesUnordered, StreamExt};
```
MajorBreakfast: The style guide suggests to put `use crate`/`use super` on top (https://github.com/rust-lang-nursery/fmt-rfcs/blob/master/guide/items.md#ordering-list-import)

futures-util/src/stream/for_each_concurrent.rs
MajorBreakfast: This should be consistent with `Buffered`.

I think that the making it possible to impose no limit is a good idea. limit = 0 should produce a stream that does nothing because it can't poll any futures. Not the best option, but IMO better than the alternatives:
- No limit is unintuitive because that's what `None` is for.
- A panic feels a bit drastic and is also not quite logical
- Exposing `NonZeroUsize` is a bit cumbersome

futures-util/src/stream/mod.rs
```diff
@@ -552,6 +555,63 @@ pub trait StreamExt: Stream {
...
+    /// Runs this stream to completion, executing the provided asynchronous
+    /// closure for each element on the stream concurrently as elements become
+    /// available.
+    ///
+    /// This is similar to [`StreamExt::for_each`], but the futures
+    /// produced by the closure are run concurrently.
+    ///
+    /// The closure provided will be called for each item this stream produces,
+    /// yielding a future. That future will then be executed to completion
+    /// concurrently with the other futures produced by the closure.
```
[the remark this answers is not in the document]
MajorBreakfast: @Nemo157 Single task, not thread (the task may run on different threads depending on the whims of the executor) 😁

I would have understood the current phrasing, however I suppose it can't hurt to make the distinction a bit clearer

futures-util/src/stream/for_each_concurrent.rs
cramertj: I agree we should be consistent-- I think we should change `Buffered`/`BuffereUnordered` as well. IMO the size savings from `Option<NonZeroUsize` are worth it, and passing in zero to any of those functions will always be buggy / bad.

futures-util/src/stream/mod.rs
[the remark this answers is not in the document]
cramertj: +1 to clarifying "concurrent, not parallel".

futures-util/src/stream/for_each_concurrent.rs
MajorBreakfast: Hmm actually it's quite common that `0` means no limit. I suppose that's also fine. However it should be mentioned in the docs for `for_each_concurrent`

futures-util/src/try_stream/try_for_each_concurrent.rs
```diff
@@ -0,0 +1,106 @@
...
+                            stream_completed = true;
+                            None
+                        }
+                        Poll::Pending => None,
+                    }
+                } else {
+                    None
+                };
+                if stream_completed {
+                    PinMut::set(self.stream(), None);
```
MajorBreakfast: Might as well set this directly in `Poll::Ready(None)`

futures-util/src/stream/mod.rs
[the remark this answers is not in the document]
Nemo157: @MajorBreakfast "single thread _at a time_" is what I meant.

Nemo157:
Just had a thought, could this just be `stream.map(f).buffer_unordered(limit)`? Is the compiler capable of optimizing that as much as this implementation?

cramertj:
@Nemo157 yeah, it is just `.map_ok(...).buffer_unordered(limit).collect::<()>()` and I've been writing it as such, but this is much more ergonomic, and it's a common enough case that I think it deserves an ergonomic solution.

futures-util/src/try_stream/try_for_each_concurrent.rs
cramertj: I can't do that because `self` is already borrowed inside the `if let`.

futures-util/src/stream/for_each_concurrent.rs
cramertj: Done.

futures-util/src/stream/for_each_concurrent.rs
cramertj: Done.

Nemo157:
Yeah, I wasn’t denying that a named combinator for this case is useful, I just wonder if the _implementation_ could literally be that.

cramertj:
@Nemo157 I considered that, but I think it'd wind up being slightly less efficient and probably more confusing, but I don't have strong feelings against the idea.

PR #1198, opened 2018-08-14
Update for new nightly

Nemo157:
Fixes the build, but ~~I still need to do the tests and~~ the docs need another going over to ensure "executor" and "spawn"/"spawner" are described well and used consistently.


MajorBreakfast:
This looks fantastic!

PR #1208, opened 2018-08-17
feat: Avoid cloning the output of Shared on the last poll

Marwes:
If a `Shared` instance knows that it is the last instance it can now
move the output out and return it without cloning.

Followup of #1093

### DIFF
--- a/futures-util/src/future/shared.rs
+++ b/futures-util/src/future/shared.rs
@@ -1,5 +1,5 @@
 use crate::task::local_waker_ref_from_nonlocal;
-use futures_core::future::Future;
+use futures_core::future::{FusedFuture, Future};
 use futures_core::task::{LocalWaker, Poll, Wake, Waker};
 use slab::Slab;
 use std::cell::UnsafeCell;
@@ -171,6 +171,12 @@ where
             FutureOrOutput::Output(ref item) => &item,
             FutureOrOutput::Future(_) => unreachable!(),
         }
+    }
+}
+
+impl<Fut: Future> FusedFuture for Shared<Fut> {
+    fn is_terminated(&self) -> bool {
+        self.inner.is_none()
     }
 }
 
--- a/futures-util/src/future/shared.rs
+++ b/futures-util/src/future/shared.rs
@@ -97,9 +97,9 @@ where
     /// clone of this `Shared` alive then, returns its result immediately
     /// without blocking. Otherwise, returns None without triggering the work represented by
     /// this `Shared`.
-    pub fn peek(&self) -> Option<Fut::Output> {
+    pub fn peek(&self) -> Option<&Fut::Output> {
         match self.inner.as_ref().map(|inner| inner.notifier.state.load(SeqCst)) {
-            Some(COMPLETE) => unsafe { Some(self.inner().clone_output()) },
+            Some(COMPLETE) => unsafe { Some(self.inner().get_output()) },
             Some(POISONED) => panic!("inner future panicked during poll"),
             _ => None,
         }
@@ -152,7 +152,7 @@ where
                     FutureOrOutput::Future(_) => unreachable!(),
                 }
             }
-            Err(inner) => inner.clone_output(),
+            Err(inner) => inner.get_output().clone(),
         }
     }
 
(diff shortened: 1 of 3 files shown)


futures-util/src/future/shared.rs
```diff
@@ -201,38 +220,53 @@ where
...
-                    this.inner.notifier.state.store(COMPLETE, SeqCst);
-                    let wakers = &mut lock.take().unwrap();
-                    for (_key, opt_waker) in wakers {
-                        if let Some(waker) = opt_waker.take() {
-                            waker.wake();
-                        }
-                    }
-                    return Poll::Ready(output);
-                }
+        if let Some(_) = Arc::get_mut(&mut this.inner) {
```
cramertj: Nit: clippy is angry about this and wants you to write `if Arc::get_mut(&mut this.inner).is_some()` instead.

futures-util/src/future/shared.rs
```diff
@@ -133,13 +134,28 @@ where
         debug_assert!(self.waker_key != NULL_WAKER_KEY);
     }
 
+    /// Safety: callers must first ensure that `self.inner.state`
+    /// is `COMPLETE`
+    unsafe fn take_or_clone_output(&mut self) -> Fut::Output {
+        if let Some(inner) = Arc::get_mut(&mut self.inner) {
```
cramertj: This is slightly racey because it will still clone the last element if other futures have completed but not yet been dropped. One option to prevent this would be to store `self.inner` in an `Option` and drop it on completion. WDYT?

cramertj:
I think this is a good optimization to have, but i want to fix `peek` to either return `None` or panic with a more appropriate error message in the case that the last value has been taken.

Marwes:
> I think this is a good optimization to have, but i want to fix peek to either return None or panic with a more appropriate error message in the case that the last value has been taken.

Made it return None in https://github.com/rust-lang-nursery/futures-rs/pull/1208/commits/841c2c1b1aaa9f9cb1b6d86b921272669cd5ecb8

futures-util/src/future/shared.rs
Marwes: I think that is a good idea, the extra branch is unfortunate but it ensures that `Shared` will only clone the output once per completed instance of `Shared` (-1 for the original value).

I haven't documented the guarantee though.

Marwes:
Also tweaked `peek` so that it just returns a reference as there is no reason to clone the value there https://github.com/rust-lang-nursery/futures-rs/pull/1208/commits/27296c5ad6461906187f0c3e93074d6475b796d4

futures-util/src/future/shared.rs
```diff
@@ -93,22 +93,31 @@ where
     Fut: Future,
     Fut::Output: Clone,
 {
-    /// If any clone of this `Shared` has completed execution, returns its result immediately
+    /// If any clone of this `Shared` has completed execution and there are still at least one
+    /// clone of this `Shared` alive then, returns its result immediately
     /// without blocking. Otherwise, returns None without triggering the work represented by
     /// this `Shared`.
```
MajorBreakfast: I think this text doesn't match the current implementation. Here's what I suggest:

> Returns `Some` containing a reference to this `Shared`'s output if it has already been computed by a clone or `None` if it hasn't been computed yet or if this `Shared` already returned its output from `poll`.

(Also, this text break at 80 chars per line)

futures-util/src/future/shared.rs
```diff
@@ -93,10 +93,9 @@ where
     Fut: Future,
     Fut::Output: Clone,
 {
-    /// If any clone of this `Shared` has completed execution and there are still at least one
-    /// clone of this `Shared` alive then, returns its result immediately
-    /// without blocking. Otherwise, returns None without triggering the work represented by
-    /// this `Shared`.
+    /// Returns Some containing a reference to this `Shared`'s output if it has
+    /// already been computed by a clone or `None` if it hasn't been computed yet
+    /// or if this `Shared` already returned its output from poll.
```
MajorBreakfast: Can you make all these links? 🙂

MajorBreakfast:
@Marwes @cramertj I made some changes. Plz, tell me what you think

futures-util/src/future/shared.rs
```diff
@@ -150,89 +178,106 @@ where
...
-                return unsafe { Poll::Ready(this.clone_output()) };
+                // Safety: We're in the COMPLETE state
+                return unsafe { Poll::Ready(this.take_or_clone_output()) };
             }
             POISONED => panic!("inner future panicked during poll"),
             _ => unreachable!(),
         }
 
-        let waker = local_waker_from_nonlocal(this.inner.notifier.clone());
+        let waker = local_waker_ref_from_nonlocal(&inner.notifier);
```
MajorBreakfast: @cramertj Your function proved useful once again 😉

PR #1209, opened 2018-08-17
Mark test spawn implementations as !Sync

Nemo157:


Nemo157:
@MajorBreakfast as I mentioned on discord, I think this is necessary to avoid the static references making it to other threads (the wake implementations are fine since they return `LocalWaker` which is already !Sync).

PR #1210, opened 2018-08-17
Remove fields of ZSTs

MajorBreakfast:


futures-test/src/task/wake/noop.rs
```diff
@@ -18,14 +18,12 @@ use std::sync::Arc;
 /// cx.waker().wake();
 /// ```
 #[derive(Debug)]
-pub struct Noop {
-    _reserved: (),
-}
+pub struct Noop {}
```
Nemo157: Can just be `pub struct Noop;`

futures-test/src/task/wake/noop.rs
```diff
@@ -18,14 +18,12 @@ use std::sync::Arc;
...
 
 impl Noop {
     /// Create a new instance
     pub fn new() -> Self {
-        Self { _reserved: () }
+        Self {}
```
Nemo157: This might work as just `Self`, otherwise just `Noop`

Nemo157: You could also just drop the `new` function since anyone can construct a field-less ZST.

PR #1212, opened 2018-08-17
Improve futures-test helper functions

MajorBreakfast:


Nemo157:
Heh, `&mut`, `&` what's the difference 🤷‍♂️

MajorBreakfast:
@Nemo157 I added another commit that implements some major changes.

What do you think about the flat hierarchy approach? I think it's more convenient to use. 

The overview page of the `futures_test::task` module still needs to be updated.

Nemo157:
I would `s/Spawner/Spawn` in the type names to match the trait they’re implementing. But other than that looks like a good simplification.

MajorBreakfast:
@Nemo157 To me a "spawner" implements the `Spawn` trait. (Similarly a "reader" implements the `Read` trait, e.g. `BufReader`)

Nemo157:
Sure, I don't really care about the name much. :shipit: once the docs in `futures_test::task` are fixed up .

PR #1217, opened 2018-08-20
Add `StreamExt::shared()` to create a shared stream

tmccombs:
This allows sharing a Stream between multiple threads, similar to how
`FutureExt::shared()` works, and in facts this implementation uses
`FutureExt::shared()` under the covers.

An alternative implementation would be something similar to what we do for `future::Shared` for streams. The advantage of that would be avoiding allocating a new `Slab` for wakers for each item in the stream. I don't think the additional complexity is worth the possible performance improvement, but if that is a concern, I could implement it that way instead.


futures-util/src/stream/shared.rs
```diff
@@ -0,0 +1,94 @@
...
+impl<St> Future for Inner<St>
+where
+    St: Stream + Unpin,
+    St::Item: Clone,
+{
+    type Output = Option<(St::Item, SharedFuture<Self>)>;
+
+    fn poll(mut self: PinMut<Self>, cx: &mut task::Context) -> Poll<Self::Output> {
+        let (item, stream) = ready!(PinMut::new(&mut self.0).poll(cx));
+        Poll::Ready(item.map(|it| (it, Inner(stream.into_future()).shared())))
```
cramertj: Internally, calling `shared` allocates a new `Arc`, so this would result in a new allocation for every item in the stream (separately from the `Clone` that already has to happen for every item in the stream). Essentially, it builds a linked-list of the elements yielded by the stream. I'd prefer instead if the combinator kept a vector of all the elements yielded so far that have not yet been consumed by an interested waker, though I'm not exactly sure of the details. What do you think? Does this seem achievable? If not, I'd still be interested in adding support for this, but I'm concerned about the cost of all the allocations required and would want to include a note about them in the docs for the method.

cramertj:
This is great! I'd love to see something like this added. I think, however, that we need to change the underlying implementation away from `Shared<StreamFuture<..>>` in order to reduce the number of allocations. This would also allow us to support `!Unpin` streams.

futures-util/src/stream/shared.rs
tmccombs: >  Essentially, it builds a linked-list of the elements yielded by the stream. 

Yep, that's exactly how this works. My first attempt didn't use the future's shared method, but it also created a shared-structure linked list of Arcs. I eventually put that approach on ice because it was pretty complicated and used a fair amount of unsafe code.

> I'd prefer instead if the combinator kept a vector of all the elements yielded so far that have not yet been consumed by an interested waker

If I understand correctly, that would have different behavior than this though. With this implementation, every copy of the stream will get the same items in the same order. If items are only kept until they are consumed by one waker, then each item will only every be seen by one copy of the stream.

I think there are two other possibilities:
1. Keep a Vec of all results that have been retrieved, and each copy of the stream keeps track of a position in the Vec, once one stream reaches the end, it calls `poll_next`. The problem with this is that it either needs to continue to keep items in the vec after they are no longer needed, or somehow keep of all copies of the stream and where they are so that when an item is no longer needed it can be removed from the Vec, and the indexes stored in all streams updated.
2. Using a slab instead of Arcs. That would probably require re-implementing some of Arcs logic, but with slab references instead of pointers.

MajorBreakfast: There probably should the possibility to optionally specify a capacity limit for the vector / `VecDeque` that stores the items

cramertj: @tmccombs 
> If I understand correctly, that would have different behavior than this though. With this implementation, every copy of the stream will get the same items in the same order. If items are only kept until they are consumed by one waker, then each item will only every be seen by one copy of the stream.

Yeah, my statement 
> kept a vector of all the elements yielded so far that have not yet been consumed by an interested waker

could have been clearer: I meant to say "a vector of all elements yielded so far *for which there exists a consumer that has not yet seen them*". So what you'd need to do would be to track the number of consumers that have not yet seen each value, and only pull them off of the queue for good once everyone has seen them.

cramertj: @tmccombs did you have any luck getting this going?

tmccombs: I've made some progress on it. I've been fairly busy with other things, and getting this working is substantially more complicated than my first attempt. I definitely think it is feasible, though.

tmccombs: I've pushed a branch up at https://github.com/tmccombs/futures-rs/tree/shared-stream-vec  I haven't quite got it working yet, but I'd like feedback on my approach.

cramertj:
Closing due to inactivity.

PR #1218, opened 2018-08-20
Add a `StreamExt::boxed` method

tmccombs:
To make it a litle more convenient to create a PinBox of a Stream


MajorBreakfast:
Good idea!

Issue #1203, opened 2018-08-16
Clarify that examples are in async context


Nemo157:
As [pointed out on reddit](https://www.reddit.com/r/rust/comments/97dtn3/futures_030alpha3/e49gbll) we don't actually specify anywhere that our examples are in an async context. We could have a comment in the top level docs saying this, but it's likely to be missed. We could have comments in every example, but that's a lot of extra comments. Is there some way we can obviously show this without cluttering the documentation too much? (Also, this was in relation to examples in the blog, those should also be annotated somehow, probably with something louder than the doc examples).

Nemo157:
To explain this a little more, I’m not actually referring to the `task::Context` at all here, by “async context” I mean that we are within an `async fn`/async closure/async block. Maybe a less confusable name would be something like “async scope”?

Basically, all our examples in the documentation have a hidden `block_on(async { ... })` wrapping them since that’s unimportant when you’re using futures from within an `async fn`. This is not immediately obvious to new developers looking at the documentation though, if they try copy-pasting the code into a normal `fn` then it’s likely to fail to compile because `await!` doesn’t work.

cramertj:
I'd be in favor of un-hiding the `async { ... }` wrapping block or the whole thing altogether (including `block_on`), but I don't have particularly strong feelings here. @MajorBreakfast probably has some.

dbcfd:
I think between a concise example, and then documentation that most examples assume an async context, it's probably enough for a start.

I'd definitely prefer all examples to just have an async block around them, in case users don't enter the documentation at the lib.rs level (e.g. google search for and_then)

MajorBreakfast:
Well it would add a bit of boilerplate, but since the question about how to run futures is popping up again and again, I think we should just do it.

taiki-e:
Addressed by #1230

PR #1220, opened 2018-08-21
Async Context Example #1203

dbcfd:
Not ready for merge.

This example is meant to address #1203 providing clarification of how context is created and passed through macros and the async keyword.

The code in examples folder is currently meant to be temporary (quick compilation of single example), and to be removed when example is finalized.

### DIFF
--- a/futures/examples/functional.rs
+++ b/futures/examples/functional.rs
@@ -1,0 +1,57 @@
+#![feature(futures_api, async_await, await_macro)]
+#[macro_use] extern crate futures;
+
+use futures::{
+    channel::mpsc,
+    executor, //standard executors to provide a context for futures and streams
+    futures::ready,
+    StreamExt
+};
+
+fn main() {
+    let (tx, rx) = mpsc::unbounded::<i32>();
+
+    //
+    // Create a future by building an async block that will be implemented as a future,
+    // whose Future::poll will be called sometime later with a std::task::Context
+    //
+    let fut_values = async {
+        //
+        // Create another async block, again that implements future and will
+        // be provided a std::task::Context at some point
+        //
+        let fut_tx_result = async {
+            (0..100).for_each(|v| {
+                tx.unbounded_send(v).expect("Failed to send");
+            })
+        };
+
+        //
+        // Use the context that has been provided to this async block by
+        // Future::poll to spawn a future, which entails Future::poll
+        // being invoked by some std::task::Context, whose parent is the
+        // context of the current async block
+        //
+        spawn!(fut_tx_result);
+
+        let fut_values = rx
+            .map(|v| ready(v * 2))
+            .collect();
+
+        //
+        // Await the result of collecting the stream, by polling the Future using
+        // the std::task::Context of this async block.
+        //
+        await!(fut_values)
+    };
+
+    //
+    // Actually execute the above future, which will invoke Future::poll and
+    // subsequenty chain appropriate Future::poll and std::task::Context's to
+    // drive all futures, eventually driving the fut_values future to
+    // completion.
+    //
+    let values: Vec<i32> = executor::block_on(fut_values);
+
+    println!("Values={:?}", values);
+}
--- a/futures/examples/imperative.rs
+++ b/futures/examples/imperative.rs
@@ -1,0 +1,57 @@
+#![feature(futures_api, async_await, await_macro)]
+#[macro_use] extern crate futures;
+
+use futures::{
+    channel::mpsc,
+    executor, //standard executors to provide a context for futures and streams
+    StreamExt
+};
+
+fn main() {
+    let (tx, mut rx) = mpsc::unbounded::<i32>();
+
+    //
+    // Create a future by building an async block that will be implemented as a future,
+    // whose Future::poll will be called sometime later with a std::task::Context
+    //
+    let fut_values = async {
+        //
+        // Create another async block, again that implements future and will
+        // be provided a std::task::Context at some point
+        //
+        let fut_tx_result = async {
+            (0..100).for_each(|v| {
+                tx.unbounded_send(v).expect("Failed to send");
+            })
+        };
+
+        //
+        // Use the context that has been provided to this async block by
+        // Future::poll to spawn a future, which entails Future::poll
+        // being invoked by some std::task::Context, whose parent is the
+        // context of the current async block
+        //
+        spawn!(fut_tx_result);
+
+        let mut pending = vec![];
+        //
+        // Await the result of rx.next(), by polling the Future using the
+        // std::task::Context of this async block
+        //
+        while let Some(v) = await!(rx.next()) {
+            pending.push(v * 2);
+        };
+
+        pending
+    };
+
+    //
+    // Actually execute the above future, which will invoke Future::poll and
+    // subsequenty chain appropriate Future::poll and std::task::Context's to
+    // drive all futures, eventually driving the fut_values future to
+    // completion.
+    //
+    let values: Vec<i32> = executor::block_on(fut_values);
+
+    println!("Values={:?}", values);
+}
--- a/futures/src/lib.rs
+++ b/futures/src/lib.rs
@@ -20,6 +20,72 @@
 //! threading. Large asynchronous computations are built up using futures,
 //! streams and sinks, and then spawned as independent tasks that are run to
 //! completion, but *do not block* the thread running them.
+//!
+//! The following example describes how the task system context is built and used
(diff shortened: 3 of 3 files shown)


futures/examples/async_context.rs
```diff
@@ -0,0 +1,62 @@
...
+    // available within the async block for additional async/await operations.
+    //
+    // This is approximately equal to having a thread local context that is
+    // available within the future, and given to additional async/await blocks.
+    //
+    // ```text
+    //   static mut TASK_CONTEXT: Option<NonNull<std::task::Context>> = None;
+    //   let polled = unsafe {
+    //     let mut cx = TASK_CONTEXT.take().unwrap();
+    // ```
```
MajorBreakfast: - `async` is just the keyword. You mean "async block"
- The example with the thread local context is well intentioned, but I think it will unfortunately confuse newbies more than help them

futures/examples/async_context.rs
```diff
@@ -0,0 +1,62 @@
...
+    //
+    let fut_result = async {
+        //
+        // Once inside the async block and its std::task::Context, we can use
+        // additional async or await blocks, utilizing the std::task::Context
```
MajorBreakfast: "await block" doesn't exist ^^'

futures/examples/async_context.rs
```diff
@@ -0,0 +1,62 @@
...
+        // that this async block has maintained.
```
MajorBreakfast: The context can be a different one every time the future is polled

futures/examples/async_context.rs
```diff
@@ -0,0 +1,62 @@
...
+        //
+        // For async blocks, the std::task::Context of the parent block will be
+        // passed via future::poll to the child async blocks.
+        //
+        let fut_value = async {
+            rand::random::<i32>()
+        };
+
+        //
+        // For await! macros, the std::task::Context of the parent async block
```
MajorBreakfast: `await!` is currently a macro. It will be a compiler-builtin. But that's technical. Probably best to call it an `await!` expression.

futures/examples/async_context.rs
```diff
@@ -0,0 +1,62 @@
...
+        // will be passed through the macro to the inner async functionality,
+        // via the inner async functionality's future::poll method.
+        //
+        let r = await!(fut_value);
+
+        r % 2 == 0
+    };
+
+    //
+    // Now we're defining an implementor of std::task::Context that can provide
```
MajorBreakfast: `task::Context` is a `struct`. So, it can only be created, not implemented

futures/examples/async_context.rs
```diff
@@ -0,0 +1,62 @@
...
+    // the appropriate std::task::Context for running async functionality.
+    //
+    let mut exec = executor::LocalPool::new();
+    let mut spawn = exec.spawner();
+
+    //
+    // Use the executor to poll the future, and by polling the future, provide
+    // a std::task::Context to the previously defined async functionality via
```
MajorBreakfast: the term "async functionality" is strange. It's best to call it "future"

Nemo157:
I’ve added a comment to the linked issue. Documentation covering this area is definitely something we need, but isn’t quite what I was referring to in #1203. I see you do have a comment about it at the end of the example, but I think we can have a much shorter explanation specifically about how the examples are structured (basically all that most users really need to know is that the examples will work when copied into an `async fn`), and have this more detailed documentation on how the `task::Context` gets from place to place separate.

MajorBreakfast:
I added some comments. I think, however, that a more fitting location for such a guide would be the [async programming book](https://github.com/rust-lang-nursery/wg-net/tree/master/async-book). It's the place for high level guidelines, tutorials and explainers like this one.

dbcfd:
Thanks for the feedback. One of the minimal things I was trying to address at the lib.rs level is just how to actually use futures/async/await, since you can't just do ```
let f = async { ... }
await!(f)```

Was trying to figure out an example that made it more apparent that there is a context available in `async`, and that is why `await!` works, but that would be definitely too complex for lib.rs docs, and does belong in the async programming book.

I'll see if I can simplify the example a little more and improve the comments. Maybe non-working code snippets to indicate that context comes from executor or poll?

MajorBreakfast:
@dbcfd You can do that. [Playground example](https://play.rust-lang.org/?gist=35de56e4ca8c44e1139e5b63420513e7&version=nightly&mode=debug&edition=2018)

dbcfd:
I've added two examples which seem a little more concise, a more functional style and a more imperative style example. For now, I've chosen the more functional style to add to lib.rs, and put a note at the end of the docs about the majority of examples and snippets assuming async block.

futures/examples/functional.rs
```diff
@@ -0,0 +1,57 @@
+#![feature(futures_api, async_await, await_macro)]
+#[macro_use] extern crate futures;
```
MajorBreakfast: `extern crate` is so Rust 2015 🙂

futures/examples/functional.rs
```diff
@@ -0,0 +1,57 @@
...
+
+use futures::{
+    channel::mpsc,
+    executor, //standard executors to provide a context for futures and streams
+    futures::ready,
+    StreamExt
```
MajorBreakfast: Usually we use the expanded style:

```rust
use futures::channel::mpsc;
use futures::executor::block_on;
use futures::future;
use futures::spawn;
```

(We use `{...}`, but only for imports from the same module and not nested)

futures/examples/functional.rs
```diff
@@ -0,0 +1,57 @@
...
+        //
+        // Use the context that has been provided to this async block by
+        // Future::poll to spawn a future, which entails Future::poll
+        // being invoked by some std::task::Context, whose parent is the
+        // context of the current async block
+        //
+        spawn!(fut_tx_result);
+
+        let fut_values = rx
+            .map(|v| ready(v * 2))
```
MajorBreakfast: `future::ready`

We usually prefix functions with the module. (`block_on` is probably an exception to that rule, haven't decided yet because we have always hidden it up in all doc examples until now 😄)

futures/examples/functional.rs
```diff
@@ -0,0 +1,57 @@
...
+    channel::mpsc,
+    executor, //standard executors to provide a context for futures and streams
+    futures::ready,
+    StreamExt
+};
+
+fn main() {
+    let (tx, rx) = mpsc::unbounded::<i32>();
+
+    //
```
MajorBreakfast: Comments elsewhere in the library don't have this padding

futures/examples/functional.rs
```diff
@@ -0,0 +1,57 @@
...
+    let fut_values = async {
+        //
+        // Create another async block, again that implements future and will
+        // be provided a std::task::Context at some point
+        //
+        let fut_tx_result = async {
+            (0..100).for_each(|v| {
+                tx.unbounded_send(v).expect("Failed to send");
+            })
```
MajorBreakfast: There's no reason for putting this into an async block

Nemo157: It is being spawned off to a separate task though.

futures/examples/functional.rs
```diff
@@ -0,0 +1,57 @@
...
+        };
+
+        //
+        // Use the context that has been provided to this async block by
+        // Future::poll to spawn a future, which entails Future::poll
+        // being invoked by some std::task::Context, whose parent is the
```
MajorBreakfast: `task::Context` is just a mindless struct. All scheduling and execution is performed by the executor onto which the spawner inside the context spawns futures

futures/examples/functional.rs
```diff
@@ -0,0 +1,57 @@
...
+    executor, //standard executors to provide a context for futures and streams
+    futures::ready,
+    StreamExt
+};
+
+fn main() {
+    let (tx, rx) = mpsc::unbounded::<i32>();
+
+    //
+    // Create a future by building an async block that will be implemented as a future,
```
MajorBreakfast: To be precise: An async block is an expression that gives back a type that implements `Future`

MajorBreakfast: How I would phrase it: Create a future via an async block

MajorBreakfast: It's probably better to mention `Future::poll` only later

futures/examples/functional.rs
MajorBreakfast: Hmm yes, you're right

dbcfd:
PR not picking up latest changes by force push. Closing and re-opening against latest.

PR #1223, opened 2018-08-23
Fix unsound take() in FuturesUnordered

cramertj:
Fix a bug in `FuturesUnordered` that would cause the inner future to be moved before `Drop` being called, which is unsound. Added a test to ensure that we don't regress this behavior.

### DIFF
--- a/futures-test/src/future/assert_unmoved.rs
+++ b/futures-test/src/future/assert_unmoved.rs
@@ -1,0 +1,55 @@
+use futures_core::future::Future;
+use futures_core::task::{self, Poll};
+use pin_utils::{unsafe_pinned, unsafe_unpinned};
+use std::marker::Pinned;
+use std::mem::PinMut;
+use std::ptr;
+
+/// Combinator for the
+/// [`FutureTestExt::assert_unmoved`](super::FutureTestExt::assert_unmoved)
+/// method.
+#[derive(Debug, Clone)]
+#[must_use = "futures do nothing unless polled"]
+pub struct AssertUnmoved<Fut> {
+    future: Fut,
+    this_ptr: *const AssertUnmoved<Fut>,
+    _pinned: Pinned,
+}
+
+impl<Fut> AssertUnmoved<Fut> {
+    unsafe_pinned!(future: Fut);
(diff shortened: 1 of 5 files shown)


cramertj:
@MajorBreakfast Great suggestion! I've added `assert_unmoved` to the futures-test crate and changed the futures_unordered test to use it.

futures-test/src/future/assert_unmoved.rs
```diff
@@ -0,0 +1,55 @@
+use futures_core::future::Future;
+use futures_core::task::{self, Poll};
+use std::marker::Pinned;
+use std::mem::PinMut;
+use std::ptr;
+use pin_utils::{unsafe_pinned, unsafe_unpinned};
```
MajorBreakfast: alphabetic order 😁

futures/tests/futures_unordered.rs
```diff
@@ -131,38 +129,10 @@ fn iter_mut_len() {
...
-
-    let dmm = || DontMoveMe { this: ptr::null(), _pinned: Pinned };
-    let mut stream = block_on_stream(futures_unordered(vec![dmm(), dmm(), dmm()]));
+    // Future that will be ready after being polled twice,
+    // asserting that it does not move.
+    let fut = future::ready(()).pending_once().assert_unmoved();
+    let mut stream = block_on_stream(futures_unordered(vec![fut; 3]));
     assert_eq!(Some(()), stream.next());
     assert_eq!(Some(()), stream.next());
     assert_eq!(Some(()), stream.next());
```
MajorBreakfast: Should probably use [`assert_stream_next!` and `assert_stream_done!`](https://github.com/rust-lang-nursery/futures-rs/blob/6f60ab33d0cb3c148ce34af49a8af38f5487dcd4/futures-test/src/assert.rs#L61-L63)

MajorBreakfast:
Looks good!

futures-test/src/future/assert_unmoved.rs
cramertj: Done.

futures/tests/futures_unordered.rs
cramertj: Done.

PR #1224, opened 2018-08-24
A StreamObj implementation

### LINKED ISSUE #1221: is there a FutureObj equivalent for streams?
I would like to write an API where I return a struct with a stream value inside of it. Is there a `FutureObj` equivalent for `Stream`?

```
use futures::stream::{ repeat, Repeat };
use std::future::{FutureObj};

struct Response<'a> {
    stream: FutureObj<'a, Repeat<u8>>
}

pub fn main() {
    let mut s = repeat(1);
    let _r = Response {
        stream: FutureObj::new(&mut s)
    };
}
```

errors with:

```
error[E0277]: the trait bound `futures::stream::Repeat<{integer}>: futures::Future` is not satisfied
  --> src/boop.rs:11:17
   |
11 |         stream: FutureObj::new(&mut s)
   |                 ^^^^^^^^^^^^^^ the trait `futures::Future` is not implemented for `futures::stream::Repeat<{integer}>`
   |
   = note: required because of the requirements on the impl of `std::future::UnsafeFutureObj<'_, _>` for `&mut futures::stream::Repeat<{integer}>`
   = note: required by `<std::future::FutureObj<'a, T>>::new`

error: aborting due to 2 previous errors
```

I can satisfy the FutureObj by transforming the Stream in to a Future but that changes the code a lot:

```
use futures::stream::{ repeat, StreamExt, Repeat }; //, Stream };
use std::future::{FutureObj};

struct Response<'a> {
    stream: FutureObj<'a, (Option<u8>, Repeat<u8>)>
}

pub fn main() {
    let s = repeat(1);
    let mut sf = s.into_future();
    let _r = Response {
        stream: FutureObj::new(&mut sf)
    };
}
```

xrl:
I stole/borrowed the code from the `core::future::future_obj` module. Now I can store a Stream in a struct and pass it around, so that's cool.

Outstanding questions:

 - how do I test it?
 - do I need to edit these docs?

Closes #1221

### DIFF
--- a/futures-core/src/lib.rs
+++ b/futures-core/src/lib.rs
@@ -28,4 +28,6 @@ pub mod future;
 
 pub mod stream;
 
+pub mod stream_obj;
+
 pub mod task;
--- a/futures-core/src/lib.rs
+++ b/futures-core/src/lib.rs
@@ -28,6 +28,4 @@ pub mod future;
 
 pub mod stream;
 
-pub mod stream_obj;
-
 pub mod task;
--- a/futures-core/src/stream/stream_obj.rs
+++ b/futures-core/src/stream/stream_obj.rs
@@ -181,3 +181,68 @@ unsafe impl<'a, T, F> UnsafeStreamObj<'a, T> for PinMut<'a, F>
 
     unsafe fn drop(_ptr: *mut ()) {}
 }
+
+if_std! {
+    use std::boxed::{Box,PinBox};
+
+
+    unsafe impl<'a, T, F> UnsafeStreamObj<'a, T> for Box<F>
+        where F: Stream<Item = T> + 'a
+    {
+        fn into_raw(self) -> *mut () {
+            Box::into_raw(self) as *mut ()
+        }
+
+        unsafe fn poll_next(ptr: *mut (), cx: &mut task::Context) -> Poll<Option<T>> {
+            let ptr = ptr as *mut F;
+            let pin: PinMut<F> = PinMut::new_unchecked(&mut *ptr);
+            pin.poll_next(cx)
+        }
+
+        unsafe fn drop(ptr: *mut ()) {
+            drop(Box::from_raw(ptr as *mut F))
+        }
+    }
+
+    unsafe impl<'a, T, F> UnsafeStreamObj<'a, T> for PinBox<F>
+        where F: Stream<Item = T> + 'a
+    {
+        fn into_raw(self) -> *mut () {
+            PinBox::into_raw(self) as *mut ()
+        }
+
+        unsafe fn poll_next(ptr: *mut (), cx: &mut task::Context) -> Poll<Option<T>> {
+            let ptr = ptr as *mut F;
+            let pin: PinMut<F> = PinMut::new_unchecked(&mut *ptr);
+            pin.poll_next(cx)
+        }
+
+        unsafe fn drop(ptr: *mut ()) {
+            drop(PinBox::from_raw(ptr as *mut F))
+        }
+    }
+
+    impl<'a, F: Stream<Item = ()> + Send + 'a> From<PinBox<F>> for StreamObj<'a, ()> {
+        fn from(boxed: PinBox<F>) -> Self {
+            StreamObj::new(boxed)
+        }
+    }
+
+    impl<'a, F: Stream<Item = ()> + Send + 'a> From<Box<F>> for StreamObj<'a, ()> {
+        fn from(boxed: Box<F>) -> Self {
+            StreamObj::new(boxed)
+        }
+    }
+
+    impl<'a, F: Stream<Item = ()> + 'a> From<PinBox<F>> for LocalStreamObj<'a, ()> {
+        fn from(boxed: PinBox<F>) -> Self {
+            LocalStreamObj::new(boxed)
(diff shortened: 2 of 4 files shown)


futures-core/src/lib.rs
```diff
@@ -28,4 +28,6 @@ pub mod future;
 
 pub mod stream;
 
+pub mod stream_obj;
```
MajorBreakfast: `StreamObj` needs to be available under `futures_core::stream::StreamObj` for consistency with the `future` module

To do this:
- Rename `stream.rs` to `stream/mod.rs`
- Rename `stream_obj.rs` to `stream/stream_obj.rs`
- Export `StreamObj` from `stream/mod.rs`

futures-core/src/stream_obj.rs
```diff
@@ -0,0 +1,161 @@
+//! Asynchronous stream object value
```
MajorBreakfast: this module will be private -> remove this

futures-core/src/stream_obj.rs
```diff
@@ -0,0 +1,161 @@
...
+
+use std::fmt;
+use std::marker::{PhantomData,Unpin};
+use std::task::{ Context, Poll };
+use core::mem::PinMut;
```
MajorBreakfast:
```rust
use crate::task::{self, Poll};
use core::fmt;
use core::marker::{PhantomData, Unpin};
use core::mem::PinMut;
```

Note: `Context` should alway be referred to as `task::Context`

futures-core/src/stream_obj.rs
```diff
@@ -0,0 +1,161 @@
...
+
+use crate::stream::Stream;
+
+/// Stealing code from core::future::future_obj
```
MajorBreakfast: Remove that 😅

Instead adapt the API docs for `LocalFutureObj` 🙂

futures-core/src/stream_obj.rs
MajorBreakfast: Ah I see. We don't do so in libcore. Well, we should...

futures-core/src/stream_obj.rs
```diff
@@ -0,0 +1,161 @@
...
+impl<'a, T> Drop for LocalStreamObj<'a, T> {
+    fn drop(&mut self) {
+        unsafe {
+            (self.drop_fn)(self.ptr)
+        }
+    }
+}
+
+/// A custom trait object for polling futures, roughly akin to
+/// `Box<dyn Future<Output = T> + Send + 'a>`.
```
MajorBreakfast: Future ^^'

MajorBreakfast:
What's still missing:
- Add the parts from `src/libcore/mem.rs`
- `if_std! { ... }` part at the end that adds the parts from `src/liballoc/boxed.rs`

Thanks for implementing this! 🙂

futures-core/src/stream_obj.rs
MajorBreakfast: The `core` instead of `std` will also fix the broken CI which tires to compiles futures-rs for cortex-m microchips and fails because it needs `no_std`

MajorBreakfast:
About your questions:
- Yes, the docs need to be in shape
- Tests are not required because it's just a temporary solution until `Stream` becomes object safe.

futures-core/src/stream_obj.rs
MajorBreakfast: You should do a quick case-insensitive text search for "future" to find all these

xrl:
@MajorBreakfast thanks for the feedback! I've addressed everything except `if_std! {...}`. I don't follow what parts of this `stream_obj.rs` need to rely on std.

xrl:
Should I put https://github.com/rust-lang/rust/blob/0dd88c97975b2a2e4ea3e4a2bdc3670a9f40bd67/src/liballoc/boxed.rs#L939-L975 and https://github.com/rust-lang/rust/blob/0dd88c97975b2a2e4ea3e4a2bdc3670a9f40bd67/src/liballoc/boxed.rs#L993-L1019 inside of `if_std!`?

MajorBreakfast:
@xrl everything from liballoc goes into `if_std!`

xrl:
@MajorBreakfast OK, cool! I've added the code from liballoc. I think this addresses the last of the feedack.

xrl:
Any idea why the `thumbv6m-none-eabi` arch doesn't have `std::fmt`?

Nemo157:
@xrl you can use `core::fmt`, none of `std` is available on that target but basic functionality is instead exported from the `core` crate.

MajorBreakfast:
Still contains the word "future" a few times 😉

futures-core/src/stream/stream_obj.rs
```diff
@@ -0,0 +1,248 @@
+use crate::task::{self, Poll};
+use core::fmt;
+use core::marker::{PhantomData,Unpin};
+use core::mem::PinMut;
+
+use crate::stream::Stream;
```
MajorBreakfast: Move this after the first line. `use crate` first, then the others (and every line ordered alphabetically 🙂)

MajorBreakfast:
Merged. Thanks!

PR #1225, opened 2018-08-26
Use crate renaming for depending on futures 0.1

MajorBreakfast:


MajorBreakfast:
Uhm, just imagine that the branch is called `crate-renaming` 😉

futures-util/tests/foo.rs
```diff
@@ -0,0 +1,18 @@
+#![feature(async_await)]
+use futures::future::{FutureExt, TryFutureExt};
+use futures::compat::Executor01CompatExt;
+use tokio_threadpool::ThreadPool;
+
+#[test]
+fn foo () {
```
Nemo157: Could do with a better name for this file and test

MajorBreakfast: I never meant to commit this file...

futures-util/src/compat/tokio.rs
```diff
@@ -14,33 +14,27 @@ use tokio_executor::{DefaultExecutor, Executor as TokioExecutor};
...
 /// use futures::spawn;
-/// use futures::channel::oneshot;
 /// use futures::compat::TokioDefaultSpawner;
-/// use futures::executor::block_on;
-/// use futures::future::{FutureExt, TryFutureExt};
-/// use std::thread;
-///
-/// let (sender, receiver) = oneshot::channel::<i32>();
-///
-/// thread::spawn(move || {
```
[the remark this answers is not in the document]
Nemo157: Looks fine (I still think it would be good to have an example of cross-executor communication somewhere, but makes sense to not include it in the basic example here).

PR #1228, opened 2018-08-27
StreamObj should be exported by the main futures crate

xrl:
PR #1221 missed re-exporting the StreamObj values.


MajorBreakfast:
Thx!

PR #1231, opened 2018-08-29
fix futures-preview on latest nightly

tinaun:
Pin types now live in core::pin. 

~~blocked on https://github.com/rust-lang-nursery/pin-utils/pull/3~~


tinaun:
updated

cramertj:
LGTM, thanks!

PR #1235, opened 2018-08-29
Deprecate `executor::Executor` and `executor::Run`

carllerche:
These two APIs have long been deprecated, they just have not been
annotated as such.

This patch also hides a number of deprecated functions to help keep the
generated documentation cleaner.


MajorBreakfast:
@carllerche Sure. I'll publish a release after your other PR is merged 🙂

PR #1236, opened 2018-08-29
Task docs

carllerche:
Attempt to clarify the problem with using a set structure to store tasks.

This question has come up a bunch; e.g. #670.

Continuation of #689 since I no longer have access to that branch.

### DIFF
--- a/src/task_impl/mod.rs
+++ b/src/task_impl/mod.rs
@@ -68,12 +68,12 @@ fn with<F: FnOnce(&BorrowedTask) -> R, R>(f: F) -> R {
 ///
 /// In this case, the idea is that whenever a `start_send` is called on one of
 /// the channel's send handles, if the channel is at capacity, the current task
-/// is stored in a `TaskSet`. Then, when capacity is available, a task is
-/// removed from the task set and notified.
+/// is stored in a set. Then, when capacity is available, a task is removed from
+/// the set and notified.
 ///
 /// The problem with this strategy is that multiple `Sender` handles can be used
 /// on the same task. In this case, when the second handle is used and the task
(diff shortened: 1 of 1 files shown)


carllerche:
I updated the URL. I still believe this is a net win and the question still gets asked in the Tokio gitter.

src/task_impl/mod.rs
```diff
@@ -54,6 +54,31 @@ fn with<F: FnOnce(&BorrowedTask) -> R, R>(f: F) -> R {
...
+/// A valid use case for `Task` to implement these two traits has not been
+/// encountered.
+///
+/// Usually, this question is asked by someone who wants to store a `Task`
+/// instance in a `HashSet`. This seems like an obvious way to implement a
+/// future aware, multi-handle structure; e.g. a multi-producer channel.
+///
+/// In this case, the idea is that whenever a `start_send` is called on one of
+/// the channel's send handles, if the channel is at capacity, the current task
+/// is stored in a `TaskSet`. Then, when capacity is available, a task is
```
MajorBreakfast: is `TaskSet` an actual type?

carllerche: It is a hypothetical.

cramertj: Perhaps "a task set" (like you use elsewhere) would be clearer that `TaskSet` isn't an actual type?

MajorBreakfast:
Looks good

MajorBreakfast:
And about the discussion about whether to put such an explanation in the docs: IMO if people are asking this a lot, then it's appropriate to put such an explanation in the docs for the type.

PR #1239, opened 2018-08-31
impl Unpin for NotifyHandle with feature flag

carllerche:
This adds an implementation of `Unpin` for `NotifyHandle`. `Unpin` is
only available on the nightly channel, so the implementation is guarded
by a feature flag.

The implementation will always be needed as the struct contains a raw
pointer to a trait object.


carllerche:
I'm still working on this, it looks like there are more places to impl.

carllerche:
Ok, I think I got the types that matter.

MajorBreakfast:
👍

PR #1240, opened 2018-08-31
fix mpsc::Sender::poll_complete impl

carllerche:
This fixes the "bounded senders are actually unbounded" problem. The
problem is discussed here: rust-lang-nursery/futures-rs#984

The issue probably exists in the `unsync` mod. I am not sure about 0.3


MajorBreakfast:
@seanmonstar Could you review this one as well?

PR #1241, opened 2018-08-31
fix race with dropping mpsc::Receiver

### LINKED ISSUE #909: Race condition in futures::sync::mpsc
It looks like there's a race condition in the MPSC implementation in `futures 0.1.19`, unless my understanding of the intended behavior is incorrect.  If a `Sender::try_send()` happens concurrently with a `Receiver` close/drop, it's possible for `try_send()` to return `Ok(())` even though the item can never be received and will not be dropped until the `Sender` is dropped.  My expectation was to receive an error matching `Err(ref e) if e.is_disconnected()`.  The `Receiver` `Drop` implementation closes the channel and drains any items present, but this can apparently happen just before the `Sender` thinks it has successfully enqueued the item.

I wrote a small program to stress-test this scenario and demonstrate the bug:
https://github.com/simmons/mpsc-stress

I discovered this behavior while stress testing a program which implements a synchronous API by sending a command + `oneshot::Sender` to a future's MPSC, which processes the command and sends the oneshot when complete.  When the described race occurs, the sending thread would deadlock while hopelessly waiting forever for the `oneshot::Receiver` to either receive or indicate disconnection.

If I get a chance in the next few days, I'll see if I can root-cause the problem.  (Unless I'm told that my expectation is incorrect.)

carllerche:
Fix a bug where messages sent into a channel at the same time as the
`Receiver` is dropped are never removed.

The impl in 0.3 suffers the same bug.

Fixes #909

### DIFF
--- a/src/sync/mpsc/mod.rs
+++ b/src/sync/mpsc/mod.rs
@@ -813,6 +813,12 @@ impl<T> Receiver<T> {
         loop {
             match unsafe { self.inner.message_queue.pop() } {
                 PopResult::Data(msg) => {
+                    // If there are any parked task handles in the parked queue,
+                    // pop one and unpark it.
+                    self.unpark_one();
+                    // Decrement number of messages
+                    self.dec_num_messages();
+
                     return Async::Ready(msg);
                 }
                 PopResult::Empty => {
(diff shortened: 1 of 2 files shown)


MajorBreakfast:
This needs review by either @aturon or @cramertj because I'm not familiar with that part of the code base. I do however think that it's important that this gets merged

PR #1248, opened 2018-09-03
Revert usage of crate renaming features, alpha.5

MajorBreakfast:


MajorBreakfast:
I really hope that these changes remain temporary and that we can soon use crate renaming.

MajorBreakfast:
Clippy is unhappy, but that's for another PR

Nemo157:
Clippy failures are because of the tool lints namespacing, need to switch all the clippy lints to be `#[allow(clippy::foo)]` and can get rid of the `#[allow(unknown_lints)]` (I've not tested this yet, just been waiting for this feature eagerly).

PR #1250, opened 2018-09-03
Update to use namespaced tool lints

### LINKED ISSUE #1249: Fix clippy
It seems that Clippy now wants its rules prefixed with `clippy::`.

Nemo157:
Fixes #1249


MajorBreakfast:
And we're green again. Thx 🙂

PR #1253, opened 2018-09-09
remove +x permissions on some files

sylvestre:
FYI, Debian static analyzer (lintian) is complaining about that

> W: librust-futures-core-dev: executable-not-elf-or-script usr/share/cargo/registry/futures-core-0.2.1/src/task/atomic_waker.rs


cramertj:
Weird-- good catch!

Issue #1219, opened 2018-08-20
FusedFuture and FusedStream


cramertj:
[The `select` macro](https://rust-lang-nursery.github.io/futures-api-docs/0.3.0-alpha.3/futures/macro.select.html) would be far more useful if it was usable inside of loops, like this:

```rust
pin_mut!(future_1, future_2, my_stream);
let mut state = ...;
loop {
    let x = select! {
        future_1 => { ...transform state... }
        future_2 => { ...transform state... }
        next(my_stream) as elem => { ... transform state... }
        done => { ... }
    };
    ...some more code doing something with x and state...
}
```

However, the `Stream` and `Future` traits require that they are not polled after completion, and they provide no way to check if completion has already occurred. It's possible to track this extra bit of state in a boolean or something as part of the macro, but then the looping functionality has to be built into the macro, meaning that (1) there would have to be separate `select!` and `select_loop!` macros and (2) any shared code or controlflow in the "always" branch would have to be duplicated across branches:

```rust
pin_mut!(future_1, future_2, my_stream);
let mut state = ...;
select_loop! {
    future_1 => { let x = ...transform state...; ...some more code doing something with x and state... }
    future_2 => { let x = ...transform state...; ...some more code doing something with x and state... }
    next(my_stream) as elem => { let x = ...transform state...; ...some more code doing something with x and state... }
    done => { let x = ...transform state...; ...some more code doing something with x and state... }
}
```

Another option would be to add `FusedFuture` and `FusedStream` with `has_completed()` methods, implement `FusedFuture` and `FusedStream` for many of the combinators which support them (in particular the `fuse()` combinator). This is likely more ergonomic for users familiar with how the system works, but could be confusing for new users who now have one more thing to understand (fusing, pinning, etc...).

Yet another option would be to require that all futures track this bit of state and add a `has_completed` method to the Future and Stream traits. This requires less involvement from end-users, but again increases the already-high ergonomic burden on manual-futures implementors.

Personally, I lean towards the `FusedFuture`/`FusedStream` approach as it adds minimal complexity for users who don't care about the `select!` combinator while still retaining the manual control-flow flexibility of `select!` + `loop` rather than `select_loop!`. However, I'm by no means attached to this idea, and I recognize that it pushes another point of complexity onto end users, which is disappointing. I'd like to hear what y'all think.

[...]

PR #1259, opened 2018-09-12
FusedFuture, FusedStream, and select! macro changes

cramertj:
This is still very much WIP in order to improve ergonomics of fusing and using the `select!` macro, but I figured I'd open this PR to see what folks think.

Closes https://github.com/rust-lang-nursery/futures-rs/issues/1219

### DIFF
--- a/futures-core/src/future/mod.rs
+++ b/futures-core/src/future/mod.rs
@@ -8,6 +8,19 @@ pub use core::future::Future;
 mod future_obj;
 pub use self::future_obj::{FutureObj, LocalFutureObj, UnsafeFutureObj};
 
+/// A `Future` or `TryFuture` which tracks whether or not the underlying future
+/// should no longer be polled.
+///
+/// `is_terminated` will return `true` if a future should no longer be polled.
+/// Usually, this state occurs after `poll` (or `try_poll`) returned
+/// `Poll::Ready`. However, `is_terminated` may also return `true` if a future
+/// has become inactive and can no longer make progress and should be ignored
+/// or dropped rather than being `poll`ed again.
+pub trait FusedFuture {
+    /// Returns `true` if the underlying future should no longer be polled.
+    fn is_terminated(&self) -> bool;
+}
+
 /// A convenience for futures that return `Result` values that includes
 /// a variety of adapters tailored to such futures.
 pub trait TryFuture {
--- a/futures-util/src/async_await/select.rs
+++ b/futures-util/src/async_await/select.rs
@@ -4,8 +4,7 @@
 /// that finishes first.
 ///
 /// `select!` can select over futures with different output types, but each
-/// branch has to have the same return type. Inside each branch, the respective
-/// future's output is available via a variable with the same name as the future.
+/// branch has to have the same return type.
 ///
 /// This macro is only usable inside of async functions, closures, and blocks.
 ///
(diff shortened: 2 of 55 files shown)


futures-core/src/future.rs
```diff
@@ -5,6 +5,19 @@ use core::pin::PinMut;
...
+/// should no longer be polled.
+///
+/// `can_poll` will return `false` if a future should no longer be polled.
+/// Usually, this state occurs after `poll` (or `try_poll`) returned
+/// `Poll::Ready`. However, `can_poll` may also return `false` if a future
+/// has become inactive and can no longer make progress and should be ignored
+/// or dropped rather than being `poll`ed again.
+pub trait FusedFuture {
+    /// Returns `false` if the underlying future should no longer be polled.
+    fn can_poll(&self) -> bool;
```
carllerche: I do not think this is a good name as it is not clear that it implies the future being complete. `can_poll` initially read to me as a hint for being ready.

`is_complete()`, `is_final`, ... would be much better.

cramertj: I tried so many darn names for this thing! I had `has_completed()` for a while, but IMO it's misleading because it sounds like a "complete" future would be one that is done working and ready to return a value, and for streams it's even more unclear what "complete" means, because it's not the point at which it has yielded every element-- it's the point at which the stream has already yielded `None`.

There's also the interesting case that I was trying to be compatible with where we want to support `FuturesUnordered` having a mode where `can_poll` returns false when there are no elements left, rather than repeatedly resolving to `None` (which is often an uninteresting case). e.g.:

```rust
let mut running_tasks = FuturesUnordered::new();
running_tasks.no_none_yield(); // bikeshed away
loop {
    let mut recv = rx.next().fuse();
    select! {
        future(recv as msg) => {
            running_tasks.push(async { ... })
        }
        next(running_tasks as completed_task) => { ... }
        all complete => { return ... }
    }
}
```

It may be we make a thin wrapper around `FuturesUnordered` called e.g. `TaskSet` or something for this, but the point is that for whatever type this is, `can_poll` refers to something different from "has yielded `None` already".

carllerche: `is_terminated`?

cramertj: Sure, I'd be fine with `is_terminated`, where "terminated" is something like "has reached a state where polling would yield redundant or inconsistent results, or otherwise misbehave". Does that sound about right to you?

cramertj:
Updated. Please review. I'm working on a followup PR to reorganize the structure of the `select!` macro and allow taking streams directly.

.idea/misc.xml
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
carllerche: Should this file be checked in?

futures-core/src/future/mod.rs
```diff
@@ -8,6 +8,19 @@ pub use core::future::Future;
...
+/// should no longer be polled.
+///
+/// `has_terminated` will return `true` if a future should no longer be polled.
+/// Usually, this state occurs after `poll` (or `try_poll`) returned
+/// `Poll::Ready`. However, `has_terminated` may also return `false` if a future
+/// has become inactive and can no longer make progress and should be ignored
+/// or dropped rather than being `poll`ed again.
+pub trait FusedFuture {
+    /// Returns `true` if the underlying future should no longer be polled.
+    fn has_terminated(&self) -> bool;
```
carllerche: I believe `is_` is more idiomatic than `has_`

There are a number of `is_` fns in the repo already, but I did not find a fn prefixed with `has_` (this would be the first.

.idea/misc.xml
cramertj: nah, thanks for catching that.

futures-core/src/future/mod.rs
cramertj: Sure, done.

futures-core/src/future/mod.rs
```diff
@@ -8,6 +8,19 @@ pub use core::future::Future;
 mod future_obj;
 pub use self::future_obj::{FutureObj, LocalFutureObj, UnsafeFutureObj};
 
+/// A `Future` or `TryFuture` which tracks whether or not the underlying future
+/// should no longer be polled.
+///
+/// `is_terminated` will return `true` if a future should no longer be polled.
+/// Usually, this state occurs after `poll` (or `try_poll`) returned
+/// `Poll::Ready`. However, `is_terminated` may also return `false` if a future
```
Nemo157: This `false` seems like it should be `true`?

futures-util/src/async_await/select.rs
```diff
@@ -14,58 +13,122 @@
...
+/// use futures::select;
+/// let mut a = future::ready(4);
+/// let mut b = future::ready(6);
+/// let mut total = 0;
+///
+/// loop {
+///     select! {
+///         future(a as a) => total += a,
+///         future(b as b) => total += b,
+///         all complete => break,
```
Nemo157: I find `all complete` a little confusing, it makes me think of a branch where all are _currently_ complete and ready to handle, not that they all had completed and been consumed in the past. I wonder if there’s a similarly succinct way to say this that implies that better.

futures-core/src/future/mod.rs
cramertj: Yup (all the booleans got flipped around from an earlier version of this PR).

futures-util/src/async_await/select.rs
cramertj: Yeah I puzzled on this a bit, but I think we're going to want to change this in the future anyways as we add cases for `next(...)` etc.

futures-core/src/future/mod.rs
cramertj: Done.

PR #1262, opened 2018-09-17
compat: use a CurrentRef waker in 01 compat

seanmonstar:
Just to poll a `Future01`, it isn't yet necessary to call
`task01::current()`, nor store it in an `Arc::new`. All that is needed
is a "reference", and if someone calls `waker().clone()`, *then* we can
do the allocations.

### DIFF
--- a/futures-util/src/compat/compat03to01.rs
+++ b/futures-util/src/compat/compat03to01.rs
@@ -8,7 +8,7 @@ use futures_core::{
     task as task03, TryFuture as TryFuture03, TryStream as TryStream03,
 };
 use futures_sink::Sink as Sink03;
-use std::{marker::Unpin, pin::PinMut, sync::Arc};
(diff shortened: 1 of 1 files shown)


seanmonstar:
CI failure seems unrelated to this PR.

Nemo157:
CI is definitely unrelated, doc failure is the buggy rustdoc link detection (which I will open a pr to disable soon) and benchmarks are caused by https://github.com/rust-lang/rust/pull/54116, left a comment there to try and sort that out.

seanmonstar:
I've tried restarting those CI jobs, but they still fail. Seems this can't be merged with them failing...

Nemo157:
Admins can merge even with failing ci (and as far as I know only the admins have merge rights here anyway).

aturon:
Merged, sorry for the delay (and feel free to ping if things are getting stuck.)

PR #1263, opened 2018-09-18
Add `extern crate test` back in to benches

Nemo157:
Required until there's a way to specify this dependency in the `Cargo.toml`.

See https://github.com/rust-lang/rust/pull/54116#issuecomment-422303058


Nemo157:
Was done as part of #1268

PR #1264, opened 2018-09-18
Disable denying rustdoc warnings

Nemo157:
There are currently too many bugs in the intra-doc-resolution feature to rely on its failures being our fault.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -68,11 +68,7 @@ matrix:
     - name: cargo doc
       rust: nightly
       script:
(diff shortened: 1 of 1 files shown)


Nemo157:
@cramertj updated to just remove the warning denial so the job will still fail if rustdoc crashes for some reason

Nemo157:
Heh, I was really confused on the order of things that happened in this issue, seems my time somehow ended up being 10 minutes out.

Issue #1265, opened 2018-09-20
Update to latest pinning API


Thomasdezeeuw:
PR https://github.com/rust-lang/rust/pull/53877 change the pinning API, breaking futures as well. Is someone working on fixing this? Or should I make an attempt?

cramertj:
I was planning to wait on this until https://github.com/rust-lang/rust/pull/54339 landed, which includes more breaking changes.

JesseWright:
@cramertj I believe this can be closed now that #1268 and #1273 have been merged and #1266 closed.

PR #1266, opened 2018-09-20
[WIP] Follow latest pinning API

Kroisse:
There is no advanced changes with new API like using `Pin<Rc<T>>` or else. Just a direct conversion from before.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -9,3 +9,6 @@ members = [
   "futures-util",
   "futures-test",
 ]
+
+[patch.crates-io]
+pin-utils = { git = "https://github.com/Kroisse/pin-utils.git", branch = "pin" }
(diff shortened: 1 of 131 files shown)


Thomasdezeeuw:
This would close #1265.

Thomasdezeeuw:
I think https://github.com/rust-lang/rust/pull/54339 breaks even more stuff.

Kroisse:
I've assumed that there is a bug in `Box::pinned`, or `impl<'a, T, F> UnsafeFutureObj<'a, T> for Pin<Box<F>>` in stdlib.

@Thomasdezeeuw Thank you. It'll be helpful to investigate it.

Kroisse:
https://github.com/rust-lang/rust/pull/54339/files#diff-96b4af8600be84c75015ef2be9198546R115

It's getting harder than I'd thought...

PR #1267, opened 2018-09-21
Let cas + swap instead of loop cas

Pslydhh:
While state == `REGISTERING` | `WAKING`, no one could change the state value, So Just take a swap operation.
‘Store’ here is more appropriate，but I heard there are some minor problems in the ‘store‘s memory order :(

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -5,7 +5,7 @@ members = [
   "futures-core",
   "futures-channel",
 #  "futures-executor",
-  "futures-io",
+#  "futures-io",
 #  "futures-macro-async",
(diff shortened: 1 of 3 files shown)


cramertj:
This looks right to me, but I'd like others to review as well since this stuff is tricky ;)

cc @carllerche

carllerche:
This looks correct to me. The `swap` is definitely required here due to Acq / Rel.

It would be helpful to back port this to 0.1 as well 👍

cramertj:
Rebased and submitting as https://github.com/rust-lang-nursery/futures-rs/pull/1288

PR #1268, opened 2018-09-25
[WIP] Update to new nightly API

cramertj:
TODO: decide what to do w/ compat, executor trait.

Where should the executor trait go? I know @carllerche at least prefers an executor trait that is generic on the type of future it accepts, so I don't think the exact API is totally decided. `-util` seems wrong, but I hesitate to add another crate for it (especially since futures-executor is the good name, and that's already taken).

Includes some changes from https://github.com/rust-lang-nursery/futures-rs/pull/1266 (thanks, @kroisse!)


Nemo157:
@cramertj are you thinking to rename `Spawn` back to `Executor`? There's the existing https://crates.io/crates/futures-spawn crate for 0.1 for this abstraction, it even suggests that it may become part of `futures-rs` at some point.

cramertj:
@Nemo157 I've re-added it as `Spawn` and `LocalSpawn` here.

Cargo.toml
```diff
@@ -9,3 +9,6 @@ members = [
   "futures-util",
   "futures-test",
 ]
+
+[patch.crates-io]
+pin-utils = { git = "https://github.com/Kroisse/pin-utils.git", branch = "pin" }
```
aturon: Do we plan on publishing the needed changes here?

cramertj: These are actually already published-- I just have to update the path here.

PR #1269, opened 2018-09-26
Use io::ErrorKind's Into<io::Error> trait

Thomasdezeeuw:
This avoid allocating.


seanmonstar:
These error situations should be rather rare, so the allocation won't be making much of a difference.

Thomasdezeeuw:
@seanmonstar True, however I don't think it complicates the code, so I don't see any reason why not to change the code.

cramertj:
Thanks for the PR!

PR #1273, opened 2018-10-02
Stage 0.3.0-alpha.7 release

cramertj:


Nemo157:
I believe there's some doc attributes that need updating for every version as well?

futures-util/Cargo.toml
```diff
@@ -23,16 +23,16 @@ bench = []
...
-futures-sink-preview = { path = "../futures-sink", version = "0.3.0-alpha.6", default-features = false}
+futures-core-preview = { path = "../futures-core", version = "0.3.0-alpha.7", default-features = false }
+futures-channel-preview = { path = "../futures-channel", version = "0.3.0-alpha.7", default-features = false }
+futures-io-preview = { path = "../futures-io", version = "0.3.0-alpha.7", default-features = false }
+futures-sink-preview = { path = "../futures-sink", version = "0.3.0-alpha.7", default-features = false}
 either = { version = "1.4", default-features = false }
 slab = { version = "0.4", optional = true }
 futures = { version = "0.1", optional = true }
 tokio-executor = { version = "0.1.2", optional = true }
 pin-utils = "0.1.0-alpha.2"
```
Nemo157: This version doesn't build on latest nightly, should be bumped to `0.1.0-alpha.3` to force a `Cargo.lock` update. (I haven't checked whether any other crates directly depend on it).

Nemo157: Opened https://github.com/rust-lang-nursery/futures-rs/pull/1274 to update this and re-enable the minimal versions testing

cramertj: Good catch! Thanks!

cramertj:
Updated.

aturon:
LGTM, modulo travis.

JesseWright:
I might have missed something, but I think the compat feature is broken: compat03to01.rs still includes PinMut instead of Pin after #1268. Should that go out with alpha 7?

cramertj:
@JesseWright Huh, weird-- I was definitely building with it working earlier, but perhaps it got caught twice by one of my regexes. Thanks for pointing it out!

cramertj:
@aturon travis has been angry because of https://github.com/rust-lang/rust/issues/43466#issuecomment-426039999

PR #1276, opened 2018-10-03
Add io-compat feature for AsyncRead/AsyncWrite

cramertj:


cramertj:
Added a shim for converting a type which implements both `Spawn` and `SpawnExt` into an executor, which is a bit fiddly, but should still be usable.

futures-util/src/io/mod.rs
```diff
@@ -221,6 +223,16 @@ pub trait AsyncReadExt: AsyncRead {
     {
         split::split(self)
     }
+
+    /// Wraps an [`AsyncRead`] in a compatibility wrapper that allows it to be
+    /// used as a futures 0.1 / tokio-io 0.1 `AsyncRead`.
```
Nemo157: Maybe mention that this will simultaneously support `AsyncWrite` if implemented for bidirectional IO?

cramertj: Done.

PR #1277, opened 2018-10-03
0.2: fix rustdoc yanked-redirect path

seanmonstar:
It appears that the path `.cargo/registry/src/github.com-1ecc6299db9ec823/` is constant. If a different version is used to actually publish, this needs adjusting.


seanmonstar:
Looking through all the published versions of [pwnies](https://docs.rs/crate/pwnies), that path is always the same, and I just checked on my machine, and it's also the same.

cramertj:
That's... interesting. But I'm happy to try it! XD

PR #1279, opened 2018-10-05
Test with all features in CI

cramertj:


futures-util/Cargo.toml
```diff
@@ -30,11 +32,12 @@ futures-io-preview = { path = "../futures-io", version = "0.3.0-alpha.7", defaul
 futures-sink-preview = { path = "../futures-sink", version = "0.3.0-alpha.7", default-features = false}
 either = { version = "1.4", default-features = false }
 slab = { version = "0.4", optional = true }
-futures = { version = "0.1", optional = true }
+futures_01 = { version = "0.1", optional = true, package = "futures" }
```
cramertj: This rename is needed in order to allow both this and the `futures-preview` crate to be included at the same time when running tests.

Nemo157: Hopefully the index fixes for renaming should be working for the next release :)

PR #1280, opened 2018-10-05
 0.2 nuclear option to fix docs.rs build 

seanmonstar:
The [build log](https://docs.rs/crate/futures/0.2.3-docs-yank.3/builds/119437) suggests some weirdness where it was trying to use a `futures-util-preview` of `0.3.0-alpha`, even though it supposedly requires `0.2.2`. Instead of trying to fix that, just remove all the dependency stuff and simply publish an empty crate with a docs redirect.


cramertj:
lol, sgtm

Issue #1251, opened 2018-09-04
panicked at 'no Task is currently running' for tokio::timer::Delay when using block_on


redtankd:
the code is 

```
use futures::compat::{Future01CompatExt, TokioDefaultSpawner};
use futures::executor::{block_on, ThreadPoolBuilder};
use tokio::timer::Delay;

#[test]
    fn job_on_time() {
        let when = Instant::now() + Duration::from_millis(100);
        assert_eq!(
            1,
            block_on(Delay::new(when).compat().map(|x| {
                assert_eq!((), x.unwrap());
                1
            }))
        );
    }
```

the error message is 
```
---- tests::job_on_time stdout ----
thread 'tests::job_on_time' panicked at 'no Task is currently running', libcore/option.rs:1000:5
note: Run with `RUST_BACKTRACE=1` for a backtrace.
```

the  versions are
```
futures-preview = { version = "0.3.0-alpha.5", features = ["tokio-compat","compat"] }
tokio = "0.1.8"
```

Nemo157:
This is a limitation of the current compat layer, a wrapped 0.1 future still needs an ambient `futures(0.1)::task::Task` available, which 0.3 executors don't setup.

It might be possible to setup the `Task` automatically, but last time I looked at the 0.1 docs for this I didn't see any way to detect whether a task is currently available (except `std::panic::catch_unwind`, but we really don't want that on this path).

Alternatively it's probably easier to have an extra compat wrapper for 0.1 -> 0.3 that sets up the task always, but that will almost certainly panic if either run on a 0.1 executor, or nested in itself.

Nemo157:
(as a side note: even once the `Task` is available Tokio futures will still almost certainly not work on a 0.3 executor since they require an ambient `tokio::reactor::Reactor` as well, it seems like it should be possible to run a standalone reactor next to a 0.3 executor, but I don't know how ergonomic using that would be. It's much easier to just run Tokio futures on Tokio's executor for now).

redtankd:
ok, thank you very much

carllerche:
@redtankd In general, if you want to use any Tokio types, you should use a Tokio runtime.

cramertj:
@Nemo157 
> This is a limitation of the current compat layer, a wrapped 0.1 future still needs an ambient futures(0.1)::task::Task available, which 0.3 executors don't setup.

[...]

PR #1282, opened 2018-10-08
Spawn 0.1 futures when using them as 0.3 futures

cramertj:
This eliminates the need for separate configuration of a global task
when using 0.1 futures on an 0.3 executor.

Fixes https://github.com/rust-lang-nursery/futures-rs/issues/1251


cramertj:
Needs https://github.com/rust-lang-nursery/futures-rs/pull/1281 to avoid regressing `AsyncRead`/`AsyncWrite` support.

seanmonstar:
I believe this will make intermediate futures that go through `compat` much more expensive, even though they don't actually need their own task allocated.

seanmonstar:
I'm thinking it'd be better to modify futures 0.1 to allow a `with_notify` that doesn't need a task previously set.

Nemo157:
I also wonder what this means for when you're using a 0.1 executor, running some 0.3 async functions and terminating in 0.1 leaf I/O (the common case for the compat layers I assume). The 0.3 async functions -> 0.1 leaf I/O boundary will allocate a new `Spawn` separate from the one that the executor is running the task under. Wake notifications should flow back from the 0.1 leaf I/O notify ->  0.3 waker -> 0.1 executor notify, but I don't have enough familiarity with 0.1 internals to know whether the `Spawn` instance can include other "global" state that may need to be synchronized between the executor and leaf I/O.

cramertj:
> I'm thinking it'd be better to modify futures 0.1 to allow a with_notify that doesn't need a task previously set.

That wouldn't allow task-local data.

Why do you think this is much more expensive?

seanmonstar:
> That wouldn't allow task-local data.

I'm pretty sure task-local data won't survive this either. If you've gone through the compat layer once, a task-local key from higher up won't be visible in this `Spawn`'s map.

cramertj:
A task-local key from higher up will not, that is true-- this encapsulates task-local data, which I believe is the only sensible behavior when looking to run an 0.1 future on an 0.3 executor.

seanmonstar:
I re-read through the source of `Spawn`, and it seems I assumed it automatically boxed the future. I was worried about introducing dynamic dispatch when intermediate futures could probably just be inline.

cramertj:
@seanmonstar Gotcha! Yeah, that would be unfortunate. How do you feel about this PR now? Would you still prefer a different approach, or do you think this works?

cramertj:
@seanmonstar Makes sense-- thanks for the review!

PR #1284, opened 2018-10-09
Fix an invalid cast in the impl of UnsafeFutureObj for Pin<Box<F>>

andreasots:
It's currently trying to drop a `Box<Pin<Box<T>>>` from a `Box<F>`.

### DIFF
--- a/futures/tests/future_obj.rs
+++ b/futures/tests/future_obj.rs
@@ -1,0 +1,35 @@
+#![feature(pin, async_await, arbitrary_self_types, futures_api)]
+
+use futures::future::{Future, FutureExt, FutureObj};
+use std::pin::Pin;
+use futures::task::{LocalWaker, Poll};
+
+#[test]
+fn dropping_does_not_segfault() {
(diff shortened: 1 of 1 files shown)


Nemo157:
This could be improved by adding a couple of extra functions to `std` as well:

```rust
impl<T: ?Sized> Box<T> {
     pub fn pinned_into_raw(b: Pin<Box<T>>) -> *mut T;
     pub unsafe fn pinned_from_raw(raw: *mut T) -> Pin<Box<T>>;
}
```

but this fix looks correct enough for now :+1:

cramertj:
Can you add a test for this to ensure that the behavior doesn't regress? Things like this seem to keep getting misaligned during refactors of the Pin stuff and the future API.

cramertj:
Ping @andreasots -- will you be able to add the test for this?

andreasots:
Sorry for the delay. I've added two tests: one is the snippet I discovered this with and the other checks that the wrapped future's drop code runs exactly once.

PR #1286, opened 2018-10-13
Fix the stack overflow in the compat layers (#1283)

andreasots:
Since the `LocalWaker` is dropped immediately after polling the `task01::Task` can be stored on the stack.

Fixes #1283.

PR #1287, opened 2018-10-15
Remove feature(tool_lints), it is stable

Nemo157:
Fixes failure in https://travis-ci.org/rust-lang-nursery/futures-rs/jobs/441068122 (for #1286). Not tested in CI since clippy is unavailable on latest nightly, but tested locally with `nightly-2018-10-14`.


cramertj:
Thanks!

Issue #1296, opened 2018-10-19
futures-core-preview broken on latest nightly


tikue:
I think https://github.com/rust-lang/rust/pull/54671 might have broken futures-core-preview when the std feature is enabled (which it is by default).

Lots of errors spit out, e.g.:

```
error[E0433]: failed to resolve. Could not find `std` in `{{root}}`                                                                                                                                                                                                                       
   --> futures-core/src/future/future_obj.rs:194:9                                                                                                                                                                                                                                        
    |                                                                                                                                                                                                                                                                                     
194 |     use std::boxed::Box;                                                                                                                                                                                                                                                            
    |         ^^^ Could not find `std` in `{{root}}` 
```

PR #1297, opened 2018-10-19
no_std only when std feature is not enabled

cramertj:
Fix https://github.com/rust-lang-nursery/futures-rs/issues/1296


cramertj:
Well, it used to work on the 2018 edition, but then I complained and it got fixed! XD go figure

cramertj:
(see https://github.com/rust-lang/rust/issues/53166 for context)

PR #1302, opened 2018-10-24
Don't drop the Pin<Box> being converted into the UnsafeStreamObj

### LINKED ISSUE #1301: UnsafeStreamObj impl for Pin<Box<T>> has use-after-free
https://github.com/rust-lang-nursery/futures-rs/blob/0.3.0-alpha.9/futures-core/src/stream/stream_obj.rs#L219-L221

`self` is dropped inside the function, so its pointer points to garbage after it returns. I hit this in my crate that crashed when polling a `LocalStreamObj` created via `LocalStreamObj::new(Box::pinned(some_stream))`

Before https://github.com/rust-lang-nursery/futures-rs/commit/1d936e18812d7c4116a62b61e2306f36457b4f15 this used to use `PinBox::into_raw`. Since `Pin` has no equivalent now, I suppose the fix will be something like:

```diff
         fn into_raw(mut self) -> *mut () {
-            unsafe { Pin::get_mut_unchecked(Pin::as_mut(&mut self)) as *mut F as *mut () }
+            let result = unsafe { Pin::get_mut_unchecked(Pin::as_mut(&mut self)) as *mut F as *mut () };
+            core::mem::forget(self);
+            result
```

Edit: The `UnsafeFutureObj` impl for `Pin<Box<T>>` already does `mem::forget` in this way, so yes this fix should be applied to `UnsafeStreamObj` too.

Arnavion:
Fixes #1301

### DIFF
--- a/futures-core/src/stream/stream_obj.rs
+++ b/futures-core/src/stream/stream_obj.rs
@@ -2,6 +2,7 @@ use super::Stream;
 use crate::task::{LocalWaker, Poll};
 use core::fmt;
 use core::marker::{PhantomData, Unpin};
+use core::mem;
 use core::pin::Pin;
 
 /// A custom trait object for polling streams, roughly akin to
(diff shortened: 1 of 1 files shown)


seanmonstar:
Great catch!

Issue #1272, opened 2018-09-30
Does pinning allow other components to safely store pointers to a Future during poll?


Matthias247:
First the question:

If `poll(self: Pin<&mut Self>, ...)` is called, it means that the address of the the future not only constant during this call to `poll`, but also between multiple invocations to `poll`, right? Because the normal reference would already guarantee a stable address for a single call.

Does that mean we can store `self.deref() as Self*` during the call to `poll()`, and assume it to exist until the destructor of the `Future` is called?

What I am trying to understand is whether pinning allows us utilize some of the same patterns that are sometimes used in synchronous code also in asynchronous code. E.g. in synchronous multithreaded code I would sometimes use a pattern like the following, which exploits that the client thread is blocked to safely share it's memory with another processing thread without copying (sorry for pseudo C++ code):

 ```

struct MulOp {
    uint32_t* a;
    uint32_t* b;
    uint32* result;
    ManualResetEvent& event;
};

BlockingQueue<MulOp> op_queue;

void client_func() {
    ManualResetEvent ev;
    uint32_t a = 100;
    uint32_t b = 20;
    uint32_t res = 0;

    MulOp op {&a, &b, &res, &ev};
    op_queue.submit(&op);

    ev.wait();
    assert(res == 2000);
}

void service_thread_func() {
     while (true) {
         MulOp* op = op_queue.deque();
         *(op->res) = *(op->a) * *(op->b);
         op->event->signal();
     }
}

```

That is obviously a trivial example. In more real-world examples one would use this pattern to sure that e.g. a network or another shared resource is only utilized from a certain thread which has ownership over that resource. And I would sometime try to exploit things even more, e.g. by making things like `MulOp` an intrusive list element, which could then be in the service thread queued without any memory requirements (storage is on the clients threads stack).

I'm now trying to check whether the same pattern is possible with `Futures` - and to be clear: I'm not too concerned whether it requires `unsafe` code or not. The goal is to minimize the amount of data that must be "secured" with another heap allocation, be moved behind an ARC, etc.

As far as I understand it should be possible with the `Pin` guarantees I described above. We can hand out pointers to members of the future, which are equivalent to variables on the stack. We only need to guarantee that dropping the `Future` cleans up all references to itself. (This might be a harder task, since dropping needs to be synchronous and still shouldn't block, but the cancellation operation might require more complex coordination and waiting. But that's another discussion).

Nemo157:
[...]

PR #1315, opened 2018-11-07
Add a new ManualResetEvent synchronization primitive

Matthias247:
The primitive can be used to let one or multiple tasks wait for a certain event.
Compared to the existing oneshot channel, the difference is that the ManualResetEvent doesn't carry a value, and is reusable.

In addition to the thread-safe ManualResetEvent a non-allocating LocalManualResetEvent variant is provided. This can be e.g. used as a cancellation token in nested asynchronous functions.

This CR is currently a bit raw and mostly about asking for feedback. The goal was not only to implement the primitive, but also to try to see how far one get with non-allocating futures by exploiting Pinning (as described in #1272). This turned out to work quite well. Even though the synchronous cancellation requirement for `Drop()` requires the usage of a `Mutex` and thereby makes things more complicated compared to the C++ implementation at https://lewissbaker.github.io/2017/11/17/understanding-operator-co-await.

For simplicity things have been added into the channel module and into some async/await tests. However another module might be better suited. Also better documentation is missing. And the usage of the pinning APIs could certainly be improved.

Some use-cases for the primitive:
- Using it for implementing poll_ready() on higher level streams, e.g. TCP/HTTP/TLS/etc streams
- Using it as a cancellation token

### DIFF
--- a/futures-channel/src/manual_reset_event.rs
+++ b/futures-channel/src/manual_reset_event.rs
@@ -119,7 +119,7 @@ impl InnerEventState {
             self.waiters = null_mut();
 
             unsafe {
-                while waiter != null_mut() {
+                while !waiter.is_null() {
                     let task = (*waiter).task.take();
                     if let Some(ref handle) = task {
                         handle.wake();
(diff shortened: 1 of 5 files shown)


futures-channel/src/manual_reset_event.rs
```diff
@@ -0,0 +1,351 @@
...
+        self.inner.borrow_mut().reset()
+    }
+
+    /// Returns whether the event is set
+    pub fn is_set(&self) -> bool {
+        self.inner.borrow().is_set()
+    }
+
+    /// Returns a future that gets fulfilled when the event is set.
+    pub fn poll_set(&'a self) -> impl Future<Output = ()> + FusedFuture + 'a {
```
Nemo157: Generally `poll_*` functions are ones that return `Poll<_>`, like `InnerEventWaiter::poll_waiter`. Is there some other appropriate name for this? Maybe just `wait`?

Nemo157:
The intrusive list part of this looks correct to me (although, I haven't really thought about panic safety). Is it possible to pull it out to a module that provides a fully safe generic API? That would remove all unsafety except pin-projection from the actual `ManualResetEvent` code.

futures/tests/async_await_macros.rs
```diff
@@ -197,3 +200,91 @@ fn try_join_doesnt_require_unpin() {
...
+    });
+
+    block_on(async {
+        let wait1 = ev.poll_set();
+        let wait2 = ev2.poll_set();
+
+        let mut nr_selects = 0;
+
+        // Since the futures from poll_set are Unpin, they must be alias by a reference which is
+        // !Unpin for select
```
Nemo157: The inverse, `poll_set` returns an `impl Future + !Unpin` which `pin_mut!()` allows you to treat as an `impl Future + Unpin` by pre-pinning it to the current stack.

futures-channel/src/manual_reset_event.rs
Matthias247: That's right! I derived the name from the existing async code like AsyncRead. But since those are not really awaitable futures the situation is a bit different.
How about `get_awaiter()`? That's a bit aligned with the C# terminology. `get_future()` is obviously also ok, but less specific.

futures/tests/async_await_macros.rs
Matthias247: Thanks for catching it, I got confused while writing this. Changed it.

Matthias247:
> The intrusive list part of this looks correct to me (although, I haven't really thought about panic safety). Is it possible to pull it out to a module that provides a fully safe generic API? That would remove all unsafety except pin-projection from the actual `ManualResetEvent` code.

It probably is, but I'm not sure if it's beneficial as long as it's not reused by another module (like an async Semaphore for example). Currently I think I find it preferable to understand all the logic in a single place, since it isn't that much code.

Matthias247:
Replaced by #1389

PR #1323, opened 2018-11-10
Add FuturesUnordered::next_some

cramertj:


Nemo157:
Having `FusedFuture::is_terminated()` transition from `true` -> `false` seems surprising to me.

It seems like this could be provided as `StreamExt::next_some where Self: FusedStream`, then the question would be whether it’s appropriate to implement `FusedStream` for `FuturesUnordered` delegating to `is_empty` (I guess it’s about as appropriate as having it return `None` while empty, surprising but necessary for some use cases).

My other thought based on the example would be to use a channel abstraction instead of a collection so that you know once all senders have been dropped (with `buffer_unordered` to get the same concurrency). I assume that’s not wanted for performance reasons, could bringing back `futures::channel::unsync::mpsc` provide enough performance? (although, then you lose `Send` as well...)

Nemo157:
Only other thought I had on name is `next_unwrap`/`next_expect` since that’s what it’s doing other than the fusing. I think that’s probably worse than the current name though.

alexcrichton:
Ah I think I'm far enough out of the loop to not be much help on naming bikesheds :(

cramertj:
@Nemo157 
> FusedStream does not guarantee that is_terminated() will return true before Stream::poll_next returns None

Yes, this is intentional. It returns `is_terminated()` *after* `Stream::poll_next` returns `None`.

cramertj:
Superseded by #1329

PR #1329, opened 2018-11-13
Add StreamExt::next_some

Nemo157:
More general alternative to #1323, `FuturesUnordered` could use this by just adding a `FusedStream` implementation for it.

r? @cramertj

### DIFF
--- a/futures-util/src/stream/buffer_unordered.rs
+++ b/futures-util/src/stream/buffer_unordered.rs
@@ -1,6 +1,6 @@
 use crate::stream::{Fuse, FuturesUnordered};
 use futures_core::future::Future;
-use futures_core::stream::Stream;
+use futures_core::stream::{Stream, FusedStream};
 use futures_core::task::{LocalWaker, Poll};
 use futures_sink::Sink;
 use pin_utils::{unsafe_pinned, unsafe_unpinned};
@@ -131,6 +131,16 @@ where
         } else {
             Poll::Pending
         }
+    }
+}
+
+impl<St> FusedStream for BufferUnordered<St>
+where
+    St: Stream,
+    St::Item: Future,
+{
+    fn is_terminated(&self) -> bool {
+        self.in_progress_queue.is_empty() && self.stream.is_terminated()
     }
 }
 
--- a/futures-util/src/stream/buffer_unordered.rs
+++ b/futures-util/src/stream/buffer_unordered.rs
@@ -1,6 +1,6 @@
 use crate::stream::{Fuse, FuturesUnordered};
 use futures_core::future::Future;
-use futures_core::stream::{Stream, FusedStream};
+use futures_core::stream::Stream;
 use futures_core::task::{LocalWaker, Poll};
 use futures_sink::Sink;
 use pin_utils::{unsafe_pinned, unsafe_unpinned};
@@ -131,16 +131,6 @@ where
         } else {
             Poll::Pending
         }
-    }
-}
-
-impl<St> FusedStream for BufferUnordered<St>
-where
-    St: Stream,
-    St::Item: Future,
-{
-    fn is_terminated(&self) -> bool {
-        self.in_progress_queue.is_empty() && self.stream.is_terminated()
     }
 }
 
--- a/futures-util/src/stream/buffer_unordered.rs
+++ b/futures-util/src/stream/buffer_unordered.rs
@@ -1,6 +1,6 @@
 use crate::stream::{Fuse, FuturesUnordered};
 use futures_core::future::Future;
-use futures_core::stream::Stream;
+use futures_core::stream::{Stream, FusedStream};
 use futures_core::task::{LocalWaker, Poll};
 use futures_sink::Sink;
 use pin_utils::{unsafe_pinned, unsafe_unpinned};
@@ -131,6 +131,16 @@ where
         } else {
             Poll::Pending
         }
+    }
+}
+
+impl<St> FusedStream for BufferUnordered<St>
+where
+    St: Stream,
+    St::Item: Future,
+{
+    fn is_terminated(&self) -> bool {
+        self.in_progress_queue.is_terminated() && self.stream.is_terminated()
     }
 }
 
--- a/futures-util/src/stream/futures_unordered/mod.rs
+++ b/futures-util/src/stream/futures_unordered/mod.rs
@@ -49,6 +49,9 @@ pub struct FuturesUnordered<Fut> {
     ready_to_run_queue: Arc<ReadyToRunQueue<Fut>>,
     len: usize,
     head_all: *const Task<Fut>,
+    /// Track whether we have yielded `None` and can consider ourselves
+    /// terminated
+    is_terminated: bool,
 }
 
 unsafe impl<Fut: Send> Send for FuturesUnordered<Fut> {}
@@ -123,6 +126,7 @@ impl<Fut: Future> FuturesUnordered<Fut> {
             len: 0,
             head_all: ptr::null_mut(),
             ready_to_run_queue,
+            is_terminated: false,
         }
     }
 }
@@ -172,6 +176,10 @@ impl<Fut> FuturesUnordered<Fut> {
         // futures are ready. To do that we unconditionally enqueue it for
         // polling here.
         self.ready_to_run_queue.enqueue(ptr);
+
+        // If we've previously marked ourselves as terminated we need to clear
+        // that now that we will yield a new item.
+        self.is_terminated = false;
     }
 
     /// Returns an iterator that allows modifying each future in the set.
@@ -284,6 +292,9 @@ impl<Fut: Future> Stream for FuturesUnordered<Fut> {
             let task = match unsafe { self.ready_to_run_queue.dequeue() } {
                 Dequeue::Empty => {
                     if self.is_empty() {
+                        // We can only consider ourselves terminated once we
+                        // have yielded a `None`
+                        self.is_terminated = true;
                         return Poll::Ready(None);
                     } else {
                         return Poll::Pending;
@@ -472,6 +483,6 @@ where
 
 impl<Fut: Future> FusedStream for FuturesUnordered<Fut> {
     fn is_terminated(&self) -> bool {
-        self.is_empty()
+        self.is_terminated
     }
 }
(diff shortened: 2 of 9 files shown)


Nemo157:
I realized we don't need `is_terminated` able to peek to see if the stream is done, instead we can simply see the `None` then yield to allow `select!` to check `is_terminated` again.

futures-util/src/stream/buffer_unordered.rs
```diff
@@ -134,6 +134,16 @@ where
     }
 }
 
+impl<St> FusedStream for BufferUnordered<St>
+where
+    St: Stream,
+    St::Item: Future,
+{
+    fn is_terminated(&self) -> bool {
+        self.in_progress_queue.is_empty() && self.stream.is_terminated()
```
Nemo157: This can return `true` before `poll` returns a `None`. Is that valid for `FusedStream`, or does this need to track whether it's returned a `None` before it can return `true`?

cramertj:
This isn't the same as my other PR-- the behavior there is *specific* to `FuturesUnordered`, because `FuturesUnordered`can become `Some` again after yielding `None`. This may be something that could be useful in cases where you don't care to have special behavior on the end of the stream, but it's a different feature.

cramertj:
I think the behavior here is too subtle-- it only really makes sense in the context of `select!`.

futures-util/src/stream/next_some.rs
```diff
@@ -0,0 +1,41 @@
...
+
+impl<'a, St> NextSome<'a, St> {
+    pub(super) fn new(stream: &'a mut St) -> Self {
+        NextSome { stream: Some(stream) }
+    }
+}
+
+impl<'a, St: FusedStream> FusedFuture for NextSome<'a, St> {
+    fn is_terminated(&self) -> bool {
+        self.stream.as_ref().map(|s| s.is_terminated()).unwrap_or(true)
```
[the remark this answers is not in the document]
Nemo157: Hmm, I don't see how.

If the stream has yet to yield `None` then this delegates to the streams `is_terminated()` implementation (this can probably be dropped if it is a contract violation for `FusedStream::is_terminated` to return `true` before yielding `None`, it's just an optimization for streams that can know they're terminated before producing the sentinel element). For a normal stream this will return `false` and the future will go on to be polled.

Once the stream returns `None` then this will return `true`.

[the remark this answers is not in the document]
Nemo157: I guess it's possible to spin loop by constructing multiple of these futures and polling them each once.

That could be avoided by extending the panic to also occur when the stream is terminated, even without polling it.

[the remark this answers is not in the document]
cramertj: Oh, I understand now. The next time around after they are woken up, the stream will be `None` and the callers of `poll` will get a panic if they're not checking `is_terminated()`. That works, and is a correct implementation of this interface that could be useful in combination with `select!`, but it seems like a dangerous footgun everywhere else, since unless you're checking `is_terminated()` it's basically `await!(s.next()).unwrap()`. This version of `next_some` also doesn't work for the original purpose I added `FuturesUnordered::next_some`, which is special because it can become `Some` again after yielding `None`. I think I'd still be fine w/ adding it just for use w/ `select!`, but we'll have to come up w/ a good name for it to make it clear it should only be used in `select!`, as well as come up w/ a different name for the method on `FuturesUnordered`.

futures-util/src/stream/next_some.rs
```diff
@@ -0,0 +1,41 @@
...
+    type Output = St::Item;
+
+    fn poll(mut self: Pin<&mut Self>, lw: &LocalWaker) -> Poll<Self::Output> {
+        let stream = self.stream.as_mut().expect("NextSome polled after terminated");
+
+        if let Some(item) = ready!(stream.poll_next_unpin(lw)) {
+            Poll::Ready(item)
+        } else {
+            self.stream = None;
+            lw.wake();
```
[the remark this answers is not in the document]
Nemo157: We don't want them to poll, but we do want them to check `is_terminated`. This should deadlock without it when it reaches the end of a stream, but I can't seem to make that happen in a test case...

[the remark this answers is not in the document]
cramertj: Yeah, I responded below.

futures-util/src/stream/next_some.rs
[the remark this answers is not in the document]
Nemo157: Yep, it's definitely just intended for use in `select!` when you don't want to do something special on end-of-stream, and I have no idea of a decent name for it 😦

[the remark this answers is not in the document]
cramertj: Maybe `select_only_some()`? But finding two different names for this and the one on `FuturesUnordered` is tough.

futures-util/src/stream/next_some.rs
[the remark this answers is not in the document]
Nemo157: Actually, I did manage to come up with a test case that shows this is a deadlock without the waking:

```rust
llet mut future = async {                                                                                                                                                              let mut tasks = stream::iter(vec![5]).fuse();
    loop {
        select! {
            num = tasks.next_some() => {
            }
            complete => break,
        }
    }
};
assert_eq!(futures::executor::block_on(future), ());
```

My previous attempts were all using `buffer_unordered` and didn't deadlock, it seems like it might trigger some spurious wakeups.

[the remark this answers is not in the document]
cramertj: Yup, that makes sense.

Nemo157:
I see now why this isn't equivalent to your `FuturesUnordered::next_some`, crucially having to explicitly handle a sentinel `None` from a normal stream means that this does not have the "reset" behaviour of being able to yield items again after the stream has ended.

Thinking about it a bit more this is possible to support as well, if it's valid for `FusedStream::is_terminated` to transition from `true` to `false`, then this can keep the stream around after seeing a `None` and check whether the stream has resumed yielding items. Then having

```rust
impl FusedStream for FuturesUnordererd {
    fn is_terminated(&self) -> bool {
        self.is_empty()
    }
}
```

would, I believe, give identical behaviour to your PR.

Nemo157:
Pushed an update that should give that behaviour, and fixes any potential for a spin-loop if this future is reconstructed on the same stream between polls.

cramertj:
```rust
impl FusedStream for FuturesUnordererd {
    fn is_terminated(&self) -> bool {
        self.is_empty()
    }
}
```
Yes, having this impl would be enough, but this impl is also surprising in that it doesn't satisfy the usual contract of `is_terminated` on `Stream`s-- that is, it wouldn't ever yield `None` before `is_terminated` returned true. That's inconsistent with how every other `Stream` works.

futures-util/src/stream/buffer_unordered.rs
```diff
@@ -134,16 +134,6 @@ where
     }
 }
 
-impl<St> FusedStream for BufferUnordered<St>
```
cramertj: I think we can keep this-- it should just delegate to the underlying `FuturesUnordered`, right?

futures-util/src/stream/mod.rs
```diff
@@ -1095,7 +1095,7 @@ pub trait StreamExt: Stream {
     /// let (tx, mut rx) = mpsc::channel(1);
     /// let mut total = 0;
     /// let mut tx = Some(tx);
-    /// let mut async_tasks = rx.buffer_unordered(2);
+    /// let mut async_tasks = rx.buffer_unordered(2).fuse();
```
cramertj: I'd really like to show an example like the one I have [here](https://github.com/rust-lang-nursery/futures-rs/pull/1323/files#diff-bdc0724f41a9872d82003eae02367ecbR180) where we push new tasks onto a `FuturesUnordered` after the loop has started, since I think that's a key usecase that is the primary reason (IMO) for this feature.

futures-util/src/stream/mod.rs
```diff
@@ -1072,8 +1072,8 @@ pub trait StreamExt: Stream {
     /// resolve to [`None`] if used on an empty [`Stream`]. Instead, the
     /// returned future type will return [`true`] from
     /// [`FusedFuture::is_terminated`][] when the [`Stream`] is empty, allowing
```
cramertj: Looks like you're missing a link here

Nemo157: It's below, the full path is very long so I found it best to keep it out of the prose.

futures-util/tests/select_next_some.rs
```diff
@@ -7,26 +7,23 @@ use futures_test::future::FutureTestExt;
 
 #[test]
 fn futures_unordered() {
+    // Checks that even though `async_tasks` will yield a `None` and return
```
cramertj: Can you add another more directed test for this specific behavior? Due to random polling order, the `select!` in this example only tests this property with some probability.

futures-util/src/stream/mod.rs
Nemo157: Yep, I'll use that example as is, it's definitely more succinct than the channel based one.

futures-util/tests/select_next_some.rs
Nemo157: Doesn't the `pending_once` guarantee this?

First poll of the `select!` generated future must poll both futures, `fut` returns `Pending` because of the `pending_once`, `async_tasks` yields `None` so `select_next_some()` returns `Pending`, all futures are pending so the `select!` yields `Pending`.

cramertj: Ah, i missed the `pending_once`. Yes, that's  sufficient, but would you mind making a more targeted test anyways? This feels more like an integration test and it's hard to tell exactly what components are being tested.

futures-util/src/stream/mod.rs
cramertj: Thanks! :)

cramertj:
Awesome, I think this is way nicer than my previous attempt!

futures-util/src/stream/futures_unordered/mod.rs
```diff
@@ -49,6 +49,9 @@ pub struct FuturesUnordered<Fut> {
     ready_to_run_queue: Arc<ReadyToRunQueue<Fut>>,
     len: usize,
     head_all: *const Task<Fut>,
+    /// Track whether we have yielded `None` and can consider ourselves
+    /// terminated
+    is_terminated: bool,
```
cramertj: I'd really like to figure out how to pack this into the same size somehow. Maybe we could set `len` to `usize::MAX` or something as a discriminant? Seems funky, but might work okay.

Nemo157: Yep, I think that should be relatively straightforward.

PR #1331, opened 2018-11-14
Add AtomicWaker::take method

withoutboats:
This allows the user to take the waker from an AtomicWaker without waking it immediately, making taking the waker and waking it non-atomic if desired.

This has empirical use cases in the ecosystem; for example, tokio has forked `AtomicTask` (the previous version of this API) and added this method (missing out on later optimizations because of the fork).

### DIFF
--- a/futures-core/src/task/__internal/atomic_waker.rs
+++ b/futures-core/src/task/__internal/atomic_waker.rs
@@ -340,7 +340,7 @@ impl AtomicWaker {
                 // Release the lock
                 self.state.fetch_and(!(WAKING | HOLDS), Release);
 
-                return waker;
+                waker
             }
             state => {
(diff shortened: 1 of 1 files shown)


futures-core/src/task/__internal/atomic_waker.rs
```diff
@@ -326,9 +340,7 @@ impl AtomicWaker {
                 // Release the lock
                 self.state.fetch_and(!(WAKING | HOLDS), Release);
 
-                if let Some(waker) = waker {
-                    waker.wake();
-                }
+                return waker;
```
cramertj: clippy is mad about your tail-position `return`s

withoutboats: this is the Good Code but i'll change it

PR #1332, opened 2018-11-14
Resolve remaining comments from #1329

Nemo157:
r? @cramertj

### DIFF
--- a/futures/src/lib.rs
+++ b/futures/src/lib.rs
@@ -174,7 +174,7 @@ pub mod future {
     //!   immediate defined value.
 
     pub use futures_core::future::{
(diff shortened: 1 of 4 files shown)


futures/src/lib.rs
```diff
@@ -174,7 +174,7 @@ pub mod future {
     //!   immediate defined value.
 
     pub use futures_core::future::{
-        Future, TryFuture,
+        Future, TryFuture, FusedFuture,
```
Nemo157: I assume there was no reason these weren't re-exported.

cramertj:
This looks great, thanks so much!

PR #1333, opened 2018-11-15
Remove the usage of LocalPool for block_on.

Matthias247:
It should be no longer necessary, since there is no Spawner passed along the poll chain anymore.
This makes `block_on` cheaper, by avoiding the allocation for `LocalPool`. It should be allocation free after the first usage.


cramertj:
Neat! Thanks.

PR #1341, opened 2018-11-19
Port futures_util::future::join_all to the new API

tomhoule:
This is a very naive port (basically I just made sure it compiles and that a few tests pass).

Open questions:

- Are there feature flags to consider?
- I assumed we would want the futures to be `Unpin` because they move into the vector, but I can't say my understanding of pinning is crystal-clear yet.
- Do we want to be able to construct a `JoinAll` from an iterator of non-future values like before (apparently)
- Do we want also want to implement the `Try` version in this PR?

As I said this is a naive port, if this PR does not seem viable please do feel free to just close it.


Nemo157:
> I assumed we would want the futures to be `Unpin` because they move into the vector

This would be fine initially, but could be relaxed by converting the vector into a `Pin<Box<[ElemState<T>]>>` (since its length does not change after creation). That would require a little more `unsafe` code to do the pin projection though.

cramertj:
This looks like a great start! I'll go ahead and merge this, and then we can loosen the `Unpin` constraint and add `TryJoinAll` in a followup.

PR #1343, opened 2018-11-20
impl Default for Futures*ordered

jonhoo:
This implements `Default` for both `FuturesUnordered` and `FuturesOrdered`, which already have a `new` method that takes no argument.


cramertj:
Thanks!

PR #1345, opened 2018-11-21
Futures-aware Mutex

cramertj:
This mutex allows acquiring a lock asynchronously, which prevents locking from blocking the current thread. Once acquired, the resulting guard can also be sent across threads (implements `Send`) unlike `std::sync::MutexGuard`, allowing the `async { ... }` block which holds the lock to be `Send`.


futures-util/src/lock/mutex.rs
```diff
@@ -0,0 +1,246 @@
...
+            .finish()
+    }
+}
+
+impl<T> Drop for MutexGuard<'_, T> {
+    fn drop(&mut self) {
+        let old_state = self.mutex.state.fetch_and(!IS_LOCKED, Ordering::AcqRel);
+        if (old_state & HAS_WAITERS) != 0 {
+            let mut waiters = self.mutex.waiters.lock().unwrap();
+            if let Some((_i, waiter)) = waiters.iter_mut().next() {
```
cramertj: This approach is quite unfair, but we don't really have a better way to do this with `Slab`. We could perhaps use a `VecDeque` with elements that get nulled out instead as well as an offset number that is incremented each time the queue is popped (so that waiters can find themselves and update their entries appropriately). That would give approximately FIFO order, but is a bit more complex. WDYT?

cramertj: That approach would also make it possible for a single long-waiting task that didn't advance its future to cause the wait set to grow indefinitely.

stjepang: Another option is to have `next: usize` and `prev: usize` fields in `Waiter` (but really, `u32` or even `u16` should be enough), which indicate the next and previous waiter in the linked list. Then we can build a FIFO doubly linked list.

futures/src/lib.rs
```diff
@@ -237,6 +237,12 @@ pub mod io {
     };
 }
 
+#[cfg(feature = "std")]
+pub mod lock {
+    //! Futures-powered synchronization primitives.
+    pub use futures_util::lock::{Mutex, MutexAcquire, MutexGuard};
```
stjepang: Not sure how I feel about the `MutexAcquire` name. It would be perfect if the `lock()` method was named `acquire()`. What do you think about `MutexLock`?

Matthias247:
Good idea! I thought about working on the same thing. It could be implemented it in the same non-allocating way than I did with #1315. That one should now be also fair, but would make the resulting future `!Unpin`.

Some further ideas:
- Should this also have a blocking lock method, for integration with non-futures code?
- Can there be an easy variant that uses no mutex and atomic ops and is only suitable for singlethreaded event loops? 

I guess the next thing on the useful list would be an async Semaphore.

futures-util/src/lock/mutex.rs
Matthias247: So this wakes only one waiter instead of all of them, to avoid a thundering herd? That seems reasonable (apart from the fairness question).
One question that I had in my mind when I thought about implementing this is whether this release should already mark the mutex as locked again for the other waiter, even though that one doesn't yet know about it. That would avoid the issue that another call to lock can acquire the mutex in between.
But I'm not sure if it introduces other issues, since that `MutexAcquire` would now have acquired the lock without a direct call on it. However that passively acquired lock could still be released when the `MutexAcquire ` gets dropped without another `poll` call onto it which yields the `MutexGuard`.

futures-util/src/lock/mod.rs
```diff
@@ -0,0 +1,10 @@
+//! Futures-powered synchronization primitives.
```
Matthias247: Should this be in futures-util/lock, or could there be a crate like `futures-sync[-primitives]`.

The reason for this is
- channels have their own crate
- channels might be rewritten or variants added that might make use of those more basic asynchronous sync primitives
- a Mutex only only one variant of an asynchronous sync primitive. Semaphores and Events are others (and those are not necessarily locks)

futures-util/src/lock/mutex.rs
cramertj: @Matthias247 That could possibly increase fairness, but it reduces throughput since it doesn't necessarily allow the next task to poll the lock to acquire it, and it makes it harder to recover from the case where a task is cancelled and has to figure out whether or not to pass the lock along to another waiter.

futures-util/src/lock/mutex.rs
```diff
@@ -0,0 +1,246 @@
...
+    value: UnsafeCell<T>,
+    waiters: StdMutex<Slab<Waiter>>,
+}
+
+impl<T> fmt::Debug for Mutex<T> {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        let state = self.state.load(Ordering::SeqCst);
+        f.debug_struct("Mutex")
+            .field("is_locked", &((state & IS_LOCKED) != 0))
+            .field("has_waiters", &((state & IS_LOCKED) != 0))
```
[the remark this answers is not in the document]
cramertj: heh, good catch.

futures-util/src/lock/mutex.rs
```diff
@@ -0,0 +1,246 @@
...
+        if (old_state & IS_LOCKED) == 0 {
+            Some(MutexGuard { mutex: self })
+        } else {
+            None
+        }
+    }
+
+    /// Acquire the lock asynchronously.
+    ///
+    /// This method returns a future that will resolve once the loch has been
```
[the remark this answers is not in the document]
cramertj: Done.

futures-util/src/lock/mutex.rs
```diff
@@ -0,0 +1,246 @@
...
+enum Waiter {
+    Waiting(Waker),
+    Woken,
+}
+
+impl Waiter {
+    fn register(&mut self, lw: &LocalWaker) {
+        match self {
+            Waiter::Waiting(waker) => if !lw.will_wake_nonlocal(waker) {
+                *waker = lw.clone().into_waker();
```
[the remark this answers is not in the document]
cramertj: yeah, something like that will probably appear eventually

futures-util/src/lock/mutex.rs
```diff
@@ -0,0 +1,246 @@
...
+            },
+            Waiter::Woken => *self = Waiter::Waiting(lw.clone().into_waker()),
+        }
```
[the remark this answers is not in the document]
cramertj: Nice! :)

futures/src/lib.rs
cramertj: I understand wanting to name it after the method, but IMO `MutexLock` kind of sounds like you're holding the lock, which isn't right... Hm. Maybe we should just go ahead and call it `MutexLockFuture`? We don't typically include `Future` in the names of these types, but it could be helpful.

cramertj:
> What about poisoning - do we plan to implement it?

I didn't plan to-- in my experience, it's exceedingly rare to see code that properly handles the error case of `std::sync::Mutex::lock`, and I even considered pulling in the `parking_lot` `Mutex` types (which are faster than the ones in std, but don't implement poisoning) to use for the `waiters` set. I decided against it, but I think this is something we should leave room for in the future.

futures-util/src/lock/mod.rs
cramertj: I don't have a strong opinion here, but I'd like to avoid introducing another crate now. I think this makes sense to expose through `futures-util` since futures-util is sort of the "dumping ground" for new ideas that we're not sure we want to stabilize on, and which aren't ready for people to expose through public APIs. Channels are intended to be something with a more stable API that can appear in the public API of libraries, which is why they have their own crate to avoid a semver bump when something in futures-util gets a breaking change.

cramertj:
> Another option is to have next: usize and prev: usize fields in Waiter (but really, u32 or even u16 should be enough), which indicate the next and previous waiter in the linked list. Then we can build a FIFO doubly linked list.

That's not a bad idea! If it's alright with you, though, I'd like to land this version first so we can experiment and track the performance and size changes from different modifications.

futures/src/lib.rs
cramertj: Changed to `MutexLockFuture`.

cramertj:
I think all the other changes we've discussed to this are internal to the implementation and wouldn't visibly affect end-users, so I'm going to go ahead and merge this for now and we can visit potential optimizations in future PRs.

Matthias247:
> I think all the other changes we've discussed to this are internal to the implementation and wouldn't visibly affect end-users, so I'm going to go ahead and merge this for now and we can visit potential optimizations in future PRs.

I agree on that in general. One thing that however might make sense is adding a `!Unpin` implementation for the future through `Pinned`. That would avoid breaking code if it would get transformed into a intrusive implementation.

futures-util/src/lock/mutex.rs
Matthias247: I don't think the cancellation scenario gets a lot tricker, since the future already needs to track whether it got signalled and needs to wake another waiter if that's the case during `drop`.

However it might really introduce some new issues (e.g. deadlocks) during cancellation, e.g. if the future inside a task is not immediately destructed when done but hold alive due to a still waker references that are stored somewhere. As far as I understand some executors might behave that way (even though it should certainly be discouraged).

Therefore I agree that the strategy makes sense.

cramertj:
> a intrusive implementation

What do you have in mind here that would require pinning?

Matthias247:
> What do you have in mind here that would require pinning?

See the implementation of #1315

PR #1355, opened 2018-11-27
Fused channels

cramertj:
It gets pretty annoying having to fuse these manually everywhere or pass around a `Fuse<UnboundedReceiver<T>>` etc. This is basically zero-cost, and since this will save folks from some headache I figured we might as well do it by default.

### DIFF
--- a/futures-channel/src/mpsc/mod.rs
+++ b/futures-channel/src/mpsc/mod.rs
@@ -78,7 +78,7 @@
 // happens-before semantics required for the acquire / release semantics used
 // by the queue structure.
 
-use futures_core::stream::Stream;
+use futures_core::stream::{FusedStream, Stream};
 use futures_core::task::{LocalWaker, Waker, Poll};
 use futures_core::task::__internal::AtomicWaker;
 use std::any::Any;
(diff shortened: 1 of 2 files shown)


futures-channel/src/mpsc/mod.rs
```diff
@@ -820,10 +833,15 @@ impl<T> Stream for Receiver<T> {
...
+                }
+                Poll::Ready(msg)
+            },
             Poll::Pending => {
                 // There are no messages to read, in this case, park.
-                self.inner.recv_task.register(lw);
+                self.inner.as_ref().unwrap().recv_task.register(lw);
                 // Check queue again after parking to prevent race condition:
                 // a message could be added to the queue after previous `next_message`
                 // before `register` call.
```
Nemo157: The below `self.next_message()` could return `Poll::Ready(None)` and then not update `self.inner`.

Relatedly tests are failing because `Drop` will call `next_message` at least once, even if `self.inner` is `None`; and `try_next` also calls it and I'm not sure what the best semantics for it would be.

Maybe `next_message` should return `Option<Poll<Option<T>>>` to make handling a post-termination call the callers job?

cramertj: I think it's correct for `try_next` to panic in this case as well-- I'll add a note to the docs, and fix the `Drop` impl. Thanks!

PR #1226, opened 2018-08-26
WIP: Compat blog post

MajorBreakfast:
I'm going to continue writing this tomorrow if I find the time. In the meantime, here's what I have so far 🙂

PR #1357, opened 2018-11-29
Compat blog post take 2

Nemo157:
Replacement for #1226 (@MajorBreakfast I hope you're ok with me taking this over).

I still want to add some info on the spawn and IO compat that's been added since this was originally written, and I need to actually test the snippets against the new API. Should get to that over the weekend at some point.

### DIFF
--- a/_posts/2018-01-01-compatibility-layer.md
+++ b/_posts/2018-01-01-compatibility-layer.md
@@ -83,46 +83,88 @@ It converts a 0.1 `Future<Item = T, Error = E>` into a `std` `Future<Output = Re
 
 [Future01CompatExt::compat]: https://rust-lang-nursery.github.io/futures-api-docs/0.3.0-alpha.10/futures/compat/trait.Future01CompatExt.html#method.compat
 
-## Streams
+## Streams and Sinks
 
 Converting between 0.1 and 0.3 streams is possible via the [`TryStreamExt::compat`][] and [`Stream01CompatExt::compat`][] methods. Both combinators work analogously to their future equivalents.
 
+Sinks are a little different, the current compat layer only includes support for transforming a 0.3 sink into a 0.1 sink, this goes through a [`SinkExt::compat`][] method as normal. See [rust-lang-nursery/futures-rs#1362][] for more details on the inverse transform.
+
 [`TryStreamExt::compat`]: https://rust-lang-nursery.github.io/futures-api-docs/0.3.0-alpha.10/futures/prelude/trait.TryStreamExt.html#method.compat
 [`Stream01CompatExt::compat`]: https://rust-lang-nursery.github.io/futures-api-docs/0.3.0-alpha.10/futures/compat/trait.Stream01CompatExt.html#method.compat
+[`SinkExt::compat`]: https://rust-lang-nursery.github.io/futures-api-docs/0.3.0-alpha.10/futures/prelude/trait.SinkExt.html#method.compat
+[rust-lang-nursery/futures-rs#1362]: https://github.com/rust-lang-nursery/futures-rs/issues/1362
+
+## Spawning
+
+Next, there's also conversions between futures 0.1 [`Executor`][] trait and futures 0.3 [`Spawn`][] trait. Again, these function near identically to the futures compatibility functions with [`SpawnExt::compat`][] and [`Executor01CompatExt::compat`][] providing the conversions in each direction.
+
+[`Executor`]: https://docs.rs/futures/0.1.25/futures/future/trait.Executor.html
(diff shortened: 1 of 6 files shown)


_posts/2018-01-01-compatibility-layer.md
```diff
@@ -0,0 +1,129 @@
...
+[`futures@0.1`]: https://docs.rs/futures
+[`futures@0.3`]: https://rust-lang-nursery.github.io/futures-api-docs/
+[`std::future`]: https://doc.rust-lang.org/nightly/std/future/
+
+## `Cargo.toml`
+
+The compatibility layer can be enabled by setting the `compat` feature in your `Cargo.toml`:
+
+```toml
+futures-preview = { version = "0.3.0-alpha.10", features = ["compat"] }
```
cramertj: `compat` doesn't include `io-compat`, which also pulls in `tokio-io` and adds the `AsyncRead` and `AsyncWrite` compat bits.

cramertj:
Aside from the one nit and updating to add IO/spawn info, this LGTM so far!

LucioFranco:
One, thing that might be useful to add is something about sink support with tokio. As I used this post to convert my project to async/await. I was a bit confused on how to approach sinks. I ended up with using some of the tokio compat features. So maybe a mention about it would be helpful imo.

_posts/2018-01-01-compatibility-layer.md
Nemo157: Yep, I was going to mention the extra features when adding the additional IO/spawn sections.

Nemo157:
@LucioFranco yeah, probably worth adding a section about `Sink`. Currently it's possible to wrap a 0.3 `Sink` into a 0.1 `Sink`, but not the other direction. I will also open an issue about whether it's possible to add that additional conversion, I couldn't think of an easy way when implementing the current `Sink` compat stuff but maybe it's possible with a little bit of extra runtime cost.

Nemo157:
Added info on `Spawn`, IO and `Sink`, and updated the final example to use the new IO compat layer.

LucioFranco:
@Nemo157 Great! I just took a look at it again, it def clears some things up. I am still slightly struggling to understand where the `TokioDefaultSpawner` fits into the whole picture. Also, I think the fact that there is no `0.1` -> `0.3` is a bigger issue for people using tokio. So maybe might be worthwhile to be more upfront about that?

This btw is my hacky solution to using std await and `UdpFramed` from tokio. I would be interested to help build out some of the sink compat stuff.
```rust
while let Some(msg) = await!(rx.next()) {
    println!("Sending: {:?}", msg);

    {
        use tokio::prelude::SinkExt;
        if let Err(e) = await!(sink.send_async((msg, to))) {
            println!("Sending message: {}", e);
        }
    }
}
```

I would also maybe add a `cargo.toml` snippet for the example so that its easier for people to try it out.

Nemo157:
@LucioFranco thanks, those are some good points. I think an example of using `TokioDefaultSpawner` is probably the easiest way to show what it’s useful for. I can expand a bit more on the `Sink` situation as well. And having the `Cargo.toml` for the example would definitely be worth it.

Nemo157:
Sorry about the big delay on this again, I'll try and finish this off before 2019!

Nemo157:
Updated now that `Sink` is supported and `TokioDefaultSpawner` was removed, also uses the new methods from #1469 for consistency so is blocked on that being merged and a new release.

corbinu:
@Nemo157 Read through it as somebody outside of the working team and just excited about async/await. It looks great and is easy to understand. Super excited!

archseer:
How would one go about using `select!` with tokio compatibility? I ran into some problems because `select-macro` seems to call out to the stdlib/futures implementations so it doesn't hit the [`tokio` await wrapper](https://github.com/tokio-rs/tokio/blob/9c5cad037fd78adb087bc4734f7670d5732122c5/tokio-async-await/src/await.rs).

Nemo157:
@archseer I have no real experience using Tokio's `await!` macro, but you can just use the `futures` 0.1 -> 0.3 compat methods to convert Tokio's futures into `std` futures so that they work with `std::await!`, something like

```rust
use futures::future::FutureExt;
use futures::compat::Future01CompatExt;

select! {
        foo = some_tokio_future().compat().fuse() => { ... },
        bar = some_other_tokio_future().compat().fuse() => { ... },
};
```

Nemo157:
@cramertj I've checked the example code works with the new release and all the links are correct, I think this is good to publish finally 😁

cramertj:
Nice!

PR #1361, opened 2018-12-04
0.1: Avoid eagerly running executors when waiting for FutureResult or Option

rphmeier:


rphmeier:
(I would make a 0.3 PR as well but I would appreciate some pointers as to where to look in the code for `FutureResult`, `Option` and `Either`)

rphmeier:
TBH I'm not sure why the `buffer_unordered::works` test is stalling. Locally it passes instantly.

rphmeier:
That's not ideal for cases like this. A trait `BlockOn` that is implemented by default for all futures but can be overridden in the case of "already resolved" futures would be useful.

seanmonstar:
Using `Future::wait` is often times incorrect, and thus it is deprecated. While the optimizations here seem neat, these futures don't actually do much when calling `wait` anyways. There is a `fresh_task_id()` call, a TLS access, and then `poll`. It doesn't "spin up an executor".

For those reason, I'd lean towards leaving this out, leaving it uncomplicated, and encouraging more people to **not** call `Future::wait`.

rphmeier:
@seanmonstar I see, I didn't realize `wait` was deprecated. Thanks for the feedback.

Perhaps the general idea of this PR would be something more in the vein of lazily initializing the thread-local executor when polling. Perhaps only doing the `fresh_task_id` work when `task::current` is actually called.

PR #1363, opened 2018-12-04
Rewrite select! as a procedural macro

cramertj:
This commit removes the old macro_rules tt muncher parser
implementation of the select! macro and replaces it with
a procedural macro that accomplishes the same task, but which
should be dramatically simpler to read and modify.


futures-select-macro/LICENSE-APACHE
```diff
@@ -0,0 +1,202 @@
+                              Apache License
```
Nemo157: The license files can be just symlinks to the ones in the root (as long as you don't care about publishing from windows working).

dtolnay:
Haven't read the implementation of the procedural macro yet (will get to it later today) but a quick idea regarding the duplicated documentation -- you could have futures-util provide:

```rust
#[doc(hidden)]
#[macro_export]
macro_rules! with_documentation_of_select_macro {
    ($item:item) => {
        /// Polls multiple futures and streams simultaneously, ...
        $item
    };
}
```

which both it and futures use.

I forget whether doc comments can be added to an $:item this way. If not, you'll need to make it $($tt:tt)*.

futures/src/lib.rs
```diff
@@ -384,3 +384,27 @@ pub mod task {
...
+#[doc(hidden)]
+pub use futures_util::rand_reexport;
+
+#[cfg(feature = "std")]
+#[doc(hidden)]
+pub mod inner_select {
+    pub use futures_util::select;
+}
+
+futures_util::document_select_macro! {
```
Nemo157: This is missing a `#[cfg(feature = "std")]`

futures-select-macro/Cargo.toml
```diff
@@ -0,0 +1,25 @@
...
+description = """
+The `select!` macro for waiting on multiple different `Future`s at once and handling the first one to complete.
+"""
+
+[lib]
+name = "futures_select_macro"
+proc-macro = true
+
+[features]
+std = []
```
cramertj: At some point I figured we could use this to turn off `rand` support, though I've left that unimplemented for now. Perhaps we'd require the user to provide a link-time function that could be used to shuffle arrays?

Nemo157: Yep, something like that sounds possible. I was thinking I might take a look at getting it working on `nostd`, then realised it wouldn’t work without `await!` support.

cramertj: Right, `await!` is the big blocker for getting no-std.

futures-select-macro/src/lib.rs
```diff
@@ -0,0 +1,305 @@
...
+
+    // `_1(_1), _2(_2), ..., Complete`
+    let enum_variants = variant_names.iter()
+        .map(|name| parse_quote!( #name ( #name ) ))
+        .chain(if complete { Some(unit_variant("Complete")) } else { None })
+        .collect();
+
+    (
+        variant_names,
+        syn::ItemEnum {
```
dtolnay: In my code I typically avoid constructing Syn types. Syn's parsers construct Syn types while the macro only constructs token streams. That might look something like this:

```rust
// "_0", "_1", "_2"
let variant_names: Vec<Ident> =
    (0..variants)
        .map(|num| Ident::new(&format!("_{}", num), span))
        .collect();

let type_parameters = &variant_names;
let variants = &variant_names;

let complete_variant = if complete {
    Some(quote!(Complete))
} else {
    None
};

let enum_item = quote! {
    enum #result_ident<#(#type_parameters,)*> {
        #(
            #variants(#type_parameters),
        )*
        #complete_variant
    }
};

(variant_names, enum_item)
```

Up to you which you prefer, I am also fine with your current way.

futures-select-macro/src/lib.rs
```diff
@@ -0,0 +1,305 @@
...
+    let mut future_let_bindings = Vec::with_capacity(parsed.normal_fut_exprs.len());
+    let bound_future_names: Vec<_> = parsed.normal_fut_exprs.into_iter()
+        .zip(variant_names.iter())
+        .map(|(expr, variant_name)| {
+            match expr {
+                // Don't bind futures that are already a path.
+                // This prevents creating redundant stack space
+                // for them.
+                syn::Expr::Path(path) => path,
+                x => {
```
dtolnay: I would either call this `other` instead of `x`, or use `_` here and interpolate as `#expr` below.

futures-select-macro/src/lib.rs
cramertj: Oh this is much better, thank you!

futures-select-macro/src/lib.rs
```diff
@@ -0,0 +1,289 @@
...
+
+        while !input.is_empty() {
+            let case_kind = if input.peek(kw::complete) {
+                // `complete`
+                if select.complete.is_some() {
+                    return Err(input.error("multiple `complete` cases found, only one allowed"));
+                }
+                input.parse::<kw::complete>()?;
+                CaseKind::Complete
+            } else if input.peek(Token![default]) {
```
Matthias247: What's the reason that this doesn't look exactly like the `complete` branch?

futures-select-macro/src/lib.rs
```diff
@@ -0,0 +1,289 @@
...
+        parsed.complete.is_some(),
+        span,
+    );
+
+    // bind non-`Ident` future exprs w/ `let`
+    let mut future_let_bindings = Vec::with_capacity(parsed.normal_fut_exprs.len());
+    let bound_future_names: Vec<_> = parsed.normal_fut_exprs.into_iter()
+        .zip(variant_names.iter())
+        .map(|(expr, variant_name)| {
+            match expr {
```
Matthias247: Since working with a `!Unpin` futures is a bit painful at the moment (must be manually pinned before use in `select`): Can this happen automatically here?
Maybe in a restricted form only when the `Future` is an rvalue and guaranteed to not be used outside of `select`?
I have no idea whether this can be detected, but it sounds like it could simplify some things.

Matthias247: Similar question for `Future`s which are not fused.
I guess if one wants to use the same future for each iteration in a loop then it still needs to be pinned and fused manually, but if it's a new `Future` for each iteration it might be possible to do this automatically.

futures-select-macro/src/lib.rs
dtolnay: Syn has a token for `default` because it is used as a keyword in `default fn` syntax, while `complete` is not a keyword.

futures-select-macro/src/lib.rs
Matthias247: > For loops, I think select! could somehow consume the future and do that automatically as well
I think for the situations where that's ok, one would mostly be fine by creating the future as an rvalue in this expression.

However there are situations where one really needs to operate on exactly the same future instance between loop iterations: Some `Future`s, like for the mutex and manualresetevent one, carry internal state between poll() calls, and destroying and recreating them would reset this. For those types it would e.g. mean they might never be able to acquire the mutex when getting recreated instead of staying the same - or at least the fairness guarantee would be different.

Same with using a `async fn` which contains multiple `await!`s as a Future in of the expressions: If it's recreated in each iteration, it will would always start running from the beginning. If it's kept between iterations, it keeps also it's progress. Automatically hoisting this out of the loop would be interesting, but then the concern of this macro would go beyond the poor `select!` part.

cramertj: I want to leave any functional changes to `select!` out of this change, so that it's just refactoring. If there are other features you'd like to suggest, I'd be happy to review them as separate PRs.

Issue #1362, opened 2018-12-04
Investigate compat for transforming 0.1 Sink into 0.3 Sink


Nemo157:
The PR that added `Stream` and `Sink` to the compat layers (#1162) only added support for converting a `futures(0.3)::Sink` into a `futures(0.1)::Sink` because the API differences make the inverse transform non-straightforward. We should investigate whether this is possible at all, even if it requires some sort of extra runtime cost.

Nemo157:
@LucioFranco, since you mentioned you’d be interested in helping with this I’ll try and write up a bit more detail on what issues I encountered with it, tomorrow.

Nemo157:
Basically the issue is that in 0.1 `Sink` has a single method [`Sink::start_send`](https://docs.rs/futures/0.1.25/futures/sink/trait.Sink.html#tymethod.start_send) that checks for capacity then either consumes the incoming element or returns `NotReady` if there wasn't enough capacity. In 0.3 `Sink` has two separate methods for this, [`Sink::poll_ready`](https://rust-lang-nursery.github.io/futures-api-docs/0.3.0-alpha.10/futures/sink/trait.Sink.html#tymethod.poll_ready) that checks whether there is capacity and returns `NotReady` if there isn't, and [`Sink::start_send`](https://rust-lang-nursery.github.io/futures-api-docs/0.3.0-alpha.10/futures/sink/trait.Sink.html#tymethod.start_send) which takes in an element to consume and presumably drops it and returns an error if you did not check `poll_ready` and there was no capacity available.

Implementing `futures(0.1)::Sink` for `Compat<impl futures(0.3)::Sink>` is then trivial, you can just [call the two methods one after another](https://github.com/rust-lang-nursery/futures-rs/blob/master/futures-util/src/compat/compat03as01.rs#L91-L93) to get the correct behaviour. Going the other way would require you to somehow call `futures(0.1)::Sink::start_send` from `futures(0.3)::Sink::poll_ready` to check for capacity, even though you don't have an item to send.

[...]

PR #1364, opened 2018-12-07
Add Sink01As03 compat shim

LucioFranco:
This adds a compat shim to go from a 0.1 Sink to a 0.3 Sink. This follows, roughly how the stream compat works with the exception that it stores an internal buffer of one item. This is due to the fact that there is no `LocalWaker` passed to `Sink01::start_send`, so we can only actually write the item in the sink on the next `Sink03::poll_read` which gets a `LocalWaker`.

This PR is related to #1362 and cc @Nemo157.

### DIFF
--- a/futures-util/src/compat/compat01as03.rs
+++ b/futures-util/src/compat/compat01as03.rs
@@ -72,7 +72,7 @@ pub trait Sink01CompatExt: Sink01 {
     /// [`Sink<SinkItem = T, SinkError = E>`](futures::sink::Sink)
     /// into a futures 0.3
     /// [`Sink<SinkItem = T, SinkError = E>`](futures_sink::sink::Sink).
-    fn compat(self) -> Compat01As03Sink<Self, Self::SinkItem>
+    fn compat_sink(self) -> Compat01As03Sink<Self, Self::SinkItem>
     where
         Self: Sized,
     {
@@ -80,24 +80,6 @@ pub trait Sink01CompatExt: Sink01 {
     }
 }
 impl<Si: Sink01> Sink01CompatExt for Si {}
-
-/// Extension trait for futures 0.1 [`Stream`](futures::stream::Stream) and
-///  [`Sink`](futures::sink::Sink) combos
-pub trait StreamSink01CompatExt: Stream01 + Sink01 {
-    /// Converts a futures 0.1
-    /// [`Stream<Item = T, Error = E`](futures::stream::Stream) +
-    /// [`Sink<SinkItem = T, SinkError = E>`](futures::sink::Sink)
-    /// into a futures 0.3
-    /// [`Stream<Item = T, Error = E>`](futures_core::stream::Stream) +
-    /// [`Sink<SinkItem = T, SinkError = E>`](futures_sink::sink::Sink).
-    fn compat(self) -> Compat01As03Sink<Self, <Self as Sink01>::SinkItem>
-    where
-        Self: Sized,
-    {
-        Compat01As03Sink::new(self)
-    }
-}
-impl<S: Stream01 + Sink01> StreamSink01CompatExt for S {}
 
 fn poll_01_to_03<T, E>(x: Result<Async01<T>, E>) -> task03::Poll<Result<T, E>> {
     match x {
--- a/futures-util/src/compat/compat01as03.rs
+++ b/futures-util/src/compat/compat01as03.rs
@@ -1,10 +1,11 @@
 use futures_01::executor::{
-    Spawn as Spawn01, spawn as spawn01,
-    UnsafeNotify as UnsafeNotify01,
-    Notify as Notify01,
-    NotifyHandle as NotifyHandle01,
+    spawn as spawn01, Notify as Notify01, NotifyHandle as NotifyHandle01,
+    Spawn as Spawn01, UnsafeNotify as UnsafeNotify01,
 };
-use futures_01::{Async as Async01, Future as Future01, Stream as Stream01, Sink as Sink01, AsyncSink as AsyncSink01};
+use futures_01::{
+    Async as Async01, AsyncSink as AsyncSink01, Future as Future01,
+    Sink as Sink01, Stream as Stream01,
+};
 use futures_core::{task as task03, Future as Future03, Stream as Stream03};
 use std::pin::Pin;
 use std::task::Waker;
@@ -41,7 +42,10 @@ pub trait Future01CompatExt: Future01 {
     /// [`Future<Item = T, Error = E>`](futures::future::Future)
     /// into a futures 0.3
     /// [`Future<Output = Result<T, E>>`](futures_core::future::Future).
-    fn compat(self) -> Compat01As03<Self> where Self: Sized {
+    fn compat(self) -> Compat01As03<Self>
+    where
+        Self: Sized,
+    {
         Compat01As03::new(self)
     }
 }
@@ -53,7 +57,10 @@ pub trait Stream01CompatExt: Stream01 {
     /// [`Stream<Item = T, Error = E>`](futures::stream::Stream)
     /// into a futures 0.3
     /// [`Stream<Item = Result<T, E>>`](futures_core::stream::Stream).
-    fn compat(self) -> Compat01As03<Self> where Self: Sized {
+    fn compat(self) -> Compat01As03<Self>
+    where
+        Self: Sized,
+    {
         Compat01As03::new(self)
     }
 }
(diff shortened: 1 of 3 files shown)


futures-util/src/compat/compat01as03.rs
```diff
@@ -95,6 +108,83 @@ impl<St: Stream01> Stream03 for Compat01As03<St> {
...
+        self.inner.poll_fn_notify(notify, 0, f)
+    }
+}
+
+impl<S: Sink01> Sink03 for Compat01As03Sink<S> {
+    type SinkItem = S::SinkItem;
+    type SinkError = S::SinkError;
+
+    fn start_send(mut self: Pin<&mut Self>, item: Self::SinkItem) -> Result<(), Self::SinkError> {
+        self.buffer = Some(item);
```
Nemo157: Could do with an `assert!(self.buffer.is_none())` before here (maybe `debug_assert!` since it’s a non-`unsafe` programmer error).

futures-util/src/compat/compat01as03.rs
```diff
@@ -95,6 +108,83 @@ impl<St: Stream01> Stream03 for Compat01As03<St> {
...
+            Ok((Async01::NotReady, item)) => {
+                self.buffer = item;
+                task03::Poll::Pending
+            }
+            Err(e) => task03::Poll::Ready(Err(e)),
+        }
+    }
+
+     #[allow(unused_mut)]
+    fn poll_close(mut self: Pin<&mut Self>, lw: &LocalWaker) -> task03::Poll<Result<(), Self::SinkError>> {
```
Nemo157: You can drop the `mut` from `mut self` and get rid of the `allow`.

futures-util/src/compat/compat01as03.rs
```diff
@@ -58,6 +59,18 @@ pub trait Stream01CompatExt: Stream01 {
 }
 impl<St: Stream01> Stream01CompatExt for St {}
 
+/// Extension trait for futures 0.1 [`Sink`](futures::sink::Sink)
+pub trait Sink01CompatExt: Sink01 {
+    /// Converts a futures 0.1
+    /// [`Sink<SinkItem = T, SinkError = E>`](futures::sink::Sink)
+    /// into a futures 0.3
+    /// [`Sink<SinkItem = T, SinkError = E>`](futures_sink::sink::Sink).
+    fn compat(self) -> Compat01As03Sink<Self> where Self: Sized {
```
Nemo157: I wonder if this should be `compat_sink` to avoid method name conflicts with `Stream01CompatExt::compat`

LucioFranco: Do you have any example where this conflict might occur? I added a StreamSink compat trait so that types that implement both can still be compat. I think this should avoid having to name this `compat_sink`.

futures-util/src/compat/compat01as03.rs
```diff
@@ -52,15 +61,49 @@ pub trait Stream01CompatExt: Stream01 {
...
+        Self: Sized,
+    {
+        Compat01As03Sink::new(self)
+    }
+}
+impl<Si: Sink01> Sink01CompatExt for Si {}
+
+/// Extension trait for futures 0.1 [`Stream`](futures::stream::Stream) and
+///  [`Sink`](futures::sink::Sink) combos
+pub trait StreamSink01CompatExt: Stream01 + Sink01 {
```
LucioFranco: I've used as a reference impl for this https://github.com/google/tarpc/blob/master/bincode-transport/src/compat.rs

futures-util/src/compat/compat01as03.rs
```diff
@@ -95,6 +138,119 @@ impl<St: Stream01> Stream03 for Compat01As03<St> {
...
+            }
+            Err(e) => task03::Poll::Ready(Err(e)),
+        }
+    }
+
+    fn poll_close(
+        self: Pin<&mut Self>,
+        lw: &LocalWaker,
+    ) -> task03::Poll<Result<(), Self::SinkError>> {
+        self.poll_flush(lw)
```
LucioFranco: Not sure if this is the right way to `poll_close` this but from reading the docs all that is needed to close is just to flush and return `Poll::Ready`.

futures-util/src/compat/compat01as03.rs
Nemo157: The issue would be code like

```rust
use futures::compat{Stream01CompatExt, Sink01CompatExt};

let stream: impl Stream + Sink = something();
stream.compat();
```

There would be 2 different `compat` methods in scope available on `stream` which `rustc` can't decide between.

LucioFranco: 👍 Did some playing around and spotted the issue. I removed the extra compat trait for `StreamSink` and changed the sink compat trait to use `compat_sink`. Do we want to change stream also to `compat_stream`?

Nemo157: There's a small chance that there are types that implement both `Future` and `Stream` so have the same issue, but I think that would be rare enough that having the same method name is fine (there's precedent in 0.1 that `Future` and `Stream` both have `map_err` but `Sink` has `sink_map_err`).

LucioFranco: This should be resolved now then with `compat_sink`.

futures-util/src/compat/compat01as03.rs
```diff
@@ -95,6 +120,119 @@ impl<St: Stream01> Stream03 for Compat01As03<St> {
...
+        let item = self.buffer.take();
+        match self.in_notify(lw, |f| match item {
+            Some(i) => match f.start_send(i) {
+                Ok(AsyncSink01::Ready) => f.poll_complete().map(|i| (i, None)),
+                Ok(AsyncSink01::NotReady(t)) => {
+                    Ok((Async01::NotReady, Some(t)))
+                }
+                Err(e) => Err(e),
+            },
+            None => Ok((Async01::NotReady, None)),
```
Nemo157: This should be `None => f.poll_complete().map(|i| (i, None)),` here as well, right? We don't have a buffered item, but we still need to flush the underlying sink.

futures-util/src/compat/compat01as03.rs
Nemo157: There is a [`Sink01::close`](https://docs.rs/futures/0.1.25/futures/sink/trait.Sink.html#method.close) method that should be called here. I think you need to `poll_flush` _first_ then call the `close` method, 0.1 doesn't appear to make the same guarantee of flushing any remaining items when the sink is closed. Making this slightly more difficult is that once you call `Sink01::close` you can't call `poll_flush` anymore, so this probably needs an extra flag in the struct to know if you have started closing.

futures-util/src/compat/compat01as03.rs
LucioFranco: Ah yup, totally missed that, it should still attempt to `poll_complete` without calling `start_send` like above.

futures-util/src/compat/compat01as03.rs
LucioFranco: Okay, I've attempted to fix this. `poll_complete` will only be called before close but never after, this is tracked via a bool field within the struct.

futures-util/src/compat/compat01as03.rs
```diff
@@ -95,6 +120,168 @@ impl<St: Stream01> Stream03 for Compat01As03<St> {
     }
 }
 
+/// Converts a futures 0.1 Sink object to a futures 0.3-compatible version
+#[derive(Debug)]
+#[must_use = "futures do nothing unless polled"]
```
xfix: This isn't a future.

```suggestion
#[must_use = "sinks do nothing unless polled"]
```

LucioFranco: You are right, Ill add this change!

LucioFranco:
@Nemo157 do we want to still move this forward, I know you might be busy so no need to rush but want to make sure this isn't getting too far out of sync if there still is a need for it.

Nemo157:
Sorry, I managed to completely forget about this :frowning: 

I skimmed the latest changes, and they look good at first glance. That's a deep match though, so I will try to read it a bit more thoroughly this weekend.

LucioFranco:
@Nemo157 no worries! I am still around to finish this up, I will address the one comment left, see if I can ease up that match, and rebase against master.

futures-util/src/compat/compat01as03.rs
```diff
@@ -95,6 +120,168 @@ impl<St: Stream01> Stream03 for Compat01As03<St> {
...
+                        match <S as Sink01>::close(f) {
+                            Ok(i) => Ok((i, None, true)),
+                            Err(e) => Err(e)
+                        }
+                    },
+                    Ok(Async01::NotReady) => Ok((Async01::NotReady, None, close_started)),
+                    Err(e) => Err(e)
+                }
+            },
+        }) {
```
Nemo157: This looks correct, but I think using early exits for the first things tried would make this a bit easier to read, a quick (untested) sketch of what I mean:

```rust
let item = self.buffer.take();
let close_started = self.close_started;

let result = self.in_notify(lw, |f| {
    if !close_started {
        if let Some(item) = item {
            if let AsyncSink01::NotReady(item) = f.start_send(item)? {
                return Ok((Async01::NotReady, item, false));
            }
        }

        if let Async01::NotReady = f.poll_complete()? {
            return Ok((Async01::NotReady, None, false));
        }
    }

    Ok((<S as Sink01>::close(f)?, None, true))
});

match result {
    Ok((Async01::Ready(_), _, _)) => task03::Poll::Ready(Ok(())),
    Ok((Async01::NotReady, item, close_started)) => {
        self.buffer = item;
        self.close_started = close_started;
        task03::Poll::Pending
    }
    Err(e) => task03::Poll::Ready(Err(e)),
}
```

LucioFranco:
@Nemo157 sorry for the delay, I've cleaned up the poll_close logic and tested it. All seems like its good to go. Thanks.

manuels:
Great job! There is a slight inconsistency: Usually `sink` comes first in the function name (e.g. `SinkExt::sink_map_err(), SinkExt::sink_err_into()`. I would suggest to rename `compat_sink()` to `sink_compat()`.

futures-util/src/compat/mod.rs
```diff
@@ -6,7 +6,7 @@ mod executor;
 pub use self::executor::{Executor01CompatExt, Executor01Future, Executor01As03};
 
 mod compat01as03;
-pub use self::compat01as03::{Compat01As03, Future01CompatExt, Stream01CompatExt};
+pub use self::compat01as03::{Compat01As03, Future01CompatExt, Stream01CompatExt, Sink01CompatExt};
```
manuels: I'm not 100% sure, but I guess `Compat01As03Sink` should be included here, right?

LucioFranco: Yes that is probably a good idea. Will push that soon.

manuels: Sorry to be a pain, but `Compat01As03Sink` has also been included here: https://github.com/rust-lang-nursery/futures-rs/blob/master/futures/src/lib.rs#L90 Anyway, thank for your great work!

LucioFranco: No worries! Thanks for pointing these out 😄 , just pushed the change.

zargony:
I think this may need a rebase because of the `LocalWaker` to `Waker` rename.

I'd love to see this merged since it's currently a pain to work with websockets without being able to send to them. I ended up using a copy of this PR in my code which works fine so far, but official support for sinks would be great.

LucioFranco:
@zargony thats for pointing that out, I should fix this.

@Nemo157 and others are we still planning on merging this, its been sitting for a while...

Nemo157:
I’m keen to merge it, but don’t have a commit bit. @cramertj is the only active merger as far as I’m aware (maybe it’s worth spreading that responsibility a bit?).

cramertj:
I'm watching and would be happy to merge when this is ready!

LucioFranco:
sounds good, give me a few and ill fix conflict issues.

cramertj:
@Nemo157 done! you should have merge rights now.

Nemo157:
Thanks for this 😄 I should have time in a couple days to update the compat blog post with these changes and finally get that merged.

PR #1367, opened 2018-12-10
Add a helper that converts a stream of bytes into an `AsyncRead`

erickt:
This allows byte streams to be converted into an `AsyncRead` to allow more APIs to communicate with each other. It imposes a small amount of overhead, due to there being a state machine that buffers the `Stream` items before copying them to the external buffers.

### DIFF
--- a/futures-util/src/try_stream/into_async_read.rs
+++ b/futures-util/src/try_stream/into_async_read.rs
@@ -1,0 +1,152 @@
+use core::marker::Unpin;
+use core::pin::Pin;
+use futures_core::stream::TryStream;
+use futures_core::task::{LocalWaker, Poll};
+use futures_io::AsyncRead;
+use std::cmp;
+use std::io::{Error, Result};
+
+/// An `AsyncRead` for the [`into_async_read`](super::TryStreamExt::into_async_read) combinator.
+#[derive(Debug)]
+#[must_use = "streams do nothing unless polled"]
+pub struct IntoAsyncRead<St>
+where
+    St: TryStream<Error = Error> + Unpin,
+    St::Ok: AsRef<[u8]>,
+{
+    stream: St,
+    state: ReadState<St::Ok>,
+}
+
+impl<St> Unpin for IntoAsyncRead<St>
+where
+    St: TryStream<Error = Error> + Unpin,
+    St::Ok: AsRef<[u8]>,
+{
+}
+
+#[derive(Debug)]
+enum ReadState<T: AsRef<[u8]>> {
+    Ready { chunk: T, chunk_start: usize },
+    PendingChunk,
+    Eof,
+}
+
+impl<St> IntoAsyncRead<St>
+where
+    St: TryStream<Error = Error> + Unpin,
+    St::Ok: AsRef<[u8]>,
+{
+    pub(super) fn new(stream: St) -> Self {
+        IntoAsyncRead {
+            stream: stream,
+            state: ReadState::PendingChunk,
+        }
+    }
+}
+
+impl<St> AsyncRead for IntoAsyncRead<St>
+where
+    St: TryStream<Error = Error> + Unpin,
+    St::Ok: AsRef<[u8]>,
+{
+    fn poll_read(
+        &mut self,
+        lw: &LocalWaker,
+        buf: &mut [u8],
+    ) -> Poll<Result<usize>> {
+        loop {
+            match &mut self.state {
+                ReadState::Ready { chunk, chunk_start } => {
+                    let chunk = chunk.as_ref();
+                    let len = cmp::min(buf.len(), chunk.len() - *chunk_start);
+
+                    buf[..len].copy_from_slice(
+                        &chunk[*chunk_start..*chunk_start + len],
+                    );
+                    *chunk_start += len;
+
+                    if chunk.len() == *chunk_start {
+                        self.state = ReadState::PendingChunk;
+                    }
+
+                    return Poll::Ready(Ok(len));
+                }
+                ReadState::PendingChunk => {
+                    match ready!(Pin::new(&mut self.stream).try_poll_next(lw)) {
+                        Some(Ok(chunk)) => {
+                            self.state = ReadState::Ready {
+                                chunk,
+                                chunk_start: 0,
+                            };
+                            continue;
+                        }
+                        Some(Err(err)) => {
+                            self.state = ReadState::Eof;
+                            return Poll::Ready(Err(err));
+                        }
+                        None => {
+                            self.state = ReadState::Eof;
+                            return Poll::Ready(Ok(0));
+                        }
+                    }
+                }
+                ReadState::Eof => {
+                    return Poll::Ready(Ok(0));
+                }
+            }
+        }
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use futures::stream::{self, StreamExt, TryStreamExt};
+    use futures_io::AsyncRead;
+    use futures_test::task::noop_local_waker_ref;
+
+    macro_rules! assert_read {
+        ($reader:expr, $buf:expr, $item:expr) => {
+            let lw = noop_local_waker_ref();
+            match $reader.poll_read(lw, $buf) {
+                Poll::Ready(Ok(x)) => {
+                    assert_eq!(x, $item);
+                }
+                Poll::Ready(Err(err)) => {
+                    panic!("assertion failed: expected value but got {}", err);
+                }
+                Poll::Pending => {
+                    panic!("assertion failed: reader was not ready");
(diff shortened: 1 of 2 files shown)


futures-util/src/stream/async_reader.rs
```diff
@@ -0,0 +1,157 @@
...
+where
+    St: Stream + Unpin,
+    St::Item: AsRef<[u8]>,
+{
+}
+
+#[derive(Debug)]
+enum ReadState<T: AsRef<[u8]>> {
+    Ready { chunk: T, chunk_start: usize },
+    NotReady,
```
cramertj: nit: s/NotReady/Pending

futures-util/src/stream/async_reader.rs
```diff
@@ -0,0 +1,157 @@
...
+    St::Item: AsRef<[u8]>,
+{
+    fn poll_read(
+        &mut self,
+        lw: &LocalWaker,
+        buf: &mut [u8],
+    ) -> Poll<Result<usize>> {
+        loop {
+            match self.state {
+                ReadState::Ready {
```
cramertj: nit: match ergo makes this prettier ;)

```rust
match &mut self.state {
    ReadState::Ready { chunk, chunk_start } => { ... }
}
```

futures-util/src/stream/async_reader.rs
```diff
@@ -0,0 +1,157 @@
...
+                } => {
+                    let chunk = chunk.as_ref();
+                    let len = cmp::min(buf.len(), chunk.len() - *chunk_start);
+
+                    buf[..len].copy_from_slice(
+                        &chunk[*chunk_start..*chunk_start + len],
+                    );
+                    *chunk_start += len;
+
+                    if chunk.len() == *chunk_start {
```
cramertj: In this case I think we need to go ahead and pull the next item from the stream and start loading it into the buffer as well.

futures-util/src/stream/async_reader.rs
cramertj: Actually, this is more `NoChunkPresent` or something, right?

futures-util/src/stream/async_reader.rs
cramertj: (but probably only do this if the buffer provided is big enough to hold more data)

seanmonstar: This is contrary to what things like `std::io::BufReader` do. If `poll_next` returns an error, what would be returned? The number of bytes that *were* read, or the error?

futures-util/src/stream/async_reader.rs
erickt: How does `PendingChunk` sound? When we are in this state, we really are in a holding pattern until the stream produces the next item.

futures-util/src/stream/async_reader.rs
erickt: Ack, speaking of errors, I really ought to be consuming a stream of `Result<AsRef<T>, std::io::Error>`s. I'll port my stuff over to it.

I also agree with @seanmonstar here. The only way I can think of that we could handle this case is if `AsyncRead` error includes the partial length that we read. I also don't know if there'd a real performance win if we were looping inside `AsyncReader` or an outer loop.

cramertj: I'd return the number of bytes that were read and store the error for the next time, but it also seems fine to me to not do this-- if the stdlib types don't typically do this, then I suppose it'd be more consistent not to.

futures-util/src/stream/async_reader.rs
```diff
@@ -0,0 +1,157 @@
...
+                        self.state = ReadState::NotReady;
+                    }
+
+                    return Poll::Ready(Ok(len));
+                }
+                ReadState::NotReady => {
+                    match Pin::new(&mut self.stream).poll_next(lw) {
```
Nemo157: This can use the `ready!` macro to handle the `Poll::Pending` state. (Aside, I'm amazed at how much nicer handling this sort of state machine is with NLL, I'm tempted to update some of my older crates to 2018 just to see the massive readability gain).

Nemo157: Oh, also `StreamExt::poll_next_unpin` to avoid the `Pin` construction here (although I think this adapter can be updated to handle `!Unpin` streams pretty easily, but that can be added later).

erickt: Yeah I was a little shocked this worked. NLL is so wonderful.

Nice tip on `ready!`, I'll switch to that.

Regarding `poll_next_unpin`, that is unfortunately not supported by `TryStream`, which I just switched to. Shall I add it as well?

futures-util/src/try_stream/mod.rs
```diff
@@ -572,8 +576,42 @@ pub trait TryStreamExt: TryStream {
...
+    /// # Examples
+    ///
+    /// ```
+    /// use futures::executor::block_on;
+    /// use futures::future::lazy;
+    /// use futures::stream::{self, StreamExt, TryStreamExt};
+    /// use futures_io::AsyncRead;
+    /// use std::io::Error;
+    ///
+    /// let stream = stream::iter(vec![Ok::<Vec<u8>, Error>(vec![1, 2, 3, 4, 5])]);
```
cramertj: I wouldn't think you'd need the turbofish here

futures-util/src/try_stream/mod.rs
```diff
@@ -572,8 +576,42 @@ pub trait TryStreamExt: TryStream {
...
+    /// let mut reader = stream.into_async_read();
+    /// let mut buf = vec![0; 3];
+    ///
+    /// block_on(lazy(|ctx| {
```
cramertj: Consider maybe using something like `read_to_end` instead, and show it w/ `await!`? typically we try to give examples in terms of the `async`/`await!` usage

cramertj: (also an `assert!` and the end showing that the operation completed successfully would be welcome ;) )

futures-util/src/try_stream/into_async_read.rs
```diff
@@ -0,0 +1,152 @@
...
+                    panic!("assertion failed: reader was not ready");
+                }
+            }
+        };
+    }
+
+    #[test]
+    fn test_into_async_read() {
+        let stream = stream::iter(1..=3)
+            .map(|_| Ok::<Vec<u8>, Error>(vec![1, 2, 3, 4, 5]));
```
cramertj: I'd think you wouldn't need the turbofish here either

cramertj: (though maybe to get `u8` rather than `i32`?)

futures-util/src/try_stream/into_async_read.rs
```diff
@@ -0,0 +1,152 @@
...
+                ReadState::PendingChunk => {
+                    match ready!(Pin::new(&mut self.stream).try_poll_next(lw)) {
+                        Some(Ok(chunk)) => {
+                            self.state = ReadState::Ready {
+                                chunk,
+                                chunk_start: 0,
+                            };
+                            continue;
+                        }
+                        Some(Err(err)) => {
```
cramertj: I'd also switch `self.state = ReadState::Eof;` here.

futures-util/src/try_stream/mod.rs
```diff
@@ -57,6 +57,10 @@ pub use self::try_for_each_concurrent::TryForEachConcurrent;
 #[cfg(feature = "std")]
 use futures_core::future::Future;
 
+#[cfg(feature = "std")]
+mod into_async_read;
+pub use self::into_async_read::IntoAsyncRead;
```
cramertj: You need `#[cfg(feature = "std")]` here as well.

erickt:
@cramertj thanks for your comments! I think I got it all addressed. Hopefully travis will be happy too.

futures-util/src/try_stream/mod.rs
```diff
@@ -572,8 +577,44 @@ pub trait TryStreamExt: TryStream {
...
+    ///
+    /// assert!(await!(reader.read_to_end(&mut buf)).is_ok());
+    /// assert_eq!(buf, &[1, 2, 3, 4, 5]);
+    /// # })
+    /// ```
+    #[cfg(feature = "std")]
+    fn into_async_read(self) -> IntoAsyncRead<Self>
+    where
+        Self: Sized + TryStreamExt<Error = std::io::Error> + Unpin,
+        Self::Ok: AsRef<[u8]>,
```
cramertj: ugh it's so annoying that you can have this but `Self::Error = std::io::Error` doesn't work. Someday...

cramertj:
This looks great! thanks for you patience w/ all my nits :)

PR #1368, opened 2018-12-10
Add try_poll_next_unpin convenience method

erickt:
This simplifies calling `TryStream::poll_next_unpin` on `Unpin` stream types.

### DIFF
--- a/futures-util/src/try_stream/mod.rs
+++ b/futures-util/src/try_stream/mod.rs
@@ -4,8 +4,10 @@
 //! that return `Result`s, allowing for short-circuiting computations.
 
 use core::marker::Unpin;
+use core::pin::Pin;
 use futures_core::future::TryFuture;
 use futures_core::stream::TryStream;
+use futures_core::task::{LocalWaker, Poll};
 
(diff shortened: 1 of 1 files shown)


futures-util/src/try_stream/mod.rs
```diff
@@ -550,6 +552,17 @@ pub trait TryStreamExt: TryStream {
         TryBufferUnordered::new(self, n)
     }
 
+    /// A convenience method for calling [`TryStream::poll_next_unpin`] on [`Unpin`]
+    /// stream types.
+    fn try_poll_next_unpin(
+        &mut self,
+        lw: &LocalWaker
+    ) -> Poll<Option<Result<Self::Ok, Self::Error>>>
+    where Self: Unpin + Sized
```
cramertj: Do you need the `Sized` bound here?

erickt: I do not! I removed it.

erickt: I do not! I removed it.

PR #1371, opened 2018-12-14
Rename Pinned to PhantomPinned

LucioFranco:
This is to keep futures-rs updated with the nightly changes that are causing CI to fail on master.

Reference commit https://github.com/rust-lang/rust/commit/0076f58d5333f24f709aa46b4bad760ffb51b9b0


cramertj:
Thanks!

PR #1372, opened 2018-12-19
Update select_all.

turnage:
This updates select_all to work with the new API. It is implemented in this commit for `Stream + Unpin`.


cramertj:
LGTM-- I'll fix the clippy failures in another change

Issue #1374, opened 2018-12-26
Build failed with 2018-12-26 nightly


Riey:
```
error[E0599]: no function or associated item named `get_mut_unchecked` found for type `std::pin::Pin<_>` in the current scope
   --> /home/riey/.cargo/registry/src/github.com-1ecc6299db9ec823/futures-core-preview-0.3.0-alpha.10/src/future/future_obj.rs:176:40
    |
176 |         let mut_ref: &mut F = unsafe { Pin::get_mut_unchecked(Pin::as_mut(&mut self)) };
    |                                        ^^^^^^^^^^^^^^^^^^^^^^ function or associated item not found in `std::pin::Pin<_>`

error[E0599]: no function or associated item named `get_mut_unchecked` found for type `std::pin::Pin<_>` in the current scope
   --> /home/riey/.cargo/registry/src/github.com-1ecc6299db9ec823/futures-core-preview-0.3.0-alpha.10/src/future/future_obj.rs:216:44
    |
216 |             let mut_ref: &mut F = unsafe { Pin::get_mut_unchecked(Pin::as_mut(&mut self)) };
    |                                            ^^^^^^^^^^^^^^^^^^^^^^ function or associated item not found in `std::pin::Pin<_>`

error[E0277]: the trait bound `P: std::marker::Unpin` is not satisfied in `std::pin::Pin<P>`
  --> /home/riey/.cargo/registry/src/github.com-1ecc6299db9ec823/futures-core-preview-0.3.0-alpha.10/src/stream/mod.rs:83:9
   |
83 |         Pin::get_mut(self).as_mut().poll_next(lw)
   |         ^^^^^^^^^^^^ within `std::pin::Pin<P>`, the trait `std::marker::Unpin` is not implemented for `P`
   |
   = help: consider adding a `where P: std::marker::Unpin` bound
   = note: required because it appears within the type `std::pin::Pin<P>`
   = note: required by `<std::pin::Pin<&'a mut T>>::get_mut`

error[E0599]: no function or associated item named `get_mut_unchecked` found for type `std::pin::Pin<_>` in the current scope
  --> /home/riey/.cargo/registry/src/github.com-1ecc6299db9ec823/futures-core-preview-0.3.0-alpha.10/src/stream/mod.rs:96:19
   |
96 |             match Pin::get_mut_unchecked(self) {
[...]

PR #1375, opened 2018-12-26
Try to match new std api

Riey:
PR for #1374 

But _pin-utils_ crate is also not support current api

so _unsafe_unpinned!_ macro is not working and don't compiled


cramertj:
Addressed in https://github.com/rust-lang-nursery/futures-rs/pull/1377

PR #1376, opened 2018-12-26
[WIP] Update to latest Pin API

taiki-e:
This is an alternative to #1375.

This depends on rust-lang-nursery/pin-utils#13

Also, it includes fix to clippy errors that occurred with [#1372](https://github.com/rust-lang-nursery/futures-rs/pull/1372).


cramertj:
Covered by https://github.com/rust-lang-nursery/futures-rs/pull/1377.

Thanks for the PR! I went a different route in order to get the `arbitrary_self_types` and `pin` features removed.

PR #1380, opened 2018-12-28
impl Default for SelectAll

taiki-e:
This implements `Default` for `SelectAll`, which already has a new method that takes no argument.

Also, `FuturesUnordered` and `FuturesOrdered` are already implementing `Default` in #1343.


cramertj:
Thanks!

PR #1382, opened 2018-12-28
Add the minimum required version to README

taiki-e:
I think that it is preferable to display the minimum required version.

Related: rust-lang-nursery/pin-utils#19


cramertj:
Would you be able to add a test for this to .travis.yml as well? Without a test, I'm concerned that this wouldn't be updated when appropriate.

cramertj:
Thanks!

PR #1383, opened 2018-12-28
Allow joining !Unpin futures in JoinAll

Nemo157:
2 things I want to do before merging:

1. ~~Add a new `Box<T> -> Pin<Box<T>>` API to `std` to remove one use of `Pin::new_unchecked`.~~ EDIT: Possible via `<Pin<Box<T>> as From<Box<T>>>::from`, not very discoverable however...
2. Try and throw a minimal wrapper around `Pin<Box<[T]>>` to provide a safe API to iterate over it.

Just opening this up now in case anyone has any examples of doing something like the second already.

### DIFF
--- a/futures-util/src/future/join_all.rs
+++ b/futures-util/src/future/join_all.rs
@@ -15,7 +15,44 @@ where
     F: Future,
 {
     Pending(F),
-    Done(F::Output),
+    Done(Option<F::Output>),
+}
+
+impl<F> ElemState<F>
+where
+    F: Future,
+{
+    fn pending_pin_mut<'a>(self: Pin<&'a mut Self>) -> Option<Pin<&'a mut F>> {
+        // Safety: Basic enum pin projection, no drop + optionally Unpin based
+        // on the type of this variant
+        match unsafe { self.get_unchecked_mut() } {
+            ElemState::Pending(f) => Some(unsafe { Pin::new_unchecked(f) }),
+            ElemState::Done(_) => None,
+        }
+    }
+
+    fn take_done(self: Pin<&mut Self>) -> Option<F::Output> {
+        // Safety: Going from pin to a variant we never pin-project
+        match unsafe { self.get_unchecked_mut() } {
+            ElemState::Pending(_) => None,
+            ElemState::Done(output) => output.take(),
+        }
+    }
+}
+
+impl<F> Unpin for ElemState<F>
+where
+    F: Future + Unpin,
+{
+}
+
+fn iter_pin_mut<T>(slice: Pin<&mut [T]>) -> impl Iterator<Item = Pin<&mut T>> {
+    // Safety: `std` _could_ make this unsound if it were to decide Pin's
+    // invariants aren't required to transmit through slices. Otherwise this has
+    // the same safety as a normal field pin projection.
+    unsafe { slice.get_unchecked_mut() }
+        .iter_mut()
+        .map(|t| unsafe { Pin::new_unchecked(t) })
 }
 
 /// A future which takes a list of futures and resolves with a vector of the
@@ -27,7 +64,7 @@ pub struct JoinAll<F>
 where
     F: Future,
 {
-    elems: Vec<ElemState<F>>,
+    elems: Pin<Box<[ElemState<F>]>>,
 }
 
 impl<F> fmt::Debug for JoinAll<F>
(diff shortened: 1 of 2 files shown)


futures-util/src/future/join_all.rs
```diff
@@ -86,25 +101,30 @@ where
     ) -> Poll<Self::Output> {
         let mut all_done = true;
 
-        for elem in self.as_mut().elems.iter_mut() {
+        let elems: &mut [ElemState<F>] = unsafe {
```
cramertj: seems like we need `get_pin_ref(index: usize)` and `get_pin_mut(index: usize)`

futures-util/src/future/join_all.rs
```diff
@@ -53,30 +61,37 @@ where
...
+    let elems: Vec<_> = i.into_iter().map(ElemState::Pending).collect();
+    let elems = unsafe { Pin::new_unchecked(elems.into_boxed_slice()) };
     JoinAll { elems }
 }
 
+impl<F> JoinAll<F>
+where
+    F: Future,
+{
+    unsafe_pinned!(elems: Pin<Box<[ElemState<F>]>>);
```
cramertj: This isn't necessary-- `JoinAll` and `Pin<Box<...>>` are both `Unpin`, so it's safe to access `elems` directly.

futures-util/src/future/join_all.rs
```diff
@@ -53,30 +61,37 @@ where
 /// # Examples
 ///
 /// ```
-/// use futures_util::future::{FutureExt, join_all, ready};
+/// #![feature(async_await, await_macro, futures_api)]
```
cramertj: This isn't new to your PR, but this change adds yet another reason to point from `join_all` to `FuturesUnordered`. After this PR, it won't be possible to add new futures to a `JoinAll`. Additionally, the fact that all futures must all be scanned even when some have completed gives another reason to prefer `FuturesUnordered`.

Can you add a comment to this doc comment about how users with large numbers of futures or who need to dynamically `push` to the collection should use `FuturesUnordered` instead?

futures-util/src/future/join_all.rs
cramertj: (this would address your point (2) above)

Nemo157: Yep, I was mostly thinking to throw together a quick prototype here to get this done then open a PR to Rust once it's been validated a little. (For ergonomics here I was planning something like `iter_pin_mut(self: Pin<&mut [T]>) -> impl Iterator<Item = Pin<&mut T>> + '_`).

futures-util/src/future/join_all.rs
Nemo157: uh, yeah, don't know what I was thinking

futures-util/src/future/join_all.rs
Nemo157: Yep, I see this mostly as a convenience utility when you have a handful of futures from the same `async fn` and want to simply `await!` them all (that could mostly be covered by a macro version that avoids the heap allocation, but there is still the smaller usecase of an unknown length handful of futures that this could be used for). Definitely makes sense to push users towards `FuturesUnordered` pretty aggressively.

futures-util/src/future/join_all.rs
cramertj: sgtm

futures-util/src/future/join_all.rs
cramertj: Cool-- sgtm!

Nemo157:
Rebased and updated to use new API from `std`, now the unsafety is all nicely encapsulated in some pretty simple APIs 😄

futures-util/src/future/join_all.rs
```diff
@@ -86,27 +134,20 @@ where
...
 
         if all_done {
-            let elems = mem::replace(&mut self.elems, Vec::new());
-            let result = elems
-                .into_iter()
-                .map(|e| match e {
-                    ElemState::Done(t) => t,
-                    _ => unreachable!(),
-                })
+            let mut elems = mem::replace(&mut self.elems, Box::pin([]));
```
Matthias247: Isn't that causing a new allocation to keep `elems` intact, even though we don't need it anymore? If yes it would be nice to avoid it.

futures-util/src/future/join_all.rs
```diff
@@ -50,33 +87,44 @@ where
 /// collecting the results into a destination `Vec<T>` in the same order as they
 /// were provided.
 ///
+/// # Caveats
```
Matthias247: I'm not sure whether I would call it `Caveats` or whether I would mention futures-unordered. They seem to have very different use-cases.
The main use-case for `join_all` seems to be to wait for a dynamically created list of futures (opposed to `join!`, which operates on a static list).

futures-util/src/future/join_all.rs
Nemo157: Hmm, does `Box` allocate for a dynamic ZST? If so I guess `self.elems` could be an `Option`, but I'm hoping `Box` is smart enough to not actually allocate in this case (I'll see if I can come up with a testcase showing it).

futures-util/src/future/join_all.rs
Nemo157: The performance with a large number of futures is likely to be a good reason to push users towards `futures_unordered` even when `join_all` could work. Maybe "Caveats" is a bit strong though, how do "Alternatives" or "See Also" sound?

futures-util/src/future/join_all.rs
Matthias247: Oh it’s zero sized? I’m not sure. Actually it wouldn’t make sense, because malloc(0) is borderline invalid (not super well defined). 

Guess it might be checkable with global allocator?

Matthias247: Oh it’s zero sized? I’m not sure. Actually it wouldn’t make sense, because malloc(0) is borderline invalid (not super well defined). 

Guess it might be checkable with global allocator?

Nemo157: Heh, should have just checked [the docs](https://doc.rust-lang.org/nightly/std/boxed/struct.Box.html#method.new):

> This doesn't actually allocate if T is zero-sized.

So it should just be basically updating the capacity to zero.

cramertj:
Nice :)

PR #1387, opened 2018-12-31
impl Default for Mutex

arcnmx:
Implements `Default` for `Mutex<T>`.


arcnmx:
Hm the CI failure seems to be due to a clippy lint rename that happened the other day, merging #1380 will fix that.

cramertj:
Cool, thanks!

PR #1388, opened 2019-01-01
Improve error without required features for futures-test

### LINKED ISSUE #1385: Unable to compile with futures-test as dev-dependency
Adding `futures-test-preview = "0.3.0-alpha.11"` to dev-dependencies causes the following compilation error for me:

```log
error[E0425]: cannot find function `block_on` in module `futures_executor`
  --> ~/.cargo/registry/src/github.com-1ecc6299db9ec823/futures-test-preview-0.3.0-alpha.11/src/future/mod.rs:79:44
   |
79 |         thread::spawn(|| futures_executor::block_on(self));
   |                                            ^^^^^^^^ not found in `futures_executor`

```

I think this is the case because by default it doesn't enable the `std` features in futures-executor, see https://github.com/rust-lang-nursery/futures-rs/blob/71c59479b538380567907b115ec71799eb03b88d/futures-test/Cargo.toml#L20

Which means `block_on` isn't available, see https://github.com/rust-lang-nursery/futures-rs/blob/71c59479b538380567907b115ec71799eb03b88d/futures-executor/src/lib.rs#L14-L15

I assume the fix is as simple as enabling the `std` feature on future-executor by default, but I don't what other things that would cause.

Nemo157:
fixes #1385

### DIFF
--- a/futures-test/Cargo.toml
+++ b/futures-test/Cargo.toml
@@ -22,3 +22,7 @@ pin-utils = { version = "0.1.0-alpha.4", default-features = false }
 
 [dev-dependencies]
 futures-preview = { version = "0.3.0-alpha.2", path = "../futures", default-features = false, features = ["std"] }
+
+[features]
+default = ["std"]
+std = ["futures-core-preview/std", "futures-util-preview/std", "futures-executor-preview/std"]
--- a/futures-test/src/lib.rs
+++ b/futures-test/src/lib.rs
@@ -11,16 +11,24 @@
     html_root_url = "https://rust-lang-nursery.github.io/futures-doc/0.3.0-alpha.5/futures_test"
 )]
 
+#[cfg(not(feature = "std"))]
+compile_error!("`futures-test` must have the `std` feature activated, this is a default-active feature");
+
 #[doc(hidden)]
+#[cfg(feature = "std")]
 pub use std as std_reexport;
 
 #[doc(hidden)]
+#[cfg(feature = "std")]
 pub extern crate futures_core as futures_core_reexport;
 
 #[macro_use]
 #[doc(hidden)]
+#[cfg(feature = "std")]
 pub mod assert;
 
+#[cfg(feature = "std")]
 pub mod task;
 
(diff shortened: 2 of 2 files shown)


Nemo157:
https://travis-ci.org/rust-lang-nursery/futures-rs/jobs/474096048#L531 shows the failure building the new crate.

Unfortunately the same thing that requires having an extra crate to test this causes the failure in https://travis-ci.org/rust-lang-nursery/futures-rs/jobs/474097015, I don't think it's possible to actually fix this properly with Cargo's current handling of features.

I'll try adding a `std` feature and an error message if this is not active so that users can fix it easily.

Nemo157:
If someone attempts to use `futures-test = { default-features = false }` now they will get a useful error message:

```
→ cargo build --manifest-path futures-test/Cargo.toml --no-default-features
   Compiling futures-test-preview v0.3.0-alpha.11 (/Users/nemo157/sources/futures-rs/futures-test)
error: `futures-test` must have the `std` feature activated, this is a default-active feature
  --> futures-test/src/lib.rs:15:1
   |
15 | compile_error!("`futures-test` must have the `std` feature activated, this is a default-active feature");
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

Nemo157:
For what it needs from other crates yes, but around a third of the rest of the APIs depend on `Arc` and a few are using `Box::leak` with ZSTs (probably possible to do soundly some other way, but I don’t know how). If anyone is actually using a `no_std` testing framework then I’d be happy to have the feature used here properly, I just don’t think it’s worth it at this point (after 80% implementing it, then reverting back to this).

Thomasdezeeuw:
@Nemo157 Ok, then I'm happy to use the `std` feature by default, I don't have a `no_std` use case.

cramertj:
Nice, this will be much better, thanks!

PR #1389, opened 2019-01-01
Add a new futures-intrusive crate.

Matthias247:
This crate contains a set of future-based primitives which are built on top of intrusive data structures.
The main goal of this crate is to provide types which require no dynamic allocation during runtime, independent of the number of tasks and concurrent paths that are interacting with the future.
The goal should be be to provide all types in a no-std compatible form.

Currently the crate contains a Mutex and a ManualResetEvent type, in both Sync and !Sync fashion. The !Sync variants are already no-std compatible.
Other planned types would include:
- Timers
- Channels
- More primitives (e.g. Semaphores)

### DIFF
--- a/futures-intrusive/src/sync/mutex.rs
+++ b/futures-intrusive/src/sync/mutex.rs
@@ -1,0 +1,542 @@
+//! An asynchronously awaitable mutex for synchronization between concurrently
+//! executing futures.
+
+use futures_core::future::{Future, FusedFuture};
+use futures_core::task::{LocalWaker, Poll, Waker};
+use core::ops::{Deref, DerefMut};
+use core::pin::Pin;
+use core::ptr::null_mut;
+use core::cell::{UnsafeCell, RefCell};
+use crate::intrusive_list::{LinkedList, ListNode};
+
+/// Tracks how the future had interacted with the mutex
+#[derive(PartialEq)]
+enum PollState {
+    /// The task has never interacted with the mutex.
+    New,
+    /// The task was added to the wait queue at the mutex.
+    Waiting,
+    /// The task had previously waited on the mutex, but was notified
+    /// that the mutex was released in the meantime.
+    Notified,
+    /// The task had been polled to completion.
+    Done,
+}
+
+/// Tracks the MutexLockFuture waiting state.
+/// Access to this struct is synchronized through the mutex in the Event.
+struct WaitQueueEntry {
+    /// The task handle of the waiting task
+    task: Option<Waker>,
+    /// Current polling state
+    state: PollState,
+}
+
+impl WaitQueueEntry {
+    /// Creates a new WaitQueueEntry
+    fn new() -> WaitQueueEntry {
+        WaitQueueEntry {
+            task: None,
+            state: PollState::New,
+        }
+    }
+}
+
+/// Internal state of the `Mutex`
+struct MutexState {
+    is_fair: bool,
+    is_locked: bool,
+    waiters: LinkedList<WaitQueueEntry>,
+}
+
+impl MutexState {
+    fn new(is_fair: bool) -> Self {
+        MutexState {
+            is_fair,
+            is_locked: false,
+            waiters: LinkedList::new(),
+        }
+    }
+
+    /// Wakes up the last waiter and removes it from the wait queue
+    fn wakeup_last_waiter(&mut self) {
+        let last_waiter =
+            if self.is_fair {
+                self.waiters.peek_last()
+            } else {
+                self.waiters.remove_last()
+            };
+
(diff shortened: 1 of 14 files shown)


Matthias247:
This supersedes #1315

Matthias247:
@aturon, @cramertj, @Nemo157, @withoutboats  and everyone else:

Let me know what you think about adding this here, or whether it should go into a totally separate crate. I think it fits better here, since it has the potential to demonstrate true zero-costness of Rust futures, and improves the library surface for no-std.

futures-intrusive/src/sync/mutex.rs
```diff
@@ -0,0 +1,542 @@
...
+impl<T> LocalMutex<T> {
+    /// Creates a new futures-aware mutex.
+    ///
+    /// `is_fair` defines whether the `Mutex` should behave be fair regarding the
+    /// order of waiters. A fair `Mutex` will only allow the first waiter which
+    /// tried to lock but failed to lock the `Mutex` once it's available again.
+    /// Other waiters must wait until either this locking attempt completes, and
+    /// the `Mutex` gets unlocked again, or until the `MutexLockFuture` which
+    /// tried to gain the lock is dropped.
+    pub fn new(value: T, is_fair: bool) -> LocalMutex<T> {
```
Matthias247: Could be also 2 Mutex types (e.g. `FairLocalMutex`) instead of a parameterized one. Or one version could go away.
I felt that adding a parameter might make it easier to change the behavior in an application later on when required, und reduces duplication in the code. But open to all suggestions

cramertj:
This looks great, I'm really excited about further exploration into the `!Unpin` world! This will take me some time to review, though-- I'll try and grab some time later this week, and if you don't hear from me soon, feel free to pester me about it ;)

withoutboats:
I haven't reviewed the code, but this looks like a really cool crate. My only concern is that I don't think it should be a part of the futures project. For a long time, we essentially had a monolithic ecosystem in which futures and tokio were maintained by the same team and organizationally tightly coupled, and we've managed to begin moving away from that to a world where many projects can coexist built on hte same bedrock foundation (the `Future` trait). I'd like to see that trend toward a multipolar world continue, and also to see the futures crate (as the most foundational component) tend to contract toward stability and minimalism.

Why do you think this should be part of the futures project, instead of making a new project? It seems rather analogous to the proposal when parking_lot first came out to merge it into std. The libs team decided (rightly IMO) that it was too soon at that time to make the change to std. Now, of course, work is underway to start using parking_lot's implementation in std. But I think allowing for experimentation and proving things out without incorporating them into the root project creates a more sustainable initial trajectory.

Of course, I'd also like to see the networking working group do more work to promote and make visible all of the different projects people work on, so that other users can try using them out or become contributors. I hope my message isn't misunderstood as being uneasy *supporting* this work, which I really hope we can do! That's distinct from *incorporating* it.

Matthias247:
That's a very valid comment. I'm not a 100% sure either and asked for opinions therefore.  Some further thoughts:

> Why do you think this should be part of the futures project, instead of making a new project?

I think in order to answer this, we might need to answer some other questions:
- _What is the future of the futures crate, and what should generally end up in there?_
- _Should we offer users of futures a batteries included solution, or only a rough framework and some primitives, and other things are delivered via the wider ecosystem?_

I understood the futures crate to be the extended standard library around futures. 
It already has a bit of everything:
- Combinators
- Executors
- Some leaf futures (like channels and one mutex)
- Some trait definitions, that seem
- Async utility methods (like `select!` and `join!`).

So a question is whether this should get extended, or reduced.
If we want authors of async code to find solutions for most common things in a single place, we should increase the surface area of the crate, and close gaps. 

For users adding more things has the advantage that they don't need to search for other crates that fulfill their use-case, and are not sure about their status. Async IO in Rust already is considered as not easy to get started, and needing to import 4 crates (e..g for an executor, channels, timers, and network IO) won't make it easier.
I think having a bigger crate might also make maintenance easier. The amount of people who can and will write this kind of code will be very limited. Working on the same repository means there is more help for CRs, and code sees more eyes.

However working on the same crate/repository obviously also requires more alignment from all maintainers, and makes it harder to experiment in some directions. Smaller crates enable this more easily.

Therefore I'm obviously seeing benefits of both solutions.

Now back to the more specific question, whether exactly this code should go into futures:
The reason for **yes** is, that I think currently the futures-rs crate is currently mostly focused on `std` environments, and could be improved for `no-std`. Since the types here are mostly fine without dynamically allocating memory, they are well suited for this. And also to give the users who look no further than into the futures crate more options/types.

Matthias247:
I uploaded things [here](https://github.com/Matthias247/futures-intrusive), since there wasn't a response anymore.

PR #1391, opened 2019-01-03
Make AssertUnmoved and PendingOnce public

Thomasdezeeuw:
Both types are already exposed in futures_test::future::FutureTestExt, but neither are public.


Nemo157:
Ha, whoops, I wish the `private-in-public` lint actually worked for cases like this...

cramertj:
Thanks!

PR #1392, opened 2019-01-03
Support StreamExt::collect in no_std environment

xfix:
There is no reason to disallow `collect` here anymore, now that it is generic.


cramertj:
Good catch!

PR #1393, opened 2019-01-03
Apply edition idioms

xfix:
``cargo fmt --edition-idioms``

### DIFF
--- a/futures-util/src/future/flatten.rs
+++ b/futures-util/src/future/flatten.rs
@@ -35,7 +35,7 @@ impl<Fut> fmt::Debug for Flatten<Fut>
     where Fut: Future + fmt::Debug,
           Fut::Output: Future + fmt::Debug,
 {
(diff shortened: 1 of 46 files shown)


futures-util/src/future/flatten.rs
```diff
@@ -35,7 +35,7 @@ impl<Fut> fmt::Debug for Flatten<Fut>
     where Fut: Future + fmt::Debug,
           Fut::Output: Future + fmt::Debug,
 {
-    fn fmt(&self, fmt: &mut fmt::Formatter) -> fmt::Result {
+    fn fmt(&self, fmt: &mut fmt::Formatter<'_>) -> fmt::Result {
```
Nemo157: From what I recall of the "recent" discussion on https://github.com/rust-lang/rust/issues/44524 this was still potentially not going to be needed (although, if `elided_lifetimes_in_paths` is already stable :person_shrugging:).

cramertj:
Thanks!

PR #1395, opened 2019-01-03
Don't panic in AssertUnmoved while panicking

### LINKED ISSUE #1394: AssertUnmoved aborts when the future is moved.
Currently if the future is moved between calls it fail the assertion and panic, see https://github.com/rust-lang-nursery/futures-rs/blob/03c0d3d0f2363efcbd42e50104e09b9042c158de/futures-test/src/future/assert_unmoved.rs#L44

Next when the future is dropped it will panic again here: https://github.com/rust-lang-nursery/futures-rs/blob/03c0d3d0f2363efcbd42e50104e09b9042c158de/futures-test/src/future/assert_unmoved.rs#L53

And when you panic whilst panicing it causes the process to abort, which at least on macOS can be quite confusing (it prints `process didn't exit successfully: bin (signal: 4, SIGILL: illegal instruction)`).

I think the best solution is to check for panicking in the `Drop` implementation and not panic if already panicking. Any opinions? Would a pr in that direction be accepted?

Thomasdezeeuw:
Also doesn't panic anymore if the future was never polled and then dropped.

Fixes #1394.

### DIFF
--- a/futures-test/src/future/assert_unmoved.rs
+++ b/futures-test/src/future/assert_unmoved.rs
@@ -4,6 +4,7 @@ use pin_utils::{unsafe_pinned, unsafe_unpinned};
 use std::marker::PhantomPinned;
 use std::pin::Pin;
 use std::ptr;
+use std::thread::panicking;
 
 /// Combinator for the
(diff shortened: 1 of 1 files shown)


cramertj:
Nice, thanks for the fix!

PR #1397, opened 2019-01-04
Allow ?Sized implementations of Wake to local_waker_ref

cramertj:


cramertj:
This is blocked on a corresponding change to upstream Rust to add `?Sized` to some functions.

Matthias247:
Looks fine.
But out of curiosity: Why is this blocking something in Rust? The Wake trait barely shows up there.
Btw: We should really figure out what to do with this thing longer-term (see the open stabilization issue).

Issue #1396, opened 2019-01-03
Broken no_std build


Nemo157:
Related to changes from https://github.com/rust-lang-nursery/futures-rs/pull/1388 + [#1390](https://github.com/rust-lang-nursery/futures-rs/pull/1390/files#diff-41c03002d828b5290c4f39adc8d88060R35). Before the change in #1390 the tests on CI passed because they use `cargo test --all --all-features`, this would activate the `futures-test/std` feature while running the `futures` tests; but if you want to run just the `futures` tests locally with `cargo test -p futures-preview` they would fail because the `futures-test/std` feature was not activated. Removing `default-features` false from the `futures-test` dev-dependency in `futures` fixed this so that both ways of testing would work, but because of feature unification between dev and non-dev dependencies this results in the `futures-core/std` feature being force activated always.

I don't see any way to fix this with how features currently work in Cargo, but there's been no real movement on un-unifying the features of dependency types (https://github.com/rust-lang/cargo/issues/4866).

Nemo157:
I have an idea for a hacky solution to this, I'll try and get a PR up soon.

PR #1402, opened 2019-01-08
Add new_count_waker

### LINKED ISSUE #1384: Change to WakeCounter API proposal
I would like to propose a breaking change in API for the `WakerCounter` type in the futures-test crate. The current API has the counter and `LocalWaker` implementation in the same `WakerCounter` type, which can be limiting. I propose an API that splits it into two types. Since it's not a lot of code I have a working implementation below.

```rust
/// Create a new `LocalWaker` that counts the number of times it's awoken.
pub fn new_count_waker() -> (LocalWaker, AwokenCount) {
    let inner = Arc::new(WakerInner { count: AtomicUsize::new(0) });
    (local_waker_from_nonlocal(inner.clone()), AwokenCount { inner })
}

/// Number of times the waker was awoken.
///
/// See [`new_count_waker`].
pub struct AwokenCount {
    inner: Arc<WakerInner>,
}

impl AwokenCount {
    /// Get the number of times the waker was awoken.
    pub fn get(&self) -> usize {
        self.inner.count.load(Ordering::SeqCst)
    }
}

struct WakerInner {
    count: AtomicUsize,
}

impl Wake for WakerInner {
    fn wake(arc_self: &Arc<Self>) {
        let _ = arc_self.count.fetch_add(1, Ordering::SeqCst);
    }
}
```

And the usage.

```rust
let (waker, count) = new_count_waker();

// Polling is easy since we just have a `LocalWaker`.
let future = // ...
future.poll(&waker);

// And elsewhere we can check the wake count.
assert_eq!(count.get(), 1);
```

Thomasdezeeuw:
This is an alternative API to WakeCounter, see #1384.

I didn't touch the old API, but it might be a bit confusing that there are two different API's to do the same thing. I'll leave it up to someone else to decide whether to deprecate/remove/leave it.

Closes #1384.

/cc @Nemo157


Thomasdezeeuw:
Some more notes:

The atomic ordering might be overkill, but since it's a test API I don't think the performance impact matters too much. Also I didn't add a test, but the example tests the entire API already.

Thomasdezeeuw:
I think the Travis failure is unrelated, from the log:

```log
error[E0463]: can't find crate for `std`
  |
  = note: the `thumbv6m-none-eabi` target may not be installed
```

Other architectures are fine.

Nemo157:
Yeah, it’s https://github.com/rust-lang-nursery/futures-rs/issues/1396, I would fix it but I don’t really know if there’s a way to do so that doesn’t break other stuff.

Thomasdezeeuw:
I've force pushed a rebase on master, the CI should pass now as the failure was fixed in #1407.

futures-test/src/task/wake_counter.rs
```diff
@@ -68,3 +68,52 @@ impl Wake for Inner {
...
+
+/// Number of times the waker was awoken.
+///
+/// See [`new_count_waker`] for usage.
+#[derive(Debug)]
+pub struct AwokenCount {
+    inner: Arc<WakerInner>,
+}
+
+impl PartialEq<usize> for AwokenCount {
```
cramertj: I think `Deref<Target = usize>` would also be handy.

cramertj:
Seems good to me! I'd prefer we replaced the old one at the same time, though, or at least marked it as deprecated and switched our internal usage. No reason to let tech debt sit around on an alpha version of the library.

Thomasdezeeuw:
@cramertj  I'm going to address the comments now, to respond to your comment in the other issue.

cramertj:
sg, thanks!

futures-test/src/task/wake_counter.rs
Thomasdezeeuw: How can I do that? I don't think `AtomicUsize` has any methods that allow you to get a reference to the inner usize. I tried the following

```rust
impl Deref for AwokenCount {
    type Target = usize;

    fn deref(&self) -> &usize {
        &self.inner.count.load(Ordering::SeqCst)
    }
}
```

But that gives the `cannot return reference to temporary value` error, which makes sense.

I could add back the `get(&self) -> usize` method

Thomasdezeeuw:
Pushed 2 commits to changes the tests and remove the old `WakeCounter`. Should I implement the `get(&self) -> usize` method?

Thomasdezeeuw:
@cramertj I think the test will pass (they work on my machine ™), if your OK with the API then I'm going to bed now.

cramertj:
Yup, looks good, thanks for the changes!

PR #1406, opened 2019-01-11
fix build failed with rust nightly-20190110

leaxoy:
This pr fix build error with rust nightly-20190110


gilescope:
Most of the changes are cargo fmt. Looks like the only change is that Pin::set takes &mut? LGTM.

taiki-e:
Addressed in #1407

PR #1412, opened 2019-01-14
Stage 0.3.0-alpha.12

cramertj:


Nemo157:
Idea I had after seeing some issues with pinning the versions of the sub-crates on urlo and discord: should all the internal dependencies be exact version constraints during the alphas? We don't really provide any back-compat between the different alphas so there's little chance of having a mixture of versions working.

cramertj:
You mean should we set `=x.y.z` rather than `x.y.z`? I'd be fine with that.

cramertj:
I've pushed a change that pins the exact versions.

Nemo157:
:+1: Yep, that's what I meant. Changes LGTM.

PR #1414, opened 2019-01-15
Fix `Stream::select`

king6cong:
Prevent the situation where one stream is starved.
Corresponding test also re-enabled.


cramertj:
Nice, thanks!

PR #1415, opened 2019-01-15
Add #[must_use] to Stream and Sink

taiki-e:
This adds `#[must_use]` attribute to `Stream` and `Sink`, and this unifies `#[must_use]` message of Sink.

Related: rust-lang/rust#56677, rust-lang/rust#57549


cramertj:
Great! I love that we can `#[must_use]` traits now :)

PR #1418, opened 2019-01-16
Update dependencies

taiki-e:
* either 1.4 -> 1.5
* num_cpus 1.8.0 -> 1.9
* lazy_static 1.1.0 -> 1.2
* syn 0.15.22 -> 0.15.25

This PR mainly updates minor versions of dependencies.
`num_cpus` and `lazy_static` seem to be stable enough, so I think that it is probably OK without specifying a patch version.
Also, `non_camel_case_type` warning was allowed in dtolnay/syn#563, so I would like to include updating `syn`.

### DIFF
--- a/futures-core/Cargo.toml
+++ b/futures-core/Cargo.toml
@@ -20,7 +20,7 @@ std = ["either/use_std"]
 nightly = []
 
 [dependencies]
-either = { version = "1.4", default-features = false, optional = true }
+either = { version = "1.5", default-features = false, optional = true }
 
 [dev-dependencies]
 futures-preview = { path = "../futures", version = "=0.3.0-alpha.12" }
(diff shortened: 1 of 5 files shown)


taiki-e:
I thought that it might be possible to move `rand_core` to `dev-dependencies`, so I made additional tests on this.
As a result, it succeeds when building this crate with `-Zminimal-versions` ([Cargo.toml](https://github.com/taiki-e/futures-rs/blob/deps-ci/futures-util/Cargo.toml#L47), [CI](https://travis-ci.org/taiki-e/futures-rs/jobs/480372084)), but failed if it was built with `-Zminimal-versions` as a dependency of other crates ([CI](https://travis-ci.com/taiki-e/test-deps/jobs/170841447)).

[By using ">=0.2.2, <0.4" instead of "0.2.2"](https://github.com/taiki-e/futures-rs/blob/dependencies/futures-util/Cargo.toml#L36), in addition to being able to delete unnecessary compilation in the normal build ([cargo tree](https://travis-ci.com/taiki-e/test-deps/jobs/170851023)), it can be made successful even if it is built with `-Zminimal-versions` as a dependency of other crates ([CI](https://travis-ci.com/taiki-e/test-deps/jobs/170851020)).

taiki-e:
> (since users would already get the newer versions by default in a cargo build w/o a lockfile or a cargo update)

I wrote this PR because I found older versions in dependencies when I built it locally, but I seem to have forgotten this.

PR #1422, opened 2019-01-16
Add `Stream::select2` to support short-circuited select

king6cong:
As discussed in https://github.com/websockets-rs/rust-websocket/issues/136

Often, we have two streams combined by `select`:
1. the steam we care about, e.g. a steam of network events, as in the case of websocket connection.
2. an auxiliary stream, e.g. stream of timeouts or as in the discussion above, a stream of user inputs.

The stream of network events will complete when there is a network failure or something else, but the auxiliary stream often never completes. 

When network steam completes, some actions should be take, such as a retry. But because the combined stream doesn't complete, we have no means to know whether the network steam have completed.

With the added `Stream::select2` and `short_circuit` being `true`, the returned stream completes when one of the two input stream completes, which is suitable for situations like this.

### DIFF
--- a/futures-util/src/stream/mod.rs
+++ b/futures-util/src/stream/mod.rs
@@ -962,11 +962,33 @@ pub trait StreamExt: Stream {
     ///
     /// Note that this method consumes both streams and returns a wrapped
     /// version of them.
+    /// Also note that when one stream completes, it will not be `Drop`ed until
+    /// both stream complete. Resources it takes up will not be released in
+    /// the meanwhile. See also:
+    /// https://github.com/rust-lang-nursery/futures-rs/pull/1422
     fn select<St>(self, other: St) -> Select<Self, St>
         where St: Stream<Item = Self::Item>,
               Self: Sized,
     {
-        Select::new(self, other)
+        Select::new(self, other, false)
+    }
+
+    /// This combinator will attempt to pull items from both streams. Each
+    /// stream will be polled in a round-robin fashion, and whenever a stream is
+    /// ready to yield an item that item is yielded.
+    ///
+    /// With `short_circuit` being `true`, the returned stream completes when one of
+    /// the two input stream completes.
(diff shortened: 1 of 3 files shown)


cramertj:
> When network steam completes, some actions should be take, such as a retry.

The way this API works today, it's impossible to know which stream was the one that completed, which seems like it would be necessary for this sort of retry behavior. Is that not the case?

king6cong:
>  it's impossible to know which stream was the one that completed, which seems like it would be necessary for this sort of retry behavior

Yes, we don't know which one completes, maybe we should add this capability? Not sure about this, but it is not necessary in this case, because we can retry the combined `select` as a whole, without knowing which stream competes, especially when the other stream is auxiliary.

As in this case:

https://github.com/websockets-rs/rust-websocket/blob/ad7df37bd9e1b2590017cc24fb48abf1c64cc8f7/examples/async-client.rs#L64

When the websocket server is stopped, an tcp `FIN` is received on the client side and `ACK`ed immediately, but:
* in the `complete when both stream completes` version of `select`, client will hold on the tcp connection without a `FIN` sent. We are unaware of this and continue waiting on a tcp connection which will no longer receive any data.
* in the `complete when any stream completes` version of `select`,  client will finish the tcp connection with a `FIN` immediately. The combined stream completes and we can take action If needed

king6cong:
@cramertj @alexcrichton ping :)

cramertj:
TBH I'm not really sure what to do with this. This seems like a pretty uncommon case to me, and you could replicate this behavior by e.g. using `try_select` and converting the end of the stream into an error and catching that error. On the other hand, the alternatives are sort of hacky and we don't have a good syntax for ergonomically writing things like this by-hand outside of the crate (e.g. what `async/await` did for normal futures).

414owen:
This looks like it's a duplicate of https://github.com/rust-lang/futures-rs/pull/1881

@taiki-e looks like a lot of people want this feature :)

PR #1425, opened 2019-01-17
Update rand dependency to 0.6

taiki-e:
This updates rand dependency to 0.6.4 from 0.5.5

See rust-random/rand#645 for the reason why `rand_core` was added to the dependency (TL;DR: Building with `-Zminimal-versions` go well by adding this.).

This PR was split from #1418.


cramertj:
LGTM, thanks for the context on the `minimal-versions` issue.

PR #1426, opened 2019-01-17
Update rand to 0.6.4

xfix:
Doing that to lower amount of different versions of the same dependency in my program.


taiki-e:
Covered by #1425

PR #1429, opened 2019-01-18
Make io Compat (0.1) still usable with futures 0.3

dbcfd:
Implement AsyncRead and AsyncWrite for Compat, such that it is still usable within a futures 0.3 context.

### DIFF
--- a/futures-util/src/compat/compat03as01.rs
+++ b/futures-util/src/compat/compat03as01.rs
@@ -185,6 +185,26 @@ mod io {
         }
     }
 
+    impl<T: AsyncRead03> AsyncRead03 for Compat<T> {
+        fn poll_read(&mut self, lw: &std::task::LocalWaker, buf: &mut [u8])
+                     -> task03::Poll<Result<usize, std::io::Error>> {
+            self.inner.poll_read(lw, buf)
+        }
+    }
+
+    impl<T: AsyncWrite03> AsyncWrite03 for Compat<T> {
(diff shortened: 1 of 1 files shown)


cramertj:
Interesting! I don't think we have these pure-delegation impls for the other traits. Can you say more about why you needed this? I'd think that we'd want to guide folks away from using this as it indicates that they might be calling `compat` unnecessarily.

dbcfd:
Looking at stream of stream use cases, where we have mixing of different types. This came from attempting to port tokio-tls to futures 0.3: https://github.com/dbcfd/tls-async

Definitely open to other approaches, but would like to be able to write 0.3 code, only downgrading to 0.1 at the application main if possible.

Does it make more sense to make 0.3 AsyncRead/AsyncWrite compatible with Read/Write which native-tls relies on?

cramertj:
I'd instead have the application expose 0.3 types which the end-user can call `Compat` on if necessary.

dbcfd:
So you're saying to implement Read/Write for AsyncRead/AsyncWrite so that it can be used with native-tls?

I essentially have `tls_async::TlsStream<native_tls::TlsStream<romio::TcpStream>>` where tls_async::TlsStream expects AsyncR/W, and native_tls::TlsStream expects R/W, so then I end up with `tls_async::TlsStream<CompatToProvideAsyncR/W<native_tls::TlsStream<CompatToProvideR/W<romio::TcpStream>`

dbcfd:
Changes to layering made this no longer needed, works correctly without it, closing.

PR #1433, opened 2019-01-22
Remove TokioDefaultSpawner and tokio-compat

seanmonstar:
These pieces were when `Future` received a `Context` and needed to be able to spawn futures. It no longer does, so these can be removed.


cramertj:
IMO it's still useful to have an implementation of the `Spawn` trait that is cooperates with the tokio default executor, and I'd be open to keeping this around for that purpose. Is there an argument against keeping this here to be used that way?

seanmonstar:
I dunno, seems like something that would be provided by tokio someday, and it's main reason for existing *in* the futures crate no longer exists...

Is there actual concrete usage where people are using tokio and the `Spawn` trait?

cramertj:
> Is there actual concrete usage where people are using tokio and the Spawn trait?

No idea, but it doesn't seem that weird of a thing to do, e.g. if a 0.3-based crate wanted to be generic over a custom executor, one of which being the tokio executor.

seanmonstar:
It looks like the `Executor01As03` should handle that case (though looks like it's bounds could be loosened to not require `Clone + Send`).

cramertj:
Okay, SGTM! :)

PR #1435, opened 2019-01-23
Implement try_join_all() and TryJoinAll

### LINKED ISSUE #1420: try_join_all() in futures-preview 0.3
It seems that `futures-preview` 0.3 includes `future::join_all()` with similar behavior to `futures` 0.1, but there doesn't seem to be any `future::try_join_all()` for futures resolving to `Result<T, E>`. I was relying on being able to join on a `Vec` of futures while short-circuiting on `Err` in one of my projects, and it would be nice to have this in the new API as well.

ebkalderon:
### Added

* Add new `futures_util::try_future::try_join_all` submodule.
* Implement `try_join_all()` function and `TryJoinAll` future, closely mimicking the existing `JoinAll` design.
* Add doc and integration tests for `try_join_all()`.

### Changed

* Fix mention of "method" to "function" in `join_all.rs`.
* Superficial tweaks in `join_all.rs` test.

Fixes #1420.

### DIFF
--- a/futures-util/src/try_future/try_join_all.rs
+++ b/futures-util/src/try_future/try_join_all.rs
@@ -1,0 +1,182 @@
+//! Definition of the `TryJoinAll` combinator, waiting for all of a list of
+//! futures to finish with either success or error.
+
+use std::fmt;
+use std::future::Future;
+use std::iter::FromIterator;
+use std::mem;
+use std::pin::Pin;
+use std::prelude::v1::*;
+use std::task::Poll;
+use super::TryFuture;
+
+#[derive(Debug)]
+enum ElemState<F>
+where
+    F: TryFuture,
+{
+    Pending(F),
+    Done(Option<Result<F::Ok, F::Error>>),
+}
+
+impl<F> ElemState<F>
+where
+    F: TryFuture,
+{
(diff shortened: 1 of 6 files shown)


futures-util/src/try_future/try_join_all.rs
```diff
@@ -0,0 +1,182 @@
...
+use std::task::Poll;
+use super::TryFuture;
+
+#[derive(Debug)]
+enum ElemState<F>
+where
+    F: TryFuture,
+{
+    Pending(F),
+    Done(Option<Result<F::Ok, F::Error>>),
```
Nemo157: Since this is shortcircuiting errors, I don't think `ElemState::Done(Some(Err(...)))` is ever used? You could simplify this to `Done(Option<F::Ok>)`.

ebkalderon: Good point, that's certainly correct. I'll make the change real quick.

ebkalderon:
Thanks for the suggestion, @Nemo157. I've also identified and removed an unnecessary `#[derive(Debug)]` on `FinalState` which I didn't mean to include in the final PR.

Issue #1369, opened 2018-12-11
LocalPool::poll_pool is private - how does one advances the pool as much as possible without blocking ?


gnzlbg:
I'd like to advance `LocalPool` as much as possible without blocking, i.e., I don't want to use `block_on` to block until all task finish, but I want to do as much work as can be done, and when no more work can be done, I want to do some other computation and then some time later poll the pool again. 

It appears that `LocalPool::poll_pool` is what I need, but it is private. Is there another way to do this ?

SrTobi:
I think `LocalPool::poll_pool` does some internal stuff.

I myself needed a method that polls exactly one ready future and returns afterwards or if none is ready returns as well. I think those methods should be added to local pool. Like `poll_at_most(n: i32) -> i32` and `poll_available() -> i32`.

For the moment I used `FuturesUnordered`. @gnzlbg maybe this code helps you:

```rust
let mut pool = stream::FuturesUnordered::<FutureObj<'static, ()>>::new();

let f = async {
    println!("first");
    3
};

let handle = pool.spawn_with_handle(f).unwrap();

let mapped_f = async {
    let res = await!(handle);
    println!("second: {}", res);
};

pool.spawn(mapped_f);

// this never resolves and will stay in the pool forever.
// nevertheless the loop below will exit, as after the two previous futures
// have been resolved, no more ready futures remain in pool
pool.spawn(future::empty());


block_on(future::lazy(|lw| {
    loop {
        if let Poll::Ready(Some(_)) = pool.poll_next_unpin(lw) {
            // found one more
            println!("done one more");
        } else {
            // either more ready futures (or none at all)
            println!("nothing to do");
            break;
        }
    }
}));
```

prints:

```
first
done one more
second: 3
done one more
nothing to do
```

PR #1437, opened 2019-01-26
Added functions to partially progress a local pool.

SrTobi:
I added two functions (`poll` and `poll_one`) that only partially progress the local_pool (see #1369).

I think it is ok to put these functions into the `LocalPool` as they make no sense in thread pool.

Both functions are useful when wanting to run tasks in a certain thread (for example a gui thread), while the same thread has also other work to do in an event loop for instance. I myself used the *boost::asio::io_service* equivalents multiple times with very good results in c++ projects.

Currently I'm not really satisfied with the names of the functions. Not sure if the *poll* term is correct here, given that  it is a very loaded term in the context of rust futures. Other possible names would be `run_available` / `run_one_available` or `try_run_all` / `try_run_one`... not really sure. What do you think?

### DIFF
--- a/futures-executor/src/local_pool.rs
+++ b/futures-executor/src/local_pool.rs
@@ -77,6 +77,19 @@ fn run_executor<T, F: FnMut(&LocalWaker) -> Poll<T>>(mut f: F) -> T {
     })
 }
 
+// Set up and run a basic single-threaded spawner and return the result of `f`
+fn poll_executor<T, F: FnMut(&LocalWaker) -> T>(mut f: F) -> T {
+    let _enter = enter()
+        .expect("cannot execute `LocalPool` executor from within \
+                 another executor");
+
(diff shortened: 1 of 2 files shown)


futures-executor/src/local_pool.rs
```diff
@@ -157,20 +170,94 @@ impl LocalPool {
         })
     }
 
+    /// Runs all tasks and returns after completing one future or until no more progress
```
cramertj: In Fuchsia we call this function `run_until_stalled`. Maybe that would work here as well?

yilinwei:
@SrTobi Apologies, I rebased on top of master [here](https://github.com/yilinwei/futures-rs/tree/SrTobi/pool-partial-progress) but can't submit a PR on top of your branch. Would you be OK if I changed the function name and submitted it as a separate PR?

SrTobi:
Sure, go ahead.

Issue #626, opened 2017-10-29
no_std + alloc feature selection


wez:
I'd like to make use of futures in some embedded (arm) and separately in some kernel (macOS) code.  In both of these places the code is `no_std` but implements the allocator api such that `alloc::arc::Arc` and `alloc::boxed::Box` are available.

I'm relatively new to rust so I may be wrong, but it appears to be impossible from outside `futures-rs` to replicate a couple of things that would unlock using futures in these situations:

* the `ArcWrapped` `UnsafeNotify` implementation to avoid losing ones mind trying to implement notify handles
* `impl Future for Box` which appears to be necessary to implement an executor like this one: https://github.com/alexcrichton/futures-rs/blob/2b3f0f0cc6b60f2659092dd3f2f88844c3f4bd5b/tests/support/local_executor.rs#L29

If there were a crate level feature selection to enable `Box` and `Arc` related conveniences and do so in terms of `alloc` rather than the `std` counterparts, I think it would make it significantly easier to integrate futures in these environments.

I'm also interested in guidance on how to achieve this without modifying `futures-rs`.

(I'd love to try `futures-await` in these environments too, while you're thinking about this!)

alexcrichton:
Thanks for the report! Unfortunately though `alloc` isn't a stable feature of the standard library (e.g. it's not a stable crate to link against) so there's not quite a compilation target for this crate to use on stable, so I'd be hesitatnt to add support. We could perhaps, however, look to get a stable slice of the standard library for this!

wez:
Thanks; that makes sense re: `alloc`.

I did manage to solve my immediate need by not using the `Spawn`/`executor` stuff; I realized that I can just poll all of the `Future`/`Stream` instances directly from the main loop of the MCU.  It's not quite as efficient as the notify handle stuff would allow, but the cardinality of these things is typically very low in the embedded world and it is no worse than the pattern of just polling everything in a loop that is prevalent in a lot of embedded code.

I'd love to try a `no_std` `futures-await`, but at least I have a decent `futures` foundation to build on top of now :-)

alexcrichton:
Oh so actually futures-await is probably totally compatible with `no_std`, I think it just needs to disable the feature when depending on the `futures` crate actually!

carllerche:
It looks like the outcome of this is "no action" for futures-rs. I'm going to close the issue. Let me know if I am wrong.

jrobsonchase:
[...]

PR #1438, opened 2019-01-28
[WIP] Add no_std + alloc support

taiki-e:
This adds the `alloc` features to `futures`, `futures-core`, `futures-sink`, and `futures-util`.

This allows to use `Box<Future>`, `alloc::task`, many `*Ext` traits's methods etc in the `no_std` + `alloc` environment.

This PR adds [alloc-shim](https://github.com/taiki-e/alloc-shim) crate as an optional dependency. ~~This is only valid if the `alloc` feature is enabled.~~
~~`alloc-shim` will re-export items of `core` and `alloc` in the same layout as `std`. This allows adding `no_std` + `alloc` support without changing existing imports.~~ 
Edit: The current `alloc-shim` has the same layout as the alloc crate.

Some other points:

* When using the `alloc` feature, the `nightly` feature must be used at the same time (see this [comment](https://github.com/rust-lang-nursery/futures-rs/issues/626#issuecomment-457906007) in #626).

* ~~`#[cfg(alloc)]` is the same as `#[cfg(any(feature = "alloc", feature = "std"))]`, but it is simpler and easier to write (`build.rs` was added for this).~~
  Edit:  The implementation was changed to use `#[cfg(feature = "alloc")]` (thanks! @Nemo157).

* `futures::stream::ReuniteError` implements `std::error::Error` only if "std" feature is valid.

* ~~Fixed an `unused_imports` warning in `no_std` build~~. Edit2: It was split into #1447

## Reason to add `alloc-shim` to dependencies

The layout in the prelude module is different for `std` and `alloc`, so `alloc::prelude::v1::*` can not be used (and `alloc::sync` does not include `atomic` module). ~~Also, until now, we've imported everything from `std`, but from now on we'll need to import them from `alloc` and `core`, respectively. Using `alloc-shim` can avoid these problems.~~

~~`alloc` crate is still unstable, so I think `alloc-shim` will be unnecessary in the future, but until then it will be possible to support `no_std` + `alloc` with less cost by using `alloc-shim` (also, `alloc-shim` is designed so that the build will not fail even if `atomic` module is added to `alloc::sync`).~~

~~Also, I've examined several other ways ([1](https://github.com/taiki-e/futures-rs/compare/master...alloc-no-deps), [2](https://github.com/taiki-e/futures-rs/compare/master...alloc-reexport), [3](https://github.com/taiki-e/futures-rs/compare/master...alloc-shim)), but I believe this is the simplest and cleanest.~~

Edit: 

Using `alloc-shim` can avoid the first problem. 

Also, this is an implementation that does not use `alloc-shim`: [alloc-no-deps](https://github.com/taiki-e/futures-rs/compare/master...alloc-no-deps)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -54,6 +54,14 @@ matrix:
         - cargo build --manifest-path futures-sink/Cargo.toml --no-default-features
         - cargo build --manifest-path futures-util/Cargo.toml --no-default-features
 
+    - name: cargo build (alloc)
+      rust: nightly
+      script:
+        - cargo build --manifest-path futures/Cargo.toml --no-default-features --features=alloc,nightly
+        - cargo build --manifest-path futures-core/Cargo.toml --no-default-features --features=alloc,nightly
+        - cargo build --manifest-path futures-sink/Cargo.toml --no-default-features --features=alloc,nightly
+        - cargo build --manifest-path futures-util/Cargo.toml --no-default-features --features=alloc,nightly
+
     - name: cargo build (default features)
       rust: nightly
       script:
--- a/futures-core/Cargo.toml
+++ b/futures-core/Cargo.toml
@@ -16,11 +16,13 @@ name = "futures_core"
 
 [features]
 default = ["std"]
-std = ["either/use_std"]
+std = ["alloc", "either/use_std", "alloc-shim/std"]
 nightly = []
+alloc = ["alloc-shim/alloc"]
 
 [dependencies]
 either = { version = "1.4", default-features = false, optional = true }
+alloc-shim = { version = "0.3.0", features = ["futures"], optional = true }
 
 [dev-dependencies]
 futures-preview = { path = "../futures", version = "=0.3.0-alpha.12" }
--- a/futures-core/src/future/future_obj.rs
+++ b/futures-core/src/future/future_obj.rs
@@ -185,10 +185,11 @@ where
     unsafe fn drop(_ptr: *mut ()) {}
 }
 
-#[cfg(feature = "std")]
-mod if_std {
+#[cfg(feature = "alloc")]
+mod if_alloc {
     use super::*;
-    use std::mem;
+    use core::mem;
+    use alloc::boxed::Box;
 
     unsafe impl<'a, T, F> UnsafeFutureObj<'a, T> for Box<F>
         where F: Future<Output = T> + 'a
--- a/futures-core/src/future/mod.rs
+++ b/futures-core/src/future/mod.rs
@@ -33,9 +33,11 @@ impl<F: FusedFuture + ?Sized> FusedFuture for &mut F {
     }
 }
 
-#[cfg(feature = "std")]
-mod if_std {
+#[cfg(feature = "alloc")]
+mod if_alloc {
+    use alloc::boxed::Box;
     use super::*;
+
     impl<F: FusedFuture + ?Sized> FusedFuture for Box<F> {
         fn is_terminated(&self) -> bool {
             <F as FusedFuture>::is_terminated(&**self)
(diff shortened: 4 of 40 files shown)


taiki-e:
Related: #626

cc @cramertj, @Nemo157

taiki-e:
`alloc-shim` is a very simple crate as follows:

```rust
#![cfg_attr(feature = "alloc", feature(alloc, futures_api))]

#[cfg(all(feature = "alloc", not(feature = "std")))]
extern crate alloc as liballoc;

#[cfg(not(feature = "std"))]
mod shim {
    // str: https://doc.rust-lang.org/nightly/core/str/index.html vs https://doc.rust-lang.org/nightly/alloc/str/index.html
    pub use core::{str, *};

    #[cfg(feature = "alloc")]
    pub use liballoc::{alloc, borrow, fmt, slice, task, *};

    // The layout in the prelude module is different for `std` and `alloc`.
    /// The alloc Prelude
    #[cfg(feature = "alloc")]
    pub mod prelude {
        /// The alloc Prelude
        pub mod v1 {
            pub use core::prelude::v1::*;
            pub use liballoc::prelude::*;
        }
    }

    // `alloc::sync` does not include `atomic` module
    /// Synchronization primitives
    #[cfg(feature = "alloc")]
    pub mod sync {
        pub use core::sync::atomic;
        pub use liballoc::sync::*;
    }
}

#[cfg(not(feature = "std"))]
pub use self::shim::*;

#[cfg(feature = "std")]
pub use std::*;
```

The alternative to `alloc-shim` I think is to create a crate in `futures-rs` that provide equivalent features.

jrobsonchase:
>     `pub use core::{str, *};`

Probably don't really need to facade anything from `core` since it's already available everywhere.

>    `pub use liballoc::{alloc, borrow, fmt, slice, task, *};`

Why import all of these specifically and then the glob as well?

> ```
>     pub mod prelude {
>         /// The alloc Prelude
>         pub mod v1 {
> ```

Is the goal here to imitate the `alloc` crate or `std`? There's no `alloc::prelude::v1` module.

Overall I really like the idea behind `alloc-shim`, but I think we need to make up our minds about what its actual goal is. IMO, it should strive to look as close to `alloc` itself as possible and nothing more.

Nemo157:
One goal that I think this should aim for is minimal changes (and no extra dependencies) once `alloc` is stabilized. So, while this results in the least churn right now, [`alloc-no-deps`](https://github.com/taiki-e/futures-rs/compare/master...alloc-no-deps) appears to me to be moving closest to that endgoal (although because of the slight differences between `std` and `alloc`'s API this may still want to use something like `extern crate alloc_shim as alloc;` to map `std` -> `alloc`).

---

I just thought of an alternative to the build-script based detection, you can have your original plan of `std` implying `alloc`, then force the extra `nightly` feature when not using `std` with something like

```rust
#[cfg(all(feature = "alloc", not(any(feature = "std", feature = "nightly"))))]
compile_error!("The `alloc` feature without `std` requires the `nightly` feature active to explicitly opt-in to unstable features");
```

Then switch back to using `#[cfg(feature = "alloc")]` everywhere else.

taiki-e:
> I just thought of an alternative to the build-script based detection, you can have your original plan of std implying alloc, then force the extra nightly feature when not using std with something like

It's great, thanks!

---

Previously I thought that it would be better for us to use alloc like std, but now I think that this is a bad idea (@jrobsonchase, @Nemo157 Thanks for pointing that out). And I rewrote this PR and `alloc-shim` so that it was almost the same as after `alloc` crate was stabilized.

taiki-e:
When rust-lang/rust#58175 is done, I will update this PR and write a new summary.

PR #1439, opened 2019-01-28
Remove deprecated ATOMIC_USIZE_INIT

taiki-e:
This fixes deprecation warnings.


seanmonstar:
Hm, seems that CI build hung... I can't figure out how to restart it, since the link leads to an odd page.

Nemo157:
I think that Travis CI really bugged out on this one, there should only be one check instance and the "The build" link in the details pages which leads to the build on travis-ci.com leads to the same page for both...

Maybe try force-pushing to reset it?

cramertj:
Thanks for the cleanup!

PR #1441, opened 2019-01-28
Allow !Unpin Sinks in Forward combinator

### LINKED ISSUE #1440: Possible regression with mpsc channels closing early
I have a simple project which spawns a directed acyclic graph of tasks which each feed data into a `futures::channel::mpsc::Sender` while the main thread `await`s the `futures::channel::mpsc::Receiver` for progress reported from each task. The DAG is constructed by calling `future::join_all()` on a set of dependencies and then chaining the task itself with `.then()`. I have a minimal verifiable example for the code in question at [ebkalderon/sink-futures-example](https://github.com/ebkalderon/sink-futures-example).

Given the following task graph:

```
+-------+   +--------+
| first |   | second |
+-------+   +--------+
       \     /
      +-------+
      | final |
      +-------+
```

Each future forwards a `stream::once(future::ok(...))` to the `Sender`. When the code was written with `futures` 0.1.25, the `Receiver` stream results in the output:

```
Ok("first")
Ok("second")
Ok("final")
```

However, when the code is ported to `futures` 0.3.0-alpha.12, with very minimal changes, the `Receiver` stream unexpectedly closes early after the first element:

```
Ok("first")
```

I would hazard a guess that the `stream::once()` in `futures` 0.3.0-alpha.12 is causing the sink to close early, whereas the `stream::once()` in 0.1.25 does not, but I'm not familiar enough with the implementation of the `mpsc` channels to know if this is the case.

ebkalderon:
### Changed

* Update `Forward` so it drops the sink without returning it. Users wishing to access the sink again can pass a `&mut sink` to `StreamExt::forward()`.
* Update docs for both `Forward` and `StreamExt::forward()`.

### Removed

* Remove requirement for `Unpin` sinks on `Forward`.

Fixes #1440. /CC @Nemo157.

### DIFF
--- a/futures-util/src/stream/forward.rs
+++ b/futures-util/src/stream/forward.rs
@@ -9,16 +9,10 @@ use pin_utils::{unsafe_pinned, unsafe_unpinned};
 const INVALID_POLL: &str = "polled `Forward` after completion";
 
 /// Future for the `Stream::forward` combinator, which sends a stream of values
-/// to a sink and then flushes the sink.
-///
-/// Note: this is only usable with `Unpin` sinks, so `Sink`s that aren't `Unpin`
-/// will need to be pinned in order to be used with this combinator.
-//
-// This limitation is necessary in order to return the sink after the forwarding
-// has completed so that it can be used again.
+/// to a sink and then flushes and closes the sink.
 #[derive(Debug)]
 #[must_use = "steams do nothing unless polled"]
-pub struct Forward<St: Stream, Si: Sink + Unpin> {
+pub struct Forward<St: Stream, Si: Sink> {
     sink: Option<Si>,
     stream: Fuse<St>,
     buffered_item: Option<Si::SinkItem>,
@@ -28,7 +22,7 @@ impl<St: Stream + Unpin, Si: Sink + Unpin> Unpin for Forward<St, Si> {}
 
 impl<St, Si> Forward<St, Si>
 where
-    Si: Sink + Unpin,
+    Si: Sink,
     St: Stream<Item = Result<Si::SinkItem, Si::SinkError>>,
 {
     unsafe_pinned!(sink: Option<Si>);
@@ -68,7 +62,7 @@ impl<St: Stream, Si: Sink + Unpin> FusedFuture for Forward<St, Si> {
 
 impl<St, Si> Future for Forward<St, Si>
 where
-    Si: Sink + Unpin,
+    Si: Sink,
     St: Stream<Item = Result<Si::SinkItem, Si::SinkError>>,
 {
     type Output = Result<(), Si::SinkError>;
@@ -91,7 +85,7 @@ where
                 Poll::Ready(None) => {
                     try_ready!(self.as_mut().sink().as_pin_mut().expect(INVALID_POLL)
                                    .poll_close(waker));
-                    let _ = self.as_mut().sink().take().unwrap();
+                    let _ = self.as_mut().sink().as_pin_mut().take().unwrap();
                     return Poll::Ready(Ok(()))
                 }
                 Poll::Pending => {
--- a/futures-util/src/stream/forward.rs
+++ b/futures-util/src/stream/forward.rs
@@ -85,7 +85,7 @@ where
                 Poll::Ready(None) => {
                     try_ready!(self.as_mut().sink().as_pin_mut().expect(INVALID_POLL)
                                    .poll_close(waker));
-                    let _ = self.as_mut().sink().as_pin_mut().take().unwrap();
+                    self.as_mut().sink().set(None);
                     return Poll::Ready(Ok(()))
                 }
                 Poll::Pending => {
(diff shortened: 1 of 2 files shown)


cramertj:
It seems like maybe we should *require* passing `Pin<&mut impl Sink>` rather than taking a sink by-value, since the default behavior of forwarding and then dropping without closing will always be wrong (right?).

Nemo157:
Another alternative would be to just update the documentation to say this will close the provided `Sink` once the `Stream` is complete (and fix the `mpsc::Sender` to not kill the whole channel when closed like it currently does). You could then use something like an unclosable adaptor on a `Sink` that just flushes but doesn't actually close it if you want to keep it active after using `forward`.

ebkalderon:
@Nemo157 That sounds like a good idea. I can update the documentation, if you would like. Could you please elaborate on what you have in mind with an uncloseable adapter? Would this be another combinator we would add to the `SinkExt` trait?

ebkalderon:
Another thing to note: if we do not need to return the sink directly from `Forward`, we can also remove the `Unpin` sink requirement [mentioned here](https://rust-lang-nursery.github.io/futures-api-docs/0.3.0-alpha.12/src/futures_util/stream/forward.rs.html#15-19).

Nemo157:
I've started the mpsc changes in #1443.

The uncloseable adaptor would just wrap a sink and call `poll_flush` from its `poll_close`, so once the adaptor was closed the underlying sink would be flushed but left open, you could then use it like

```rust
let mut sink = foo();
await!(stream.forward(sink.by_ref().dont_close()))?;
await!(sink.send(some_final_item))?;
```

ebkalderon:
Sounds good! I've just pushed some commits updating the `Forward` docs and removing the `Unpin` sink bound as well. I would appreciate a soundness review of my changes here.

ebkalderon:
Pinging @cramertj @Nemo157..Do the docs and the removal of the `Unpin` sink bound look correct to you?

ebkalderon:
Pinging @cramertj @Nemo157. Do the docs and the removal of the `Unpin` sink bound from the last two commits look correct to you?

futures-util/src/stream/forward.rs
```diff
@@ -90,8 +84,9 @@ where
                 Poll::Ready(Some(Err(e))) => return Poll::Ready(Err(e)),
                 Poll::Ready(None) => {
                     try_ready!(self.as_mut().sink().as_pin_mut().expect(INVALID_POLL)
-                                   .poll_close(lw));
-                    return Poll::Ready(Ok(self.as_mut().sink().take().unwrap()))
+                                   .poll_flush(lw));
```
Nemo157: This should be back to `poll_close`, right?

futures-util/src/stream/forward.rs
```diff
@@ -90,8 +84,9 @@ where
...
+                    let _ = self.as_mut().sink().as_pin_mut().take().unwrap();
```
Nemo157: I think this will have to be `self.as_mut().sink().set(None)`, currently you're calling `take` on a temporary `Option<Pin<&mut Si>>` which won't cause the actual `Sink` held to be dropped.

futures-util/src/stream/forward.rs
ebkalderon: If I switch this back to `poll_close()`, the example code in the repo on #1445 doesn't work as expected.

futures-util/src/stream/forward.rs
ebkalderon: Ah, good catch! Thanks, I'll fix that right away.

futures-util/src/stream/forward.rs
ebkalderon: When I switch this back to `poll_close()`, then the use of `forward()` in the example code shown in #1440 no longer works as expected.

ebkalderon: Thanks for merging #1443, @cramertj! I've switched back to `poll_close()`, as I can confirm on my machine that `StreamExt::forward()` now behaves as expected, identically to 0.1.25.

ebkalderon:
@cramertj @Nemo157 Now that #1443 has been merged, I've updated my PR to switch back to using `poll_close()` over `poll_flush()`, and I have confirmed that the example code between 0.3.0 and 0.1.25 now behaves correctly. Ready for review again!

ebkalderon:
Interesting, `futures-core-preview` no longer compiles on some toolchains after my rebase. Will need to look into it further.

Nemo157:
That’s #1453, there are new changes in nightly that are waiting on #1445

ebkalderon:
Ah, yes. I just noticed this as well when trying to navigate to `std::task::LocalWaker` on the Rust API docs page and it wasn't found. Standing by until #1445 gets merged.

ebkalderon:
@cramertj @Nemo157 Awesome, looks like CI passes again thanks to #1445 being merged. 🎉 Anyone available for a final review of this PR?

cramertj:
LGTM-- would you mind flattening your commits to remove the previous changes to `Forward` that have since been removed?

ebkalderon:
@cramertj Okay, I've flattened away the redundant commits! I hope the commit history looks satisfactory now.

cramertj:
Thanks!

PR #1443, opened 2019-01-29
Don't close the whole mpsc channel when one sender is closed

### LINKED ISSUE #1440: Possible regression with mpsc channels closing early

Nemo157:
I've confirmed locally that this fixes #1440, all senders are given a chance to send before the channel is closed.

There's probably a nicer way to do this with less unwrapping, suggestions welcome otherwise I might take another look whether I can clean this up a bit tomorrow.

### DIFF
--- a/futures-channel/src/mpsc/mod.rs
+++ b/futures-channel/src/mpsc/mod.rs
@@ -94,11 +94,8 @@ use crate::mpsc::queue::Queue;
 
 mod queue;
 
-/// The transmission end of a bounded mpsc channel.
-///
-/// This value is created by the [`channel`](channel) function.
 #[derive(Debug)]
-pub struct Sender<T> {
+struct SenderInner<T> {
     // Channel state shared between the sender and receiver.
     inner: Arc<Inner<T>>,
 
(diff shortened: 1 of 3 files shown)


carllerche:
Could the change include tests?

ebkalderon:
Looks pretty good. Should this PR also include similar changes to `UnboundedSender` as well? I haven't tested myself to see if it fails in a similar fashion to #1440.

Nemo157:
It does, both have a `disconnect` method added and both `Sink` implementations are updated to use it. I'll add test cases covering both as well.

I have thought of a cleaner way to do this, move all the internal implementation details into an `InnerSender` struct (although maybe a different name to avoid having `.inner.inner` fields) which `Sender` and `UnboundedSender` contain in an `Option`, that way disconnecting a specific sender is just dropping that inner struct, and there should be no unwrapping in the internal functions.

ebkalderon:
Looks like this PR is done, with test cases added and the `SenderInner` implemented. Anyone free for review?

ebkalderon:
Looks like this PR is done, with test cases added and `SenderInner` implemented. Anyone ready to review?

ebkalderon:
Seems like this PR is ready for review, with test cases added and the `SenderInner` implemented.

ebkalderon:
@Nemo157 Just wondering, is this PR still being worked on? It appears to be finished.

Nemo157:
It's all done from my side (but with the All Hands happening this coming week, there's a good chance there's no one available to review for a bit).

cramertj:
Thanks for the fix!

PR #1446, opened 2019-02-02
Implement Clone for oneshot Sender

the20login:
Use additional Arc to track number of `Sender`, use additional struct `SenderInnerWrapper` to call `inner.drop_tx` when all `Sender` are gone.

Usage example(won't compile): https://play.rust-lang.org/?version=stable&mode=release&edition=2018&gist=6425707852edc26be77a60dae65ac2d0

### DIFF
--- a/futures-channel/src/oneshot.rs
+++ b/futures-channel/src/oneshot.rs
@@ -25,6 +25,12 @@ pub struct Receiver<T> {
 /// This is created by the [`channel`](channel) function.
 #[derive(Debug)]
 pub struct Sender<T> {
+    wrapper: Arc<SenderInnerWrapper<T>>,
+}
+
+// Use intermediate struct to to track number of Senders
+#[derive(Debug)]
+struct SenderInnerWrapper<T> {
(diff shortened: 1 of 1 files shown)


Matthias247:
What's the use-case for this? Since a Oneshot can only be completed once, having multiple senders won't make sense, and would mainly introduce one more possible error that is possible (another sender has already completed the Oneshot).

For the rare events where this is necessary, once can just wrap the current `Sender` implementation inside the application in something like `Arc<Mutex<Option<Sender>>>`, and implement send by moving the inner sender out and calling send on it.

The proposed change makes the type more expensive for all users due to 2 `Arc`s. I guess the second `Arc` could be avoided, but then there might be other overhead since the current atomic ops won't be possible anymore.

the20login:
Thanks for the tip about types composition, I'm new to Rust and not catch common patterns yet.

I would disagree about rarity of use case, but again, I'm new to Rust.
If you think this change have more cons than pros, I'm ok this this and will not insist.

PR #1448, opened 2019-02-05
Use #![warn(rust_2018_idioms)]

taiki-e:
Currently, only `#![bare_trait_objects)]` is used (bare_trait_objects is included in rust_2018_idioms), but since `#![deny(rust_2018_idioms)]` is used in rust-lang/rust (see rust-lang/rust#58099), I would like to use ~~`#![deny(rust_2018_idioms)]`~~`#![warn(rust_2018_idioms)]` even in futures-rs.

### DIFF
--- a/futures-channel/src/lib.rs
+++ b/futures-channel/src/lib.rs
@@ -8,7 +8,7 @@
 #![cfg_attr(not(feature = "std"), no_std)]
 
 #![warn(missing_docs, missing_debug_implementations)]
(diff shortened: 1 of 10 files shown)


Nemo157:
`warn` is nicer for developers, and will be denied by CI when a PR is made.

taiki-e:
> `warn` is nicer for developers, and will be denied by CI when a PR is made.

That makes sense.

PR #1445, opened 2019-01-30
Implements the waker changes for task API stabilization.

Matthias247:
This removes LocalWaker, and adds ArcWake to the library.

PR #1456, opened 2019-02-19
Fix `no_std` build

### LINKED ISSUE #1396: Broken no_std build

Nemo157:
Fixes #1396

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -64,17 +64,26 @@ matrix:
       script:
         - cargo build --manifest-path futures/Cargo.toml --features io-compat
 
-    # Allow build fail until #1396 is fixed.
     - name: cargo build --target=thumbv6m-none-eabi
       rust: nightly
-      env: ALLOW_FAILURES=true
       install:
         - rustup target add thumbv6m-none-eabi
       script:
+        - cargo run --manifest-path ci/remove-dev-dependencies/Cargo.toml */Cargo.toml
         - cargo build --manifest-path futures/Cargo.toml
             --target thumbv6m-none-eabi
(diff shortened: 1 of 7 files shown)


Nemo157:
This is based off #1445, hopefully GitHub will strip those commits once it's merged... I was hoping to provide a comparison link until then but it doesn't look like cross-fork compare pages work.

The last two commits are pretty simple, but that workaround to test the `no_std` build is not the nicest... it's the only solution I could think of until Cargo's handling of features across normal and dev dependencies is improved.

Nemo157:
No auto-update ☹️ re-rebased onto master.

cramertj:
LGTM-- will merge once travis is happy

PR #1457, opened 2019-02-20
Cleanup

cramertj:
Fixes some leftover bits from https://github.com/rust-lang-nursery/futures-rs/pull/1445

### DIFF
--- a/futures-util/src/compat/compat03as01.rs
+++ b/futures-util/src/compat/compat03as01.rs
@@ -4,11 +4,21 @@ use futures_01::{
     StartSend as StartSend01, Stream as Stream01,
 };
 use futures_core::{
-    task as task03, TryFuture as TryFuture03, TryStream as TryStream03,
+    task::{
+        self as task03,
+        RawWaker,
+        RawWakerVTable,
+    },
+    TryFuture as TryFuture03,
+    TryStream as TryStream03,
 };
 use futures_sink::Sink as Sink03;
-use crate::task::ArcWake as ArcWake03;
-use std::{pin::Pin, sync::Arc};
+use crate::task::{ArcWake as ArcWake03, WakerRef};
+use std::{
+    mem,
+    pin::Pin,
+    sync::Arc,
+};
 
 /// Converts a futures 0.3 [`TryFuture`](futures_core::future::TryFuture),
 /// [`TryStream`](futures_core::stream::TryStream) or
@@ -108,6 +118,7 @@ where
     }
 }
 
+#[derive(Clone)]
 struct Current(task01::Task);
 
 impl Current {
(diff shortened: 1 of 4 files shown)


cramertj:
cc @Matthias247

futures-util/src/task/waker_ref.rs
```diff
@@ -38,28 +38,8 @@ impl<'a> Deref for WakerRef<'a> {
...
-macro_rules! owned_vtable {
-    ($ty:ident) => {
-        &RawWakerVTable {
-            clone: clone_arc_raw::<$ty>,
-            drop: drop_arc_raw::<$ty>,
-            wake: wake_arc_raw::<$ty>,
-        }
-    };
-}
+#[inline]
```
Matthias247: I doubt this has any effect, since the method is always virtually dispatched.

futures-util/src/task/arc_wake.rs
```diff
@@ -32,39 +23,43 @@ pub trait ArcWake {
...
+// code here. We should guard against this by aborting.
+
 unsafe fn increase_refcount<T: ArcWake>(data: *const()) {
     // Retain Arc by creating a copy
     let arc: Arc<T> = Arc::from_raw(data as *const T);
     let arc_clone = arc.clone();
     // Forget the Arcs again, so that the refcount isn't decrased
-    let _ = Arc::into_raw(arc);
-    let _ = Arc::into_raw(arc_clone);
+    mem::forget(arc);
```
Matthias247: I used `Arc::into_raw` since that's the version where the API guarantees that it's the opposite of `Arc::from_raw`.
forget will work too, but I thought if there is a more specific API available it might be nice to use it.
Anyway, both is fine.

futures-util/src/task/arc_wake.rs
```diff
@@ -32,39 +23,43 @@ pub trait ArcWake {
...
+    mem::forget(arc_clone);
 }
 
-unsafe fn clone_arc_raw<T: ArcWake>(data: *const()) -> RawWaker {
+// used by `waker_ref`
+pub(super) unsafe fn clone_arc_raw<T: ArcWake>(data: *const()) -> RawWaker {
```
Matthias247: Where are those things exported to now? Only one level above, or even further?
It's totally to fine to use those within other parts of futures-util, but I don't think the library should publicly export these.

futures-util/src/compat/compat03as01.rs
```diff
@@ -108,19 +118,40 @@ where
...
+        }
+        fn current_to_ptr(current: &Current) -> *const () {
+            current as *const Current as *const ()
+        }
+
+        unsafe fn clone(ptr: *const ()) -> RawWaker {
+            // Lazily create the `Arc` only when the waker is actually cloned.
+            // FIXME: remove `transmute` when a `Waker` -> `RawWaker` conversion
+            // function is landed in `core`.
+            mem::transmute::<task03::Waker, RawWaker>(
```
Matthias247: A method on ArcWake or the same module that returns a RawWaker seems cleaner than transmute.

The ideal solution might be to directly reference and reuse a refcounted thing within Task01. But after looking up the definition of that I don't think it's possible without the extra allocation and wrapping.

futures-util/src/task/arc_wake.rs
cramertj: They're just `pub(super)`-- so only available to the futures_util::task module.

Issue #1459, opened 2019-02-21
no api docs


tinaun:
alpha.13 is on too new a nightly for docs.rs to build correctly, and the website is still on alpha.12

taiki-e:
@cramertj It seems that stage2 (publish docs) was canceled because clippy failed https://github.com/rust-lang-nursery/futures-rs/runs/66484822

Nemo157:
Just published https://rust-lang-nursery.github.io/futures-api-docs/0.3.0-alpha.13/futures/

PR #1460, opened 2019-02-21
Fix clippy errors

taiki-e:
https://travis-ci.com/rust-lang-nursery/futures-rs/jobs/179360667

Related: #1459, https://github.com/rust-lang/rust-clippy/pull/3756

### DIFF
--- a/futures-util/src/compat/compat01as03.rs
+++ b/futures-util/src/compat/compat01as03.rs
@@ -75,7 +75,7 @@ impl<Fut: Future01> Future03 for Compat01As03<Fut> {
         mut self: Pin<&mut Self>,
         waker: &Waker,
     ) -> task03::Poll<Self::Output> {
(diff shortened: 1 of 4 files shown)


futures-util/src/compat/compat01as03.rs
```diff
@@ -163,13 +163,13 @@ mod io {
         fn poll_flush(&mut self, waker: &task03::Waker)
             -> task03::Poll<Result<(), Error>>
         {
-            poll_01_to_03(self.in_notify(waker, |x| x.poll_flush()))
+            poll_01_to_03(self.in_notify(waker, AsyncWrite01::poll_flush))
```
cramertj: IMO it's cleaner to have these match above-- I think this clippy lint seems a bit overzealous, but also not particularly worth arguing over...

PR #1465, opened 2019-03-01
Use 'self: Pin<&mut Self>' on AsyncRead/AsyncWrite

### LINKED ISSUE #1454: Should AsyncRead and AsyncWrite take self by Pin?
We didn't do this because there's not a lot of motivation, but on reflection there's also not a lot of downside. Implementations of these traits are a) usually concrete (and therefore `Unpin`) and b) usually don't need to worry too hard about moving their receiver around in poll_read/poll_write.

The motivation for this change would be that it would enable users to write mock AsyncRead sources from async fns or possibly async generators without boxing them. Possibly someday the same will be true for AsyncWrite once generators with resume arguments are a thing (along the same line as Sink).

cc @cramertj

taiki-e:
Closes #1454

concerns:

* Most methods of `AsyncReadExt`/`AsyncWriteExt` require `Self: Unpin`.

* ~~Add a new dependency ([pin-project](https://github.com/taiki-e/pin-project)).~~ - Removed

### DIFF
--- a/futures-io/src/lib.rs
+++ b/futures-io/src/lib.rs
@@ -18,6 +18,7 @@ mod if_std {
     use std::boxed::Box;
     use std::cmp;
     use std::io as StdIo;
+    use std::pin::Pin;
     use std::ptr;
 
     // Re-export IoVec for convenience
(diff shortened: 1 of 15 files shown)


taiki-e:
"cycle detected" error is related to https://github.com/rust-lang/rust/issues/55997.

cramertj:
This looks great, thanks so much! I'll go ahead and merge this for now, since it allows us to avoid some unnecessary allocation in the Stream -> AsyncRead shim and would allow the same in a hypothetical Sink/AsyncWrite shim, which could be extra-valuable when we get async streams / generators. If we discover people are hitting usability problems (which I don't anticipate) then we can revisit.

Issue #1455, opened 2019-02-18
Use std IoVec types in AsyncRead and AsyncWrite


withoutboats:
This issue is to track using the std `IoVec` types in our public API, instead of having the `iovec` crate as a public dependency. Currently these types are not in std, but they will be when rust-lang/rust#58357 lands. We'll have to decide when to make this shift: when they're available in nightly, or when they're available in stable.

Matthias247:
How about `no-std` support? AsyncRead/Write traits are so universal that it would be very nice to have them available on those environments.

`IoVec` seems to be one more type which isn't available in that environment and moves us away from that goal. A generic `&[&[u8]]` parameter might be nicer. Which can be either directly cast to a OS specific `IoVec` on some platforms, or must be copied into another vector in the implementation. In the bad case, the copy (of e.g. 64 ptr/length tuples) is not that expensive compared to the overall cost of that huge IO operation.

withoutboats:
I don't believe futures-io is intended to be no-std compatible any more than `std::io` is intended to be. The no-std ecosystem should develop its own abstractions appropriate to the kinds of guarantees that can be made about certain no-std environments. My understanding of futures-io is that it is modeled after `std::io` but async, and that it assumes a std-supporting platform.

Nemo157:
> The no-std ecosystem should develop its own abstractions

I really don't think we should have a split ecosystem at this level, `io::{Read,Write}` work fine as abstractions on `no_std` devices in a lot of (probably even the majority of) cases. The main issue is `io::Error`, but that is something that is [already being tracked by the portability WG](https://github.com/rust-lang-nursery/portability-wg/issues/12).

> I don't believe futures-io is intended to be no-std compatible any more than `std::io` is intended to be.

This I agree with though. As long as `futures-io` doesn't force more use of `std` than already exists in `std::io` then it should be able to start supporting `no_std` at the same time. To that end, if `std::io` is introducing a dependency on some platform specific `IoVec` type, then that will have to be sorted out if `std::io` abstractions are to be used on `no_std`, so `futures-io` also using this type shouldn't make the transition any harder.

PR #1476, opened 2019-03-14
Use std IoVec on AsyncRead/AsyncWrite

taiki-e:
By using this at an early stage, I think there is a high possibility of getting more tests and feedback early. I think that will have a positive impact on std `IoVec` API improvements and stabilization.

(Personally, at the time of stabilizing std `IoVec` API, it is preferable we are somewhat used to this API.)

Closes https://github.com/rust-lang-nursery/futures-rs/issues/1455

### DIFF
--- a/futures-io/Cargo.toml
+++ b/futures-io/Cargo.toml
@@ -15,12 +15,11 @@ The `AsyncRead` and `AsyncWrite` traits for the futures-rs library.
 name = "futures_io"
 
 [features]
-std = ["futures-core-preview/std", "iovec"]
+std = ["futures-core-preview/std"]
 default = ["std"]
 
 [dependencies]
 futures-core-preview = { path = "../futures-core", version = "=0.3.0-alpha.14", default-features = false }
-iovec = { version = "0.1", optional = true }
 
 [dev-dependencies]
 futures-preview = { path = "../futures", version = "=0.3.0-alpha.14" }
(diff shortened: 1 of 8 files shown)


seanmonstar:
Not to say that I'm against this, but just to call out explicitly: it's possible this could have these traits (and by extension all of futures-util) stuck on nightly longer, depending on if futures stabilize much sooner.

taiki-e:
Yeah, I think the concern is right. 

Current futures-util relies on the following unstable features:
```
#![feature(futures_api)]
#![cfg_attr(feature = "std", feature(async_await, await_macro, box_into_pin))]
```
I think futures_api in which the stabilization RFC is merged, and box_into_pin is likely to stabilize earlier than iovec. On the other hand, async_await and await_macro still have many [blockers](https://github.com/rust-lang/rust/labels/AsyncAwait-Blocking) and [unsolved problems](https://github.com/rust-lang/rust/issues/50547), and I think stabilization of these features is likely to be slower than iovec, so I opened this PR.

(This is my personal thinking (I'm not participating in the WG discussions), so it may not be correct.)

taiki-e:
Also, if a decision is made to close this PR to reduce uncertainty in the future, I will favor it.

Nemo157:
> On the other hand, async_await and await_macro still have many blockers and unsolved problems, and I think stabilization of these features is likely to be slower than iovec

Depending on stabilisation timing I’m pretty sure these parts of `futures` can be placed behind feature flags to allow moving the main part of the crate to stable.

Since there are default implementations of the vectored IO methods it seems like that could be possible for `iovec` as well.

cramertj:
I agree with all the above comments-- I think this is a good direction to move, but I'd cfg-out the methods for now unless both the `unstable` feature is enabled, and avoid using `iovec` otherwise.

PR #1479, opened 2019-03-18
Add no_std + alloc environment support take 2

taiki-e:
This replaces https://github.com/rust-lang-nursery/futures-rs/pull/1438. 

This allows to use `Box<dyn Future>`, many `*Ext` traits's methods etc... in the `no_std` + `alloc` environment.

* This adds the `alloc` features to `futures`, `futures-core`, `futures-sink`, and `futures-util`.
* When using the `alloc` feature, the nightly feature must be used at the same time.

cc @Nemo157

### DIFF
--- a/futures-core/src/lib.rs
+++ b/futures-core/src/lib.rs
@@ -2,6 +2,7 @@
 
 #![feature(futures_api)]
 #![cfg_attr(feature = "cfg-target-has-atomic", feature(cfg_target_has_atomic))]
+#![cfg_attr(all(feature = "alloc", not(feature = "std")), feature(alloc))]
 
 #![cfg_attr(not(feature = "std"), no_std)]
 
@@ -11,6 +12,14 @@
 
 #[cfg(all(feature = "cfg-target-has-atomic", not(feature = "nightly")))]
 compile_error!("The `cfg-target-has-atomic` feature requires the `nightly` feature as an explicit opt-in to unstable features");
+
+#[cfg(all(feature = "alloc", not(any(feature = "std", feature = "nightly"))))]
+compile_error!("The `alloc` feature without `std` requires the `nightly` feature active to explicitly opt-in to unstable features");
+
+#[cfg(all(feature = "alloc", not(feature = "std")))]
+extern crate alloc;
+#[cfg(feature = "std")]
+extern crate std as alloc;
 
 pub mod future;
 #[doc(hidden)] pub use self::future::{Future, FusedFuture, TryFuture};
--- a/futures-util/src/macros/mod.rs
+++ b/futures-util/src/macros/mod.rs
@@ -1,2 +1,12 @@
 #[macro_use]
 mod poll;
+
+macro_rules! cfg_target_has_atomic {
+    ($($item:item)*) => {$(
+        #[cfg_attr(
+            feature = "cfg-target-has-atomic",
+            cfg(all(target_has_atomic = "cas", target_has_atomic = "ptr"))
+        )]
+        $item
+    )*};
+}
--- a/.travis.yml
+++ b/.travis.yml
@@ -54,6 +54,14 @@ matrix:
         - cargo build --manifest-path futures-sink/Cargo.toml --no-default-features
         - cargo build --manifest-path futures-util/Cargo.toml --no-default-features
 
+    - name: cargo build (alloc)
+      rust: nightly
+      script:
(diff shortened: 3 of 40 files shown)


futures-core/src/lib.rs
```diff
@@ -12,6 +13,14 @@
 #[cfg(all(feature = "cfg-target-has-atomic", not(feature = "nightly")))]
 compile_error!("The `cfg-target-has-atomic` feature requires the `nightly` feature as an explicit opt-in to unstable features");
 
+#[cfg(all(feature = "alloc", not(any(feature = "std", feature = "nightly"))))]
+compile_error!("The `alloc` feature without `std` requires the `nightly` feature active to explicitly opt-in to unstable features");
+
+#[cfg(all(feature = "alloc", not(feature = "std")))]
+extern crate alloc;
+#[cfg(feature = "std")]
+extern crate std as alloc;
```
cramertj: Clarifying: because the portions of `std` that appear in `alloc` use an identical API, this declaration makes it possible for everything to compile on stable when the `alloc` feature is disabled but the `std` feature is enabled, or when both features are enabled (in which case the `alloc` crate won't be used, only the `std` crate).

futures-util/src/macros/mod.rs
```diff
@@ -1,2 +1,12 @@
 #[macro_use]
 mod poll;
+
+macro_rules! cfg_target_has_atomic {
```
cramertj: There are a bunch of spots in e.g. `futures-util/src/stream/mod.rs` that seem like they could benefit greatly from this-- or does that not work out because they're trait items?

cramertj:
This looks great! left a couple of small comments, but otherwise this seems like a great approach towards supporting `alloc`-only environments.

futures-core/src/lib.rs
taiki-e: Yes.

futures-util/src/macros/mod.rs
taiki-e: > -- or does that not work out because they're trait items?

Yeah, it doesn't work well with trait items... [playground](https://play.rust-lang.org/?version=nightly&mode=debug&edition=2018&gist=dd94d620920242baac820f0120e915bc)

taiki-e: Also, we should be able to work around this problem by using procedural macros, but in that case, I would like to compare with the approach using `proc_macro_attribute`.

1. `proc_macro`

```rust
// definition
#[proc_macro]
pub fn cfg_target_has_atomic(input: TokenStream) -> TokenStream {
    let mut file: File = parse_macro_input!(input);
    let attr = parse_quote! {
       #[cfg_attr(
            feature = "cfg-target-has-atomic",
            cfg(all(target_has_atomic = "cas", target_has_atomic = "ptr"))
        )]
    };
    file.items.iter_mut().for_each(|item| {
        match item {
            Item::Fn(ItemFn { attrs, .. }) 
            | Item::Use(ItemUse { attrs, .. }) 
            | Item::Mod(ItemMod { attrs, .. }) => attrs.push(attr.clone()),
            _ => {}        
        }
    });
    TokenStream::from(file.into_token_stream())
}

// usage
cfg_target_has_atomic! {
    #[cfg(feature = "alloc")]
    mod buffer_unordered;
    #[cfg(feature = "alloc")]
    pub use self::buffer_unordered::BufferUnordered;
}
```

2. `proc_macro_attribute`

```rust
// definition
#[proc_macro_attribute]
pub fn cfg_target_has_atomic(_args: TokenStream, input: TokenStream) -> TokenStream {
    let mut tts = TokenStream::from(quote! {
       #[cfg_attr(
            feature = "cfg-target-has-atomic",
            cfg(all(target_has_atomic = "cas", target_has_atomic = "ptr"))
        )]
    });
    tts.extend(input);
    tts
}

// usage
#[cfg_target_has_atomic]
#[cfg(feature = "alloc")]
mod buffer_unordered;
#[cfg_target_has_atomic]
#[cfg(feature = "alloc")]
pub use self::buffer_unordered::BufferUnordered;
```

`1` is especially useful when many items are in one place, but the indent is deeper. 
(I have no strong opinion on which one to use.)

cramertj: Yeah, that'd work-- I don't have particularly strong thoughts either way. We can leave that for a future PR, though, since it doesn't seem like the current situation is too bad (just some extra boilerplate in a few spots).

PR #1481, opened 2019-03-21
Change Sink's associated item type to a generic Sink<Item>

### LINKED ISSUE #623: Make Sink::SinkItem a generic parameter
See https://github.com/tokio-rs/tokio-core/issues/273.

This would allow `Sink`s to accept values of multiple different types. I'm sure there's a reason it's an associated type right now, but I can't quite figure out why that is.

Example:

```rust
pub trait Sink<Item> {
    type SinkError;
    fn start_send(
        &mut self, 
        item: Item
    ) -> StartSend<Item, Self::SinkError>;
    fn poll_complete(&mut self) -> Poll<(), Self::SinkError>;
    ...
}
```
The same goes for the tokio-io [Encoder](https://docs.rs/tokio-io/0.1/tokio_io/codec/trait.Encoder.html) trait:
```rust
pub trait Encoder<Item> {
    type Error: From<Error>;
    fn encode(
        &mut self, 
        item: Item, 
        dst: &mut BytesMut
    ) -> Result<(), Self::Error>;
}
```
This would solve [this tokio-io issue](https://github.com/tokio-rs/tokio-io/issues/54), because users could write an impl like:

```rust
struct Bytes;

impl<'a> Encoder<&'a [u8]> for Bytes {
    type Error = io::Error;
    fn encode(&mut self, data: &'a [u8]>, buf: &mut BytesMut) -> io::Result<()> {
        buf.put(data);
        Ok(())
    }
}

impl Encoder<Vec<u8>> for Bytes {
    type Error = io::Error;
    fn encode(&mut self, data: Vec<u8>, buf: &mut BytesMut) -> io::Result<()> {
        buf.put(&data[..]);
        Ok(())
    }
}
```
Conceptually, `Item` is an input to a `Sync` or `Encoder`, rather than an output like it is in `Stream`, `Future`, and `Iterator`. As an input, it makes sense for it to be variant. For other examples of this design, see the `Fn<...>` traits, which use `Args` as a generic parameter, and `Output` as an associated type.

seanmonstar:
Closes #623

### DIFF
--- a/futures-sink/src/channel_impls.rs
+++ b/futures-sink/src/channel_impls.rs
@@ -3,8 +3,7 @@ use futures_core::task::Waker;
 use futures_channel::mpsc::{Sender, SendError, TrySendError, UnboundedSender};
 use std::pin::Pin;
 
-impl<T> Sink for Sender<T> {
-    type SinkItem = T;
+impl<T> Sink<T> for Sender<T> {
     type SinkError = SendError;
 
     fn poll_ready(mut self: Pin<&mut Self>, waker: &Waker) -> Poll<Result<(), Self::SinkError>> {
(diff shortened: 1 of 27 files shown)


cramertj:
Nice! Works for me. Thanks for the change :)

Issue #793, opened 2018-02-23
Rename or remove `futures_unordered` module in 0.2


aturon:
The fact that we have a function and module both named `futures_unordered` is confusing both to people and to rustdoc.

alexcrichton:
I'd also be fine removing iteration as it [didn't solve the original problem](https://github.com/rust-lang-nursery/futures-rs/issues/605#issuecomment-343706577) anyway

PR #1482, opened 2019-03-21
Remove futures_[un]ordered functions

### LINKED ISSUE #1477: Iterators for FuturesUnordered are private
The `IterMut` and `IterPinMut` types returned by [`futures::stream::FuturesUnordered::iter_mut`](https://rust-lang-nursery.github.io/futures-api-docs/0.3.0-alpha.13/futures/stream/struct.FuturesUnordered.html#method.iter_mut) and [`futures::stream::FuturesUnordered::iter_pin_mut`](https://rust-lang-nursery.github.io/futures-api-docs/0.3.0-alpha.13/futures/stream/struct.FuturesUnordered.html#method.iter_pin_mut) are not exposed.

taiki-e:
Closes #1477

### DIFF
--- a/futures-util/src/stream/mod.rs
+++ b/futures-util/src/stream/mod.rs
@@ -123,12 +123,13 @@ cfg_target_has_atomic! {
     #[cfg(feature = "alloc")]
     mod futures_ordered;
     #[cfg(feature = "alloc")]
-    pub use self::futures_ordered::{futures_ordered, FuturesOrdered};
+    pub use self::futures_ordered::FuturesOrdered;
 
     #[cfg(feature = "alloc")]
-    mod futures_unordered;
+    pub mod futures_unordered;
     #[cfg(feature = "alloc")]
-    pub use self::futures_unordered::{futures_unordered, FuturesUnordered};
+    #[doc(inline)]
+    pub use self::futures_unordered::FuturesUnordered;
 
     #[cfg(feature = "alloc")]
     mod split;
--- a/futures-util/src/stream/futures_ordered.rs
+++ b/futures-util/src/stream/futures_ordered.rs
@@ -85,8 +85,8 @@ impl<T> Future for OrderWrapper<T>
 /// some of the later futures have already completed.
 ///
 /// Note that you can create a ready-made `FuturesOrdered` via the
-/// `futures_ordered` function in the `stream` module, or you can start with an
-/// empty queue with the `FuturesOrdered::new` constructor.
+/// [`collect`](Iterator::collect) method, or you can start with an empty queue
+/// with the `FuturesOrdered::new` constructor.
 #[must_use = "streams do nothing unless polled"]
 pub struct FuturesOrdered<T: Future> {
     in_progress_queue: FuturesUnordered<OrderWrapper<T>>,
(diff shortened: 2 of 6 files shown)


futures-util/src/stream/mod.rs
```diff
@@ -128,7 +128,7 @@ cfg_target_has_atomic! {
     #[cfg(feature = "alloc")]
     mod futures_unordered;
     #[cfg(feature = "alloc")]
-    pub use self::futures_unordered::{futures_unordered, FuturesUnordered};
+    pub use self::futures_unordered::{futures_unordered, FuturesUnordered, IterMut, IterPinMut};
```
cramertj: If we're going to expose these at top-level, IMO they need less generic names (`FuturesUnorderedIterMut` or something like that)

taiki-e: Indeed!

futures-util/src/stream/mod.rs
```diff
@@ -126,8 +126,9 @@ cfg_target_has_atomic! {
     pub use self::futures_ordered::{futures_ordered, FuturesOrdered};
 
     #[cfg(feature = "alloc")]
-    mod futures_unordered;
+    pub mod futures_unordered;
```
taiki-e: It was not possible to overwrite `FuturesUnordered::iter_pin_mut` and `FuturesUnordered::iter_mut` return types in the document simply by renaming on export, so I changed the module itself to be exposed like [`proc_macro2::token_stream`](https://docs.rs/proc-macro2/0.4/proc_macro2/token_stream/index.html).

taiki-e: futures 0.1 also exposed `futures_unordered` module: https://docs.rs/futures/0.1.25/futures/stream/futures_unordered/index.html

taiki-e: I overlooked the discussion in https://github.com/rust-lang-nursery/futures-rs/issues/793.

Nemo157:
Personally I'd go for [@cramertj's proposal](https://github.com/rust-lang-nursery/futures-rs/issues/793#issuecomment-368576067) in the linked issue, remove the `futures_unordered` function and use `FuturesUnordered::from_iter` instead. Then the iterators can be namespaced away since they're going to be very rarely used types.

Probably good to wait and see if @cramertj's opinion has changed since then though.

cramertj:
I don't have a strong opinion either way. @Nemo157 feel free to review this and merge if it matches your expectations.

Nemo157:
Thanks, this looks great.

Benjamin-L:
I don't know if I'm missing something, but what should people be doing instead of `futures_ordered` now?

Nemo157:
@Benjamin-L you can use the [`FromIterator` implementation of `FuturesOrdered`](https://rust-lang-nursery.github.io/futures-api-docs/0.3.0-alpha.13/futures/stream/struct.FuturesOrdered.html#impl-FromIterator%3CFut%3E). Either by using `some_futures.into_iter().collect::<FuturesOrdered>()` or `use std::iter::FromIterator; FuturesOrdered::from_iter(some_futures);` (I really wish that `FromIterator` made it into the prelude so that that could be the goto recommendation).

Arnavion:
Needs to be mentioned in the alpha.14 changelog since it's a breaking change. (Took me a bunch of git-blame-ing to discover this.)

PR #1484, opened 2019-03-21
0.1: Correct the behavior of Chunks when items are ZST

taiki-e:
See #1483


cramertj:
Thanks! Do you need me to do an 0.1 release?

taiki-e:
That will be great. Thanks!

PR #1485, opened 2019-03-21
Remove unused lazy static GLOBAL_POOL

seanmonstar:


futures-executor/src/local_pool.rs
```diff
@@ -192,17 +191,9 @@ impl Default for LocalPool {
     }
 }
 
-lazy_static! {
-    static ref GLOBAL_POOL: ThreadPool = ThreadPool::builder()
```
cramertj: huh-- too bad this didn't generate a warning :(

taiki-e: @cramertj `lazy_static` applies `#[allow(dead_code)]` to the code: https://github.com/rust-lang-nursery/lazy-static.rs/blob/1.3.0/src/lib.rs#L153

PR #1491, opened 2019-03-24
Don't wake oneshot sender after it has been dropped

### LINKED ISSUE #1490: oneshot channel wakes sender after it has completed
Noticed in https://github.com/rust-lang-nursery/futures-rs/issues/1471

When the receiver for a oneshot channel is dropped it wakes the sender's task up in case it cares about the channel being canceled. In normal operation of `spawn_with_handle` this means that there will be a spurious wakeup of the already completed task that was spawned when the handle is dropped. It seems likely that clearing the registered `tx_task` when the sender is dropped will be cheaper than having the executor deal with the spurious wakeup.

Nemo157:
Fixes #1490


cramertj:
Good catch! r=me once updated.

PR #1492, opened 2019-03-24
Never error

### LINKED ISSUE #1399: Add .uninhabited_error() next to .unit_error()
I recently rewrote my code to use futures 0.3. One of the issues I ran into was that hyper requires a future 0.1 where the error type implements `std::error::Error`. The natural solution to this would be to use `.unit_error()`, but this does not implement the required trait.

For my own purposes I basically copy-pasted the implementation of `UnitError`, but with the error type set to `!`, which implements every trait.

I would suggest deprecating `.unit_error()` in favour of `.uninhabited_error()` since it's more representative of what is actually happening (the `Future::Output` will always be `Result::Ok`), but since for example `tokio` currently requires spawned futures to have a unit error this would be unergonomic.

Nemo157:
Fixes #1399 

@PvdBerg1998 I went with the name `never_error` for consistency with the "never type", alternate opinions welcome.

This does introduce another unstable feature, but one that [will hopefully be stabilized soon](https://github.com/rust-lang/rust/issues/57012).


cramertj:
I'd be happy to merge this once it's updated.

PR #1493, opened 2019-03-24
Add sink::for_each function

### LINKED ISSUE #1360: Add a sink::for_each function to create a Sink from a closure
I want to create a `Sink` from a closure that takes values and asynchronously processes them. There should be a function in `futures::sink` which does this. I imagine the function signature looking something like this:

```rust
fn for_each<F, Fut, T, E>(f: F) -> ForEach<T, E>
where
    F: FnMut(T) -> Fut,
    Fut: Future<Output = Result<(), E>>,
{ ... }

impl Sink for ForEach<T, E> {
    type SinkItem = T;
    type SinkError = E;
    ...
}
```

Nemo157:
Fixes #1360

Not the easiest function to use when capturing an environment because of the async closure limitations, but with some `Rc<RefCell>` usage it's possible.

cc @canndrew

### DIFF
--- a/futures-util/src/sink/for_each.rs
+++ b/futures-util/src/sink/for_each.rs
@@ -1,0 +1,120 @@
+use core::marker::PhantomData;
+use core::pin::Pin;
+use futures_core::future::Future;
+use futures_core::task::{Waker, Poll};
+use futures_sink::Sink;
+use pin_utils::{unsafe_pinned, unsafe_unpinned};
+
+/// [`Sink`] for the [`for_each`] function.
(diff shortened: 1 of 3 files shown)


canndrew:
Ah, cool! Thanks for that, though I don't remember what I needed it for now :p

cramertj:
I'd prefer to hold off on adding this until we have a concrete usecase, especially since it's pretty hard to use correctly at the moment.

cramertj:
(feel free to re-open if a compelling use-case arises)

PR #1494, opened 2019-03-24
Remove StreamObj, caution against using FutureObj

### LINKED ISSUE #1352: Remove StreamObj and soft-deprecate FutureObj
For some prior context see https://github.com/rust-lang-nursery/futures-rs/pull/1348

Now that futures and streams are object safe there should be no reason to use `FutureObj`/`StreamObj` directly. There are currently two reasons I know of to use these:

1. Passing around owned dynamically typed futures/streams. This can be replaced with use of the aliases below.
2. Spawning futures via [`Spawn::spawn_obj`](https://rust-lang-nursery.github.io/futures-api-docs/0.3.0-alpha.9/futures/task/trait.Spawn.html#tymethod.spawn_obj). Rather than using this directly most users can use [`SpawnExt::spawn`](https://rust-lang-nursery.github.io/futures-api-docs/0.3.0-alpha.9/futures/task/trait.SpawnExt.html#method.spawn) instead to automatically convert the future into a `FutureObj` (the only exception I know of is `no_std` use).

(am I missing any other use cases? I admit I've personally never used `FutureObj`)

Proposed additions to `futures`:

```rust
type BoxFuture<'a, T> = Pin<Box<dyn Future<Output = T> + Send + 'a>>;
type BoxStream<'a, T> = Pin<Box<dyn Stream<Item = T> + Send + 'a>>;
```

Using `SpawnExt::spawn` for a `BoxFuture<'static, ()>` will cause a double indirection, this can be mitigated by either specialization (maybe, I don't know whether this is specializable) or an additional `SpawnExt::spawn_box` method that directly converts the boxed future into a `FutureObj`.

```rust
trait SpawnExt {
    fn spawn_box(&mut self, future: BoxFuture<'static, ()>) -> Result<(), SpawnError> { ... }
}
```

Nemo157:
Fixes #1352


futures-core/src/future/mod.rs
```diff
@@ -8,6 +8,11 @@ pub use core::future::Future;
 mod future_obj;
 pub use self::future_obj::{FutureObj, LocalFutureObj, UnsafeFutureObj};
 
+#[cfg(feature = "alloc")]
+/// An owned dynamically typed [`Future`] for use in cases where you can't
+/// statically type your result or need to add some indirection.
+pub type BoxFuture<'a, T> = Pin<alloc::boxed::Box<dyn Future<Output = T> + Send + 'a>>;
```
cramertj: I'd also like to see non-`Send` versions of these-- maybe `BoxFutureLocal`? unsure-- we can sort that out in a separate change :)

cramertj: (similarly, adding a `boxed_local` would be great)

cramertj:
Why not remove `FutureObj` as well?

futures/tests/futures_unordered.rs
```diff
@@ -68,8 +67,8 @@ fn finished_future() {
     let (c_tx, c_rx) = oneshot::channel::<i32>();
 
     let mut stream = futures_unordered(vec![
-        FutureObj::new(Box::new(a_rx)),
-        //FutureObj::new(Box::new(b_rx.select(c_rx))),
+        a_rx.boxed() as BoxFuture<_>,
```
cramertj: hmm... I kind of think we should change the default return of `boxed` to the trait-object version to make this cast unnecessary. WDYT?

cramertj:
Apologies for the delay in reviewing this!

cramertj:
(I need to keep a better eye on the repo-- or perhaps we should set up something bors-like so that I can get notifications-- I guess pinging me on the thread would also accomplish this goal)

Nemo157:
`FutureObj` is still needed for `no_std` support with `Spawn`.

--

Having bors here would also be useful for merging multiple PRs (I assume you're using the admin override to merge without CI passing, I can only merge a PR if it's up to date with master because of CI limitations).

futures/tests/futures_unordered.rs
Nemo157: Sounds good to me, if users want a non-trait-object they can call `Box::new` themselves.

Nemo157: One issue with this is that `BoxFuture` can't actually be directly converted to `FutureObj` because of #1535, which is why https://github.com/rust-lang-nursery/futures-rs/pull/1494/files#diff-57ad78f21e9043ebe29659ff5d335774R9 was necessary. I've got a fix for #1535 ready to open once this is merged (it requires an enhancement to the `Pin` API so will be blocked on a `rust` PR).

cramertj: Sorry, can you clarify-- is #1535 blocked on this change, or is this change now blocked on #1535 ? What's the change to `Pin`? Do you have a PR open / issue open already for that?

Nemo157: This change exacerbates the issue from #1535 since `FutureExt::boxed` now returns `BoxFuture` and `FutureObj::new` can’t take a `BoxFuture` since it doesn’t support boxed unsized values.

https://github.com/Nemo157/futures-rs/commit/b4361780faf764c7cb046ed75f863a6fcfd44800 is the change I’m referring to which will allow converting a `BoxFuture` to a `FutureObj`, that’s the one using the `Pin` api changes.

Nemo157: Oh, the API changes I was referring to are https://github.com/rust-lang/rust/pull/60165, they can be worked around similar to the old code if we want to get a stable release out for 1.36.

cramertj: In that case, would you like to go ahead and land this change, or would you prefer to hold off?

Nemo157: I think landing this as is is fine, I'll open a PR for that other change straight away so it can be merged once https://github.com/rust-lang/rust/pull/60165 is (or updated to work around it), I just didn't do it as a parallel PR as it would definitely have merge conflicts with this one.

cramertj: sg, merged this one.

PR #1496, opened 2019-03-24
Add no_std environment support to futures-channel

taiki-e:
In the no_std environment, use own spinlock and spin_loop_hint because the OS provided mutex and the OS scheduler cannot be used (this change does not affect the behavior when the std function is enabled.).

(It seems that no other currently opened PRs will add other std items to futures-channel, so I opened this.)

cc @cramertj @Nemo157

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -59,6 +59,7 @@ matrix:
       script:
         - cargo build --manifest-path futures/Cargo.toml --no-default-features --features alloc,nightly
         - cargo build --manifest-path futures-core/Cargo.toml --no-default-features --features alloc,nightly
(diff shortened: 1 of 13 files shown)


Nemo157:
A spinlock is not going to work if one end of the channel is moved into an interrupt handler (the main source of "threading" I know of in `no_std`). I guess this might be useful if you're running `no_std` on a multi-core chip of some sort, but at that point I assume you'd have a more efficient mutex implementation that you'd prefer to inject somehow (and I have no experience to evaluate whether this implementation is something that would be wanted).

taiki-e:
On second thought, given that alloc support is just being added, and the feature this PR will add is not at the request of someone who actually uses it, currently I think this feature should at least be postponed.

PR #1499, opened 2019-03-25
Remove unnecessary todo

taiki-e:
These are already implemented as `try_join*`.


Nemo157:
:frown: why doesn’t github have a way to rebase open PRs onto master automatically yet.

PR #1501, opened 2019-03-26
Unify the style of combinator type descriptions

### LINKED ISSUE #1214: Standardize combinator type descriptions
Currently the descriptions for the combinator types vary greatly. Some of them link to their combinator method, others include a short description about what they do and some do it in a different order:

For `FutureExt`'s `Map`:
https://github.com/rust-lang-nursery/futures-rs/blob/c9627b8d50a8a8561025577c90b6f8409040af55/futures-util/src/future/map.rs#L7-L9

For `TryFutureExt`'s `FlattenSink`:
https://github.com/rust-lang-nursery/futures-rs/blob/c9627b8d50a8a8561025577c90b6f8409040af55/futures-util/src/try_future/flatten_sink.rs#L15

Here's what I suggest:
Let's rewrite all descriptions in the style of the `TryFutureExt` combinator types:
- DRY (do not repeat yourself): Let's *not* include a short description. The combinator method always already has a more detailed, better phrased description. Linking to that is enough/better.
- Always link to the combinator method. The user should be able to click on a link that takes them to the combinator method.

taiki-e:
Although I did not read each one in detail, I excluded those that might have explained in more detail than the functions/methods and those that would be unified if other PR(#1492) landed.

Closes #1214

### DIFF
--- a/futures-util/src/future/empty.rs
+++ b/futures-util/src/future/empty.rs
@@ -3,9 +3,7 @@ use core::pin::Pin;
 use futures_core::future::{Future, FusedFuture};
 use futures_core::task::{Waker, Poll};
 
-/// A future which is never resolved.
-///
-/// This future can be created with the [`empty()`] function.
+/// Future for the [`empty`] combinator.
 #[derive(Debug)]
 #[must_use = "futures do nothing unless polled"]
 pub struct Empty<T> {
--- a/futures-util/src/future/empty.rs
+++ b/futures-util/src/future/empty.rs
@@ -3,7 +3,7 @@ use core::pin::Pin;
 use futures_core::future::{Future, FusedFuture};
 use futures_core::task::{Waker, Poll};
 
-/// Future for the [`empty`] combinator.
+/// Future for the [`empty`] function.
 #[derive(Debug)]
 #[must_use = "futures do nothing unless polled"]
 pub struct Empty<T> {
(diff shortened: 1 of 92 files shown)


Nemo157:
😁 I was thinking about doing exactly this soon after re-noticing the inconsistency in #1492.

futures-util/src/future/empty.rs
```diff
@@ -3,9 +3,7 @@ use core::pin::Pin;
 use futures_core::future::{Future, FusedFuture};
 use futures_core::task::{Waker, Poll};
 
-/// A future which is never resolved.
-///
-/// This future can be created with the [`empty()`] function.
+/// Future for the [`empty`] combinator.
```
Nemo157: Are the free functions still "combinator"s? Maybe a better term here would be "constructor"?

taiki-e: -- or drop "combinator"?

Nemo157: Dropping "combinator" and rephrasing slightly to

> Future produced by [`empty`].

seems ok to me.

taiki-e: It may be easier to understand using "function"s and "method"s rather than "combinator"s (I do not think that functions like `join_all` is "constructor").

Nemo157: That makes sense as well, so just replace "combinator" directly with "function" or "method" depending on which it is (or even go with just "function" everywhere since method is a sub-category of function).

Nemo157: Checking the `std` docs they do differentiate between "function" or "method" for the iterator types, and don't mention "combinator" at all:

> This `struct` is created by the `once` function.

> This `struct` is created by the `filter` method on `Iterator`.

Nemo157:
Cool, thanks a lot for cleaning this up ❤️

PR #1505, opened 2019-03-28
Propagate cfg-target-has-atomic feature

taiki-e:
Currently `cfg-target-has-atomic` feature is not propagate from util to core, the E0599 error occurs when executing `cargo build --manifest-path futures-util/Cargo.toml  --target thumbv6m-none-eabi  --no-default-features --features nightly,cfg-target-has-atomic`.
```text
error[E0599]: no method named `compare_and_swap` found for type `core::sync::atomic::AtomicUsize` in the current scope
   --> futures-core/src/task/__internal/atomic_waker.rs:201:26
    |
201 |         match self.state.compare_and_swap(WAITING, REGISTERING, Acquire) {
```


cramertj:
Thanks! :)

PR #1507, opened 2019-04-01
Remove unused sink/wait.rs

taiki-e:
This was added in https://github.com/rust-lang-nursery/futures-rs/pull/406, but is currently disabled and `Stream::wait` is also removed at the stage of version 0.2.


cramertj:
Yeah, that was replaced with [`executor::block_on_stream`](https://docs.rs/futures-preview/0.3.0-alpha.13/futures/executor/fn.block_on_stream.html). `executor::block_on_sink` seems like perhaps a reasonable thing to have as well, but in the meantime this seems fine to remove.

PR #1509, opened 2019-04-02
s/Async::/Poll::/

yoshuawuyts:
Updates all remaining mentions of `Async::` to use `Poll::` instead. This includes the temporarily disabled tests, which I assume is fine because they are also using the new `Waker` API.

Thanks!


Nemo157:
Thanks 👍

PR #1511, opened 2019-04-03
impl AsyncWrite for Vec<u8>

Nemo157:
Mostly useful for test purposes


cramertj:
Surprised this wasn't already there tbh

PR #1514, opened 2019-04-07
Update to new futures_api

taiki-e:
This PR covers changes by https://github.com/rust-lang/rust/pull/59119 and https://github.com/rust-lang/rust/pull/59733.

cc #1518 (I initially intended to close it with this PR, but in fact it might be better to close it when the next version is released.)

### DIFF
--- a/futures-util/src/task/arc_wake.rs
+++ b/futures-util/src/task/arc_wake.rs
@@ -17,7 +17,22 @@ pub trait ArcWake {
     ///
     /// Executors generally maintain a queue of "ready" tasks; `wake` should place
     /// the associated task onto this queue.
-    fn wake(arc_self: &Arc<Self>);
+    fn wake(self: Arc<Self>) {
+        Self::wake_by_ref(&self)
+    }
+
+    /// Indicates that the associated task is ready to make progress and should
+    /// be `poll`ed.
+    ///
+    /// This function can be called from an arbitrary thread, including threads which
+    /// did not create the `ArcWake` based `Waker`.
+    ///
+    /// Executors generally maintain a queue of "ready" tasks; `wake_by_ref` should place
+    /// the associated task onto this queue.
+    ///
+    /// This function is similar to `wake`, but must not consume the provided data
+    /// pointer.
+    fn wake_by_ref(arc_self: &Arc<Self>);
 
     /// Creates a `Waker` from an Arc<T>, if T implements `ArcWake`.
     ///
@@ -25,10 +40,10 @@ pub trait ArcWake {
     /// the `wake()` function that is defined inside this trait will get called.
     fn into_waker(self: Arc<Self>) -> Waker where Self: Sized
     {
-        let ptr = Arc::into_raw(self) as *const();
+        let ptr = Arc::into_raw(self) as *const ();
 
         unsafe {
-            Waker::new_unchecked(RawWaker::new(ptr, waker_vtable!(Self)))
+            Waker::from_raw(RawWaker::new(ptr, waker_vtable!(Self)))
         }
     }
 }
@@ -36,7 +51,7 @@ pub trait ArcWake {
 // FIXME: panics on Arc::clone / refcount changes could wreak havoc on the
 // code here. We should guard against this by aborting.
 
-unsafe fn increase_refcount<T: ArcWake>(data: *const()) {
+unsafe fn increase_refcount<T: ArcWake>(data: *const ()) {
     // Retain Arc by creating a copy
     let arc: Arc<T> = Arc::from_raw(data as *const T);
     let arc_clone = arc.clone();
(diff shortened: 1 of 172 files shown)


yoshuawuyts:
Is there an ETA when we could land this? The changes have landed in the latest nightly, and currently a good amount of packages are broken until we can land this.

taiki-e:
(Landing it now does not mean that we can immediately release the next version.)

taiki-e:
Also, https://github.com/rust-lang/rust/pull/59733 is already approved, so It will break again in a few days.

davidbarsky:
Possibly a ignorant question, but would this fix land (pending time/availability from people, naturally) when/if https://github.com/rust-lang/rust/commit/df25d79a33b0c82b716699a75a41bfdc6089850a is part of a nightly release?

taiki-e:
@cramertj @Nemo157 Could you review this PR?

Considering FCP takes more than a week, I think that it is preferable to land this at this timing.

najamelan:
Updating to nightly today still gave me another error: 
```
error[E0507]: cannot move out of borrowed content
   --> /futures-rs/futures-core/src/task/__internal/atomic_waker.rs:244:17
    |
244 |                 waker.wake();
    |                 ^^^^^ cannot move out of borrowed content

```

Not entirely sure that this should be dealt with in this pull request, but I think it's because of the waker by ref last minute changes before stabelization. Cloning it here was my fast braindead fix, but I don't know whether that's the best way to solve this.

taiki-e:
@najamelan See https://github.com/rust-lang-nursery/futures-rs/pull/1514/commits/265ca12195c56e7e6b3968b58ce659dc1c3c7736.

najamelan:
@taiki-e Oh, sorry, you're awesome. I had missed the extra commit... compiles fine. Thanks

futures-util/src/task/arc_wake.rs
```diff
@@ -17,26 +17,41 @@ pub trait ArcWake {
...
+    /// This function can be called from an arbitrary thread, including threads which
+    /// did not create the `ArcWake` based `Waker`.
+    ///
+    /// Executors generally maintain a queue of "ready" tasks; `wake_by_ref` should place
+    /// the associated task onto this queue.
+    ///
+    /// This function is similar to `wake`, but must not consume the provided data
+    /// pointer.
+    fn wake_by_ref(arc_self: &Arc<Self>) {
+        arc_self.clone().wake()
```
Nemo157: I'm not certain whether it makes sense to keep the separation of waking by-value and by-ref for `ArcWake`, it looks like every implementation of `ArcWake::wake` delegates to `ArcWake::wake_by_ref`. But maybe the thread pool implementation can be updated to take advantage of it.

I think it makes more sense to provide the `ArcWake::wake` implementation as delegating to `ArcWake::wake_by_ref`, that would allow dropping all the current delegating implementations, and means that implementors wouldn't accidentally cause an unnecessary clone by only implementing `ArcWake::wake` (they may still cause an unnecessary clone by only implementing `ArcWake::wake_by_ref` and not taking advantage of being able to reuse the `Arc` in `ArcWake::wake`, but that seems like a more obvious performance issue since the clone will be happening in their code instead of the provided method).

Nemo157:
(for others that want to start experimenting off this, here's a block you can copy-paste into your `Cargo.toml` to set all the overrides):

```toml
[patch.crates-io]
futures-preview = { git = "https://github.com/taiki-e/futures-rs", branch = "futures_api" }
futures-core-preview = { git = "https://github.com/taiki-e/futures-rs", branch = "futures_api" }
futures-util-preview = { git = "https://github.com/taiki-e/futures-rs", branch = "futures_api" }
futures-sink-preview = { git = "https://github.com/taiki-e/futures-rs", branch = "futures_api" }
futures-channel-preview = { git = "https://github.com/taiki-e/futures-rs", branch = "futures_api" }
futures-executor-preview = { git = "https://github.com/taiki-e/futures-rs", branch = "futures_api" }
futures-io-preview = { git = "https://github.com/taiki-e/futures-rs", branch = "futures_api" }
futures-test-preview = { git = "https://github.com/taiki-e/futures-rs", branch = "futures_api" }
```

futures-util/src/task/arc_wake.rs
taiki-e: > I think it makes more sense to provide the `ArcWake::wake` implementation as delegating to `ArcWake::wake_by_ref`, that would allow dropping all the current delegating implementations, and means that implementors wouldn't accidentally cause an unnecessary clone by only implementing `ArcWake::wake` (they may still cause an unnecessary clone by only implementing `ArcWake::wake_by_ref` and not taking advantage of being able to reuse the `Arc` in `ArcWake::wake`, but that seems like a more obvious performance issue since the clone will be happening in their code instead of the provided method).

Thanks for pointing that out, that makes a lot of sense.

Nemo157:
Thanks again :heart:, hopefully we can get a release out soon with these fixes in (maybe even one worth promoting to `beta`).

cramertj:
Thanks for working to get this landed! All the help is extremely appreciated. I'll get a new release out ASAP.

vorot93:
@cramertj Could you please take a look at #1524 at first? google/tarpc may miss the new release otherwise.

cramertj:
@vorot93 merged! :) (In the future, cc'ing me or adding me as a reviewer will get my attention quicker, though I try to scan the repo for PRs every once in a while)

PR #1515, opened 2019-04-07
Stabilize alloc feature

taiki-e:
Currently, this feature requires the `nightly` feature active to explicitly opt-in to unstable features, but this is no longer needed because rust-lang/rust#59675 has been merged.


cramertj:
Wooot! :)

PR #1523, opened 2019-04-15
Call Pin::set and Pin::as_mut as method

taiki-e:


cramertj:
Thanks for the cleanup! So glad we got these changed to take `&mut self` :)

PR #1524, opened 2019-04-15
Add get_ref to Compat01As03 and Compat01As03Sink

vorot93:
Allows immutable access to the wrapped 0.1 Future/Stream/Sink.

ref [google/tarpc#216 (comment)](https://github.com/google/tarpc/pull/216#issuecomment-483299729)


cramertj:
Thanks!

PR #1526, opened 2019-04-16
Remove #[allow(clippy::useless_let_if_seq)]

taiki-e:


Nemo157:
Whoo, there is even a clippy available on the latest nightly to actually check this too (I think there's a test covering it that would have been tripped).

Thanks

Issue #1152, opened 2018-07-31
either dependency is non-optional for futures-util


Nemo157:
While doing https://github.com/rust-lang-nursery/futures-rs/pull/1151 I noticed that `futures-util` has a non-optional dependency on `either`. It would be good to make this optional and enable the `futures-core` and `futures-sink` features when enabled, but that's impossible to do without using a different feature name like `use_either` because of Cargo's interaction between optional dependencies and features. Once https://github.com/rust-lang/cargo/issues/5753 is fixed it may be possible to use a renamed dependency in `futures-util` to get around this, but that seems pretty hacky. There has been some work on this in Cargo: https://github.com/rust-lang/cargo/issues/1286 and https://github.com/rust-lang/cargo/issues/5565, hopefully there'll be a nicer solution soon.

PR #1527, opened 2019-04-16
Use our own either type

taiki-e:
Closes https://github.com/rust-lang-nursery/futures-rs/issues/1152


futures-util/src/future/either.rs
```diff
@@ -0,0 +1,117 @@
...
+/// Combines two different futures, streams, or sinks having the same associated types into a single
+/// type.
+#[derive(Debug, Clone)]
+pub enum Either<A, B> {
+    /// First branch of the type
+    Left(A),
+    /// Second branch of the type
+    Right(B),
+}
+
```
cramertj: Would you mind also adding an `into_inner` for `Either<T, T>`?

futures-util/src/future/mod.rs
```diff
@@ -186,60 +187,19 @@ pub trait FutureExt: Future {
     /// wrapped version of them.
     ///
     /// Also note that if both this and the second future have the same
-    /// success/error type you can use the `Either::split` method to
+    /// output type you can use the `Either::factor_first` method to
     /// conveniently extract out the value at the end.
-    ///
-    /// # Examples
```
cramertj: Why remove these doc examples?

cramertj:
works for me, modulo the removal of doc examples and the introduction of `into_inner`

futures-util/src/future/mod.rs
```diff
@@ -191,40 +192,35 @@ pub trait FutureExt: Future {
...
+    ///             Either::Left((x, b)) => b.map(move |y| (x, y)).left_future(),
+    ///             Either::Right((y, a)) => a.map(move |x| (x, y)).right_future(),
     ///         }
-    ///     }))
+    ///     })
     /// }
     /// ```
-    fn select<B>(self, other: B) -> Select<Self, B::Future>
-        where B: IntoFuture, Self: Sized
+    fn select<B>(self, other: B) -> Select<Self, B>
```
cramertj: We just changed the `.join` methods into free functions-- seems like that would make sense for `select` as well?

taiki-e: Thanks for pointing that out, it makes sense. And I remembered that I forgot to write PR for changing `StreamExt::select` to a function 😅 .

cramertj:
Nice! thanks for this :)

PR #1528, opened 2019-04-16
Use #![warn(clippy::all)]

taiki-e:
All lints included in `clippy::all` are enabled by default when doing `cargo clippy`.
But we can only check this by actually running `cargo clippy`.

In some editors, using `#![warn(clippy::all)]` makes it possible to receive warnings as well as lints provided by the language (at least with the vscode I use).

We will be happy because we do not have to check each time in local.

### DIFF
--- a/futures-channel/src/lib.rs
+++ b/futures-channel/src/lib.rs
@@ -6,6 +6,7 @@
 #![cfg_attr(not(feature = "std"), no_std)]
 
 #![warn(missing_docs, missing_debug_implementations, rust_2018_idioms)]
+#![warn(clippy::all)]
 
 #![doc(html_root_url = "https://rust-lang-nursery.github.io/futures-api-docs/0.3.0-alpha.15/futures_channel")]
 
--- a/futures-core/src/lib.rs
+++ b/futures-core/src/lib.rs
@@ -5,6 +5,7 @@
 #![cfg_attr(not(feature = "std"), no_std)]
 
 #![warn(missing_docs, missing_debug_implementations, rust_2018_idioms)]
+#![warn(clippy::all)]
 
 #![doc(html_root_url = "https://rust-lang-nursery.github.io/futures-api-docs/0.3.0-alpha.15/futures_core")]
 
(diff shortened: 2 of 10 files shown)


futures-channel/src/lib.rs
```diff
@@ -8,6 +8,7 @@
 #![cfg_attr(not(feature = "std"), no_std)]
 
 #![warn(missing_docs, missing_debug_implementations, rust_2018_idioms)]
+#![warn(clippy::all)]
```
taiki-e: This is separated from other lints because the providers are different.

Nemo157:
How does this work with vscode? I would expect it to run either `cargo check` or `cargo clippy` against the codebase, the former will ignore this attribute since it's a tool-scoped lint, the latter would have included clippy lints already. Maybe it's using RLS and RLS has some kind of "this project uses clippy" detection based on the lint attribute and pulls in clippy automatically?

Nemo157:
Ah, yes, looks like RLS defaults to detecting clippy lint config unless you override the `clippy_preference` setting.

I dislike the fact that `clippy::all` actually means the default warning set, but otherwise I think this makes sense to add (until RLS allows a workspaced scoped configuration for this at least, having to add it to every single crate is annoying).

taiki-e:
> How does this work with vscode? I would expect it to run either `cargo check` or `cargo clippy` against the codebase, the former will ignore this attribute since it's a tool-scoped lint, the latter would have included clippy lints already. Maybe it's using RLS and RLS has some kind of "this project uses clippy" detection based on the lint attribute and pulls in clippy automatically?

Ah, as you say that's a feature of rls.

taiki-e:
> I dislike the fact that `clippy::all` actually means the default warning set,

Certainly "all" is a misleading name. Maybe `clippy::default` is a better name?

Nemo157:
Yeah `clippy::default` sounds better, but that's unlikely to change soon. I think merging this as-is to get it working now for RLS users is good (unfortunately I'm not one, maybe I should get round to looking at setting it up). I'd like @cramertj's feedback as well though.

cramertj:
I don't have a strong opinion-- I'm not an RLS user either. +1 that `default` would be much clearer than `all`.

taiki-e:
Hmm, `default` seems to have already been discussed in https://github.com/rust-lang/rust-clippy/issues/2964 and rejected.

PR #1530, opened 2019-04-17
Add StreamExt::enumerate

taiki-e:
In the future 0.1, [this method is provided by tokio](https://docs.rs/tokio/0.1.18/tokio/util/trait.StreamExt.html#method.enumerate).

### DIFF
--- a/futures-util/src/stream/enumerate.rs
+++ b/futures-util/src/stream/enumerate.rs
@@ -1,0 +1,61 @@
+use core::pin::Pin;
+use futures_core::stream::{FusedStream, Stream};
+use futures_core::task::{Context, Poll};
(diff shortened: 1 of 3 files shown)


cramertj:
Nice! Seems handy :)

ds84182:
Just wondering, wouldn't it be better to use a u64 than a usize here? While unlikely, you can definitely process more than 4 billion events on a 32-bit platform

cramertj:
@ds84182 `std::iter::Enumerate` [uses `usize`](https://doc.rust-lang.org/src/core/iter/adapters/mod.rs.html#911-914), so I'd be inclined to be consistent, although I don't really have a strong opinion.

PR #1533, opened 2019-04-18
Use std::io::IoSlice[Mut] on AsyncRead/AsyncWrite take2

### LINKED ISSUE #1455: Use std IoVec types in AsyncRead and AsyncWrite

taiki-e:
This replaces #1476.

~~`iovec` feature already in the FCP, so I think it's good to wait until it's stable.~~
Stabilized in https://github.com/rust-lang/rust/pull/60334.

Also, this renames `poll_vectored_*` to `poll_*_vectored` to make it in the same order as std.

Closes #1455

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,7 +25,7 @@ matrix:
 
     # When updating this, the reminder to update the minimum required version in README.md.
     - name: cargo test (minimum required version)
-      rust: nightly-2019-04-25
+      rust: nightly-2019-04-30
 
     - name: cargo clippy
       rust: nightly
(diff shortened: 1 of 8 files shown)


taiki-e:
rust-lang/rust#60334 merged and nightly-2019-04-30 released! 🎉

cramertj:
woot! so excited about this :)

PR #1534, opened 2019-04-18
Remove #[allow(clippy::needless_lifetimes)]

taiki-e:
https://github.com/rust-lang/rust/issues/52675 has not been fixed, but the current Clippy does not seem to warn we of this.


taiki-e:
Rebased.

PR #1545, opened 2019-04-20
impl traits for more pinned pointers and ?Sized types

taiki-e:
Some traits were not implemented to `Pin<Box<&mut T>>`, and some impls did not allow `?Sized` types.

### DIFF
--- a/futures-core/src/future/mod.rs
+++ b/futures-core/src/future/mod.rs
@@ -1,5 +1,6 @@
 //! Futures.
 
+use core::ops::DerefMut;
 use core::pin::Pin;
 use core::task::{Context, Poll};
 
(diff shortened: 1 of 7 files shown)


futures-core/src/future/mod.rs
```diff
@@ -21,40 +22,19 @@ pub trait FusedFuture {
...
-            <F as FusedFuture>::is_terminated(&**self)
-        }
-    }
-
-    #[cfg(feature = "std")]
-    impl<F: FusedFuture> FusedFuture for std::panic::AssertUnwindSafe<F> {
-        fn is_terminated(&self) -> bool {
-            <F as FusedFuture>::is_terminated(&**self)
-        }
+impl<P> FusedFuture for Pin<P>
```
cramertj: oh, neat! this is much better :)

cramertj:
This is awesome, thanks so much!

PR #1546, opened 2019-04-24
Update to stabilized futures_api

taiki-e:
https://github.com/rust-lang/rust/pull/59739 was merged! 🎉

The changes will be included in the new nightly to be released tomorrow.

### DIFF
--- a/futures-test/src/lib.rs
+++ b/futures-test/src/lib.rs
@@ -1,10 +1,6 @@
 //! Utilities to make testing [`Future`s](futures_core::Future) easier
 
-#![feature(
-    async_await,
-    await_macro,
-    futures_api,
-)]
+#![feature(async_await, await_macro)]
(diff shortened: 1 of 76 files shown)


futures-test/src/lib.rs
```diff
@@ -1,10 +1,6 @@
 //! Utilities to make testing [`Future`s](futures_core::Future) easier
 
-#![feature(
-    async_await,
-    await_macro,
-    futures_api,
-)]
+#![feature(async_await, await_macro, futures_api_const_fn_ptr)]
```
cramertj: futures_api_const_fn_ptr

^this shouldn't be necessary. What error did you hit without this?

taiki-e: This patch was created when rust-lang/rust#59739 was opened, so I may have missed that is unnecessary.

taiki-e:
nightly-2019-04-25 released!

Issue #1547, opened 2019-04-24
`SinkExt::send` and `send_all`, and `StreamExt::next` should not need `Self: Unpin`


Ekleog:
There is, as far as I understand, no theoretical reason for `send` or `send_all` to require `Self: Unpin`.

The only reason being that the current implementation requires it, changing the implementation to not require `Self: Unpin` would be great.

taiki-e:
If you want to remove `Self: Unpin` from `send` or `send_all`, it needs to change that method to consume `self`. IMO, I do not think it is convenient for those methods to consume `self` (If it does not want to consume, it need to take a mutable reference, that requires `Self: Unpin`).

Nemo157:
Yeah, this is sort of the same situation as `StreamExt::next` that you will want to manually pin an `Unpin` stream so that you have the choice between the more-powerful-less-performant heap pinning and the less-powerful-more-performant stack pinning. These methods should probably have a paragraph like that included in `next`'s documentation:

> Note that because next doesn't take ownership over the stream, the Stream type must be Unpin. If you want to use next with a !Unpin stream, you'll first have to pin the stream. This can be done by boxing the stream using Box::pin or pinning it to the stack using the pin_mut! macro from the pin_utils crate.

Although, since this is a relatively common requirement it could be good to have a more expanded section in some central docs about pinning that all these methods link to?

Ekleog:
Couldn't these functions just take `Pin<&mut Self>`, and not consume ownership of the stream/sink?

Nemo157:
That's more annoying to work with than taking `&mut Self`. With the current signature you can call it with an `Unpin` sink without pinning it, or with a `?Unpin` sink by pinning it and calling the method on the pinned reference (which causes a double indirection, but this should be easily eliminated by the compiler). If it took `Pin<&mut Self>` you would still need to pin a `?Unpin` sink to call it (and then use `sink.as_mut().send(foo)` every time to re-borrow the pin), but you would also need to pin an `Unpin` sink to call it. You would reduce the compilers work by eliminating the double indirection, but that seems very easy to optimize.

Ekleog:
Pinning an `Unpin` sink is almost free, though: it's a safe `Pin::new(&mut sink)` away.

OTOH, it's basically impossible to use these functions from a `!Unpin` sink if one does not want to be adding meaningless indirections all around.

Nemo157:
It still adds code-readability overhead

```rust
[...]

PR #1548, opened 2019-04-24
Add async_stream to create streams via generators

taiki-e:
This adds a proc-macro crate for experimenting with the async stream API that we may introduce as a feature of language in the future.

~~The implementation is basically similar to `futures-await`'s `#[async] for` and `#[async_stream]`. (re-added some of the files deleted by https://github.com/rust-lang-nursery/futures-rs/pull/1050 and updated it.)~~
EDIT: After introducing `.await` syntax, it was not possible to keep the same implementation in stream scope (`#[async_stream]`) and future scope (`async fn`), so it was necessary to adjust it.

cc https://github.com/alexcrichton/futures-await/issues/111

[Rendered](https://github.com/taiki-e/futures-rs/tree/async-stream/futures-async-macro)


futures-util/src/async_stream/mod.rs
```diff
@@ -0,0 +1,83 @@
...
+struct GenStream<U, T: Generator<Yield = Poll<U>, Return = ()>> {
+    gen: T,
+    // If resuming after Complete, std generators panic. This is natural when using the generators,
+    // but in the streams, we may want to call `poll_next` after `poll_next` returns `None`
+    // because it is possible to call `next` after `next` returns `None` in many iterators.
+    done: bool,
+    _phantom: PhantomData<U>,
+    // We rely on the fact that async/await streams are immovable in order to create
+    // self-referential borrows in the underlying generator.
+    _pin: PhantomPinned,
```
taiki-e: This is [the same as async fn](https://github.com/rust-lang/rust/blob/bfb443eb1de484fde141fa9090a9f4291cbe60a5/src/libstd/future.rs#L29-L31).

Nemo157: You shouldn't need this, `T: ?Unpin` so `Self: ?Unpin` via the structural rules (although because of `PhantomData<U>` you might be getting the `Unpin`ness of `U` as well), if the generator transform ever allows non-self-referential `static`-marked generators to implement `Unpin` automatically then this could potentially generate `impl Stream + Unpin`.

futures-util/src/async_stream/await_item.rs
```diff
@@ -0,0 +1,46 @@
...
+///             return first + second;
+///         }
+///     }
+///     0
+/// }
+///
+/// assert_eq!(10, block_on(eventually_ten()));
+/// ```
+#[macro_export]
+macro_rules! await_item {
```
Nemo157: Is this necessary now that we have `await!(stream.next())`?

futures-async-macro/src/lib.rs
```diff
@@ -0,0 +1,404 @@
...
+                };
+
+                #body
+            }
+        }}
+    }
+
+    /* TODO: If this is enabled, it can be used for `yield` instead of `stream_yield!`.
+             Raw `yield` is simpler, but it may be preferable to distinguish it from `yield` called
+             in other scopes.
```
Nemo157: The main downside of using `yield` expressions instead of a `stream_yield!` macro is that the proc-macro can't see through any other macros called in the passed `TokenStream`, so it becomes impossible for users to define their own macros that expand to `yield` expressions intended to be used as part of an `async_stream`.

That said, when I implemented `embrio_async::async_stream_block` I went for just matching `yield` expressions, because it makes it much cleaner and it's unlikely that users will really need to use macros expanding to `yield`.

Nemo157: Actually, it looks like `stream_yield!` is just a token being matched by the `async_stream` macro, not a `macro_rules!`-macro? Then you don't get the upside of allowing users to define macros that expand to `stream_yield!` and just the downside of being weird and harder to read, so I'd definitely suggest changing to just expand `yield` expressions.

Nemo157: Oh, it's both, so yeah, not sure which downside is the bigger one.

Nemo157:
One potential future addition to this would be [`#[async_sink]`](https://github.com/Nemo157/embrio-rs/blob/d58623803b8b2aa807e154c58a464fa2afb7177a/embrio-async/tests/no-prelude-leak.rs#L55-L59), it seems _much_ more niche than `#[async_stream]` though.

futures-util/src/async_stream/await_item.rs
taiki-e: I was concerned about `Self: Unpin`, but https://github.com/rust-lang-nursery/futures-rs/issues/1547#issuecomment-486202839 would be a better solution. So I will remove this macro.

Nemo157: Within the context of a `static`-generator stack-pinning works just like normal. Currently all `static` generators are `!Unpin`, but if they could be automatically made `Unpin` if they didn't have self-references, then the self-reference used in `Next` would make sure they are correctly `Unpin` or not.

futures-util/src/async_stream/mod.rs
Nemo157: Damn, that's my bad, I should have removed the explicit `!Unpin` implementation as part of https://github.com/rust-lang/rust/pull/55704 (or transformed it into an explicit conditional implementation).

Ekleog:
FWIW, here is something that would likely benefit from `#[async_sink]`, which I have rage-quit trying to write as a manual `Sink` wrapper for now: https://github.com/Ekleog/yuubind/blob/master/smtp-message/src/data/sink.rs#L18 (it's almost `with_flat_map` except it's also adding some data after the end of the incoming stream)

Nemo157:
(somewhat OT: @Ekleog it took a few upgrades to my `async_sink_block` implementation to workaround type inference issues, and I'm not sure how to deal with errors yet, but I got that example compiling with it: https://gist.github.com/Nemo157/0344ba1aa1d1d89ce2f5ae26ea7b1d46, so maybe it is worth considering adding something like it here in the future)

futures-async-macro/src/lib.rs
taiki-e: I changed it to use the `yield` expression. Because we can analyze what is in the scope (but it takes some effort) of the [attr](https://github.com/taiki-e/auto_enums/blob/3c88acc88ecb1debf8f02976fa1fcf2ef3c592a4/core/src/auto_enum/attrs.rs#L9)/[macro](https://github.com/taiki-e/futures-rs/blob/aca5bc00cfd26b0d5a786eeeee80077a10db54ab/futures-async-macro/src/lib.rs#L297), so I think it is better to use something that is easy for the user to use.

yoshuawuyts:
I would love to see this included; this would allow us to make our [tcp-echo](https://github.com/rustasync/runtime/blob/master/examples/tcp-echo.rs) and [tcp-proxy](https://github.com/rustasync/runtime/blob/master/examples/tcp-proxy.rs) examples in Runtime so much nicer!

futures-async-macro/README.md
```diff
@@ -0,0 +1,51 @@
...
+
+This is a reimplement of [futures-await]'s `#[async_stream]` for futures 0.3 and is an experimental implementation of [the idea listed as the next step of async/await](https://github.com/rust-lang/rfcs/blob/master/text/2394-async_await.md#generators-and-streams).
+
+```rust
+use futures::prelude::*;
+use futures::async_stream;
+
+// Returns a stream of i32
+#[async_stream]
+fn foo(stream: impl Stream<Item = String>) -> i32 {
```
taiki-e: One of the concerns with current implementations is that there is an Item type at the "return" position. In the `futures-await` implementation this was specified in the form of an attribute argument `#[async_stream(Item = i32)]`.

yoshuawuyts: @taiki-e that approach seems reasonable to me; are there any reasons why we can't / shouldn't follow `future-await`'s approach?

taiki-e: @yoshuawuyts There is no particular reason. Also, it is easy to change the implementation to that.

(Because the stream has no final return value, the return position is always empty. The current implementation is just trying to reuse this blank.)

JuanPotato:
I really appreciate this feature and would be happy to help with anything. I haven't been working with futures 0.3 for too long but am very willing to help.

futures-async-macro/README.md
kgv: In [futures-await](https://github.com/alexcrichton/futures-await) we can use `?` operator inside stream. If I understand correctly, with new syntax it's impossible. Is it true?

```rust
/// Fetch all provided urls one at a time
#[async_stream(item = String)]
fn fetch_all(client: hyper::Client, urls: Vec<&'static str>) -> io::Result<()> {
    for url in urls {
        let s = await!(fetch(client, url))?;
        stream_yield!(s);
    }
    Ok(())
}
```

Nemo157: @kgv yes, because of streams no longer being required to have an error and item type that no longer works.

You can probably do something mostly equivalent with a `try` block though:

```rust
/// Fetch all provided urls one at a time
#[async_stream(item = io::Result<String>)]
fn fetch_all(client: hyper::Client, urls: Vec<&'static str>) {
    let res = try {
        for url in urls {
            yield Ok(fetch(client, url).await?);
        }
    };
    if let Err(err) = res {
         yield Err(err);
    }
}
```

and it may be possible to add more sugar for this in the future (an `async_try_stream` maybe?).

futures-async-macro/src/lib.rs
```diff
@@ -0,0 +1,408 @@
...
+            Expr::Macro(expr)
+        }
+    }
+
+    /// Expands `expr.await` in `async_stream` scope.
+    ///
+    /// It needs to adjust the type yielded by the macro because generators used internally by
+    /// async fn yield `()` type, but generators used internally by `async_stream` yield
+    /// `Poll<U>` type.
+    fn expand_await(&mut self, expr: ExprField) -> Expr {
```
taiki-e: It is not possible to replace `.await` in macros such as `assert!(foo.await)` with this.

Full support for this can not be done without compiler support, but it can support for some kinds of macros (however, this is not necessarily preferable, as a macro may call an async block etc): https://github.com/taiki-e/futures-rs/commit/fb33504229db314fb97759c3170ee3ad73785c08

Related: https://github.com/Nemo157/embrio-rs/issues/14

futures-async-macro/README.md
taiki-e: @Nemo157
> and it may be possible to add more sugar for this in the future (an `async_try_stream` maybe?).

I will try it after this PR is merged. And I think it will be something like the following:

```rust
#[async_try_stream(Ok = String, Error = io::Error)]
fn fetch_all(client: hyper::Client, urls: Vec<&'static str>) {
    for url in urls {
         yield fetch(client, url).await?;
    }
}

// desugaring

fn fetch_all(client: hyper::Client, urls: Vec<&'static str>) -> impl Stream<Item = Result<String, io::Error>> {
    from_generator(static move || { // impl Generator<Yield = String, Return = Result<(), io::Error>>
        for url in urls {
             yield Poll::Ready(fetch(client, url).await?);
        }
        return Ok(()); // It is the same behavior as try_blocks that this is added automatically.
        loop { yield Poll::Pending }
    })
}

```

taiki-e: (Also, if we add `async_try_stream`, it may be preferable to write `Item` at the position of the attribute argument, not at the position of return, so I will revert it)

taiki-e:
List of features that may be added in the future as an extension of this feature (there is no plan to handle in this PR):
  * `async_try_stream` (https://github.com/rust-lang-nursery/futures-rs/pull/1548#discussion_r287558350)
  * `async_sink` (https://github.com/rust-lang-nursery/futures-rs/pull/1548#issuecomment-486205382)
  * Support `.await` in macro (https://github.com/rust-lang-nursery/futures-rs/pull/1548#discussion_r285341883)
  * Parallel version of `for_await` (https://github.com/rustasync/runtime/pull/25)

taiki-e:
Now, this PR should be ready for final review! 
cc @cramertj @Nemo157

vorot93:
@taiki-e Have you checked whether your procedural macros work with IDE autocompletion engines (chiefly, Intellij-Rust's one)? They did not in futures-await.

taiki-e:
@vorot93 
As far as I know, it doesn't play nice yet in both IntelliJ-Rust and RLS. 
(note: I checked only IntelliJ-Rust+CLion and RLS + VSCode. Also I may have missed optional features)

IntelliJ-Rust seems doesn't understand that `#[async_stream]`ed function returns `impl Stream`.
RLS seems understand that `#[async_stream]`ed function returns `impl Stream`, but auto-completion of the traits method doesn't work.

FWIW IntelliJ-Rust seems that it can avoid the problem by using type annotation.
```rust
#![feature(impl_trait_in_binding)]

#[async_stream(item = ())]
fn foo() {}

let s1 = foo(); // IntelliJ-Rust doesn't understand `s1` is `impl Stream<Item = ?>`
let s2: impl Stream<Item = _> = foo(); // IntelliJ-Rust understand `s2` is `impl Stream<Item = ?>` and auto-completion works well.
```

kpp:
@taiki-e I can test your branch in my [pet project](https://github.com/kpp/futures-async-combinators) to reimplement some of stream combinators like map, take, filter. Give me some time.

kpp:
@taiki-e I can use `#[for_await]` only inside `async_stream_block!` or inside `#[async_stream(item = ...)]`. However in the docs it is told that I can use in in `async fn`. Is there a way to implement stream::collect with for_await? Take a look at:

```
pub async fn collect<St, C>(stream: St) -> C
    where St: Stream,
          C: Default + Extend<St::Item>
{
    pin_mut!(stream);
    let mut collection = C::default();
    while let Some(item) = stream.next().await {
        collection.extend(Some(item));
    }
    collection
}
```

taiki-e:
@kpp `#[for_await]` can be used in async fn and async block. like the following(this is a test included in this pr):
https://github.com/rust-lang-nursery/futures-rs/blob/14003d1b1f149e96caf6d0979548380f4ba8c2db/futures/tests/async_stream/smoke.rs#L9-L16

So, perhaps `collect` can be implemented as follows.
```rust
#![feature(async_await, proc_macro_hygiene, stmt_expr_attributes)]
pub async fn collect<St, C>(stream: St) -> C
    where St: Stream,
          C: Default + Extend<St::Item>
{
    let mut collection = C::default();
    #[for_await]
    for item in stream {
        collection.extend(Some(item));
    }
    collection
}
```

kpp:
@taiki-e I got this error:

```
error[E0658]: The attribute `for_await` is currently unknown to the compiler and may have meaning added to it in the future
  --> src/stream.rs:23:7
   |
23 |     #[for_await]
   |       ^^^^^^^^^
```

So I have to add `#![feature(custom_attribute)]`. And still I got this error:

```
error[E0277]: `St` is not an iterator
  --> src/stream.rs:24:17
   |
24 |     for item in stream {
   |                 ^^^^^^ `St` is not an iterator
   |
   = help: the trait `std::iter::Iterator` is not implemented for `St`
   = help: consider adding a `where St: std::iter::Iterator` bound
   = note: required by `std::iter::IntoIterator::into_iter`

For more information about this error, try `rustc --explain E0277`.
```

taiki-e:
@kpp Oops, I'm sorry. I forgot to write import: `use futures::for_await`.

kpp:
Yep. It worked. Thanks!

taiki-e:
@cramertj Any thoughts about this pr?

taiki-e:
Published as [futures-async-stream](https://crates.io/crates/futures-async-stream/0.1.0-alpha.1) -- Thanks everybody for your patience. It may not be what was expected, but I believe it would be better to polish it as a separate project and then decide if it should eventually merge.

PR #1550, opened 2019-04-24
Change FutureObj to use a `*mut dyn Future` fat pointer

Nemo157:
r? @cramertj

### DIFF
--- a/futures-core/src/future/future_obj.rs
+++ b/futures-core/src/future/future_obj.rs
@@ -1,4 +1,5 @@
 use core::{
+    mem,
     fmt,
     future::Future,
     marker::PhantomData,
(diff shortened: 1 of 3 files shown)


yoshuawuyts:
@Nemo157 looks like this has a merge conflict.

Nemo157:
Conflict is probably just around the `feature` changes, I'll fix it soon. But this is also blocked on a new nightly being released with https://github.com/rust-lang/rust/pull/60165 included.

cramertj:
@Nemo157 `Pin` is `repr(transparent)`, so it'd be sound to add those conversions in an `unsafe` manner to futures-rs. I'd personally prefer to do that now rather than to take a dependency on a new unstable API.

Nemo157:
👍 sure, I'll push an update switching back to doing the conversion manually.

Issue #1519, opened 2019-04-12
AsyncBufReader / AsyncBufWriter


yoshuawuyts:
Similar to https://github.com/rust-lang-nursery/futures-rs/issues/1512, we've been missing an async equivalent of std's [io::BufReader](https://doc.rust-lang.org/nightly/std/io/struct.BufReader.html) and [io::BufWriter](https://doc.rust-lang.org/nightly/std/io/struct.BufWriter.html). This would be useful to reduce the amount of system calls, and speed up throughput.

Nemo157:
I've experimented with this in embrio with [`trait AsyncBufRead`](https://github.com/Nemo157/embrio-rs/blob/master/embrio-core/src/io/buf_read.rs), an [async `struct BufReader`](https://github.com/Nemo157/embrio-rs/blob/master/embrio-util/src/io/buf_reader.rs) and an implementation of [`async fn read_until`](https://github.com/Nemo157/embrio-rs/blob/master/embrio-util/src/io/read_until.rs) (the latter can definitely be simplified by providing `async fn fill_buf()` on top of `fn poll_fill_buf()` and actually using `async fn` syntax). These have definitely worked well, and it should be straightforward to port `std`-like variants into `futures` (these were a little different than the ones provided in `std` since embrio is `alloc`-less).

taiki-e:
[`std::io::BufWriter` implements `Drop`](https://github.com/rust-lang/rust/blob/3d720d7e617442aaf65df0910e8c9f0fdb0cf31c/src/libstd/io/buffered.rs#L653-L660), and it is convenient. Do you think that async `BufWriter` can also implement it? ([functions that call `flush_buf` (flushes the internal buffer) internally need to receive a context](https://github.com/taiki-e/futures-rs/commit/bcd61df70843e57da1145d91652238c8f339a91f#diff-5c81d940fce4005fbd5fcd27e1c6554eR272). `into_inner` also has the same problem.)

Nemo157:
IMO anyone using an async writer should be calling `close` on it to ensure it's fully flushed and closed, dropping a writer without closing it is allowed to just abandon any outstanding writes (which is something that probably needs documenting on `AsyncWrite`). Having `std::io::Write` implementations set the precedent that dropping will flush outstanding writes and swallow errors is quite unfortunate though.

PR #1552, opened 2019-04-25
Add AsyncBufRead trait

taiki-e:
This adds `AsyncBufRead` trait and `AsyncBufRead::read_until` method.

(I'm writing async `BufWriter` now, but ~~I want to add some methods and trait implementations that require `AsyncSeek`, so~~ I would like to add traits ahead.)

cc https://github.com/rust-lang-nursery/futures-rs/issues/1519


futures-io/src/lib.rs
```diff
@@ -241,6 +241,63 @@ mod if_std {
...
+        /// its buffer, returned from [`poll_fill_buf`], has been consumed and should
+        /// no longer be returned. As such, this function may do odd things if
+        /// [`poll_fill_buf`] isn't called before calling it.
+        ///
+        /// The `amt` must be `<=` the number of bytes in the buffer returned by
+        /// [`poll_fill_buf`].
+        ///
+        /// [`poll_read`]: AsyncRead::poll_read
+        /// [`poll_fill_buf`]: AsyncBufRead::poll_fill_buf
+        fn consume(&mut self, amt: usize);
```
Nemo157: Even though it doesn't need it, I think this function should take `Pin<&mut Self>`, otherwise it doesn't seem possible to pin a `AsyncBufRead + !Unpin`, call `poll_fill_buf`, then _unpin_ it to call `consume`.

taiki-e: That makes sense.

futures-util/src/io/mod.rs
```diff
@@ -324,3 +322,61 @@ pub trait AsyncWriteExt: AsyncWrite {
 }
 
 impl<W: AsyncWrite + ?Sized> AsyncWriteExt for W {}
+
+/// An extension trait which adds utility methods to `AsyncBufRead` types.
+pub trait AsyncBufReadExt: AsyncBufRead {
```
cramertj: I imagine we'll also want `read_line`, `lines` etc. but those can wait for a followup

futures-util/src/io/mod.rs
```diff
@@ -26,6 +18,9 @@ pub use self::copy_into::CopyInto;
 mod flush;
 pub use self::flush::Flush;
 
+// mod lines;
```
cramertj: Mind adding a `TODO` or something here to indicate that this is wanted, but not yet implemented?

futures-io/src/lib.rs
```diff
@@ -429,6 +490,70 @@ mod if_std {
...
+        delegate_async_buf_read_to_stdio!();
+    }
+
+    impl AsyncBufRead for StdIo::Empty {
+        delegate_async_buf_read_to_stdio!();
+    }
+
+    impl<T: AsRef<[u8]> + Unpin> AsyncBufRead for StdIo::Cursor<T> {
+        delegate_async_buf_read_to_stdio!();
+    }
```
cramertj: Could you also add an implementation for `AllowStdIo`?

futures-util/src/io/allow_std.rs
```diff
@@ -130,3 +128,25 @@ impl<T> AsyncSeek for AllowStdIo<T> where T: io::Seek {
...
+        self.0.consume(amt)
+    }
+}
+
+impl<T> AsyncBufRead for AllowStdIo<T> where T: io::BufRead {
+    fn poll_fill_buf<'a>(self: Pin<&'a mut Self>, _: &mut Context<'_>)
+        -> Poll<io::Result<&'a [u8]>>
+    {
+        // FIXME: We should use `try_with_interrupt` here.
+        Poll::Ready(io::BufRead::fill_buf(&mut self.get_mut().0))
```
taiki-e: I wanted to use `try_with_interrupt!` macro here but there was a lifetime error.

```text
error[E0499]: cannot borrow `self` as mutable more than once at a time
   --> futures-util\src\io\allow_std.rs:147:44
    |
142 |     fn poll_fill_buf<'a>(mut self: Pin<&'a mut Self>, _: &mut Context<'_>)
    |                      -- lifetime `'a` defined here
...
147 |         Poll::Ready(Ok(try_with_interrupt!(self.0.fill_buf())))
    |         -----------------------------------^^^^----------------
    |         |                                  |
    |         |                                  mutable borrow starts here in previous iteration of loop
    |         returning this value requires that `self` is borrowed for `'a`
```

cramertj: oh holy cow [that's incredibly annoying](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=40a691ad15204739454d942a540bac50)

You can do it with `unsafe`: https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=f3c6fa0b2a118e31420de40e90f64d90

taiki-e: Nice, thanks!

cramertj: @nikomatsakis says this is fixed by polonious XD

taiki-e: Great, I'll look forward to that!

futures-util/src/io/read_until.rs
```diff
@@ -0,0 +1,47 @@
...
+        Self { reader, byte, buf }
+    }
+}
+
+impl<R: AsyncBufRead + ?Sized + Unpin> Future for ReadUntil<'_, R> {
+    type Output = io::Result<usize>;
+
+    fn poll(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
+        let this = &mut *self;
+        let mut read = 0;
```
taiki-e: Oh, this is bad, sorry. When `Pending` is returned, it will forget the number of bytes reads so far.

taiki-e: https://github.com/rust-lang-nursery/futures-rs/pull/1566 will fix this.

PR #1555, opened 2019-04-25
Move non-stable Rust features behind Cargo features

Nemo157:
This gets the crates to a point where they should compile on stable 1.36 when it is released. The travis jobs have been updated to make the compiler act like a stable compiler where possible to catch any regressions.

### DIFF
--- a/futures-util/Cargo.toml
+++ b/futures-util/Cargo.toml
@@ -15,8 +15,9 @@ Common utilities and extension traits for the futures-rs library.
 name = "futures_util"
 
 [features]
-std = ["alloc", "futures-core-preview/std", "futures-io-preview/std", "futures-sink-preview/std", "futures-select-macro-preview/std", "rand", "rand_core", "slab", "memchr"]
+std = ["alloc", "futures-core-preview/std", "futures-io-preview/std", "futures-sink-preview/std", "slab", "memchr"]
 default = ["std"]
+async-await = ["std", "futures-select-macro-preview", "proc-macro-hack", "proc-macro-nested", "rand", "rand_core"]
 compat = ["std", "futures_01"]
 io-compat = ["compat", "tokio-io"]
 bench = []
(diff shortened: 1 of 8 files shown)


futures-util/Cargo.toml
```diff
@@ -15,8 +15,9 @@ Common utilities and extension traits for the futures-rs library.
 name = "futures_util"
 
 [features]
-std = ["alloc", "futures-core-preview/std", "futures-io-preview/std", "futures-sink-preview/std", "futures-select-macro-preview/std", "rand", "rand_core", "slab"]
+std = ["alloc", "futures-core-preview/std", "futures-io-preview/std", "futures-sink-preview/std", "futures-select-macro-preview/std", "slab"]
 default = ["std"]
+async-await = ["rand", "rand_core"]
```
cramertj: You can also make `futures-select-macro-preview`, `proc-macro-hack`, and `proc-macro-nested` optional based on the `async-await` feature.

cramertj: This feature also depends on `std`, right?

Nemo157: `futures-select-macro-preview` has the `futures-select-macro-preview/std` feature and Cargo doesn't really support optionally activating features of optional dependencies, but since that's currently unused I can just remove it and make these optional.

Nemo157: Yes, it should (for now).

cramertj: Yeah, I don't think `futures-select-macro-preview` makes any sense without `std`-- I think the feature is just there to match other libs. Certainly `select!` doesn't work on no_std today because `await!` doesn't work on `no_std` today.

futures-util/src/lib.rs
```diff
@@ -2,7 +2,7 @@
 //! and the `AsyncRead` and `AsyncWrite` traits.
 
 #![cfg_attr(feature = "alloc", feature(box_into_pin))]
```
taiki-e: `box_into_pin` is also an unstable feature.

Nemo157: 😲 why is this not failing in travis then.

Oh, `allow-features` is only for language features. Is there any way to deny library features as well?

cramertj: No, I don't believe so-- seems like a reasonable thing to add to `allow-features` (cc @tmandry).

tmandry: Opened https://github.com/rust-lang/rust/pull/60289

cramertj:
Looks like there are some conflicts with master-- will merge when those are resolved. Thanks again for this-- it's super exciting! :)

Nemo157:
@cramertj rebased

sdroege:
Is there a plan to put a new alpha release on crates.io once this is merged?

cramertj:
@sdroege Yup, I'll get a new alpha release out today.

PR #1556, opened 2019-04-26
Add AsyncBufReadExt::read_line

taiki-e:
Also, `AsyncBufReadExt::lines` can be implemented using this.

### DIFF
--- a/futures-util/src/io/read_line.rs
+++ b/futures-util/src/io/read_line.rs
@@ -1,0 +1,48 @@
+use futures_core::future::Future;
+use futures_core::task::{Context, Poll};
+use futures_io::AsyncBufRead;
+use std::io;
+use std::mem;
+use std::pin::Pin;
+use std::str;
+use super::read_until::read_until_internal;
+
+/// Future for the [`read_line`](super::AsyncBufReadExt::read_line) method.
+#[derive(Debug)]
+pub struct ReadLine<'a, R: ?Sized + Unpin> {
+    reader: &'a mut R,
+    buf: &'a mut String,
+    bytes: Vec<u8>,
+    read: usize,
+}
+
+impl<R: ?Sized + Unpin> Unpin for ReadLine<'_, R> {}
+
+impl<'a, R: AsyncBufRead + ?Sized + Unpin> ReadLine<'a, R> {
+    pub(super) fn new(reader: &'a mut R, buf: &'a mut String) -> Self {
+        Self {
+            reader,
+            bytes: unsafe { mem::replace(buf.as_mut_vec(), Vec::new()) },
+            buf,
+            read: 0,
+        }
+    }
+}
+
+impl<R: AsyncBufRead + ?Sized + Unpin> Future for ReadLine<'_, R> {
+    type Output = io::Result<usize>;
+
+    fn poll(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
+        let Self { reader, buf, bytes, read } = &mut *self;
+        let ret = ready!(read_until_internal(Pin::new(reader), b'\n', bytes, read, cx));
+        if str::from_utf8(&bytes).is_err() {
+            Poll::Ready(ret.and_then(|_| {
+                Err(io::Error::new(io::ErrorKind::InvalidData,
+                            "stream did not contain valid UTF-8"))
+            }))
+        } else {
+            unsafe { mem::swap(buf.as_mut_vec(), bytes); }
+            Poll::Ready(ret)
+        }
+    }
+}
--- a/futures/tests/io_read_line.rs
+++ b/futures/tests/io_read_line.rs
@@ -1,0 +1,58 @@
+use futures::executor::block_on;
+use futures::future::Future;
+use futures::io::AsyncBufReadExt;
+use futures::task::Poll;
+use futures_test::io::AsyncReadTestExt;
+use futures_test::task::noop_context;
+use std::io::Cursor;
+use std::pin::Pin;
+
+#[test]
+fn read_line() {
+    let mut buf = Cursor::new(b"12");
+    let mut v = String::new();
+    assert_eq!(block_on(buf.read_line(&mut v)).unwrap(), 2);
+    assert_eq!(v, "12");
+
+    let mut buf = Cursor::new(b"12\n\n");
+    let mut v = String::new();
+    assert_eq!(block_on(buf.read_line(&mut v)).unwrap(), 3);
+    assert_eq!(v, "12\n");
+    v.clear();
+    assert_eq!(block_on(buf.read_line(&mut v)).unwrap(), 1);
+    assert_eq!(v, "\n");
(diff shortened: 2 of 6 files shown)


futures-util/src/io/read_line.rs
```diff
@@ -0,0 +1,51 @@
...
+        unsafe { self.buf.set_len(self.len); }
+    }
+}
+
+impl<R: AsyncBufRead + ?Sized + Unpin> Future for ReadLine<'_, R> {
+    type Output = io::Result<usize>;
+
+    fn poll(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
+        let Self { reader, buf } = &mut *self;
+        unsafe {
```
cramertj: Can you include a link to https://doc.rust-lang.org/src/std/io/mod.rs.html#310 / comment explaining what's going on this code? It looks correct to me, but it's definitely not simple :)

futures/tests/io_read_line.rs
```diff
@@ -0,0 +1,22 @@
+use futures::executor::block_on;
+use futures::io::AsyncBufReadExt;
+use std::io::Cursor;
+
+#[test]
+fn read_line() {
```
cramertj: Can you also include a test with a reader which isn't always `Ready` in order to make sure the readiness bits are being plumbed through correctly?

taiki-e: Thanks to you pointing it out, I was able to find https://github.com/rust-lang-nursery/futures-rs/pull/1566.

futures-util/src/io/read_line.rs
taiki-e: That makes sense.

taiki-e:
Blocked on https://github.com/rust-lang-nursery/futures-rs/pull/1566.

futures/tests/io_read_line.rs
taiki-e: In addition, I found some issues with the previous read_line implementation. (They have been fixed.)

However, I'd like to copy and rewrite it instead of calling read_until's implementation, to make it safer.

futures-util/src/io/read_line.rs
```diff
@@ -0,0 +1,55 @@
...
+impl<R: AsyncBufRead + ?Sized + Unpin> Future for ReadLine<'_, R> {
+    type Output = io::Result<usize>;
+
+    fn poll(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
+        let Self { reader, buf, read } = &mut *self;
+        unsafe {
+            // safety: https://doc.rust-lang.org/src/std/io/mod.rs.html#310
+            let mut g = Guard { len: buf.len(), buf: buf.as_mut_vec() };
+            let poll = read_until_internal(Pin::new(reader), b'\n', g.buf, read, cx);
+            if str::from_utf8(&g.buf[g.len..]).is_err() {
```
Nemo157: `read_until_internal` could have read the first bytes of a multi-byte character then returned `Pending` at this point. I don’t think you can really look at the content of the buffer until you get a `Ready`.

taiki-e: Hmm, I think you're right.

AFAIK the problem with the previous implementation was that `g.len` was updated when `Pending` was called. So perhaps it can resolve it by recording the first `buf.len()` and getting `g.len` from the field of `ReadLine` (i.e., `self.len`). And when `Ready` returned, it updates `self.len` and `g.len` only when it is a valid string, and if it is an invalid string, it just drops `g`. Also, when `Pending` is returned, update only `g.len`.

Nemo157: An alternative would be to add an extra `bytes: Vec<u8>` field to the struct, do `bytes: mem::replace(buf.as_mut_vec(), Vec::new())` in the constructor, then once you receive `Ready` validate the bytes and `mem::swap(buf.as_mut_vec(), bytes)`. That way you avoid the whole per-poll guard thing.

taiki-e: Oh, that's the safer way. I like that, thanks.

taiki-e:
I opened https://github.com/rust-lang/rust/issues/60498 about a strange error occurring in the test.

cramertj:
This looks good to me, but has conflicts with master-- will merge after rebase.

taiki-e:
Rebased.

cramertj:
Nice!

futures/tests/io_read_until.rs
```diff
@@ -35,53 +35,14 @@ fn run<F: Future + Unpin>(mut f: F) -> F::Output {
...
 #[test]
 fn maybe_pending() {
-    let mut buf = MaybePending::new(b"12");
+    let mut buf = b"12".interleave_pending();
     let mut v = Vec::new();
     assert_eq!(run(buf.read_until(b'3', &mut v)).unwrap(), 2);
     assert_eq!(v, b"12");
 
-    let mut buf = MaybePending::new(b"12333");
+    let mut buf = b"12333".interleave_pending();
```
Nemo157: Note that this won't add a pending in between reads that actually return data like the old implementation, it will be like `Pending`, `Ready(b"123")`, `Pending`, `Ready(b"")`.

If you want to test reading data multiple times my plan elsewhere was to use `TryStreamExt::into_async_read`, so something like `stream::iter(vec![b"12", b"33", b"3"]).map(Ok).into_async_read().interleave_pending()`.

taiki-e: Thanks for pointing that out. I updated this PR after checking locally that the new test can detect problems like https://github.com/rust-lang-nursery/futures-rs/pull/1566, but the code you suggested is even better. I will open a PR

taiki-e: Hmm... `thread 'maybe_pending' panicked at 'Attempted to consume from IntoAsyncRead without chunk'` at
https://github.com/rust-lang-nursery/futures-rs/blob/f4dad6654b65be08bb0b59f2a274ad18d96e436a/futures/tests/io_read_until.rs#L56

This is because `read_until` runs `consume` unconditionally even if `used` is zero.
https://github.com/rust-lang-nursery/futures-rs/blob/f4dad6654b65be08bb0b59f2a274ad18d96e436a/futures-util/src/io/read_until.rs#L33-L43
https://github.com/rust-lang-nursery/futures-rs/blob/f4dad6654b65be08bb0b59f2a274ad18d96e436a/futures-util/src/try_stream/into_async_read.rs#L149

This can be solved by consuming it only when `used` is non-zero.

However, the current implementation is the same behavior as std.
https://github.com/rust-lang/rust/blob/55c48b4e823ad3518bb8e6d3fd7da4f2fc0c5fc2/src/libstd/io/mod.rs#L1475-L1492

Nemo157: Sounds like `IntoAsyncRead` should special-case `consume(0)` to be a noop then.

taiki-e: Sounds good. Opened #1584.

PR #1557, opened 2019-04-26
Re-add try_future::select_ok

taiki-e:
This is basically the same as 0.1's `future::select_ok`.


cramertj:
Code looks good, but I think I personally prefer the previous name. The `try_` prefix usually suggests that the combinator returns early if any of the children return an error, but that's already the behavior of `select_all`. IMO `select_ok` is clearer about the intent,

taiki-e:
> The `try_` prefix usually suggests that the combinator returns early if any of the children return an error, but that's already the behavior of `select_all`.

Indeed!

PR #1558, opened 2019-04-26
Add TryStreamExt::try_filter

svartalf:


futures-util/src/try_stream/mod.rs
```diff
@@ -444,6 +447,46 @@ pub trait TryStreamExt: TryStream {
...
+    ///     future::ready(x % 2 == 0)
+    /// });
+    ///
+    /// assert_eq!(await!(evens.next()), Some(Ok(2)));
+    /// assert_eq!(await!(evens.next()), Some(Err("error")));
+    /// # })
+    /// ```
+    fn try_filter<Fut, F>(self, f: F) -> TryFilter<Self, Fut, F>
+        where Fut: Future<Output = bool>,
+              F: FnMut(&Self::Ok) -> Fut,
```
Nemo157: One downside of this is that the returned future can't keep hold of the reference, if it needs to look at it after doing some async operation it will need to clone it. But `try_skip_while` has the same issue, and I guess it's blocked on being able to add async closure bounds, so should be fine for now.

futures-util/src/try_stream/try_filter.rs
```diff
@@ -0,0 +1,118 @@
...
+        &mut self.stream
+    }
+
+    /// Acquires a pinned mutable reference to the underlying stream that this
+    /// combinator is pulling from.
+    ///
+    /// Note that care must be taken to avoid tampering with the state of the
+    /// stream which may otherwise confuse this combinator.
+    pub fn get_pin_mut<'a>(self: Pin<&'a mut Self>) -> Pin<&'a mut St> {
+        unsafe { Pin::map_unchecked_mut(self, Self::get_mut) }
```
Nemo157: This should just call `self.stream()` to avoid the `unsafe` block.

svartalf: This one was copied from the [TryFilterMap::get_pin_mut](https://github.com/rust-lang-nursery/futures-rs/blob/5a6b1b32329b3f4a0ad1df3df1576fb36fdc5361/futures-util/src/try_stream/try_filter_map.rs#L46-L53) as is.

Maybe @taiki-e can help us here?

taiki-e: It is preferable to replace the unsafe code with the safe code.
```diff
- unsafe { Pin::map_unchecked_mut(self, Self::get_mut) }
+ self.stream()
```
(The other combinators are processed by https://github.com/rust-lang-nursery/futures-rs/pull/1559.)

futures-util/src/try_stream/try_filter.rs
```diff
@@ -0,0 +1,118 @@
...
+          F: FnMut(&St::Ok) -> Fut,
+{
+    type Item = Result<St::Ok, St::Error>;
+
+    fn poll_next(
+        mut self: Pin<&mut Self>,
+        cx: &mut Context<'_>,
+    ) -> Poll<Option<Result<St::Ok, St::Error>>> {
+        loop {
+            if self.as_mut().pending_fut().as_pin_mut().is_none() {
```
taiki-e:
```suggestion
            if self.pending_fut.is_none() {
```
It is unnecessary to do pin projection here because `Pin` implements `Deref` unconditionally.

svartalf: I was using the same implementation as a [Stream::filter](https://github.com/rust-lang-nursery/futures-rs/blob/5a6b1b32329b3f4a0ad1df3df1576fb36fdc5361/futures-util/src/stream/filter.rs#L102).

This line can be reduced down to this:

```rust
if self.as_mut().pending_fut().is_none()
```
, but the change you suggested is failing with the following error:
```rust
error[E0382]: use of moved value: `self`
  --> futures-util/src/try_stream/try_filter.rs:87:16
   |
83 |         mut self: Pin<&mut Self>,
   |         -------- move occurs because `self` has type `std::pin::Pin<&mut try_stream::try_filter::TryFilter<St, Fut, F>>`, which does not implement the `Copy` trait
...
87 |             if self.pending_fut().is_none() {
   |                ^^^^ value moved here, in previous iteration of loop

```

svartalf:
I have found right now that [TryFilterMap](https://github.com/rust-lang-nursery/futures-rs/blob/5a6b1b32329b3f4a0ad1df3df1576fb36fdc5361/futures-util/src/try_stream/try_filter_map.rs) type lacks the `FusedStream` implementation at the moment, smth similar to the following code:

```rust
impl<St: FusedStream, F> FusedStream for TryFilterMap<St, F> {}
```

Since all other types returned by `TryStreamExt` methods are implementing it, I presume that it should be implemented both for `TryFilterMap` and `TryFilter` from this PR?

futures-util/src/try_stream/try_filter.rs
taiki-e: Seems to be calling a method. It needs to remove "`()`" like this.
```rust
if self.pending_fut.is_none() {
```

svartalf: My bad, fixed.

taiki-e:
> Since all other types returned by `TryStreamExt` methods are implementing it, I presume that it should be implemented both for `TryFilterMap` and `TryFilter` from this PR?

Yeah, I think they should have `FusedStream` implementations similar to `FilterMap` and `Filter` respectively.

svartalf:
> Yeah, I think they should have `FusedStream` implementations similar to `FilterMap` and `Filter` respectively.

Implementation for the `TryFilter` was added in this PR and I've opened the [additional one](https://github.com/rust-lang-nursery/futures-rs/pull/1561) for the `TryFilterMap`

Nemo157:
Thanks :)

PR #1563, opened 2019-04-27
Non-Send boxed()

chmln:
This PR adds non-`Send` versions of `BoxFuture` and `boxed()`, prompted by the `Send` requirement in #1494 (and thus the latest release) breaking some of our code. This allows avoiding unnecessary `Send` bounds in generic-heavy code.

### DIFF
--- a/futures-core/src/future/mod.rs
+++ b/futures-core/src/future/mod.rs
@@ -16,7 +16,7 @@ pub type BoxFuture<'a, T> = Pin<alloc::boxed::Box<dyn Future<Output = T> + Send 
 
 #[cfg(feature = "alloc")]
 /// `BoxFuture`, but without the `Send` requirement.
-pub type BoxFutureLocal<'a, T> = Pin<alloc::boxed::Box<dyn Future<Output = T> + 'a>>;
+pub type LocalBoxFuture<'a, T> = Pin<alloc::boxed::Box<dyn Future<Output = T> + 'a>>;
 
 /// A `Future` or `TryFuture` which tracks whether or not the underlying future
 /// should no longer be polled.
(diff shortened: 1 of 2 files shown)


overdrivenpotato:
Dealing with the same issue here.

@cramertj Is it possible to simply revert the `boxed` changes in #1494? The new alpha 15 release today broke several of my projects, requiring hefty propagation of `Send` throughout many modules. Specifically, is there a reason why [this change](https://github.com/rust-lang-nursery/futures-rs/pull/1494/files#diff-63548cac00f39ef099b69f6caa1bdcdbL494) was necessary?

Nemo157:
The CI failure is related to #1555 and https://github.com/rust-lang/rust/pull/60289, I'm looking into a fix for it.

futures-core/src/future/mod.rs
```diff
@@ -14,6 +14,10 @@ pub use self::future_obj::{FutureObj, LocalFutureObj, UnsafeFutureObj};
 /// statically type your result or need to add some indirection.
 pub type BoxFuture<'a, T> = Pin<alloc::boxed::Box<dyn Future<Output = T> + Send + 'a>>;
 
+#[cfg(feature = "alloc")]
+/// `BoxFuture`, but without the `Send` requirement.
+pub type BoxFutureLocal<'a, T> = Pin<alloc::boxed::Box<dyn Future<Output = T> + 'a>>;
```
Nemo157: The other `Local` types use it as a prefix on the type name (`LocalFutureObj`, `LocalSpawn`, `LocalPool`) but the functions do seem to be a suffix, so maybe this should be `fn boxed_local() -> LocalBoxFuture` for consistency? (I don't know if there's other convention for this pattern outside this library?)

chmln: @Nemo157 the type is now renamed to `LocalBoxFuture`.

PR #1565, opened 2019-04-27
Allow proc_macro_span feature

### LINKED ISSUE #1564: `proc_macro_span` feature used on "stable" builds
With the changes from #1555 we are now attempting to build with something equivalent to what will be the 1.36 compiler. Unfortunately https://github.com/rust-lang/rust/pull/60289 has revealed that `proc-macro2` is using an unstable feature when it's built on nightly, which has broken those CI builds.

I've opened https://github.com/alexcrichton/proc-macro2/issues/175 to see if we can get a way to disable that feature in `proc-macro2`, but if we can't do that soon I'll probably revert the Travis changes from #1555 until it happens.

Nemo157:
Workaround for https://github.com/alexcrichton/proc-macro2/issues/175

Related to #1564

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -46,7 +46,7 @@ matrix:
     - name: cargo +stable build --no-default-features
       rust: nightly
       env:
-        - RUSTFLAGS="-Z allow-features="
+        - RUSTFLAGS="-Z allow-features=proc_macro_span" # TODO: https://github.com/rust-lang-nursery/futures-rs/issues/1564
       script:
         - cargo run --manifest-path ci/remove-dev-dependencies/Cargo.toml */Cargo.toml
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -46,7 +46,7 @@ matrix:
     - name: cargo +stable build --no-default-features
       rust: nightly
       env:
-        - RUSTFLAGS="-Z allow-features="
+        - RUSTFLAGS="-Z allow-features=proc_macro_span"
```
taiki-e: It may be preferable to include a link to #1564 here.

Nemo157: Put a TODO with link, I'll reuse that issue to track removing this once `proc_macro2` allows.

PR #1566, opened 2019-04-27
Correct the behavior of read_until when Pending is returned

taiki-e:
The current implementation did not record information about the number of bytes read so far, so it returned an incorrect value when `Pending` returned after `Ready` returned.


taiki-e:
Rebased.

PR #1570, opened 2019-04-28
Remove redundant code from TryStreamExt::{and_then, or_else}

taiki-e:


futures-util/src/try_stream/and_then.rs
```diff
@@ -82,17 +82,9 @@ impl<St, Fut, F> Stream for AndThen<St, Fut, F>
             self.as_mut().future().set(Some(fut));
         }
 
-        assert!(self.future.is_some());
```
taiki-e: IMO, these are [the same as `then`](https://github.com/rust-lang-nursery/futures-rs/blob/3872328c419e9fdd8d074c23e738fa2eef5d1b4a/futures-util/src/stream/then.rs#L85-L96), so no assertion is required.

Issue #1398, opened 2019-01-06
Rename (Trait, TraitExt) pairs to (TraitCore, Trait)


HyeonuPark:
In current API, most traits of futures-rs are splitted into 2 variants - one for implementors, and one for consumers. Formers are named without suffix, and latters are named with `Ext` suffix.

Ironically those pair traits are mostly disjoint in usage pattern. For example, a library which expose a type that impls `Stream` doesn't need to know about `StreamExt` trait in their code at all. And if I want to write a binary with heavy business logic, I don't want to care about `poll_next()` method so will only use `StreamExt` trait.

But in practice, there's way more library consumers than library writers, so most users only need `StreamExt` not `Stream`. I can easily imagin SO questions like "How can I `map` or `filter` on `Stream` type?" and correct answer for it is to replace `use futures::stream::Stream` with `futures::stream::StreamExt`. Of course we can recommend to `use futures::prelude::*` but some of them will complain about it as we all know the use-all pattern is considered bad in general.

That's why I'm suggesting here to remove suffix from consumer-side traits. The `Core` suffix I referred in title needs more bikeshedding though.

This issue is more like a RFC but I couldn't find a better place for it. Please let me know if this is not a right place for it!

Matthias247:
The `Ext` things are not generally the for-consumer version. They are extension methods on the traits, and the convention for extension traits is to use `Ext` (https://github.com/rust-lang/rfcs/pull/445). You can consume/use the traits just fine without the extension methods - then you don't have the methods available. In many cases that is sufficient.

The `Stream/Sink/AsyncRead/AsyncWrite` traits you mention are somewhat special in that regard, since they don't yet 100% fit into the async/await world natively by not returning `Future`s. Therefore those extensions might be required by more users.

Nemo157:
Adding on to what @Matthias247 mentions, the "core" trait will still be heavily used in consumers. All generic APIs should be based on the core trait, and the extension is just brought into scope to provide extra methods on it. Because of that having the short name for the core trait is better since it will commonly be mentioned in multiple places per-file, while the extension trait will only ever be `use`d once at the top of each file.

cramertj:
Closing-- the current design is intentional.

PR #1573, opened 2019-04-30
Add async BufReader

taiki-e:
cc https://github.com/rust-lang-nursery/futures-rs/issues/1519

### DIFF
--- a/futures-util/src/io/buf_reader.rs
+++ b/futures-util/src/io/buf_reader.rs
@@ -1,0 +1,258 @@
+use futures_core::task::{Context, Poll};
+use futures_io::{AsyncBufRead, AsyncRead, AsyncSeek, Initializer, IoSliceMut, SeekFrom};
+use pin_utils::{unsafe_pinned, unsafe_unpinned};
+use std::io::{self, Read};
+use std::pin::Pin;
+use std::{cmp, fmt};
+use super::DEFAULT_BUF_SIZE;
+
+/// The `BufReader` struct adds buffering to any reader.
+///
+/// It can be excessively inefficient to work directly with a [`AsyncRead`]
+/// instance. A `BufReader` performs large, infrequent reads on the underlying
+/// [`AsyncRead`] and maintains an in-memory buffer of the results.
+///
+/// `BufReader` can improve the speed of programs that make *small* and
+/// *repeated* read calls to the same file or network socket. It does not
+/// help when reading very large amounts at once, or reading just one or a few
+/// times. It also provides no advantage when reading from a source that is
+/// already in memory, like a `Vec<u8>`.
+///
+/// When the `BufReader` is dropped, the contents of its buffer will be
+/// discarded. Creating multiple instances of a `BufReader` on the same
+/// stream can cause data loss.
+///
+/// [`AsyncRead`]: futures_io::AsyncRead
+///
+// TODO: Examples
+pub struct BufReader<R> {
+    inner: R,
+    buf: Box<[u8]>,
+    pos: usize,
+    cap: usize,
+}
+
+impl<R: AsyncRead> BufReader<R> {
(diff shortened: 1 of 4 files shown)


taiki-e:
Blocked on https://github.com/rust-lang-nursery/futures-rs/pull/1533.

futures-util/src/io/buf_reader.rs
```diff
@@ -0,0 +1,259 @@
+use futures_core::task::{Context, Poll};
+use futures_io::{AsyncBufRead, AsyncRead, AsyncSeek, Initializer, IoVec, SeekFrom};
+use pin_utils::{unsafe_pinned, unsafe_unpinned};
+use std::io::{self, Read};
+use std::pin::Pin;
+use std::{cmp, fmt};
+use super::DEFAULT_BUF_SIZE;
+
+/// The `BufReader` struct adds buffering to any reader.
```
taiki-e: `BufReader` vs `AsyncBufReader`

(I have no strong opinion.)

yoshuawuyts: Given it's a concrete type, and not a trait, I think `BufReader` is fine. It's similar to the naming scheme in `Romio`, `Tokio`, and `Runtime` :+1:

futures-util/src/io/buf_reader.rs
```diff
@@ -0,0 +1,257 @@
...
+            // means we managed to allocate 8 exbibytes and that's absurd.
+            // But it's not out of the realm of possibility for some weird underlying reader to
+            // support seeking by i64::min_value() so we need to handle underflow when subtracting
+            // remainder.
+            if let Some(offset) = n.checked_sub(remainder) {
+                result = ready!(self.as_mut().inner().poll_seek(cx, SeekFrom::Current(offset)))?;
+            } else {
+                // seek backwards by our remainder, and then by the offset
+                ready!(self.as_mut().inner().poll_seek(cx, SeekFrom::Current(-remainder)))?;
+                result = ready!(self.as_mut().inner().poll_seek(cx, SeekFrom::Current(n)))?;
```
Nemo157: If you get `Poll::Ready` then `Poll::Pending` from these two calls I'm pretty sure the outer code calling `poll_seek` again will result in seeking to the wrong position. Maybe the `discard_buffer` call should happen in between these, since seeking by the remainder should cancel it out (and then next time `remainder` will be 0 so you can't enter this branch)?

futures-util/src/io/buf_reader.rs
stjepang: I wonder if `AsyncRead` and `AsyncWrite` should've been named `PollRead` and `PollWrite` (because they only have polling read/write traits).

Then we could rename `AsyncReadExt` and `AsyncWriteExt` to simply `Read` and `Write`.

I don't see much point in the `Async` prefix if literally no other trait or type has it. Also, the  `Async*` traits with polling methods are not indented to be used by end-users so it feels wrong for them to have simpler names than the `*Ext` traits which are used more often.

yoshuawuyts: @stjepang that's an interesting point; perhaps opening a separate issue might be good for a discussion?

Nemo157: @stjepang the `Async*` traits are what should be used in bounds so likely used multiple times per file, whereas the `*Ext` traits should be imported and then be used via method syntax so only mentioned once per file.

Nemo157: Most relevant issue: https://github.com/rust-lang-nursery/futures-rs/issues/1398

futures-util/src/io/buf_reader.rs
taiki-e: Oh, I think you are right, thanks! (I should add a test for checking that)

PR #1576, opened 2019-05-01
Add Stream and AsyncRead pending test adaptors

Nemo157:
I just implemented these in another project to make sure that tests using in-memory `AsyncRead` or `stream::iter` based `Stream` will still pass in `Pending`, may as well have them here so that more projects can easily use them.

I'm not really sure on the name, it doesn't entirely suggest the purpose of it, but I can't think of a better one that doesn't get really long (`PendingBetween` maybe?).

### DIFF
--- a/futures-test/Cargo.toml
+++ b/futures-test/Cargo.toml
@@ -16,6 +16,7 @@ name = "futures_test"
 
 [dependencies]
 futures-core-preview = { version = "=0.3.0-alpha.15", path = "../futures-core", default-features = false }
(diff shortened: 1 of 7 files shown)


cramertj:
LGTM aside from a nitpick about the naming-- perhaps `alternate_pending`? `pending` to me sounds like it could mean many things.

Nemo157:
`interleave_pending`?

cramertj:
:shipit:

Nemo157:
@cramertj renamed.

PR #1583, opened 2019-05-07
Replace await! macro with await syntax

taiki-e:
`await` syntax was implemented in https://github.com/rust-lang/rust/pull/60586, and nightly-2019-05-09 has been released with the changes. Also, `await!` macro will be removed in the future.

Refs: [migration tool](https://github.com/taiki-e/replace-await)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,7 +25,7 @@ matrix:
 
     # When updating this, the reminder to update the minimum required version in README.md.
     - name: cargo test (minimum required version)
-      rust: nightly-2019-04-30
+      rust: nightly-2019-05-09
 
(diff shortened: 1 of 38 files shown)


buesima:
Any hints how you replaced this? That would be great to know for projects already using await macros.

Nemo157:
There _might_ be a rust-fixable suggestion available for a little while (see https://github.com/rust-lang/rust/pull/60586#discussion_r281596671)

taiki-e:
@buesima https://github.com/taiki-e/replace-await

taiki-e:
nightly-2019-05-09 released! 🎉

PR #1585, opened 2019-05-07
Add AsyncBufReadExt::lines

taiki-e:


futures-util/src/io/read_line.rs
```diff
@@ -29,20 +29,31 @@ impl<'a, R: AsyncBufRead + ?Sized + Unpin> ReadLine<'a, R> {
...
+pub(super) fn read_line_internal<R: AsyncBufRead + ?Sized>(
+    reader: Pin<&mut R>,
+    buf: &mut String,
+    bytes: &mut Vec<u8>,
+    read: &mut usize,
+    cx: &mut Context<'_>,
+) -> Poll<io::Result<usize>> {
+    let ret = ready!(read_until_internal(reader, b'\n', bytes, read, cx));
+    if str::from_utf8(&bytes).is_err() {
+        Poll::Ready(ret.and_then(|_| {
```
cramertj: nit: `map_err`?

futures-util/src/io/read_line.rs
```diff
@@ -29,20 +29,31 @@ impl<'a, R: AsyncBufRead + ?Sized + Unpin> ReadLine<'a, R> {
...
+            Err(io::Error::new(io::ErrorKind::InvalidData, "stream did not contain valid UTF-8"))
+        }))
+    } else {
+        debug_assert!(buf.is_empty());
+        debug_assert_eq!(*read, 0);
+        unsafe { mem::swap(buf.as_mut_vec(), bytes); }
```
cramertj: nit: would you mind moving the `unsafe { ... }` just around the `as_mut_vec()` (inside the `mem::swap`) and leaving a comment explaining that this is safe because we've already validated that `bytes` is valid UTF8 above?

futures/tests/io_lines.rs
```diff
@@ -0,0 +1,52 @@
...
+use futures_test::io::AsyncReadTestExt;
+use futures_test::task::noop_context;
+use std::io::Cursor;
+use std::pin::Pin;
+
+#[test]
+fn lines() {
+    let buf = Cursor::new(&b"12\r"[..]);
+    let mut s = buf.lines();
+    assert_eq!(block_on(s.next()).unwrap().unwrap(), "12\r".to_string());
```
cramertj: nit: would you mind moving the `block_on(<arg>.next()).unwrap().unwrap()` into a function or macro so that it doesn't clutter the main test logic?

cramertj:
LGTM aside from a few small nits.

futures-util/src/io/read_line.rs
taiki-e: If that was `or_else`, we could use `map_err` instead (`and_then` does not touch the error.).

cramertj: Oh, sorry, I missed that this is actually the inversion-- it's `and_then`-ing and mapping the success value case to an error value. Ignore.

PR #1589, opened 2019-05-09
Add is_in_task

illicitonion:
This returns whether the caller is in a Task.

If a `task_local!` is accessed from outside a Task, it panics. Right now,
there is no way for a caller to know whether it's safe to look in a
task_local!.

For my use-case, I have a logger implementation which uses some task-local information when logging, if logging from a task; the logger is also used from non-tasks, and I don't have a way to guard accessing the task-local information on whether the caller is a task.

If there's a better way to do this, or a change to `task_local!` which would be better, I'd be happy to run with that instead.

### DIFF
--- a/src/task.rs
+++ b/src/task.rs
@@ -33,7 +33,7 @@
 #[allow(deprecated)]
 pub use task_impl::{Spawn, spawn, Unpark, Executor, Run, park};
 
-pub use task_impl::{Task, AtomicTask, current, init};
+pub use task_impl::{Task, AtomicTask, current, init, is_in_task};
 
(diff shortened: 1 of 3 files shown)


cramertj:
cc @carllerche who might have some recommendations. Otherwise, I personally have no opposition to this. I'm imagining you'd like a new release of 0.1 after merging?

carllerche:
I don't have any opinion either way.

cramertj:
Released 0.1.27 with `is_in_task`.

illicitonion:
Fantastic, thanks for your help!

PR #1591, opened 2019-05-10
Prepare 0.3.0-alpha.16

taiki-e:
In relation to https://github.com/rust-lang/rust/issues/60660, it seems that `select!` macro of 0.3.0-alpha.15 cannot compile ([it uses `r#await!`](https://github.com/rust-lang-nursery/futures-rs/blob/0.3.0-alpha.15/futures-select-macro/src/lib.rs#L249)).

cc @cramertj @Nemo157 @seanmonstar

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,9 +1,20 @@
+# 0.3.0-alpha.16 - 2019-5-10
+* Updated to new nightly `async_await`.
+* Changed `AsyncRead::poll_vectored_read` and `AsyncWrite::poll_vectored_write` to use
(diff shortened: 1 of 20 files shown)


cramertj:
@taiki-e published 0.3.0-alpha.16

taiki-e:
@cramertj Thanks!

seanmonstar:
@taiki-e I really appreciate all the work and energy you've been putting into this repo! Thank you!!

cramertj:
Yeah, seconded a million times over! :)

PR #1597, opened 2019-05-11
Remove proc_macro_span from allow-features

taiki-e:
Closes https://github.com/rust-lang-nursery/futures-rs/issues/1564

cc @Nemo157


Nemo157:
Thanks!

PR #1598, opened 2019-05-11
Add an implementation of `TryStream::try_zip`

### LINKED ISSUE #1593: Should TryStream::try_zip exist?
Something like:

```rust
fn try_zip<St>(self, other: St) -> TryZip<Self, St>
where
    St: TryStream<Err = Self::Err>, 
```

Happy to send a PR if it seems reasonable.

shepmaster:
Closes #1593

### DIFF
--- a/futures-util/src/try_stream/try_zip.rs
+++ b/futures-util/src/try_stream/try_zip.rs
@@ -15,7 +15,13 @@ pub struct TryZip<St1: TryStream, St2: TryStream> {
     queued2: Option<Result<St2::Ok, St2::Error>>,
 }
 
-impl<St1: TryStream + Unpin, St2: TryStream + Unpin> Unpin for TryZip<St1, St2> {}
+impl<St1, St2> Unpin for TryZip<St1, St2>
+where
+    St1: TryStream,
+    Fuse<IntoStream<St1>>: Unpin,
+    St2: TryStream,
+    Fuse<IntoStream<St2>>: Unpin,
+{}
 
 impl<St1: TryStream, St2: TryStream> TryZip<St1, St2> {
     unsafe_pinned!(stream1: Fuse<IntoStream<St1>>);
--- a/futures-util/src/try_stream/try_zip.rs
+++ b/futures-util/src/try_stream/try_zip.rs
@@ -1,0 +1,122 @@
+use crate::stream::{StreamExt, Fuse};
+use crate::try_stream::{TryStreamExt, IntoStream};
+use core::pin::Pin;
+use futures_core::stream::{FusedStream, TryStream};
+use futures_core::task::{Context, Poll};
+use pin_utils::{unsafe_pinned, unsafe_unpinned};
+
+/// Stream for the [`try_zip`](super::TryStreamExt::try_zip) method.
+#[derive(Debug)]
+#[must_use = "streams do nothing unless polled"]
+pub struct TryZip<St1: TryStream, St2: TryStream> {
+    stream1: Fuse<IntoStream<St1>>,
+    stream2: Fuse<IntoStream<St2>>,
+    queued1: Option<Result<St1::Ok, St1::Error>>,
+    queued2: Option<Result<St2::Ok, St2::Error>>,
+}
+
+impl<St1: TryStream + Unpin, St2: TryStream + Unpin> Unpin for TryZip<St1, St2> {}
+
+impl<St1: TryStream, St2: TryStream> TryZip<St1, St2> {
+    unsafe_pinned!(stream1: Fuse<IntoStream<St1>>);
+    unsafe_pinned!(stream2: Fuse<IntoStream<St2>>);
+    unsafe_unpinned!(queued1: Option<Result<St1::Ok, St1::Error>>);
+    unsafe_unpinned!(queued2: Option<Result<St2::Ok, St2::Error>>);
+
+    pub(super) fn new(stream1: St1, stream2: St2) -> TryZip<St1, St2> {
+        TryZip {
+            stream1: stream1.into_stream().fuse(),
+            stream2: stream2.into_stream().fuse(),
+            queued1: None,
+            queued2: None,
+        }
+    }
+
+    /// Acquires a reference to the underlying streams that this combinator is
+    /// pulling from.
+    pub fn get_ref(&self) -> (&St1, &St2) {
+        (self.stream1.get_ref().get_ref(), self.stream2.get_ref().get_ref())
+    }
+
+    /// Acquires a mutable reference to the underlying streams that this
+    /// combinator is pulling from.
+    ///
+    /// Note that care must be taken to avoid tampering with the state of the
+    /// stream which may otherwise confuse this combinator.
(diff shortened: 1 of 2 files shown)


shepmaster:
This makes the decision to "throw away" one of the errors if both streams report a failure for the same zipper tooth. This may or may not be desired... ?

futures-util/src/try_stream/try_zip.rs
```diff
@@ -0,0 +1,122 @@
...
+{
+    type Ok = (St1::Ok, St2::Ok);
+    type Error = St1::Error;
+
+    fn try_poll_next(
+        mut self: Pin<&mut Self>,
+        cx: &mut Context<'_>,
+    ) -> Poll<Option<Result<Self::Ok, Self::Error>>> {
+        if self.queued1.is_none() {
+            match self.as_mut().stream1().try_poll_next(cx) {
```
taiki-e: I think using `self.as_mut().stream1().try_poll_next(cx)?` here will avoid the problem you mentioned.

futures-util/src/try_stream/try_zip.rs
```diff
@@ -0,0 +1,122 @@
...
+#[derive(Debug)]
+#[must_use = "streams do nothing unless polled"]
+pub struct TryZip<St1: TryStream, St2: TryStream> {
+    stream1: Fuse<IntoStream<St1>>,
+    stream2: Fuse<IntoStream<St2>>,
+    queued1: Option<Result<St1::Ok, St1::Error>>,
+    queued2: Option<Result<St2::Ok, St2::Error>>,
+}
+
+impl<St1: TryStream + Unpin, St2: TryStream + Unpin> Unpin for TryZip<St1, St2> {}
```
Nemo157: This should technically be

```rust
impl<St1: TryStream, St2: TryStream> Unpin for TryZip<St1, St2>
where
    Fuse<IntoStream<St1>>: Unpin,
    Fuse<IntoStream<St2>>: Unpin,
{}
```

In case either `Fuse` or `IntoStream` don’t have the equivalent `Unpin` implementations.

futures-util/src/try_stream/try_zip.rs
shepmaster: Wouldn't that cause the two halves of the zipper to become unaligned? That is,

```
A: Ok(1), Err(2), Ok(3)
B: Ok(4), Ok(5), Ok(6)
```

Would result in 

```
Ok((1, 4)), Err(2), Ok((3,5))
```

Is that desired? (Perhaps this also suggests why this combinator hasn't bee created before now).

futures-util/src/try_stream/try_zip.rs
shepmaster: Makes sense, but it's unfortunate that the internal implementation details need to leak out like that.

Should the `impl Unpin for Zip` be updated to use `Fuse` as well? Is this a broader issue?

futures-util/src/try_stream/try_zip.rs
taiki-e: Hmm, but that's the same behavior as `Stream::zip` of futures 0.1.

https://github.com/rust-lang-nursery/futures-rs/blob/5654784d5b677dadc4b86affc575d12aa737e4a7/src/stream/zip.rs#L37

futures-util/src/try_stream/try_zip.rs
Nemo157: > it's unfortunate that the internal implementation details need to leak out like that.

Yeah, that was my feeling as well as I wrote the comment. I rationalized it as being the same as the auto-trait's structural rules, if `TryZip` were to rely on the implicit `Unpin` you would still get error messages saying "`Fuse` does not implement `Unpin`".

> Is this a broader issue?

Probably, I'll try to remember to open a ticket about it. (This is one of the reasons I prefer [`pin-project`](https://docs.rs/pin-project/) for projection, it can derive the `Unpin` implementation for you so that you don't have to keep it consistent with the current field definitions).

futures-util/src/try_stream/try_zip.rs
shepmaster: > the same behavior as `Stream::zip` of futures 0.1

It sure is:

```rust
use futures::{future, stream, Future, Stream}; // 0.1.26

fn main() {
    let a: Vec<Result<i32, i32>> = vec![Ok(1), Err(2), Ok(3)];
    let b: Vec<Result<i32, i32>> = vec![Ok(4), Ok(5), Ok(6)];

    stream::iter_result(a)
        .zip(stream::iter_result(b))
        .then(|x| future::ok::<_, ()>(eprintln!("{:?}", x)))
        .for_each(|_| Ok(()))
        .wait();
}
```

([Playground](https://play.integer32.com/?version=stable&mode=debug&edition=2018&gist=bbedf213c602be358159fce4f6ca5493))

Output:

```
Ok((1, 4))
Err(2)
Ok((3, 5))
```

But I think the question remains: is this the expected and desired behavior?

shepmaster: > is this the expected and desired behavior?

If we are open to changing behaviors, the most comprehensive solution is to have

```rust
enum ZipError<A, B> {
    A(A),
    B(B),
    Both(A, B)
}
```

taiki-e: Hmm... I have no clear opinion on which is better, for now.

FWIW, as another thing we may need to consider, the behavior of current this PR's TryZip can be written using Zip, but the behavior of 0.1's Zip cannot be written using either 0.3's Zip or current this PR's TryZip.

cc @cramertj @Nemo157 Do you have any thoughts?

futures-util/src/try_stream/try_zip.rs
Nemo157: I was just looking into this a little more as I encountered the situation in another library where the `Unpin` definition depends on a crate internal type. It looks like rustdoc has special handling for the automatic implementations that looks through the field types and shows the bounds that result from them instead of revealing the internal types, it would be nice if there were some way to request that to happen for manual implementations as well.

For example

```rust
struct Foo<T>(T);

pub struct Bar<T>(Foo<T>);

unsafe impl<T: Send + Iterator> Send for Foo<T> {}
```

shows in the docs

```rust
impl<T> Send for Bar<T>
where
    T: Iterator + Send, 
```

futures-util/src/try_stream/try_zip.rs
cramertj: I would've personally expected that `try_zip` would stop returning at all after the first error, but I don't have a strong opinion here.

cramertj: Bump-- @shepmaster what are your thoughts here?

shepmaster: Sadly, I don't even remember what my original use case was for this, and I don't recall what my thoughts were. 😊 

I've posted the open questions back on the issue. The fact that there aren't people clamoring for this feature is perhaps a sign that it's not useful, yet.

cramertj: Okay! Let me know if you're interested in revisiting at some point.

PR #1600, opened 2019-05-11
Add #[must_use] to more types

taiki-e:
Also fixes some `#[must_use]` messages.

EDIT: also updates `#[must_use]` message of impl `Future` types.

### DIFF
--- a/futures-util/src/try_stream/try_for_each.rs
+++ b/futures-util/src/try_stream/try_for_each.rs
@@ -6,7 +6,7 @@ use pin_utils::{unsafe_pinned, unsafe_unpinned};
 
 /// Future for the [`try_for_each`](super::TryStreamExt::try_for_each) method.
 #[derive(Debug)]
-#[must_use = "streams do nothing unless polled"]
+#[must_use = "futures do nothing unless polled"]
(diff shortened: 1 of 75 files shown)


futures-util/src/try_stream/try_for_each.rs
```diff
@@ -6,7 +6,7 @@ use pin_utils::{unsafe_pinned, unsafe_unpinned};
 
 /// Future for the [`try_for_each`](super::TryStreamExt::try_for_each) method.
 #[derive(Debug)]
-#[must_use = "streams do nothing unless polled"]
```
cramertj: I'd still expect the streams to say "streams do nothing ... "

taiki-e: `TryForEach` implements the future, not the stream: https://rust-lang-nursery.github.io/futures-api-docs/0.3.0-alpha.16/futures/stream/struct.TryForEach.html#impl-Future

(The types that implement the stream are keeping that description.)

cramertj: hoh boy thank you

PR #1601, opened 2019-05-11
Correct the Unpin bounds for Zip

shepmaster:


cramertj:
Can you clarify what you'd expect this to change? at face value it's "more correct" in a sense, but `Fuse<T>` should be `Unpin` iff `T: Unpin`, so I'd expect this bound to be equivalent.

shepmaster:
> expect this to change

Nothing, myself. This was in response to [a parallel comment](https://github.com/rust-lang-nursery/futures-rs/pull/1598#discussion_r283102458).

The benefit of this commit is that it's resilient in the face of changes to changes to `Fuse`'s implementation details.

PR #1604, opened 2019-05-13
Implement std::error::Error for SpawnError

jrobsonchase:
Pretty straightforward - just a `Display` and `Error` implementation for `SpawnError`.

### DIFF
--- a/futures-core/src/task/spawn.rs
+++ b/futures-core/src/task/spawn.rs
@@ -72,6 +72,7 @@ impl fmt::Display for SpawnError {
     }
 }
 
+#[cfg(feature = "std")]
 impl std::error::Error for SpawnError {}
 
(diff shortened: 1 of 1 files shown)


futures-core/src/task/spawn.rs
```diff
@@ -66,6 +66,14 @@ impl fmt::Debug for SpawnError {
     }
 }
 
+impl fmt::Display for SpawnError {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "Executor is shutdown")
+    }
+}
+
+impl std::error::Error for SpawnError {}
```
taiki-e: Could you add `#[cfg(feature = "std")]` to this impl.

Issue #882, opened 2018-03-20
Docs on what can be done without std?


robomancer-or:
The readme says that without std there "a significantly reduced API surface", but I haven't found anywhere that lays out what features require std and what features don't.  Does that information exist anywhere besides the source?

Nemo157:
Easy way to see it is `cd futures-rs/futures && cargo doc --no-default-features --open`. There _was_ [a ticket](https://github.com/rust-lang/rust/issues/35903) about adding support for displaying feature dependencies to `rustdoc`, which would allow viewing this on docs.rs as well, but no idea if something like that will ever actually happen.

robomancer-or:
Having to clone a repo just to get a sense of if it could work for me seems inelegant (I suppose I'm preaching to the choir on that one), but it's certainly better than nothing, thanks!  (I'm going to have to keep that trick in my back pocket for other crates)

PR #1607, opened 2019-05-15
Document feature gate for std/alloc features

taiki-e:
Closes https://github.com/rust-lang-nursery/futures-rs/issues/882


futures-util/src/stream/futures_unordered/mod.rs
```diff
@@ -1,4 +1,7 @@
 //! An unbounded set of futures.
+//!
+/// This module is only available when the `std` or `alloc` feature of this
+/// library is activated, and it is activated by default.
```
Nemo157: `s_///_//!`

taiki-e: Oops, good catch!

Issue #1510, opened 2019-04-03
impl AsyncWrite for Cursor<Vec<u8>> doesn't extend vector as needed


Nemo157:
This implementation is currently provided by [`impl<T: AsMut<[u8]>> AsyncWrite for Cursor<T>`](https://rust-lang-nursery.github.io/futures-api-docs/0.3.0-alpha.13/futures/io/trait.AsyncWrite.html#impl-AsyncWrite-1) which doesn't allow it to know that the vector is extendable. `std` provides a separate [`impl io::Write for Cursor<Vec<u8>>`](https://doc.rust-lang.org/nightly/std/io/trait.Write.html#impl-Write-2) which extends the vector as needed, but this means it can't provide a generic version.

We're also missing the direct `impl AsyncWrite for Vec<u8>` which could be used instead (EDIT: Added in #1511). It almost seems like we should drop the `Cursor` wrapped implementations until we get an `AsyncSeek` trait available?

Nemo157:
Actually the `Cursor` implementations are still useful as they allow reading off how many bytes were read/written after operating on it.

---

One thing that may be worth looking into is whether `std` plans to use specialization to get both generic implementations and an override that extends `Vec` as needed, that seems to me to be the best endpoint, but maybe there are reasons not to do it.

yoshuawuyts:
>  until we get an AsyncSeek trait available?

Chiming in here: I've def been missing an `AsyncSeek` impl. We've been looking to implement a filesystem API for Romio, and that's one of the traits we need. Tokio has [a counterpart](https://docs.rs/tokio-fs/0.1.6/tokio_fs/file/struct.File.html#method.poll_seek) directly implemented on the File struct, but it seems being able to generalize it would be ideal.

Nemo157:
Test commit providing this behaviour via specialization: https://github.com/rust-lang-nursery/futures-rs/commit/7997066642e5120db44f9eb7ceac612ca7411316.

cramertj:
@Nemo157 
> It almost seems like we should drop the Cursor wrapped implementations until we get an AsyncSeek trait available?

https://github.com/rust-lang-nursery/futures-rs/pull/1553 added `AsyncSeek`-- what would you like to see happen on this issue?

Nemo157:
[...]

PR #1608, opened 2019-05-15
Add async BufWriter

taiki-e:
Closes https://github.com/rust-lang-nursery/futures-rs/issues/1519

### DIFF
--- a/futures-util/src/io/buf_writer.rs
+++ b/futures-util/src/io/buf_writer.rs
@@ -1,0 +1,181 @@
+use futures_core::task::{Context, Poll};
+use futures_io::{AsyncSeek, AsyncWrite, IoSlice, SeekFrom};
+use pin_utils::{unsafe_pinned, unsafe_unpinned};
+use std::fmt;
+use std::io::{self, Write};
+use std::pin::Pin;
+use super::DEFAULT_BUF_SIZE;
+
+/// Wraps a writer and buffers its output.
+///
+/// It can be excessively inefficient to work directly with something that
+/// implements [`AsyncWrite`]. A `BufWriter` keeps an in-memory buffer of data and
+/// writes it to an underlying writer in large, infrequent batches.
+///
+/// `BufWriter` can improve the speed of programs that make *small* and
+/// *repeated* write calls to the same file or network socket. It does not
+/// help when writing very large amounts at once, or writing just one or a few
+/// times. It also provides no advantage when writing to a destination that is
+/// in memory, like a `Vec<u8>`.
+///
+/// When the `BufWriter` is dropped, the contents of its buffer will be
+/// discarded. Creating multiple instances of a `BufWriter` on the same
+/// stream can cause data loss. If you need to write out the contents of its
+/// buffer, you must manually call flush before the writer is dropped.
+///
+/// [`AsyncWrite`]: futures_io::AsyncWrite
+/// [`flush`]: super::AsyncWriteExt::flush
+///
+// TODO: Examples
+pub struct BufWriter<W> {
+    inner: W,
+    buf: Vec<u8>,
+    written: usize,
+}
+
+impl<W: AsyncWrite> BufWriter<W> {
+    unsafe_pinned!(inner: W);
+    unsafe_unpinned!(buf: Vec<u8>);
+
+    /// Creates a new `BufWriter` with a default buffer capacity. The default is currently 8 KB,
+    /// but may change in the future.
+    pub fn new(inner: W) -> Self {
+        Self::with_capacity(DEFAULT_BUF_SIZE, inner)
+    }
+
+    /// Creates a new `BufWriter` with the specified buffer capacity.
+    pub fn with_capacity(cap: usize, inner: W) -> Self {
+        Self {
+            inner,
+            buf: Vec::with_capacity(cap),
+            written: 0,
+        }
+    }
+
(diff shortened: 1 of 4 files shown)


futures-util/src/io/buf_writer.rs
```diff
@@ -0,0 +1,181 @@
...
+        ready!(self.as_mut().flush_buf(cx))?;
+        self.inner().poll_close(cx)
+    }
+}
+
+impl<W: AsyncWrite + fmt::Debug> fmt::Debug for BufWriter<W> {
+    fn fmt(&self, fmt: &mut fmt::Formatter<'_>) -> fmt::Result {
+        fmt.debug_struct("BufWriter")
+            .field("writer", &self.inner)
+            .field("buffer", &format_args!("{}/{}", self.buf.len(), self.buf.capacity()))
```
cramertj: nit: any reason for omitting `written` from the debug output?

futures/tests/io_buf_writer.rs
```diff
@@ -0,0 +1,247 @@
+use futures::executor::block_on;
+use futures::future::Future;
+use futures::io::{AsyncSeek, AsyncSeekExt, AsyncWrite, AsyncWriteExt, BufWriter, SeekFrom};
+use futures::task::{Context, Poll};
+use futures_test::task::noop_context;
+use std::io::{self, Cursor};
+use std::pin::Pin;
+
+macro_rules! run_write {
```
cramertj: Is there a reason for this rather than `writer.write(&mut buf).await`?

futures/tests/io_buf_writer.rs
```diff
@@ -0,0 +1,247 @@
...
+        } else {
+            self.ready_seek = true;
+            Poll::Pending
+        }
+    }
+}
+
+#[test]
+fn maybe_pending_buf_writer_seek() {
+    // FIXME: when https://github.com/rust-lang-nursery/futures-rs/issues/1510 fixed,
```
cramertj: Now that `AsyncSeek` has been merged, this can probably be fixed, right?

futures/tests/io_buf_writer.rs
taiki-e: We don't have `AsyncWriteExt::write` yet (but I think it makes sense to have it).

futures-util/src/io/buf_writer.rs
```diff
@@ -0,0 +1,182 @@
...
+            match ready!(inner.as_mut().poll_write(cx, &buf[*written..])) {
+                Ok(0) => {
+                    ret = Err(io::Error::new(
+                        io::ErrorKind::WriteZero,
+                        "failed to write the buffered data",
+                    ));
+                    break;
+                }
+                Ok(n) => *written += n,
+                Err(ref e) if e.kind() == io::ErrorKind::Interrupted => {}
```
Nemo157: [`poll_write` cannot return `Interrupted`](https://rust-lang-nursery.github.io/futures-api-docs/0.3.0-alpha.16/futures/io/trait.AsyncWrite.html#implementation)

taiki-e: Should we fix `delegate_async_*_to_stdio`'s implementation?
https://github.com/rust-lang-nursery/futures-rs/blob/d75f6efd064a30dd8564b7a6927c63606c7ad419/futures-io/src/lib.rs#L476-L482

futures/tests/io_buf_writer.rs
taiki-e: Opened https://github.com/rust-lang-nursery/futures-rs/pull/1612.

futures-util/src/io/buf_writer.rs
Nemo157: That only appears to be used for "infallible" types, `AllowStdIo` correctly handles it:

https://github.com/rust-lang-nursery/futures-rs/blob/d75f6efd064a30dd8564b7a6927c63606c7ad419/futures-util/src/io/allow_std.rs#L21-L37

taiki-e: > That only appears to be used for "infallible" types

Indeed!

futures/tests/io_buf_writer.rs
cramertj: Merged

taiki-e: Thanks!

futures/tests/io_buf_writer.rs
taiki-e: As far as I tried locally, it still didn't work...

Nemo157: Ah, yeah, the choice currently is to have a `Vec<u8>` which is `AsyncWrite` and extends as needed or `Cursor<Vec<u8>>` which is `AsyncWrite + AsyncSeek` but doesn't extend as needed #1510 is about whether we can have a type that implements both without having to drop the genericness of the current `Cursor` implementation.

Nemo157: I threw together a test commit and this can definitely be done via specialization: https://github.com/rust-lang-nursery/futures-rs/commit/7997066642e5120db44f9eb7ceac612ca7411316, if only it were close to stabilization.

taiki-e:
This seems to be my 104th commit to this repository.
Thanks to the maintainers who have reviewed my many PRs.

cramertj:
\o/

PR #1615, opened 2019-05-17
Use #[warn(unreachable_pub)]

taiki-e:
I think it helps to avoid problems like https://github.com/rust-lang-nursery/futures-rs/issues/1477. (Seems there are still some false positives, but I think their impact is small)

### DIFF
--- a/futures-io/src/lib.rs
+++ b/futures-io/src/lib.rs
@@ -10,7 +10,7 @@
 
 #![cfg_attr(not(feature = "std"), no_std)]
 
-#![warn(missing_docs, missing_debug_implementations, rust_2018_idioms)]
+#![warn(missing_docs, missing_debug_implementations, rust_2018_idioms, unreachable_pub)]
 #![warn(clippy::all)]
 
 #![doc(html_root_url = "https://rust-lang-nursery.github.io/futures-api-docs/0.3.0-alpha.16/futures_io")]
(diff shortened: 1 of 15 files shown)


futures-io/src/lib.rs
```diff
@@ -27,12 +27,16 @@ mod if_std {
 
     // Re-export some types from `std::io` so that users don't have to deal
     // with conflicts when `use`ing `futures::io` and `std::io`.
-    pub use self::StdIo::Error as Error;
-    pub use self::StdIo::ErrorKind as ErrorKind;
-    pub use self::StdIo::Result as Result;
-    pub use self::StdIo::IoSlice as IoSlice;
-    pub use self::StdIo::IoSliceMut as IoSliceMut;
-    pub use self::StdIo::SeekFrom as SeekFrom;
+    #[allow(clippy::useless_attribute)] // https://github.com/rust-lang/rust-clippy/issues/4106
```
taiki-e: This false positive has already been fixed on the master of clippy (https://github.com/rust-lang/rust-clippy/pull/4107), but it may take some time for nightly to be released with this change.

cramertj:
nice! i'm actually surprised this required so few changes. I'll happily merge once conflicts are resolved here.

taiki-e:
Rebased (and Travis passed).

Issue #1588, opened 2019-05-08
Make Sender<T> struct comparable


LinkTed:
I would be great to add the `PartialEq` trait to the `Sender<T>` struct, because I want to remove all `Sender<T>` from a `Vec<Sender<T>>` which point to a specific channel, like:
```rust
fn remove_from_list(list: &mut Vec<Sender<u32>>, other: &Sender<u32>) {
    list.retain(|s| s != other);
}
```
I do not know if the following is correct:
```rust
impl<T> PartialEq for Sender<T> {
    fn eq(&self, other: &Sender<T>) -> bool {
        Arc::eq_ptr(&self.inner, &other.inner)
    }
}
```
Can you please verify the code and add it to the repository. If you want i can make PR.

taiki-e:
Related: https://github.com/crossbeam-rs/crossbeam/issues/319

azastrael:
@LinkTed The `same_receiver` has been added in #1617 to do that.

LinkTed:
Thanks

PR #1617, opened 2019-05-17
Added a "same_receiver" method to (Unbounded)Sender

azastrael:
Hello!
This was asked in #1588 and I wanted to make a first contribution :smile:. As proposed by the issue's creator, I used `Arc::eq_ptr` to compare the `inner` field of `SenderInner`. I also added a ~`partial_eq` (though, I'm not really sure about its name)~ `same_receiver` test which compares a `Sender` with a clone of it when both are connected to the channel, when one of them is disconnected and when the channel is closed; it also compares two different channels' `Sender`s.

### DIFF
--- a/futures-channel/src/mpsc/mod.rs
+++ b/futures-channel/src/mpsc/mod.rs
@@ -758,6 +758,10 @@ impl<T> PartialEq for SenderInner<T> {
     }
 }
 
+impl<T> Eq for Sender<T> {}
+impl<T> Eq for UnboundedSender<T> {}
+impl<T> Eq for SenderInner<T> {}
+
 impl<T> Clone for Sender<T> {
     fn clone(&self) -> Sender<T> {
         Sender(self.0.clone())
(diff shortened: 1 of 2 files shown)


stjepang:
In `crossbeam-channel`, we were considering adding the same impls but, after a long discussion, concluded it would be better to have methods that test whether senders/receivers are equivalent: https://github.com/crossbeam-rs/crossbeam/issues/319

azastrael:
Oh, right, so I read the changes from https://github.com/crossbeam-rs/crossbeam/pull/320/ and it seems that the way it compares is similar to `Arc::ptr_eq` (in some cases it even is how it does it). So, if I understand correctly, removing the two impls and replacing them with a `identical_to` (or `same_channel`) method would be OK?

cramertj:
Perhaps `fn same_receiver(&self, other: &Self) -> bool`?

cramertj:
Great, thanks!

Issue #1444, opened 2019-01-30
try_flatten_stream() in futures-preview 0.3.0


ebkalderon:
With `futures` 0.1, it was possible to have a `Future` that resolves to a `Stream` and use `flatten_stream()` to get access to the underlying stream, but if the `Future` resulted in an `Error`, the resulting stream would short-circuit and resolve to an `Error` as well.

When working with `futures-preview` 0.3.0, handling this situation is pretty cumbersome:

```rust
let future = async {
    let response = await!(client.get(uri))?;
    Ok(response.into_body())
};

// Worse
let stream = future
    .then(|result| match result {
        Ok(stream) => future::ready(Box::pin(stream) as Pin<Box<dyn Stream<Item = _> + Send>>),
        Err(err) => future::ready(Box::pin(stream::once(future::err(err))) as Pin<Box<dyn Stream<Item = _> + Send>>),
    }).flatten_stream();

// Slightly better
let stream = future
    .map_ok(|stream| Box::pin(stream) as Pin<Box<dyn Stream<Item = _> + Send>>)
    .unwrap_or_else(|err| Box::pin(stream::once(future::err(err))))
    .flatten_stream();
```

It would be nice if there was a `TryFutureExt::try_flatten_stream()` method which could simplify the code above like so:

```rust
let future = async {
    let response = await!(client.get(uri))?;
    Ok(response.into_body())
};

let stream = future.try_flatten_stream();
```

PR #1618, opened 2019-05-18
Add TryFutureExt::try_flatten_stream

taiki-e:
Closes https://github.com/rust-lang-nursery/futures-rs/issues/1444

### DIFF
--- a/futures-util/src/future/flatten_stream.rs
+++ b/futures-util/src/future/flatten_stream.rs
@@ -8,11 +8,11 @@ use pin_utils::unsafe_pinned;
 /// Stream for the [`flatten_stream`](super::FutureExt::flatten_stream) method.
 #[must_use = "streams do nothing unless polled"]
 pub struct FlattenStream<Fut: Future> {
-    state: State<Fut>
+    state: State<Fut, Fut::Output>,
 }
 
 impl<Fut: Future> FlattenStream<Fut> {
-    unsafe_pinned!(state: State<Fut>);
+    unsafe_pinned!(state: State<Fut, Fut::Output>);
 
     pub(super) fn new(future: Fut) -> FlattenStream<Fut> {
         FlattenStream {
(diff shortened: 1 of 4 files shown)


futures-util/src/future/flatten_stream.rs
```diff
@@ -58,34 +61,22 @@ impl<Fut> Stream for FlattenStream<Fut>
     fn poll_next(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
         loop {
             // safety: data is never moved via the resulting &mut reference
-            let stream = match &mut unsafe { Pin::get_unchecked_mut(self.as_mut()) }.state {
+            match &mut unsafe { Pin::get_unchecked_mut(self.as_mut()) }.state {
```
Nemo157: To encapsulate this unsafety you may be able to add something like

```rust
impl<Fut: Future> State<Fut> {
    fn get_pin_mut(self: Pin<&mut Self>) -> State<Pin<&mut Fut>>
}
```

I think that might work since `<Fut as Future>::Output == <Pin<&mut Fut> as Future>::Output`, but `rustc` also might not be able to determine that I guess.

taiki-e: Perhaps that doesn't work well.
https://play.rust-lang.org/?version=nightly&mode=debug&edition=2018&gist=5149d533324f5fad6c4056b230f8117e

Nemo157: Actually, that's a different issue than what I was afraid of, it's definitely not possible to have that API because of the `Stream` variant holding it by value.

You could split it into a generic `Either` like https://play.rust-lang.org/?version=nightly&mode=debug&edition=2018&gist=0c27eb07e5693dab683e855a30aa961b, but I'm not certain it's worth it.

taiki-e: Another way I'm thinking is to use generics in `State`, not associated types.
https://play.rust-lang.org/?version=nightly&mode=debug&edition=2018&gist=36bf9cde7b51f3ac63355fb2fb3145a9

futures-util/src/future/flatten_stream.rs
```diff
@@ -30,11 +33,25 @@ impl<Fut> fmt::Debug for FlattenStream<Fut>
...
+enum State<Fut, St> {
     // future is not yet called or called and not ready
     Future(Fut),
     // future resolved to Stream
-    Stream(Fut::Output),
+    Stream(St),
+}
+
+impl<Fut, St> State<Fut, St> {
+    fn get_pin_mut<'a>(self: Pin<&'a mut Self>) -> State<Pin<&'a mut Fut>, Pin<&'a mut St>> {
```
taiki-e: `'a` can be removed when https://github.com/rust-lang/rust/pull/60944 is merged.

cramertj:
Looks great, thanks for this! Will merge once conflicts are resolved.

taiki-e:
Rebased (and Travis passed).

PR #1620, opened 2019-05-18
Add some AsyncWrite test utilities

Nemo157:
I wrote these as part of https://github.com/rustasync/async-compression/pull/22 to ensure the `AsyncWrite` wrapper there correctly handled short-writes and `Pending` while using an in-memory writer in tests.


cramertj:
Nice! LGTM.

PR #1621, opened 2019-05-18
Move futures-util tests into futures/tests

taiki-e:
Currently, most combinator tests are in futures/tests, and some remaining tests are in some files of futures-util. It is probably better to put this in one folder (it will be easier to find).


cramertj:
Thanks! This has a conflict with master, but I'll merge when that's resolved.

taiki-e:
Rebased (and Travis passed).

cramertj:
merged

PR #1622, opened 2019-05-19
Add future::try_select

taiki-e:
The same as [0.1's select2](https://docs.rs/futures/0.1.27/futures/future/trait.Future.html#method.select2).

### DIFF
--- a/futures-util/src/try_future/try_select.rs
+++ b/futures-util/src/try_future/try_select.rs
@@ -1,0 +1,80 @@
+use core::pin::Pin;
+use futures_core::future::{Future, TryFuture};
+use futures_core::task::{Context, Poll};
+use crate::future::Either;
+
+/// Future for the [`try_select()`] function.
+#[must_use = "futures do nothing unless you `.await` or poll them"]
+#[derive(Debug)]
+pub struct TrySelect<A: Unpin, B: Unpin> {
+    inner: Option<(A, B)>,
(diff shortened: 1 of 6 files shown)


Matthias247:
Is there any situation you would prefer this over `select!`?
I would actually try to encourage people to really go the idiomatic async/await way, and not to offer multiple ways to do the same thing (that all need to be maintained).

taiki-e:
`TrySelect` is `TryFuture` but `Select` is not.

futures-util/src/try_future/try_select.rs
```diff
@@ -0,0 +1,80 @@
...
+pub fn try_select<A, B>(future1: A, future2: B) -> TrySelect<A, B>
+    where A: TryFuture + Unpin, B: TryFuture + Unpin
+{
+    TrySelect { inner: Some((future1, future2)) }
+}
+
+impl<A: Unpin, B: Unpin> Future for TrySelect<A, B>
+    where A: TryFuture, B: TryFuture
+{
+    #[allow(clippy::type_complexity)]
```
cramertj: lol-- no kidding

cramertj:
seems basically fine to me, and could help people move over code directly without going all the way to async/await. It is unfortunate to have all of these "multiple ways to do things" but after all that's sort of why `futures-util` exists.

PR #1627, opened 2019-05-20
Do not terminate stream::SelectAll until the completion of all streams

### LINKED ISSUE #1626: Misleading documentation of `SelectAll`?
Documentation for the [futures::stream::SelectAll](https://rust-lang-nursery.github.io/futures-api-docs/0.3.0-alpha.16/futures/stream/struct.SelectAll.html) of the `futures = "0.3.0-alpha.16"` says that `SelectAll` "maintain a set of streams and drive them all to completion".

With the following example

```rust
#![feature(async_await)]

use futures::*;

#[runtime::main]
async fn main() {
    let a = stream::iter((0..5));
    let b = stream::iter((10..20));

    let mut s = stream::select_all(vec![a, b]);

    while let Some(res) = s.next().await {
        dbg!(res);
    }
}
```
and used dependencies
```toml
runtime = "0.3.0-alpha.3"
futures-preview = "0.3.0-alpha.16"
```

`SelectAll` stops when the first inner stream exhausts:

```
[src/main.rs:13] res = 0
[src/main.rs:13] res = 10
[src/main.rs:13] res = 1
[src/main.rs:13] res = 11
[src/main.rs:13] res = 2
[src/main.rs:13] res = 12
[src/main.rs:13] res = 3
[src/main.rs:13] res = 13
[src/main.rs:13] res = 4
[src/main.rs:13] res = 14
```

Based on the documentation I was expecting it to continue driving stream `b` to completion, but it does not.
It seems that either documentation is wrong or `SelectAll` implementation is broken, I'm not sure which case is it.

UPD: `futures::stream::select` fn exhausts both streams and from my point of view the same behavior can be expected from the `select_all` fn.

taiki-e:
Fixes #1626

### DIFF
--- a/futures-util/src/stream/select_all.rs
+++ b/futures-util/src/stream/select_all.rs
@@ -77,20 +77,20 @@ impl<St: Stream + Unpin> Stream for SelectAll<St> {
         mut self: Pin<&mut Self>,
         cx: &mut Context<'_>,
     ) -> Poll<Option<Self::Item>> {
-        match self.inner.poll_next_unpin(cx) {
-            Poll::Pending => Poll::Pending,
-            Poll::Ready(Some((Some(item), remaining))) => {
-                self.push(remaining);
-                Poll::Ready(Some(item))
+        loop {
+            match ready!(self.inner.poll_next_unpin(cx)) {
+                Some((Some(item), remaining)) => {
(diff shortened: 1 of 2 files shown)


cramertj:
Thanks for the quick fix!

PR #1630, opened 2019-05-22
Add TryFutureExt::{inspect_ok, inspect_err}

taiki-e:
`inspect_ok` is the same as 0.1's `Future::inspect_ok`. `inspect_err` is the same as `Future::inspect_err` added in #908


cramertj:
LGTM-- are you planning to add these to `TryStreamExt` as well?

taiki-e:
We added them in #1569.

PR #1632, opened 2019-05-23
Add extension traits for adapting an Option<T> into a Future or Stream

udoprog:
This is a work in progress and I'm setting it up to garner feedback.

I recently needed to have optional branches in a `select!` macro and ended up building this solution locally in the project for it. It's general enough that it could be useful to others.

The following is a motivating example for why you might need this:
```rust
use futures::OptionExt as _;

let timeout = None;
let events = events_stream();

loop {
    select! {
        _ = timeout.current() => {
            println!("timeout expired");
        }
        event = events.next() => {
            match event {
                Some(Event::Clear) => timeout = None,
                Some(Event::Set) => timeout = Some(async {
                    /* work */
                }.boxed()),
            }
        }
    }
}
```

This is similar to `OptionFuture`, but I've decided to piggyback on `Option` directly through an extension trait to get its associated plumbing "for free" (like `as_mut`, and `take`). See [this example](https://github.com/udoprog/setmod/blob/async-await/bot/src/module/countdown.rs#L100) that makes use of this.

Would it be interesting to incorporate something like this into the project?

### DIFF
--- a/futures-util/src/option/mod.rs
+++ b/futures-util/src/option/mod.rs
@@ -1,0 +1,180 @@
+//! Extension traits for dealing with [`Option`]s as [`Future`]s or [`Stream`]s.
+//!
+//! # Examples
+//!
+//! ```
+//! #![feature(async_await)]
+//! use futures::{future::{self, FusedFuture as _}, OptionExt as _};
+//! # futures::executor::block_on(async {
+//! let mut f = Some(future::ready::<u32>(1));
+//! assert!(f.is_some());
+//! assert_eq!(f.current().await, 1);
+//! assert!(f.is_none());
+//! assert!(f.current().is_terminated());
+//! # });
+//! ```
+//!
+//! This is useful when you want to implement optional branches using the
+//! `select!` macro.
+//!
+//! ```
+//! #![feature(async_await)]
+//! #![recursion_limit="128"]
+//!
+//! use futures::{future, stream, OptionExt as _, StreamExt as _};
+//! # futures::executor::block_on(async {
+//! let mut value = None;
+//! let mut values = Some(stream::iter(vec![1u32, 2u32, 4u32].into_iter()).fuse());
+//! let mut parked = None;
+//!
+//! let mut sum = 0;
+//!
+//! loop {
+//!     futures::select! {
+//!         value = value.current() => {
+//!             sum += value;
+//!             std::mem::swap(&mut parked, &mut values);
+//!         }
+//!         v = values.next() => {
+//!             match v {
+//!                 Some(v) => {
+//!                     value = Some(future::ready(v));
+//!                     std::mem::swap(&mut parked, &mut values);
+//!                 },
+//!                 None => break,
+//!             }
+//!         }
+//!     }
+//! }
+//!
+//! assert_eq!(7, sum);
+//! # });
+//! ```
+//!
+//! [`Option`]: Option
+//! [`Stream`]: futures_core::stream::Stream
+//! [`Future`]: futures_core::future::Future
+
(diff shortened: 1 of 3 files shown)


cramertj:
FWIW you can model this with `Fuse::terminated` today, although I grant that that's non-obvious.

udoprog:
This was the only blocker in terms of abstractions I faced when migrating a project from mostly manually implemented futures to 0.3 with async/await. It wouldn't hurt to improve its ergonomics somehow.

What would you recommend?

udoprog:
Since I'm still a bit unsure about how to interact with `Pin` I'd appreciate if someone could take a look at these `unsafe` to do pin projection:

https://github.com/rust-lang-nursery/futures-rs/pull/1632/files#diff-df5387eedccc97f58626482b674f3e1aR120

https://github.com/rust-lang-nursery/futures-rs/pull/1632/files#diff-df5387eedccc97f58626482b674f3e1aR166

futures/src/lib.rs
```diff
@@ -58,7 +58,14 @@ compile_error!("The `never-type` feature requires the `nightly` feature as an ex
 #[cfg(feature = "std")]
 #[doc(hidden)] pub use futures_io::{AsyncRead, AsyncWrite, AsyncSeek, AsyncBufRead};
 #[cfg(feature = "std")]
-#[doc(hidden)] pub use futures_util::{AsyncReadExt, AsyncWriteExt, AsyncSeekExt, AsyncBufReadExt};
```
taiki-e: Please put this back and write `#[doc(hidden)] pub use futures_util::OptionExt;` on a new line (this line is only compiled if the `std` feature is enabled).

udoprog: Ah, good catch. Thanks.

futures-util/src/option/mod.rs
```diff
@@ -0,0 +1,193 @@
...
+/// Adapter future for `Option` to get the next value of the stored future.
+///
+/// Resolves to `None` if no future is present.
+#[derive(Debug)]
+pub struct Next<'a, T> {
+    pub(crate) stream: &'a mut Option<T>,
+}
+
+impl<'a, T> Next<'a, T> {
+    fn stream<'b>(self: Pin<&'b mut Self>) -> Pin<&'b mut Option<T>> {
```
taiki-e: This seems to be getting `Pin<Field>` from `Pin<&mut Struct>`. I don't think it is safe to do this without the `Field: Unpin` bounds.

udoprog: Note that `Field` in this case is an `&mut Option<T> where T: Future`. The fact that we have hold mutable reference to the field should prevent it from moving for the duration of the projected `Pin`, no?

I'm very much unsure about this though.

udoprog: Indeed, this projection isn't OK. I've rebased and reverted the `Unpin` change. Thanks!

udoprog:
Any interest in incorporating an abstraction like this? Otherwise I'll move it to an external crate.

cramertj:
I think i'd prefer to leave this out as-is at the moment, but I'd be interested in exploring ideas for making this pattern more discoverable if you have them!

udoprog:
@cramertj discoverable? Do you mean with using `Fuse`?

cramertj:
Either using fuse or some similar mechanism, yes.

udoprog:
Published as https://crates.io/crates/futures-option to give it some time to mature (and allow me to keep using it in my projects for now). I'll keep thinking if there's a way to improve the fuse-based pattern.

cramertj:
Awesome, thanks!

PR #1636, opened 2019-05-24
Fix error in Sink documentation

jrobsonchase:
The use of "proceeded" makes it sound like the opposite of "precede," like how "proscribe" is the opposite of "prescribe," but I don't think that was the intended meaning, since checking for readiness shouldn't come *after* starting a send.


Nemo157:
Thanks!

Issue #1640, opened 2019-05-28
Document precedence of Select and SelectAll when multiple futures ready


betamos:
When `future::Select` and `future::SelectAll` have >1 futures ready, the first one will be picked (meaning the leftmost "future A" in the former, and the zero:th fut in the latter case). I'm not sure if this is a feature.

In the `select!` macro however, order is randomized at runtime (this is certainly a feature).

(There's a separate discussion to be had about the behavior, let's keep this one about the documentation).

I suggest we either document that a specific order can be assumed, or that no order can be assumed.

cramertj:
@betamos Thanks for sending #1641! Are you also interested in sending PRs for the existing `Select` and `SelectAll` combinator docs?

PR #1641, opened 2019-05-28
Document branch precedence of `select!`

betamos:
See #1640. (This is just the macro part, not Select and SelectAll)


cramertj:
Looks great, thanks!

PR #1647, opened 2019-06-03
Allow ?Sized types in some methods and structs

taiki-e:


futures-util/src/lock/mutex.rs
```diff
@@ -9,13 +9,13 @@ use std::sync::Mutex as StdMutex;
 use std::sync::atomic::{AtomicUsize, Ordering};
 
 /// A futures-aware mutex.
-pub struct Mutex<T> {
+pub struct Mutex<T: ?Sized> {
```
cramertj: I'm not sure if we want to reveal that this is allocated separately-- I had imagined at some point we might want to switch to using e.g. `parking_lot`'s `Mutex` type which prevents an allocation here. I don't feel super strongly about it, though, since `futures-util` should be okay to make breaking changes to in the future. Your choice :)

cramertj:
Nice! One question re `Mutex` but otherwise the rest of the changes LGTM.

futures-util/src/lock/mutex.rs
taiki-e: @cramertj 
I'm sorry, I'm not sure what is the problem.
* In this pr's implementation, if `T` is `!Sized`, then `Mutex` itself becomes `!Sized`.
  ```rust
  use futures::lock::Mutex;
  use std::sync::Arc;
  fn foo() -> Arc<Foo<dyn Fn()>> {
      let m = Foo::new(|| {});
      Arc::new(m) as Arc<_>
  }
  fn bar() -> Foo<dyn Fn()> { //~ ERROR the size for values of type `(dyn std::ops::Fn() + 'static)` cannot be known at compilation time
      let m = Foo::new(|| {});
      m as _
  }
  ```
* `parking_lot`'s `Mutex` allows `T: ?Sized`, and if `T` is `!Sized`, the behavior is the same as this pr's implementation: https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=a4ecc3eecc9fa5cdce41bebd5f97cf5e.

cramertj: Ah, I see-- I was thinking that the `Mutex` here was still `Sized` even if `T: ?Sized`.

PR #1648, opened 2019-06-03
Add Compat::get_ref method

jrobsonchase:
It's currently possible to unwrap the `Compat` wrapper via the `into_inner` method, but not to simply get a reference to the wrapped object as it is with the `Compat01As03` wrappers.

This adds a `get_ref` method to bring it in-line with the others.


Nemo157:
Thanks!

Issue #1650, opened 2019-06-04
Consider flatten_stream_sink method


DenisKolodin:
How about to add `flatten_stream_sink` method that allows to extract a `Stream + Sink` instance from `Future`?

Related: https://github.com/rust-lang-nursery/futures-rs/pull/965

cramertj:
I actually think this is just a matter of adding a `Stream` impl to the existing `FlattenSink` combinator, and I would be happy to accept a PR!

PR #1651, opened 2019-06-04
impl Stream/Sink for FlattenSink/TryFlattenStream

taiki-e:
This is an alternative to #1650.
I have no strong opinion on whether it is preferable to have a method for `Self::Ok: TryStream + Sink`...


Nemo157:
👍 I was going to suggest something like this as the solution for that issue.

cramertj:
Works for me!

PR #1652, opened 2019-06-04
Change WakerRef's wake function to panic

seanmonstar:
While no unsoundness was exposed, thanks to `WakerRef` only exposing a `&Waker`, this just changes its `wake` function to panic, in case some later changes ever accidentally expose the `Waker`.


seanmonstar:
Looks like rand stuff is breaking CI...

cramertj:
Nice, this is great! Will merge after a rebase (after which I think CI should be happy)

PR #1654, opened 2019-06-05
Add Send + Sync to ArcWake trait to fix unsoundness

seanmonstar:
Neither `Arc::new` nor `Arc<T>` itself *requires* `Send + Sync`, but the `into_waker` method assumes them when converting into a `Waker`. So without this change, it's possible to trigger UB with something like `Arc::new(Boom(some_rc)).into_waker()`.

### DIFF
--- a/futures-util/src/task/arc_wake.rs
+++ b/futures-util/src/task/arc_wake.rs
@@ -8,7 +8,9 @@ use alloc::sync::Arc;
 /// can be converted into `Waker` objects.
 /// Those Wakers can be used to signal executors that a task it owns
 /// is ready to be `poll`ed again.
-pub trait ArcWake {
+// Note: Send + Sync required because `Arc<T>` doesn't automatically imply
+// those bounds, but `Waker` implements them.
+pub trait ArcWake: Send + Sync {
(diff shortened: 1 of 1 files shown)


Nemo157:
I guess `Send` is required because of `Arc::try_unwrap`?

(Just to check I threw together [a playground](https://play.rust-lang.org/?version=nightly&mode=debug&edition=2018&gist=e35fbab6af1da793552fad0207bd5f16) with a full demonstration of the unsoundness, including cloning/dropping an `Rc` on multiple threads).

seanmonstar:
With an Rc, you could make some clones before putting in the Arc, and even making some clones on the other thread, making unsynchronized ref count changes.

Similar with RefCell and it's borrow count. Etcetera.

Shnatsel:
Unsynchronized reference count changes could potentially lead to exploitable memory corruption. Please file a security advisory at https://github.com/RustSec/advisory-db so that dependents of this crate could check if they're affected and upgrade.

Nemo157:
As far as I’m aware the unsoundness was only present in alpha releases of a preview crate, that should be sufficient “not production quality code” to not require a security advisory.

PR #1655, opened 2019-06-05
Document clarifying StreamExt::forward usage

dekellum:
Having just stumbled on this, coming from futures 0.1, I'd like to help the next person.

I was only able to figure this out via hints in #1441.


cramertj:
Thanks, this looks great! :)

PR #1656, opened 2019-06-06
Fix unused_must_use warning

taiki-e:


taiki-e:
Hmm... ICEs in doc tests. this is probably https://github.com/rust-lang/rust/issues/61442...

PR #1662, opened 2019-06-06
Make all extension traits unnamed in the prelude

Nemo157:
This is useful for avoiding accidental shadowing of glob imported preludes, the methods will be brought into scope but the trait itself will have no name so can't be shadowed by things like `use futures_timer::FutureExt;`.

Unfortunately `rustdoc` doesn't render these well, but hopefully that can get fixed at some point

Screenshot 2019-06-06 at 18 50 37

### DIFF
--- a/futures/src/lib.rs
+++ b/futures/src/lib.rs
@@ -326,14 +326,14 @@ pub mod prelude {
     //!
     //! The prelude may grow over time as additional items see ubiquitous use.
 
-    pub use crate::future::{self, Future, TryFuture, FutureExt, TryFutureExt};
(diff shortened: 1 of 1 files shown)


Nemo157:
Opened https://github.com/rust-lang/rust/issues/61592 to try and have nicer docs eventually.

taiki-e:
CI fail is because a new minor version of rand_core has been released. I am writing a fix for this now.

taiki-e:
Opened #1664 for fixing CI fail.

Nemo157:
Updated to use `doc(no_inline)` so it now renders like this:

cramertj:
NIce, that's a good trick! :)

PR #1658, opened 2019-06-06
Ignore doc tests which "Broken MIR" ICE occurred

taiki-e:
cc https://github.com/rust-lang-nursery/futures-rs/issues/1657

PR #1664, opened 2019-06-06
Update rand(_core)

taiki-e:
This may not be completely fixable. (It seems that some versions have already been yanked...)

### DIFF
--- a/futures-util/Cargo.toml
+++ b/futures-util/Cargo.toml
@@ -34,8 +34,8 @@ futures-sink-preview = { path = "../futures-sink", version = "=0.3.0-alpha.16", 
 futures-select-macro-preview = { path = "../futures-select-macro", version = "=0.3.0-alpha.16", default-features = false, optional = true }
 proc-macro-hack = { version = "0.5", optional = true }
 proc-macro-nested = { version = "0.1.2", optional = true }
-rand = { version = "0.6.4", optional = true }
-rand_core = { version = ">=0.2.2, <0.4", optional = true } # See https://github.com/rust-random/rand/issues/645
+rand = { version = "0.6.5", optional = true }
+rand_core = { version = "0.4", optional = true } # See https://github.com/rust-lang-nursery/futures-rs/pull/1664
 slab = { version = "0.4", optional = true }
 memchr = { version = "2.2", optional = true }
 futures_01 = { version = "0.1.25", optional = true, package = "futures" }


seanmonstar:
I don't really understand why it's recommended to use such version ranges for `rand_core`. Since it defines the traits, a program can't compile if there's multiple `rand_core` versions in the tree, as the traits don't match...

futures-util/Cargo.toml
```diff
@@ -34,8 +34,8 @@ futures-sink-preview = { path = "../futures-sink", version = "=0.3.0-alpha.16",
 futures-select-macro-preview = { path = "../futures-select-macro", version = "=0.3.0-alpha.16", default-features = false, optional = true }
 proc-macro-hack = { version = "0.5", optional = true }
 proc-macro-nested = { version = "0.1.2", optional = true }
-rand = { version = "0.6.4", optional = true }
-rand_core = { version = ">=0.2.2, <0.4", optional = true } # See https://github.com/rust-random/rand/issues/645
+rand = { version = "0.6.5", optional = true }
+rand_core = { version = ">=0.4, <0.5", optional = true } # See https://github.com/rust-random/rand/issues/645
```
Nemo157: This should be just `0.4` now.

taiki-e:
`rand` has a lot of sub crates, which are not always updated at the same time. 
And when releasing a new minor version of `rand_core`, they also publish the old version, which has been changed to re-export the new version. 
So, this may cause compilation to fail if Cargo.toml is not completely up-to-date. This can also be caused by the dependency using an older version.

futures-util/Cargo.toml
taiki-e: Yes, `0.4.1` has been yanked.

Nemo157: I just meant that `0.4` is an equivalent bound to `>=0.4, <0.5`.

Nemo157:
https://github.com/rust-random/rand/pull/819 looks like 0.5 may be yanked soon.

futures-util/Cargo.toml
taiki-e: Oh, I was thinking that somehow I wrote `">=0.4.1, <0.5"` and opened this PR.

taiki-e:
Hmm... still fail. ICE is will be avoided in #1658. 
#1658 is `out-of-date`, so I rebased it. looks like some `rand*` crates have been yanked, so it may work well without `rand` updates.

taiki-e:
#1658 works well.

taiki-e:
Postpone this PR until there is something upstream.

PR #1667, opened 2019-06-09
Add io::Window::set

taiki-e:
And removed `set_start` and `set_end`.

### DIFF
--- a/futures-util/src/io/window.rs
+++ b/futures-util/src/io/window.rs
@@ -1,4 +1,4 @@
-use std::ops;
+use std::ops::{Bound, Range, RangeBounds};
 
 /// A owned window around an underlying buffer.
 ///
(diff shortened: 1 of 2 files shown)


futures-util/src/io/window.rs
```diff
@@ -65,43 +65,36 @@ impl<T: AsRef<[u8]>> Window<T> {
         self.range.end
     }
 
-    /// Changes the starting index of this window to the index specified.
+    /// Changes the range of this window to the range specified.
     ///
     /// Returns the windows back to chain multiple calls to this method.
```
cramertj: not new to this change, but why would you want to chain multiple calls to this method?

taiki-e: That's certainly not necessary because we can now set the range at once with `set`.

PR #645, opened 2017-11-16
Tokio reform changes

carllerche:
**Work in progress** do not merge.

This PR represents the changes initially discussed in the [Tokio reform RFC](https://github.com/tokio-rs/tokio-rfcs/pull/3). Changes are not directly applied here but go through PRs that then get merged into this one. This enables granular changes to be reviewed / discussed in a more focused context.

## Progress

* [x] Deprecate `Future::wait` (#634)
* [x] Current thread executor (#639)
* [x] `executor::Enter` (#647)

## Unanswered questions

* `execute` vs. `spawn` naming ([original discussion](https://github.com/alexcrichton/futures-rs/pull/639#issuecomment-344760969)).

Some believe that `spawn` represents better terminology for submitting a future to an executor. However, the library already contains an [`Executor`](https://docs.rs/futures/0.1.17/futures/executor/trait.Executor.html) trait that provides an `execute` function. Using disjoint terminology is not ideal either.

PR #1668, opened 2019-06-10
Park trait, ParkThread implementation and using it in LocalPool

stbuehler:
Hi,

I think it would be good to have a `Park` abstration (like the one in tokio) in `futures-executor` (this has already been proposed and merged in #665 for the tokio-reform branch, which has been abonded in #645).

In #665 `Park` was renamed to `Sleep`, but I think `Park` is the better choice:
- `std` lib provides [`thread::park`](https://doc.rust-lang.org/std/thread/fn.park.html) for a sleeping mechanism that can be cancelled ("unparked").
- tokio users already know it by that name
- it is not just a "sleep" mechanism due to the "wakeup" part

Apart from the names `Sleep` from #665 is basically the same as tokio `Park`; further changes to the API (compared to tokio):
- merge `Park::park_timeout` into `Park::park`, now taking a `ParkDuration` enum: most implementations I've seen used shared code anyway.
- `Park::park` now also includes a `&mut Enter` argument
- Removed `Unpark` trait, `Park` no longer has an associated `Unpark` type.  Using `futures_util::task::WakerRef` instead for the result of the `waker` function (renamed from `unpark`).

I think reusing the `Waker` abstraction is rather obvious; but there are probably two open questions:
- Use `Waker` instead of `WakerRef`?
- Remove `ParkDuration::Poll` and interpret a zero duration in `ParkDuration::Duration(..)` as "poll only but don't yield"?

This pull request also contains a `ParkThread` implementation of `Park` (based on `std::thread::park` and similar to what `LocalPool` was using) and supports using `LocalPool` with custom `Park` implentations (defaulting to `ParkThread`).  This way `LocalPool` can be combined into a "local-thread runtime" with a timer and a reactor.

References:
- [`tokio_executor::park`](https://docs.rs/tokio-executor/0.1.7/tokio_executor/park/index.html)
- [`std::thread::park`](https://doc.rust-lang.org/std/thread/fn.park.html)

### DIFF
--- a/futures-executor/src/lib.rs
+++ b/futures-executor/src/lib.rs
@@ -73,3 +73,8 @@ pub use crate::thread_pool::{ThreadPool, ThreadPoolBuilder};
 mod enter;
 #[cfg(feature = "std")]
 pub use crate::enter::{Enter, EnterError, enter};
+
+#[cfg(feature = "std")]
+mod park;
+#[cfg(feature = "std")]
+pub use crate::park::{Park, ParkDuration, ParkPoll};
--- a/futures-executor/src/lib.rs
+++ b/futures-executor/src/lib.rs
@@ -55,7 +55,9 @@ extern crate std;
 #[cfg(feature = "std")]
 mod local_pool;
 #[cfg(feature = "std")]
-pub use crate::local_pool::{BlockingStream, LocalPool, LocalSpawner, block_on, block_on_stream};
+pub use crate::local_pool::{
+    BlockingStream, LocalPool, LocalPoolGen, LocalSpawner, block_on, block_on_stream,
+};
 
 #[cfg(feature = "thread-pool")]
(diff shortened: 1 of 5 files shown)


stbuehler:
Just rebased to current master.  Is there anything I can do to help moving this (and maybe other pull requests) forward?

stbuehler:
So, I basically rewrote this PR - I still (7 years later :) ) think this is would be an interesting and useful addition to the `LocalPool` API.

I also added a separate `ParkPoll` trait to handle the non-blocking additions from #1437 / #1537 - I don't really like `try_run_one` and `run_until_stalled`, as you basically need to busy-loop poll them (and `ParkPoll` obviously has the same problem). (Perhaps whatever they were used for could actually be fixed by some `Park` implementation.)

Generic default params are not working well, so I renamed the actual type to `LocalPoolGen` and aliased it with the default `ParkThread` to `LocalPool`.

PR #1671, opened 2019-06-12
Provide global back-pressure on mpsc::Sender::send

0-o-0:
Previously the bounded mpsc channels don't provide back-pressure for every new cloned sender where the capacity check is bypassed. It's an issue when use it with combinators, like `buffer_unordered` and causes OOM eventually.
There are great discussions in https://github.com/rust-lang-nursery/futures-rs/pull/984 about global vs local limits and more. From what I can see, it seems we all agree on the solution to call `poll_ready` in `poll_flush` so that it does the capacity check in every `send` call. The fix got merged into 0.1 in https://github.com/rust-lang-nursery/futures-rs/commit/4b0e21785a567b40cdf6b8c083625c88626b2a5d and merged into 0.2 in https://github.com/rust-lang-nursery/futures-rs/commit/62827f1776e833a948cb81582988cb7a70a8b700. But the latest 0.3 still has the issue. This PR is to fix that and also add tests to cover the back-pressure cases.

### DIFF
--- a/futures-channel/benches/sync_mpsc.rs
+++ b/futures-channel/benches/sync_mpsc.rs
@@ -93,7 +93,7 @@ impl Stream for TestSender {
         ready!(tx.as_mut().poll_ready(cx)).unwrap();
         tx.as_mut().start_send(this.last + 1).unwrap();
         this.last += 1;
-        assert_eq!(Poll::Ready(Ok(())), tx.as_mut().poll_flush(cx));
+        assert_eq!(Poll::Pending, tx.as_mut().poll_flush(cx));
         Poll::Ready(Some(this.last))
     }
 }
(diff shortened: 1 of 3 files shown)


cramertj:
Thanks for the backport! Sorry this got missed.

0-o-0:
@cramertj I followed what you did in https://github.com/rust-lang-nursery/futures-rs/commit/62827f1776e833a948cb81582988cb7a70a8b700, but I'm not very confident about the disconnection check in `poll_flush` because it might hide some real issues. What do you think?

cramertj:
That makes sense to me! The only alternative I see would be to return an error.

kurnevsky:
Should the [documentstion](https://docs.rs/futures/0.3.17/futures/channel/mpsc/fn.channel.html) about channel's capacity be fixed or it's still correct?

PR #1672, opened 2019-06-13
Deny warnings for doc tests

taiki-e:
We cleaned this up with some PRs (#1590, #1381, etc.), ~~so we are not warned at the moment~~, but it will be useful to maintain the quality of the examples. (EDIT: I missed `unused` being allowed by default.)

It would have been nice if this was supported as a `cargo doc` or `clippy` option (maybe like `cargo doc --test -- --Dwarnings`?), but it does not seem to be supported yet (https://github.com/rust-lang/rust/issues/56232).

### DIFF
--- a/futures-util/src/future/mod.rs
+++ b/futures-util/src/future/mod.rs
@@ -397,9 +397,7 @@ pub trait FutureExt: Future {
     ///
     /// # Examples
     ///
-    // TODO: minimize and open rust-lang/rust ticket, currently errors:
-    //       'assertion failed: !value.has_escaping_regions()'
-    /// ```ignore
+    /// ```
     /// #![feature(async_await)]
     /// # futures::executor::block_on(async {
     /// use futures::future::{self, FutureExt, Ready};
--- a/futures-channel/src/lib.rs
+++ b/futures-channel/src/lib.rs
@@ -11,6 +11,8 @@
 #![warn(missing_docs, missing_debug_implementations, rust_2018_idioms, unreachable_pub)]
 #![warn(clippy::all)]
 
+#![doc(test(attr(deny(warnings), allow(dead_code, unused_assignments, unused_variables))))]
+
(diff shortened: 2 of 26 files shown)


futures-util/src/future/mod.rs
```diff
@@ -397,9 +405,8 @@ pub trait FutureExt: Future {
     ///
     /// # Examples
     ///
-    // TODO: minimize and open rust-lang/rust ticket, currently errors:
-    //       'assertion failed: !value.has_escaping_regions()'
```
taiki-e: Also, this seems to have already been fixed.

cramertj:
> It would have been nice if this was supported as a cargo doc or clippy option (maybe like cargo doc --test -- --Dwarnings?)

Could we maybe bypass it by setting `RUSTFLAGS` directly?

taiki-e:
> Could we maybe bypass it by setting `RUSTFLAGS` directly?

I'll try it.

Nemo157:
I found this while looking through the rustdoc source trying to figure out why `RUSTFLAGS` doesn't actually apply ([docs here](https://doc.rust-lang.org/nightly/rustdoc/the-doc-attribute.html#testattr)):

```rust
#![doc(test(attr(deny(warnings))))]
```

adding that at the crate root will inject the contained attribute into all tests automatically.

taiki-e:
@Nemo157 Thanks for finding it out.

> ```rust
> #![doc(test(attr(deny(warnings))))]
> ```

It seems that it also denies the lints that rustdoc [allows by default](https://doc.rust-lang.org/nightly/rustdoc/documentation-tests.html#pre-processing-examples), so I feel that it is better to allow `dead_code` and so on. Probably like:

```rust
#![doc(test(attr(deny(warnings), allow(dead_code, unused_assignments, unused_variables))))]
```

I also agree that it is preferable to deny everywhere until it's possible to run the lints against them independently.

cramertj:
Nice! Thanks to both of you for the investigation and the fixes.

PR #1673, opened 2019-06-14
 Add AsyncReadTestExt::limited 

Thomasdezeeuw:
Also merges `InterleavePending` and `InterleavePendingWrite` to reduce code duplication. And renamed `LimitedWrite` to `Limited` and implemented `AsyncRead` on it for `AsyncReadTestExt::limited `.


futures-test/src/io/interleave_pending.rs
```diff
@@ -0,0 +1,161 @@
...
+};
+
+/// I/O wrapper for interleaving [`Poll::Pending`] in calls to read or write.
+///
+/// See the [`interleave_pending`] and [`interleave_pending_write`] methods.
+///
+/// [`interleave_pending`]: super::AsyncReadTestExt::interleave_pending
+/// [`interleave_pending_write`]: super::AsyncWriteTestExt::interleave_pending_write
+#[derive(Debug)]
+pub struct InterleavePending<IO> {
```
Nemo157: I'm not sure what the standard is for 2 character acronyms in Rust's version of PascalCase/UpperCamelCase. As far as I know there's no special-case for them so they should follow the standard acronym rule of just treating it as a single word, which would make this `Io`. (If I remember my .NET conventions correctly they had a special case for < 3 character acronyms to be kept uppercase like this).

Thomasdezeeuw: Done.

futures-test/src/io/interleave_pending.rs
```diff
@@ -0,0 +1,161 @@
...
+};
+
+/// I/O wrapper for interleaving [`Poll::Pending`] in calls to read or write.
+///
+/// See the [`interleave_pending`] and [`interleave_pending_write`] methods.
+///
+/// [`interleave_pending`]: super::AsyncReadTestExt::interleave_pending
+/// [`interleave_pending_write`]: super::AsyncWriteTestExt::interleave_pending_write
+#[derive(Debug)]
+pub struct InterleavePending<Io> {
```
cramertj: Maybe a silly question: can we also use this for `Future`s / `Stream`s / `Sink`s?

Thomasdezeeuw: Sure, want me to that in this pr?

cramertj: That would be great, thanks! :)

cramertj:
Great, thanks!

PR #1674, opened 2019-06-14
Implement AsyncBufReadExt::copy_buf_into

### LINKED ISSUE #1659: Move copy_into from AsyncRead to AsyncBufRead
Currently if you want to copy a buffered reader into a writer the easiest way is `reader.copy_into(writer)`, this won't use the buffered readers buffer and will instead use an internal buffer in the `CopyInto` future.

This can be ok for some buffered readers like the provided `BufReader` as they will bypass their internal buffer when doing large reads (although this assumes that `CopyInto` will use at least the same buffersize as the `BufReader`), but for other buffered readers like an `IntoAsyncRead` wrapping a `Stream<Item = Vec<u8>>` there will be definite performance loss from extra copying required.

We could provide a similar function on `AsyncBufRead` and keep the existing one, but I think it would be better to just move it to `AsyncBufRead` and require users to wrap their reader in a `BufReader` if they want to copy from a non-buffered reader (which will also allow them to choose the buffer size for their application if they want).

Nemo157:
closes #1659

### DIFF
--- a/futures-util/src/io/copy_into.rs
+++ b/futures-util/src/io/copy_into.rs
@@ -9,13 +9,13 @@ use pin_utils::unsafe_pinned;
 /// Future for the [`copy_into`](super::AsyncReadExt::copy_into) method.
 #[derive(Debug)]
 #[must_use = "futures do nothing unless you `.await` or poll them"]
-pub struct CopyInto<'a, R: AsyncRead, W> {
+pub struct CopyInto<'a, R: AsyncRead, W: ?Sized> {
     inner: CopyBufInto<'a, BufReader<R>, W>,
 }
 
-impl<'a, R: AsyncRead, W> Unpin for CopyInto<'a, R, W> where CopyBufInto<'a, BufReader<R>, W>: Unpin {}
+impl<'a, R: AsyncRead, W: ?Sized> Unpin for CopyInto<'a, R, W> where CopyBufInto<'a, BufReader<R>, W>: Unpin {}
 
-impl<'a, R: AsyncRead, W> CopyInto<'a, R, W> {
+impl<'a, R: AsyncRead, W: ?Sized> CopyInto<'a, R, W> {
     unsafe_pinned!(inner: CopyBufInto<'a, BufReader<R>, W>);
 
     pub(super) fn new(reader: R, writer: &mut W) -> CopyInto<'_, R, W> {
@@ -25,7 +25,7 @@ impl<'a, R: AsyncRead, W> CopyInto<'a, R, W> {
     }
 }
 
-impl<R: AsyncRead, W: AsyncWrite + Unpin> Future for CopyInto<'_, R, W> {
+impl<R: AsyncRead, W: AsyncWrite + Unpin + ?Sized> Future for CopyInto<'_, R, W> {
     type Output = io::Result<u64>;
 
     fn poll(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
--- a/futures-util/src/io/copy_into.rs
+++ b/futures-util/src/io/copy_into.rs
@@ -3,75 +3,32 @@ use futures_core::task::{Context, Poll};
 use futures_io::{AsyncRead, AsyncWrite};
 use std::io;
 use std::pin::Pin;
+use super::{BufReader, CopyBufInto};
+use pin_utils::unsafe_pinned;
 
 /// Future for the [`copy_into`](super::AsyncReadExt::copy_into) method.
 #[derive(Debug)]
 #[must_use = "futures do nothing unless you `.await` or poll them"]
-pub struct CopyInto<'a, R: ?Sized + Unpin, W: ?Sized + Unpin> {
-    reader: &'a mut R,
-    read_done: bool,
-    writer: &'a mut W,
-    pos: usize,
-    cap: usize,
-    amt: u64,
-    buf: Box<[u8]>,
+pub struct CopyInto<R: AsyncRead, W> {
+    inner: CopyBufInto<BufReader<R>, W>,
 }
 
-impl<R: ?Sized + Unpin, W: ?Sized + Unpin> Unpin for CopyInto<'_, R, W> {}
+impl<R: AsyncRead, W> Unpin for CopyInto<R, W> where CopyBufInto<BufReader<R>, W>: Unpin {}
 
-impl<'a, R: ?Sized + Unpin, W: ?Sized + Unpin> CopyInto<'a, R, W> {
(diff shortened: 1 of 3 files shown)


futures-util/src/io/mod.rs
```diff
@@ -104,12 +107,7 @@ pub trait AsyncReadExt: AsyncRead {
     /// assert_eq!(writer.into_inner(), [1, 2, 3, 4, 0]);
     /// # Ok::<(), Box<std::error::Error>>(()) }).unwrap();
     /// ```
-    fn copy_into<'a, W>(
-        &'a mut self,
-        writer: &'a mut W,
-    ) -> CopyInto<'a, Self, W>
-        where Self: Unpin, W: AsyncWrite + Unpin,
-    {
+    fn copy_into<W: AsyncWrite>(self, writer: W) -> CopyInto<Self, W> where Self: Sized {
```
Nemo157: To be able to re-use `CopyBufInto` as the implementation of `CopyInto` I had to make that take the reader and writer by value, it seemed bad to have the two methods be inconsistent so I changed it here as well.

It seems like reusing the reader after copying out of it is unlikely to be a common case, and the only overhead to do so is you will have to `(&mut reader).copy_into(&mut writer).await?;`, being able to keep the writer accessible after copying doesn't change at all as `&mut AsyncWrite + Unpin: AsyncWrite`.

futures-util/src/io/copy_into.rs
```diff
@@ -3,75 +3,32 @@ use futures_core::task::{Context, Poll};
...
+    pub(super) fn new(reader: R, writer: W) -> Self {
         CopyInto {
-            reader,
-            read_done: false,
-            writer,
-            amt: 0,
-            pos: 0,
-            cap: 0,
-            buf: Box::new([0; 2048]),
+            inner: CopyBufInto::new(BufReader::with_capacity(2048, reader), writer),
```
Nemo157: I used the same buffer size here as what the old code used, I have no idea why this isn't just using `DEFAULT_BUF_SIZE` though...

futures-util/src/io/mod.rs
cramertj: can you clarify why you had to take it by-value in order to do this? It's somewhat surprising to me that the writer would be by-value (the reader makes decent enough sense, though), especially since we don't `poll_close`.

Nemo157: Taking the reader by-ref in both cases would look something like:

```rust
impl<R: AsyncBufRead + Unpin, W: AsyncWrite + Unpin> CopyBufInto<'_, R, W> {
    fn new(reader: &mut R, writer: &mut W) -> Self { ... }
}

struct CopyInto<'a, R, W> {
    inner: CopyBufInto<'a, R, W>,
}

impl<R: AsyncRead + Unpin, W: AsyncWrite + Unpin> CopyInto<'_, R, W> {
    fn new(reader: &mut R, writer: &mut W) -> Self {
        Self { inner: CopyBufInto::new(&mut BufReader::new(reader), writer) }
    }
}
```

we want to wrap the reader in a `BufReader` to allow the reuse, but then we don't have anywhere with ownership of that `BufReader` when we pass the reference into `CopyBufInto::new` (unless we make `CopyInto` a self-referential struct and only create the inner `CopyBufInto` when we're first polled, but I'd prefer to not deal with manually implemented self-referential structs).

> It's somewhat surprising to me that the writer would be by-value

Yeah, I could change the writer back to being by-ref since we don't need to wrap that, but because of `impl AsyncWrite for &mut impl AsyncWrite + Unpin` you can just pass it by reference anyway (you can see that the example didn't have to change: https://github.com/rust-lang-nursery/futures-rs/pull/1674/files#diff-58a0189b65158d401601a19cc9f17257R104)

taiki-e: > > It's somewhat surprising to me that the writer would be by-value
> 
> Yeah, I could change the writer back to being by-ref since we don't need to wrap that, but because of `impl AsyncWrite for &mut impl AsyncWrite + Unpin` you can just pass it by reference anyway

It may be nice to have an explanation like "Note that this method consumes writer but does not close it.".

futures-util/src/io/copy_into.rs
taiki-e: I don't know why it is used in old code, I think it should be preferable to use `DEFAULT_BUF_SIZE` [because `std::io::copy` uses `DEFAULT_BUF_SIZE`](https://github.com/rust-lang/rust/blob/00859e3e653973120006aaf3227823062dde1ba7/src/libstd/io/util.rs#L47) (I probably missed it [when defining `DEFAULT_BUF_SIZE`](https://github.com/rust-lang-nursery/futures-rs/pull/1573)).

taiki-e: > I don't know why `2048` is used in old code

It may be related to `DEFAULT_BUF_SIZE` was once large: https://github.com/rust-lang/rust/pull/32695

futures-util/src/io/mod.rs
Nemo157: Added a note.

futures-util/src/io/copy_into.rs
Nemo157: Changed it to just use the default.

futures-util/src/io/mod.rs
cramertj: I'm still skeptical that this is the correct route here-- it seems to me like passing in a non-`Pin<&mut T>` or `&mut T` `AsyncWrite` by-value is basically always going to be a bug. I don't think it's worth changing the API for the sake of consistency with another when the usage should always be different.

Nemo157: :+1: I'll change it back to taking the writer by reference then.

cramertj: Thanks! Sorry for the confusing back-and-forth.

futures-util/src/io/mod.rs
```diff
@@ -441,6 +444,44 @@ impl<S: AsyncSeek + ?Sized> AsyncSeekExt for S {}
...
+    /// Creates a future which copies all the bytes from one object to another.
+    ///
+    /// The returned future will copy all the bytes read from this `AsyncBufRead` into the
+    /// `writer` specified. This future will only complete once the `reader` has hit
+    /// EOF and all bytes have been written to and flushed from the `writer`
+    /// provided.
+    ///
+    /// On success the number of bytes is returned.
+    ///
+    /// Note that this method consumes `writer` but does not close it, you will likely want to pass
```
cramertj: This comment is now outdated since the method no longer consumes `writer`

PR #1675, opened 2019-06-14
Add AsyncWriteExt::into_sink

### LINKED ISSUE #1660: Add AsyncWriteExt::into_sink
We can currently convert an `impl Stream<Item = io::Result<impl AsRef<[u8]>>>` into an `impl AsyncRead + AsyncBufRead` via `TryStreamExt::into_async_read`. It could also be useful to be able to do the inverse conversion at the other end, converting an `impl AsyncWrite` into an `impl Sink<impl AsRef<[u8]>, SinkError = io::Error>` to improve compatibility between these two different IO abstraction pairs.

Nemo157:
closes #1660

### DIFF
--- a/futures-util/src/io/into_sink.rs
+++ b/futures-util/src/io/into_sink.rs
@@ -1,38 +1,38 @@
 use futures_core::task::{Context, Poll};
 use futures_io::AsyncWrite;
 use futures_sink::Sink;
-use std::fmt;
 use std::io;
 use std::pin::Pin;
 use std::marker::Unpin;
 use pin_utils::{unsafe_pinned, unsafe_unpinned};
 
-struct Block {
+#[derive(Debug)]
+struct Block<Item> {
     offset: usize,
-    bytes: Box<dyn AsRef<[u8]>>,
+    bytes: Item,
 }
 
 /// Sink for the [`into_sink`](super::AsyncWriteExt::into_sink) method.
 #[must_use = "sinks do nothing unless polled"]
 #[derive(Debug)]
-pub struct IntoSink<W> {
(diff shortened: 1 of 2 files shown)


futures-util/src/io/into_sink.rs
```diff
@@ -0,0 +1,114 @@
...
+use futures_sink::Sink;
+use std::fmt;
+use std::io;
+use std::pin::Pin;
+use std::marker::Unpin;
+use pin_utils::{unsafe_pinned, unsafe_unpinned};
+
+struct Block {
+    offset: usize,
+    bytes: Box<dyn AsRef<[u8]>>,
```
cramertj: I'd make `Block` and `IntoSink` generic on the kind of `Bytes: AsRef<[u8]>` they accept. This would allow folks to send in e.g. references to buffers, as sending vectors without double-allocating them. If folks want to send multiple different kinds of `Bytes`, they can set `Box<dyn AsRef[u8]>` as the generic parameter.

Note that this would allow dropping the `'static` bounds below.

cramertj:
LGTM aside from switching to a type parameter to allow the caller to specify the type of the `SinkItem`.

futures-util/src/io/into_sink.rs
Nemo157: Done, it doesn't actually allow using `Box<dyn AsRef<[u8]>>` as there is `impl<T: ?Sized> AsRef<T> for Box<T>` instead of `impl<U, T: AsRef<U> + ?Sized> AsRef<U> for Box<T>`, but it should be possible to make a custom newtyped-`Box` with that impl if someone does need that behaviour.

PR #1676, opened 2019-06-15
Change ArcWake::into_waker to a free function

taiki-e:
I feel this to be more consistent.

```rust
pub fn waker<W>(wake: Arc<W>) -> Waker where W: ArcWake;

pub fn waker_ref<W>(wake: &Arc<W>) -> WakerRef<'_> where W: ArcWake;
```

### DIFF
--- a/futures-test/src/task/wake_counter.rs
+++ b/futures-test/src/task/wake_counter.rs
@@ -1,7 +1,7 @@
-use futures_core::task::{Waker};
+use futures_core::task::Waker;
+use futures_util::task::{self, ArcWake};
 use std::sync::Arc;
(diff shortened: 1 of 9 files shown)


cramertj:
Part of the reason for doing it this way was to make the method more discoverable. I don't care particularly strongly about this change either way, but I think we need some extra documentation on the trait if we're going to make this a free function.

taiki-e:
@cramertj Sorry for the late reply. I added documents for `ArcWake` -> `Waker` conversion in 1a3e8a4.

cramertj:
Works for me, then!

PR #1680, opened 2019-06-19
Use #![warn(single_use_lifetimes)]

taiki-e:
I tried to do this at the time of #1531, but I couldn't do it because of https://github.com/rust-lang/rust/issues/53738 (It needed too much `#[allow(single_use_lifetimes)]`...).


taiki-e:
Hmm... this causes many warnings in older versions: https://travis-ci.com/rust-lang-nursery/futures-rs/jobs/209181035

futures-channel/src/lib.rs
```diff
@@ -9,6 +9,8 @@
 #![cfg_attr(not(feature = "std"), no_std)]
 
 #![warn(missing_docs, missing_debug_implementations, rust_2018_idioms, unreachable_pub)]
+// It cannot be included in the published code because this lints have false positives in the minimum required version.
+#![cfg_attr(test, warn(single_use_lifetimes))]
```
taiki-e: This enables `single_use_lifetimes` lint when running `cargo test`, `cargo check --tests` etc. (FWIW, I often run `cargo check --all-features --tests --all` locally before opening PR).

taiki-e:
There are still warnings at `cargo +nightly test (minimum required version)` and `cargo test (with minimal versions)`: https://travis-ci.com/rust-lang-nursery/futures-rs/builds/116074146

taiki-e:
> `cargo +nightly test (minimum required version)` and `cargo test (with minimal versions)`

Given the reason for checking them, `cargo build` is enough, IMO.

futures-select-macro/Cargo.toml
```diff
@@ -22,4 +22,4 @@ std = []
 proc-macro2 = "0.4"
 proc-macro-hack = "0.5.3"
 quote = "0.6"
-syn = { version = "0.15.22", features = ["full"] }
+syn = { version = "0.15.25", features = ["full"] }
```
cramertj: Why was this change required?

taiki-e: A warning occurred when building with minimal versions: https://travis-ci.com/rust-lang-nursery/futures-rs/jobs/209211855. (that warning was fixed in https://github.com/dtolnay/syn/pull/563)

(Context: After the first commit, I realized that [CI did not catch(and reject) the warnings](https://github.com/rust-lang-nursery/futures-rs/pull/1680#issuecomment-503437829), so I added eb4cbfa and changed the warnings to be denied for all builds.)

PR #1681, opened 2019-06-19
Replace empty enums with Never type

taiki-e:
We have two empty enums. It should be preferable to unify this.

### DIFF
--- a/futures-core/src/never.rs
+++ b/futures-core/src/never.rs
@@ -1,0 +1,17 @@
+//! Definition of the `Never` type,
+//! a stand-in for the `!` type until it becomes stable.
+
+/// A type with no possible values.
+///
+/// This is used to indicate values which can never be created, such as the
+/// error type of infallible futures.
+///
+/// This type is a stable equivalent to the `!` type from `std`.
+///
+/// This is currently an alias for [`std::convert::Infallible`], but in
+/// the future it may be an alias for [`!`][never].
+/// See ["Future compatibility" section of `std::convert::Infallible`][infallible] for more.
+///
+/// [never]: https://doc.rust-lang.org/nightly/std/primitive.never.html
+/// [infallible]: https://doc.rust-lang.org/nightly/std/convert/enum.Infallible.html#future-compatibility
+pub type Never = core::convert::Infallible;
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,7 @@
 language: rust
 sudo: false
-cache: cargo
+# TODO: https://github.com/rust-lang-nursery/futures-rs/pull/1681#issuecomment-507791279
+# cache: cargo
 
 stages:
   - name: test
--- a/futures-core/src/lib.rs
+++ b/futures-core/src/lib.rs
@@ -29,6 +29,9 @@ pub mod stream;
 pub mod task;
 #[doc(hidden)] pub use self::task::Poll;
 
+pub mod never;
+#[doc(hidden)] pub use self::never::Never;
+
 #[doc(hidden)]
 pub mod core_reexport {
     #[doc(hidden)]
--- a/futures-sink/src/lib.rs
+++ b/futures-sink/src/lib.rs
@@ -165,14 +165,10 @@ mod channel_impls;
 #[cfg(feature = "alloc")]
 mod if_alloc {
     use super::*;
+    use futures_core::never::Never;
 
-    /// The error type for `Vec` and `VecDequeue` when used as `Sink`s.
-    /// Values of this type can never be created.
-    #[derive(Copy, Clone, Debug)]
-    pub enum VecSinkError {}
-
-    impl<T> Sink<T> for ::alloc::vec::Vec<T> {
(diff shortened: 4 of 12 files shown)


taiki-e:
Depending on the state of https://github.com/rust-lang/rust/issues/57012, I think we can use `Never` instead of real `!` type in [`FutureExt::never_error`](https://github.com/rust-lang-nursery/futures-rs/pull/1492). 

cc @Nemo157 Any thoughts?

Nemo157:
https://github.com/rust-lang/rust/issues/58733 is tempting...

futures-core/src/never.rs
```diff
@@ -0,0 +1,29 @@
...
+/// This is used to indicate values which can never be created, such as the
+/// error type of infallible futures.
+///
+/// This type is a stable equivalent to the `!` type from `std`.
+#[derive(Debug, Copy, Clone, Eq, PartialEq, Ord, PartialOrd)]
+pub enum Never {}
+
+impl Never {
+    /// Convert the `Never` type into any other type.
+    pub fn into_any<T>(self) -> T {
```
Nemo157: Having this public function would block changing to `type Never = !;` in the future, similar to [`std::convert::Infallible`](https://doc.rust-lang.org/nightly/std/convert/enum.Infallible.html#future-compatibility) I think it would make sense to plan to do this once `!` is stabilized.

taiki-e: That makes sense. I'll remove this function.

taiki-e:
> [rust-lang/rust#58733](https://github.com/rust-lang/rust/issues/58733) is tempting...

It is certainly tempting but feels potentially unstable because it seems that some versions cannot compile (although they are not versions supported by this crate...).

---

I was thinking of copying "Future compatibility" section of `std::convart::Infallible`'s document, but in the first place, we can use `Infallible` instead of having our own `Never` enum here.

Use `Infallible` directly or `type Never = core::convart::Infallible` may be better, considering that compatibility is taken into account when never type stabilizes.
Then when `Infallible` becomes `type Infallible = !`, we can switch to `#[cfg(never_type_stabilized)] type Never = !`.

taiki-e:
> Use `Infallible` directly or `type Never = core::convert::Infallible` may be better, considering that compatibility is taken into account when never type stabilizes.
> Then when `Infallible` becomes `type Infallible = !`, we can switch to `#[cfg(never_type_stabilized)] type Never = !`.

Currently this PR does this.

taiki-e:
😕 This seems to happen frequently today.
* https://travis-ci.com/rust-lang-nursery/futures-rs/jobs/212712832
* https://travis-ci.com/rust-lang-nursery/futures-rs/jobs/212712842
* https://github.com/rust-lang-nursery/futures-rs/pull/1705#issuecomment-507605324
```text
No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.
Check the details on how to adjust your build configuration on: https://docs.travis-ci.com/user/common-build-problems/#Build-times-out-because-no-output-was-received
```

taiki-e:
It seems that the total build time has been greatly reduced with a943d65.
before(master): 2 hrs 42 min 44 sec (https://travis-ci.com/rust-lang-nursery/futures-rs/builds/117603909)
after(this pr): 43 min 9 sec (https://travis-ci.com/rust-lang-nursery/futures-rs/builds/117710514)

futures-core/src/never.rs
```diff
@@ -0,0 +1,17 @@
...
+///
+/// This type is a stable equivalent to the `!` type from `std`.
+///
+/// This is currently an alias for [`std::convert::Infallible`], but in
+/// the future it may be an alias for [`!`][never].
+/// See ["Future compatibility" section of `std::convert::Infallible`][infallible] for more.
+///
+/// [never]: https://doc.rust-lang.org/nightly/std/primitive.never.html
+/// [infallible]: https://doc.rust-lang.org/nightly/std/convert/enum.Infallible.html#future-compatibility
+pub type Never = core::convert::Infallible;
```
LucioFranco: would it make sense to impl `std::error::Error` for this so that it can be used as a Never error type? I know so far we've used a bunch of bounds in tower like so `S::Error: Into<Box<std::error::Error>>` and this would make it compat with that.

taiki-e: @LucioFranco `Infallible` does not implement `std::error::Error`, but `Into<Box<std::error::Error>>` you mentioned seems to work. https://play.rust-lang.org/?version=nightly&mode=debug&edition=2018&gist=cda064b0d4459549673d30c79d70c886

LucioFranco: Awesome, thank you!

taiki-e: `Infallible` seems to implement `std::error::Error` via `std::string::ParseError`.
* [playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=da5cc6914193a2d97223c5b77590db87)
* [`std::string::ParseError`](https://doc.rust-lang.org/nightly/std/string/type.ParseError.html)

PR #1689, opened 2019-06-27
feat: Rename `future::empty` to `pending` and add `stream::pending`

### LINKED ISSUE #1624: AsyncRead::poll_read() does not expose the EOF concept
As of 85162dc, `poll_read()` is defined as:

```rust
fn poll_read(self: Pin<&mut Self>, cx: &mut Context<'_>, buf: &mut [u8])
    -> Poll<Result<usize>>;
```

In tokio-io (which seems to be the canonical definition for futures-0.1) it is defined as:

```rust
fn poll_read(&mut self, buf: &mut [u8]) -> Poll<usize, std_io::Error>;
```

In addition, the tokio-io documentation mentions:

```rust
/// * `Ok(Async::Ready(n))` means that `n` bytes of data was immediately read
///   and placed into the output buffer, where `n` == 0 implies that EOF has
///   been reached.
```

I don't much like the sentinel value of `0` for EOF, and I'm confused by the notion of EOF not being reflected in the futures-0.3 version of `AsyncRead` -- is that a conscious decision?

It seems to me that EOF is a real thing for many kinds of readers (both files and also network streams like TCP or QUIC streams). I was thinking that it might make sense to define the return type as `Poll<Result<Option<usize>>>` instead, so that there is the analogy with a `Stream` and a clearly defined marker for end of stream.

Marwes:
Fixes the inconsistency between `future::empty` and `stream::empty`.
I went with `pending` over `never` since there are already a future
called `Never` in the library.

Don't know where to put tests for this as the one place that tested `empty` seems to be disabled?

Closes #1624

### DIFF
--- a/futures-util/src/future/pending.rs
+++ b/futures-util/src/future/pending.rs
@@ -1,17 +1,19 @@
 use core::marker;
 use core::pin::Pin;
-use futures_core::future::{Future, FusedFuture};
+use futures_core::future::{FusedFuture, Future};
 use futures_core::task::{Context, Poll};
 
-/// Future for the [`empty`] function.
+/// Future for the [`pending`] function.
 #[derive(Debug)]
 #[must_use = "futures do nothing unless you `.await` or poll them"]
-pub struct Empty<T> {
+pub struct Pending<T> {
     _data: marker::PhantomData<T>,
 }
 
-impl<T> FusedFuture for Empty<T> {
-    fn is_terminated(&self) -> bool { false }
+impl<T> FusedFuture for Pending<T> {
+    fn is_terminated(&self) -> bool {
+        false
+    }
 }
 
 /// Creates a future which never resolves, representing a computation that never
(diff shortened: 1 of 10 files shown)


futures-util/src/stream/pending.rs
```diff
@@ -0,0 +1,29 @@
+use core::marker;
```
taiki-e: This module does not seem to be defined...

futures-util/src/stream/pending.rs
```diff
@@ -0,0 +1,29 @@
...
+use core::pin::Pin;
+
+use futures_core::{Stream, Poll};
+use futures_core::task;
+
+/// A stream which never returns any elements.
+///
+/// This stream can be created with the `stream::pending` function.
```
taiki-e:
```rust
/// A stream which never returns any elements.
///
/// This stream can be created with the `stream::pending` function.
```

nit: Could you change above doc-comment as follows? (because doc-comment of `pending` function gives a more detailed explanation. see also https://github.com/rust-lang-nursery/futures-rs/issues/1214)

```rust
/// Stream for the [`pending()`] function.
```

futures-util/src/future/pending.rs
```diff
@@ -1,17 +1,19 @@
 use core::marker;
 use core::pin::Pin;
-use futures_core::future::{Future, FusedFuture};
+use futures_core::future::{FusedFuture, Future};
 use futures_core::task::{Context, Poll};
 
-/// Future for the [`empty`] function.
+/// Future for the [`pending`] function.
```
taiki-e:
```suggestion
/// Future for the [`pending()`] function.
```

futures-util/src/stream/pending.rs
Marwes: :man_facepalming: Fixed...

cramertj:
Thanks!

PR #1693, opened 2019-06-29
impl FusedStream for stream::{Empty, Pending}

taiki-e:
`Empty` is always terminated. `Pending` is never returning `Poll::Ready`.

### DIFF
--- a/futures-util/src/future/pending.rs
+++ b/futures-util/src/future/pending.rs
@@ -12,7 +12,7 @@ pub struct Pending<T> {
 
 impl<T> FusedFuture for Pending<T> {
     fn is_terminated(&self) -> bool {
-        false
+        true
     }
 }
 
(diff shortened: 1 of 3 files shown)


cramertj:
I actually think that `Pending` should always be terminated-- from the docs:

> is_terminated will return true if a future should no longer be polled. Usually, this state occurs after poll (or try_poll) returned Poll::Ready. However, is_terminated may also return true if a future has become inactive and can no longer make progress and should be ignored or dropped rather than being polled again.

`Pending` is always inactive and will never make progress-- polling it further won't help. This will also cause it do to "the right thing" when used in a `select!` and panic rather than deadlocking if a `completed => ...` branch wasn't provided.

taiki-e:
> `Pending` is always inactive and will never make progress-- polling it further won't help. This will also cause it do to "the right thing" when used in a `select!` and panic rather than deadlocking if a `completed => ...` branch wasn't provided.

Oh, you are right. It seems I did not read the document properly...

PR #1697, opened 2019-06-30
Minor clean up futures-io

taiki-e:
Non-controversial part of #1688.


futures/Cargo.toml
```diff
@@ -34,6 +34,7 @@ futures-util-preview = { path = "../futures-util", version = "=0.3.0-alpha.16",
 pin-utils = "0.1.0-alpha.4"
 futures-test-preview = { path = "../futures-test", version = "=0.3.0-alpha.16" }
 tokio = "0.1.11"
+assert_matches = "1.3.0"
```
cramertj: Where is this needed?

taiki-e: It is used by `futures/tests/io_cursor.rs` that renamed without changes.

https://github.com/rust-lang-nursery/futures-rs/blob/329bd72334ee24c7a6b395e96710acd622e9f228/futures/tests/io_cursor.rs#L1

cramertj: Ah, thanks, sorry, I missed that!

PR #1706, opened 2019-07-02
Remove channel_impl from futures-sink

### LINKED ISSUE #1700: Allow depending on Sink without channel
There's valid cases of wanting to use the `Sink` trait without needing the channels, it'd be useful to allow that. Either:

- The `futures-sink` crate could make its `futures-channel` dependency optional.
- The `futures-channel` crate could optionally depend on `futures-sink` to provide the impls.

My opinion is the latter (changing it so `futures-channel` optionally depends on `futures-sink`) is cleaner, but either one would accomplish the goal.

seanmonstar:
Add an optional `sink` feature to futures-channel that implements `Sink`
for the mpsc channel.

Closes #1700

### DIFF
--- a/futures-channel/Cargo.toml
+++ b/futures-channel/Cargo.toml
@@ -16,10 +16,12 @@ name = "futures_channel"
 
 [features]
 std = ["futures-core-preview/std"]
+sink = ["futures-sink-preview"]
 default = ["std"]
 
 [dependencies]
 futures-core-preview = { path = "../futures-core", version = "=0.3.0-alpha.16", default-features = false }
+futures-sink-preview = { path = "../futures-sink", version = "=0.3.0-alpha.16", default-features = false, optional = true }
 
 [dev-dependencies]
 futures-preview = { path = "../futures", version = "=0.3.0-alpha.16", default-features = true }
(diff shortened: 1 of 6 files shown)


cramertj:
The reversed dependency here was intentional-- my feeling was that the `Sink` trait is less stable than the interface exposed by the channel types, so by keeping channels as a dependency of `Sink` rather than the other way around, we make it possible for libraries to expose channels in their public APIs while still upgrading the version of the `Sink` trait they use.

cramertj:
i'm not terribly strongly attached to this approach, though, so if you feel there's strong motivation for doing it this way I'm happy to discuss.

seanmonstar:
I get that. As mentioned in #1700, there can be use-cases that want to implement or use `Sink` but don't need the channel crate. This just one option for solving it. If new versions of `Sink` are made, a new optional feature could be added.

cramertj:
An alternative approach would be to make the channel impls and the dependency on channels optionally gated on a feature flag in the `Sink` crate.

seanmonstar:
Yea, I can do the other option if others prefer it. If there's no preference, I find this way a little cleaner 🤷‍♂

LucioFranco:
I personally find this approach a bit easier to approach and reason about. I say go with this version.

cramertj:
SG, will merge once travis is happy.

PR #1709, opened 2019-07-03
Make more of futures-util dependencies optional

### LINKED ISSUE #1489: Circular dependency with rand
Here is a reduced test case:

```toml
[package]
name = "foo"
version = "0.1.0"
authors = ["Pauan <pcxunlimited@gmail.com>"]
edition = "2018"

[dependencies]
stdweb = { version = "0.4.10", features = ["experimental_features_which_may_break_on_minor_version_bumps"] }
rand = { version = "0.6.5", features = ["stdweb"] }
```

When I run `cargo build` I get this error:

```
error: cyclic package dependency: package `rand v0.6.5` depends on itself. Cycle:
package `rand v0.6.5`
    ... which is depended on by `futures-util-preview v0.3.0-alpha.13`
    ... which is depended on by `stdweb v0.4.15`
    ... which is depended on by `rand_os v0.1.3`
```

After scratching my head at this for quite a while, I figured it out:

* The `rand` crate has optional support for generating random numbers on wasm32 (that's what the `stdweb` feature enables).

* The `stdweb` crate has optional support for using JS Promises as Futures (that's what the `experimental_features_which_may_break_on_minor_version_bumps` feature enables).

* The `futures_util` crate has *non-optional* support for using `rand` (this is used for the `select` macro).

This then creates a cycle: `futures_util` needs `rand`, which needs `stdweb`, which needs `futures_util`.

I don't have any great solutions for this. Removing `rand` from the `std` feature will fix the immediate problem, but then what if somebody *wants* to use the `select` macro...?

### LINKED ISSUE #1701: Make more of futures-utils dependencies optional
The `futures-util` crate provides a whole bunch of extra goodies. A user may wish to be able to use `FutureExt` or `StreamExt`, without most of the other goodies. It'd help with compilations if we could disable those pieces.

- `futures-channel` is used for `RemoteHandle` (just checking, does this type still pull its weight)?
- `futures-sink` and `futures-io` might not always be necessary.
- `futures-select-macro` includes some heavy dependencies itself, and if we don't use the macro, we could save much here (and could prevent needing to compile `rand` as well). The select macro is already optional, but is required if the `async_await` feature is enabled.

taiki-e:
Closes #1489
Closes #1701

cc @seanmonstar

### DIFF
--- a/futures-util/Cargo.toml
+++ b/futures-util/Cargo.toml
@@ -15,21 +15,25 @@ Common utilities and extension traits for the futures-rs library.
 name = "futures_util"
 
 [features]
-std = ["alloc", "futures-core-preview/std", "futures-io-preview/std", "futures-sink-preview/std", "slab", "memchr"]
 default = ["std"]
-async-await = ["std", "futures-select-macro-preview", "proc-macro-hack", "proc-macro-nested", "rand"]
+std = ["alloc", "futures-core-preview/std", "slab"]
+alloc = ["futures-core-preview/alloc"]
+async-await = ["std"]
 compat = ["std", "futures_01"]
-io-compat = ["compat", "tokio-io"]
+io-compat = ["io", "compat", "tokio-io"]
 bench = []
-nightly = ["futures-core-preview/nightly", "futures-sink-preview/nightly"]
+nightly = ["futures-core-preview/nightly"]
 cfg-target-has-atomic = ["futures-core-preview/cfg-target-has-atomic"]
-alloc = ["futures-core-preview/alloc", "futures-sink-preview/alloc"]
+sink = ["futures-sink-preview"]
+io = ["std", "futures-io-preview", "memchr"]
+channel = ["std", "futures-channel-preview"]
+select-macro = ["async-await", "futures-select-macro-preview", "proc-macro-hack", "proc-macro-nested", "rand"]
 
 [dependencies]
 futures-core-preview = { path = "../futures-core", version = "=0.3.0-alpha.16", default-features = false }
-futures-channel-preview = { path = "../futures-channel", version = "=0.3.0-alpha.16", default-features = false }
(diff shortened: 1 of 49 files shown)


futures-util/src/async_await/mod.rs
```diff
@@ -24,7 +24,9 @@ pub use self::pending::*;
 mod join;
 
 // Primary export is a macro
+#[cfg(feture = "select-macro")]
```
seanmonstar:
```suggestion
#[cfg(feature = "select-macro")]
```

futures-util/src/async_await/mod.rs
```diff
@@ -24,7 +24,9 @@ pub use self::pending::*;
...
 mod select_mod;
+#[cfg(feture = "select-macro")]
```
seanmonstar:
```suggestion
#[cfg(feature = "select-macro")]
```

futures-util/Cargo.toml
```diff
@@ -17,19 +17,23 @@ name = "futures_util"
...
-async-await = ["std", "futures-select-macro-preview", "proc-macro-hack", "proc-macro-nested", "rand"]
+async-await = ["std"]
 compat = ["std", "futures_01"]
-io-compat = ["compat", "tokio-io"]
+io-compat = ["io", "compat", "tokio-io"]
 bench = []
 nightly = ["futures-core-preview/nightly", "futures-sink-preview/nightly"]
 cfg-target-has-atomic = ["futures-core-preview/cfg-target-has-atomic"]
 alloc = ["futures-core-preview/alloc", "futures-sink-preview/alloc"]
+sink = ["futures-sink-preview", "futures-channel-preview/sink"]
```
carllerche: I would like to depend on futures-util's sink helpers w/o pulling in `futures-channel`.

taiki-e: Oh, that makes sense. Fixed.

PR #1710, opened 2019-07-03
Fix assertion fails in generator size tests

taiki-e:
Generator size has been improved by rust-lang/rust#61922 🎉 .


taiki-e:
Closing this PR in favor of #1711

PR #1716, opened 2019-07-06
Add #[feature(async_closure)] to examples

taiki-e:
Since `async_closure` was separated from `async_await`, two examples are broken.


MajorBreakfast:
Thanks! Just ran into this

@cramertj I hope you don't mind me merging this

cramertj:
i'd prefer these moved to stop using async closures, since those won't be stable any time soon and are equivalent to closures which return async blocks. I'm happy to take care of this in a follow-up.

Issue #1661, opened 2019-06-06
Add StreamExt::copy_into


yoshuawuyts:
## Problem definition
Currently in order to move data from an `AsyncRead` into an `AsyncWrite` you can use the [`copy_into`](https://docs.rs/futures-preview/0.3.0-alpha.16/futures/io/trait.AsyncReadExt.html#method.copy_into) method.

However when wanting to move data from a `Stream<Item = Vec<u8>>` into an `AsyncWrite`, we have to perform the slightly awkward dance of:

```rust
stream.map(io::Result::Ok).into_async_read().copy_into(writer)
```
This is not only hard to remember, but has a double buffering problem (see https://github.com/rust-lang-nursery/futures-rs/issues/1659 for a proposal to address this).

## Proposal
It'd be nice if `copy_into` would work for byte streams too:
```rust
stream.copy_into(writer)
```

This could be implemented along the lines of:
```rust
impl Stream<Item = AsRef<[u8]>> {
    pub fn copy_into(&mut self, writer: impl AsyncWrite) -> /* ret */ {
        self.for_each(move |buf| writer.write(buf))
    }
}

impl Stream<Item = io::Result<AsRef<[u8]>>> {
    pub fn copy_into(&mut self, writer: impl AsyncWrite) -> /* ret */ {
        self.try_for_each(move |buf| writer.write(buf))
    }
}
```

## Other Considerations
So far I've only encountered example uses for this in the wild moving data from `Stream -> AsyncWrite`. It might be possible to add similar conversions for other traits too, but I think it would be best to keep the conversation about `Stream<[u8]> -> AsyncWrite` first, to limit the scope of this proposal.

## Related Issues
- https://github.com/rust-lang-nursery/futures-rs/issues/1659

yoshuawuyts:
A better solution in the future would be using async iteration syntax:
```rust
for buf.await? in stream {
    writer.write(buf)?;
}
```

This is far more flexible than an adapter-based approach and works past the problems outlined in this issue. It can already be emulated using `while let Some` loops.

PR #1723, opened 2019-07-08
Stream::copy_into

yoshuawuyts:
Implementation of https://github.com/rust-lang-nursery/futures-rs/issues/1661. Based on an older impl of [`AsyncRead::copy_into`](https://github.com/Nemo157/futures-rs/blob/93223f5a2c8d34d1ba7bfc96479ae77f3580d399/futures-util/src/io/copy_into.rs).

Input would be quite helpful at this point, especially towards tackling the todos. Thanks heaps!

## Todos
- [ ] Implement over `item = AsRef<[u8]>` instead of `item = Vec<u8>`
- [ ] Implement for a `item = Result<T>` in addition to `item = T`
- [ ] Loosen stream constraint to `S: ?Sized + Unpin`

## Notes
I'm not quite sure how to change the impl from `Vec<u8>` to `AsRef<[u8]>`. I've tried a few things, but it's not quite working out.

Similarly I'm not sure how to best approach being able to iterate over both `Item = T` and `Item = Result<T, E>`. Should we implement this method for `TryStream` too? Or is there perhaps a way to do an automatic mapping of one type to the other?

### DIFF
--- a/futures-util/src/stream/copy_into.rs
+++ b/futures-util/src/stream/copy_into.rs
@@ -1,5 +1,3 @@
-// https://github.com/Nemo157/futures-rs/blob/93223f5a2c8d34d1ba7bfc96479ae77f3580d399/futures-util/src/io/copy_into.rs
-
 use futures_core::future::Future;
 use futures_core::task::{Context, Poll};
 use futures_core::Stream;
--- a/futures-util/src/stream/copy_into.rs
+++ b/futures-util/src/stream/copy_into.rs
@@ -1,0 +1,83 @@
+// https://github.com/Nemo157/futures-rs/blob/93223f5a2c8d34d1ba7bfc96479ae77f3580d399/futures-util/src/io/copy_into.rs
+
+use futures_core::future::Future;
+use futures_core::task::{Context, Poll};
+use futures_core::Stream;
+use futures_io::AsyncWrite;
+use std::io;
+use std::pin::Pin;
+
+/// Future for the [`copy_into`](super::StreamExt::copy_into) method.
+#[derive(Debug)]
+#[must_use = "futures do nothing unless you `.await` or poll them"]
+pub struct CopyInto<'a, S: Unpin, W: ?Sized + Unpin> {
+    stream: S,
+    read_done: bool,
+    writer: &'a mut W,
+    pos: usize,
+    cap: usize,
+    amt: u64,
+    buf: Option<Vec<u8>>,
+}
+
+impl<S: Unpin, W: ?Sized + Unpin> Unpin for CopyInto<'_, S, W> {}
+
+impl<'a, S: Unpin, W: ?Sized + Unpin> CopyInto<'a, S, W> {
+    pub(super) fn new(stream: S, writer: &'a mut W) -> Self {
+        CopyInto {
+            stream,
+            read_done: false,
+            writer,
+            amt: 0,
+            pos: 0,
+            cap: 0,
+            buf: None,
+        }
+    }
+}
+
+impl<S, W> Future for CopyInto<'_, S, W>
+    where S: Stream<Item = Vec<u8>> + Unpin,
+          W: AsyncWrite + ?Sized + Unpin,
+{
+    type Output = io::Result<u64>;
+
+    fn poll(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
+        let this = &mut *self;
+        loop {
+            // If our buffer is empty, then we need to read some data to
+            // continue.
+            if this.pos == this.cap && !this.read_done {
+                match ready!(Pin::new(&mut this.stream).poll_next(cx)) {
+                    Some(buf) => {
+                        this.pos = 0;
+                        this.cap = buf.len();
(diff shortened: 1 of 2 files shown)


futures-util/src/stream/mod.rs
```diff
@@ -30,6 +32,9 @@ pub use self::collect::Collect;
 mod concat;
 pub use self::concat::Concat;
 
+mod copy_into;
+pub use self::copy_into::CopyInto;
```
LucioFranco: curious, since this is an IO util type does this need to be put behind a `io` feature flag like `AsyncWrite` is?

futures-util/src/stream/mod.rs
```diff
@@ -1206,4 +1211,40 @@ pub trait StreamExt: Stream {
...
+    /// let bytes = stream1.copy_into(&mut writer).await?;
+    /// writer.close().await?;
+    ///
+    /// assert_eq!(bytes, 10);
+    /// assert_eq!(&writer.into_inner(), b"helloworld");
+    /// # Ok::<(), Box<dyn std::error::Error>>(()) }).unwrap();
+    /// ```
+    fn copy_into<W, E>(self, writer: &mut W) -> CopyInto<'_, Self, W, E>
+    where
+        Self: Sized + Unpin,
```
cramertj: I'd include the bounds required for the resulting type to be usable as a future as well (`Self::Item: AsRef<[u8]>`)

futures-util/src/stream/copy_into.rs
```diff
@@ -0,0 +1,81 @@
...
+            amt: 0,
+            pos: 0,
+            cap: 0,
+            buf: None,
+        }
+    }
+}
+
+impl<S, W> Future for CopyInto<'_, S, W>
+    where S: Stream<Item = Vec<u8>> + Unpin,
```
cramertj: I'd do this over `Self: TryStream<Ok: AsRef<[u8]>>, io::Error: Into<<Self as TryStream>::Error>` to allow yielding an error from the stream.

futures-util/src/stream/copy_into.rs
```diff
@@ -0,0 +1,81 @@
...
+use futures_core::task::{Context, Poll};
+use futures_core::Stream;
+use futures_io::AsyncWrite;
+use std::io;
+use std::pin::Pin;
+
+/// Future for the [`copy_into`](super::StreamExt::copy_into) method.
+#[derive(Debug)]
+#[must_use = "futures do nothing unless you `.await` or poll them"]
+pub struct CopyInto<'a, S: Unpin, W: ?Sized + Unpin> {
```
cramertj: Why require these to be `Unpin`?

futures-util/src/stream/copy_into.rs
```diff
@@ -0,0 +1,81 @@
...
+    buf: Option<Vec<u8>>,
+}
+
+impl<S: Unpin, W: ?Sized + Unpin> Unpin for CopyInto<'_, S, W> {}
+
+impl<'a, S: Unpin, W: ?Sized + Unpin> CopyInto<'a, S, W> {
+    pub(super) fn new(stream: S, writer: &'a mut W) -> Self {
+        CopyInto {
+            stream,
+            read_done: false,
```
cramertj: I'd keep an `Option<Stream>` rather than a separate boolean, which will allow dropping the stream sooner as well as using less space if the stream type happens to have a free niche.

futures-util/src/stream/copy_into.rs
```diff
@@ -0,0 +1,81 @@
...
+#[derive(Debug)]
+#[must_use = "futures do nothing unless you `.await` or poll them"]
+pub struct CopyInto<'a, S: Unpin, W: ?Sized + Unpin> {
+    stream: S,
+    read_done: bool,
+    writer: &'a mut W,
+    pos: usize,
+    cap: usize,
+    amt: u64,
+    buf: Option<Vec<u8>>,
```
cramertj: Not sure exactly where you got stuck trying to make this `AsRef<[u8]>` instead, but I'd expect the type definition to look like this:

```rust
pub struct CopyInto<'a, S, W: ?Sized>
where
    S: TryStream,
    S::Ok: AsRef<[u8]>,
{
    stream: Option<S>,
    writer: &'a mut W,
    pos: usize,
    cap: usize,
    amt: u64,
    buf: Option<S::Ok>
}
```

cramertj:
Hey @yoshuawuyts are you planning to pick this back up? If not, would you mind if I did?

yoshuawuyts:
@cramertj ah dang, I meant to close this PR. Feel free to, but I also think it's okay to close (:

In async-std we now have `io::copy` modeled after the stdlib, which largely fills the same role this PR was intended to.

cramertj:
Closing to revisit in a new PR. Thanks!

thomjs:
Just dropping in to see if this was ever revisited? I still think it would be a great addition to the language.

cramertj:
@thomjs I don't believe it was-- if you feel like picking it up, the code and comments above should provide a decent starting point, and if you ping me here or zulip I'll be happy to answer questions.

PR #1732, opened 2019-07-14
Add check for object safety

taiki-e:
I think most might be rejected at the review, but it should be nice to ensure that object safety of traits is not accidentally changed.

cc @cramertj

### DIFF
--- a/futures/tests/object_safety.rs
+++ b/futures/tests/object_safety.rs
@@ -1,0 +1,49 @@
+fn assert_is_object_safe<T>() {}
+
+#[test]
+fn future() {
+    // `FutureExt`, `TryFutureExt` and `UnsafeFutureObj` are not object safe.
+    use futures::future::{FusedFuture, Future, TryFuture};
+
+    assert_is_object_safe::<&dyn Future<Output = ()>>();
+    assert_is_object_safe::<&dyn TryFuture<Ok = (), Error = ()>>();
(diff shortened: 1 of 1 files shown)


futures/tests/object_safety.rs
```diff
@@ -0,0 +1,49 @@
...
+    use futures::future::{FusedFuture, Future, TryFuture};
+
+    assert_is_object_safe::<&dyn Future<Output = ()>>();
+    assert_is_object_safe::<&dyn TryFuture<Ok = (), Error = ()>>();
+    assert_is_object_safe::<&dyn FusedFuture>();
+}
+
+#[test]
+fn stream() {
+    // `StreamExt` and `StreamExt` are not object safe.
```
taiki-e: @cramertj Do you think `*Ext` traits should also be object safe? 

(I think it is easy to change they to object safe)

cramertj: I don't think it matters particularly either way. IMO it'd be kind of weird to have a `dyn *Ext` and could indicate user error, so perhaps it'd be best to leave them as non-object-safe.

taiki-e: > IMO it'd be kind of weird to have a `dyn *Ext` and could indicate user error, so perhaps it'd be best to leave them as non-object-safe.

I agree with that.

PR #1734, opened 2019-07-14
Add AsyncWrite implementation for IntoAsyncRead

najamelan:
This implements AsyncWrite for IntoAsyncRead when the inner stream implements it,
forwarding all method calls.

This allows working on Types that keep both impls, and allows framing
them with codecs where those require both traits (like tokio codec).

r? @cramertj

### DIFF
--- a/futures-util/src/try_stream/into_async_read.rs
+++ b/futures-util/src/try_stream/into_async_read.rs
@@ -2,7 +2,7 @@ use crate::try_stream::TryStreamExt;
 use core::pin::Pin;
 use futures_core::stream::TryStream;
 use futures_core::task::{Context, Poll};
-use futures_io::{AsyncRead, AsyncBufRead};
+use futures_io::{AsyncRead, AsyncWrite, AsyncBufRead};
 use std::cmp;
 use std::io::{Error, Result};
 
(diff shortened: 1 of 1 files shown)


cramertj:
This seems like a fine change to me, but did you mean to have a different commit message? The commit message and PR name seem to suggest a different thing than what is implemented here.

najamelan:
@cramertj We can change the commit message to whatever you like. From the user perspective it's like you have a Type which implements AsyncWrite and when you call into_async_read, you lose the AsyncWrite functionality. That's why I called it like this. This doesn't actually provide any new functionality, just makes the inner AsyncWrite available on the wrapper to avoid losing the functionality, and to be able to work with API's that require both. Especially since IntoAsyncRead will consume the inner stream.

Which part does not correspond to the code? I'm a bit  confused. It looks like the first paragraph describes the implementation exactly.

taiki-e:
"Add AsyncWrite implementation for IntoAsyncRead" or "impl AsyncWrite for IntoAsyncRead" seems good commit message to me. (refs: #1575, #1503)

najamelan:
I changed the pull request title and the commit message.

najamelan:
I changed the pull request title and the commit message.

PR #1738, opened 2019-07-16
process subsequent futures when partially progressing a pool

### LINKED ISSUE #1737: try_run_one/run_until_stalled cannot handle subsequently spawned futures
In #1537/#1437 to answer #1369, we introduced `LocalPool::try_run_one`/`LocalPool::run_until_stalled` to partially progress a LocalPool.

Now, I used them myself and noticed that they do not worked as described when the spawned futures spawn futures themselves.

In the following example it only prints `begin`. Instead of also printing out `inner` and `end 5`.

```rust
let mut pool = LocalPool::new();
let mut spawner = pool.spawner();

spawner.clone().spawn_local(async move {
    println!("begin");

    let inner = spawner.spawn_local_with_handle(async {
        println!("inner");
        5
    }).unwrap();

    println!("end {}", inner.await);
}).unwrap();

pool.run_until_stalled();
```

This is because `inner.await` makes the internal `poll` function of the outer future return `Pending`. `run_until_stalled` will see this `Pending` and return because it thinks it can not make any more progress, which is wrong because there are new futures which have been enqueued and which can be advanced.

The same happens with `pool.try_run_one`.

For both function, this behavior is against what their documentation describes. Because in both cases the function returns __even if progress can be made__!

```rust
/// Runs all tasks and returns after completing one future or until no more progress
/// can be made. Returns `true` if one future was completed, `false` otherwise.
/// [...]
/// This function will not block the calling thread and will return the moment
/// that there are no tasks left for which progress can be made or after exactly one
/// task was completed; [...]
pub fn try_run_one(&mut self) -> bool;

/// Runs all tasks in the pool and returns if no more progress can be made
/// on any task.
/// [...]
/// This function will not block the calling thread and will return the moment
/// that there are no tasks left for which progress can be made;
/// [...]
pub fn run_until_stalled(&mut self);
```

SrTobi:
fixes #1737

### DIFF
--- a/futures-executor/src/local_pool.rs
+++ b/futures-executor/src/local_pool.rs
@@ -193,12 +193,20 @@ impl LocalPool {
     /// Though only one task will be completed, progress may be made on multiple tasks.
     pub fn try_run_one(&mut self) -> bool {
         poll_executor(|ctx| {
-            let ret = self.poll_pool_once(ctx);
+            loop {
+                let ret = self.poll_pool_once(ctx);
 
-            // return if we really have executed a future
-            match ret {
-                Poll::Ready(Some(_)) => true,
-                _ => false
+                // return if we have executed a future
+                if let Poll::Ready(Some(_)) = ret {
+                    return true;
+                }
+
+                // if there are no new incoming futures
+                // then there is no feature that can make progress
+                // and we can return without having completed a single future
+                if self.incoming.borrow().is_empty() {
+                    return false;
+                }
             }
         })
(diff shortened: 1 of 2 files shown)


SrTobi:
What about merging this? It's a vital fix for two affected functions.

cramertj:
Thanks for the fix here, and sorry for missing this PR. In the future, cc'ing me or one of the other maintainers if you don't get a response quickly should help.

PR #1739, opened 2019-07-16
enable ThreadPool to join worker threads

### LINKED ISSUE #1470: Waiting for threadpool threads to drop
I have some shared data in an `Arc` that I want to drop when my program exits.
However, when I drop the `ThreadPool`, this only sends a `Close` message to the worker threads but does not join them. This means I have no way to block until I know the threads/shared data is dropped.

SrTobi:
Adds an option to ThreadPoolBuilder to configure the resulting ThreadPool
to join the worker threads when dropped.

fixes #1470

### DIFF
--- a/futures-executor/src/thread_pool.rs
+++ b/futures-executor/src/thread_pool.rs
@@ -27,6 +27,7 @@ pub struct ThreadPool {
 pub struct ThreadPoolBuilder {
     pool_size: usize,
     stack_size: usize,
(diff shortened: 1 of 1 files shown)


SrTobi:
We might also (additionally or instead!) add a shutdown method that does the same.

SrTobi:
This PR is garbage :scream: Sorry for that!

I didn't quite understood how the pool is shared with every task.

PR #1741, opened 2019-07-16
Add AsyncWriteExt::write_all_vectored utility

Thomasdezeeuw:
Since vectored I/O has been stabilised I was missing this. Two things to note;

* this takes `&mut [IoSlice]`, where the call in `io::Read` takes `&[IoSlice]`. This is needed to advance the buffer inside the `IoSlice`.
* this uses some unsafe code, I couldn't get around without it.

TODO:
- [x] Docs.

### DIFF
--- a/futures-util/Cargo.toml
+++ b/futures-util/Cargo.toml
@@ -30,6 +30,7 @@ unstable = ["futures-core/unstable", "futures-task/unstable"]
 cfg-target-has-atomic = ["futures-core/cfg-target-has-atomic", "futures-task/cfg-target-has-atomic"]
 bilock = []
 read-initializer = ["io", "futures-io/read-initializer", "futures-io/unstable"]
+write-all-vectored = ["io"]
 
 [dependencies]
 futures-core = { path = "../futures-core", version = "0.3.4", default-features = false }
--- a/futures-util/src/io/mod.rs
+++ b/futures-util/src/io/mod.rs
@@ -124,6 +124,11 @@ pub use self::write_vectored::WriteVectored;
 mod write_all;
 pub use self::write_all::WriteAll;
 
+#[cfg(feature = "write_all_vectored")]
+mod write_all_vectored;
+#[cfg(feature = "write_all_vectored")]
+pub use self::write_all_vectored::WriteAllVectored;
+
 /// An extension trait which adds utility methods to `AsyncRead` types.
 pub trait AsyncReadExt: AsyncRead {
     /// Creates an adaptor which will chain this stream with another.
@@ -460,6 +465,60 @@ pub trait AsyncWriteExt: AsyncWrite {
         WriteAll::new(self, buf)
     }
 
+    /// Attempts to write multiple buffers into this writer.
+    ///
+    /// Creates a future that will write the entire contents of `bufs` into this
+    /// `AsyncWrite` using [vectored writes].
+    ///
+    /// The returned future will not complete until all the data has been
+    /// written.
+    ///
+    /// [vectored writes]: std::io::Write::write_vectored
+    ///
+    /// # Notes
+    ///
+    /// Unlike `io::Write::write_vectored`, this takes a *mutable* reference to
+    /// a slice of `IoSlice`s, not an immutable one. That's because we need to
+    /// modify the slice to keep track of the bytes already written.
+    ///
+    /// Once this futures returns, the contents of `bufs` are unspecified, as
+    /// this depends on how many calls to `write_vectored` were necessary. It is
+    /// best to understand this function as taking ownership of `bufs` and to
+    /// not use `bufs` afterwards. The underlying buffers, to which the
+    /// `IoSlice`s point (but not the `IoSlice`s themselves), are unchanged and
+    /// can be reused.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// # futures::executor::block_on(async {
+    /// use futures::io::AsyncWriteExt;
+    /// use std::io::{Cursor, IoSlice};
+    ///
+    /// let mut writer = Cursor::new([0u8; 7]);
+    /// let bufs = &mut [
+    ///     IoSlice::new(&[1]),
+    ///     IoSlice::new(&[2, 3]),
+    ///     IoSlice::new(&[4, 5, 6]),
+    /// ];
+    ///
+    /// writer.write_all_vectored(bufs).await?;
+    /// // Note: the contents of `bufs` is now undefined, see the Notes section.
+    ///
+    /// assert_eq!(writer.into_inner(), [1, 2, 3, 4, 5, 6, 0]);
+    /// # Ok::<(), Box<dyn std::error::Error>>(()) }).unwrap();
+    /// ```
+    #[cfg(feature = "write_all_vectored")]
+    fn write_all_vectored<'a>(
+        &'a mut self,
+        bufs: &'a mut [IoSlice<'a>],
+    ) -> WriteAllVectored<'a, Self>
+    where
+        Self: Unpin,
+    {
+        WriteAllVectored::new(self, bufs)
+    }
+
     /// Wraps an [`AsyncWrite`] in a compatibility wrapper that allows it to be
     /// used as a futures 0.1 / tokio-io 0.1 `AsyncWrite`.
     /// Requires the `io-compat` feature to enable.
@@ -469,7 +528,6 @@ pub trait AsyncWriteExt: AsyncWrite {
     {
         Compat::new(self)
     }
-
 
     /// Allow using an [`AsyncWrite`] as a [`Sink`](futures_sink::Sink)`<Item: AsRef<[u8]>>`.
     ///
--- a/futures-util/src/io/write_all.rs
+++ b/futures-util/src/io/write_all.rs
@@ -33,7 +33,7 @@ impl<W: AsyncWrite + ?Sized + Unpin> Future for WriteAll<'_, W> {
                 this.buf = rest;
             }
             if n == 0 {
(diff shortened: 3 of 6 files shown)


futures-util/src/io/write_all.rs
```diff
@@ -40,3 +41,239 @@ impl<W: AsyncWrite + ?Sized + Unpin> Future for WriteAll<'_, W> {
...
+                    let bytes = unsafe {
+                        // We create a new slice unsafely to get around lifetime
+                        // issues. Since all buffers need to life at least as long
+                        // as WriteAllVectored, by definition of lifetime 'a, this
+                        // should be safe.
+                        slice::from_raw_parts(
+                            this.bufs[0].as_ptr().add(bytes_written),
+                            this.bufs[0].len() - bytes_written,
+                        )
+                    };
```
Nemo157: This seems to indicate a missing `IoSlice` API; either something like `into_slice` so that you could get the slice with correct lifetime back and use `split_at`, or having `split_at` directly on `IoSlice`. Would be good to open an issue about adding something to move this unsafety to `std`.

Thomasdezeeuw: Or better yet and `advance` method that would do this for use.

Thomasdezeeuw: Created https://github.com/rust-lang/rust/issues/62726.

Thomasdezeeuw:
@Nemo157 I've created {IoSlice, IoSliceMut}::advance which does the most heavy lifting of this change in https://github.com/rust-lang/rust/pull/62987. However it's an unstable feature, now that `Future` is also available on stable how are nightly features supposed to be handled?

Thomasdezeeuw:
@Nemo157 sorry about the delay. I've rebased this on master, updated the code to use the newly created `io_slice_advance` Rust feature and added documentation.

futures-util/src/io/write_all.rs
Thomasdezeeuw: This is now available using the `io_slice_advance` Rust feature, tracking issue: https://github.com/rust-lang/rust/issues/62726.

futures-util/src/io/mod.rs
```diff
@@ -500,6 +502,61 @@ pub trait AsyncWriteExt: AsyncWrite {
...
+    ///
+    /// Creates a future that will write the entire contents of `bufs` into this
+    /// `AsyncWrite` using [vectored writes].
+    ///
+    /// The returned future will not complete until all the data has been
+    /// written.
+    ///
+    /// [vectored writes]: std::io::Write::write_vectored
+    ///
+    /// # Notes
```
Thomasdezeeuw: Please carefully review this section, I don't if it clearly states what I mean.

futures-util/src/io/write_all.rs
```diff
@@ -33,10 +35,208 @@ impl<W: AsyncWrite + ?Sized + Unpin> Future for WriteAll<'_, W> {
...
 
         Poll::Ready(Ok(()))
     }
 }
+
+/// Future for the
+/// [`write_all_vectored`](super::AsyncWriteExt::write_all_vectored) method.
+#[derive(Debug)]
+#[must_use = "futures do nothing unless you `.await` or poll them"]
+#[cfg(feature = "write_all_vectored")]
```
taiki-e: Could you split these into a new file(`futures-util/src/io/write_all_vectored.rs`) instead of using cfg for each item?

futures-util/src/io/mod.rs
Nemo157: _If_ the future successfully completes, `bufs` is an empty slice, right?

On failure `bufs` may have been modified, but is not guaranteed to match with what has actually been written out (because I believe the underlying IO doesn't guarantee that it either writes some data or returns an error, it could have written some of the data _and then_ returned an error). That's pretty much equivalent to what `write_all` guarantees, if it returns an error some part of the data will have been written with no way to know how much.

I think this should try to avoid the "undefined" term since that has such strong memory-safety connotations, and the failure cases are relatively well-defined.

Thomasdezeeuw: > _If_ the future successfully completes, `bufs` is an empty slice, right?

That is correct, but in the future we could optimise to just return ready if the entire (remaining) buffer is written, not modifying `bufs`.

> On failure `bufs` may have been modified, but is not guaranteed to match with what has actually been written out (because I believe the underlying IO doesn't guarantee that it either writes some data or returns an error, it could have written some of the data _and then_ returned an error). That's pretty much equivalent to what `write_all` guarantees, if it returns an error some part of the data will have been written with no way to know how much.

Do you want this to be reflected in the documentation somehow, or leave it as is?

> I think this should try to avoid the "undefined" term since that has such strong memory-safety connotations, and the failure cases are relatively well-defined.

Hence I added the it was not about memory safety, but I do understand the concern. Should I replace "undefined" with "unknown" (or something else) or you want to provide the guarantee that `bufs` will be empty after it successfully returns (also see my first remark in this comment)?

futures-util/src/io/write_all.rs
Thomasdezeeuw: Done.

Thomasdezeeuw:
Let me know if you want me to rebase this on the master branch.

Thomasdezeeuw:
@Nemo157 this pr has fallen out of date, but can I get an OK/not OK before spending more time on this?

Thomasdezeeuw:
@Nemo157 @taiki-e ping :)

Thomasdezeeuw:
@cramertj do you have time to review this?

cramertj:
Sure, sorry for the delay! Thanks for the ping.

futures-util/Cargo.toml
```diff
@@ -34,6 +34,7 @@ unstable = ["futures-core-preview/unstable"]
 cfg-target-has-atomic = ["futures-core-preview/cfg-target-has-atomic"]
 bilock = []
 read_initializer = ["io", "futures-io-preview/read_initializer", "futures-io-preview/unstable"]
+write_all_vectored = ["io"]
```
cramertj: Can you add a corresponding feature to the top-level `futures` crate which turns this on, as well as a conditional re-export of the public types introduced?

futures-util/src/io/mod.rs
```diff
@@ -500,6 +505,61 @@ pub trait AsyncWriteExt: AsyncWrite {
...
+    ///
+    /// Different to `io::Write::write_vectored` this takes a *mutable*
+    /// reference to a slice of `IoSlice`s, not a non-mutable reference, because
+    /// we need to modify the slice to keep track of the bytes already written.
```
cramertj: Who needs to? The future can store that state internally. Is this for the user to know? IMO we should try to match the API of `write_all` as closely as possible. Actually, for that matter, what we should be trying to match would be the yet-to-exist `std::io::Write::write_all_vectored`. If there isn't yet an issue for creating that API, would you mind creating one and linking it here? I'd really like to keep these extension traits from diverging from the API in the standard library, but I'm fine landing it here if it's already under a feature gate, so long as we make sure to tag that feature gate as `_unstable` / have it require an `unstable` feature flag in order to compile. Does that make sense?

futures-util/src/io/write_all_vectored.rs
```diff
@@ -0,0 +1,200 @@
+use futures_core::future::Future;
```
cramertj: Implementation and tests here look good to me, thanks!

futures-util/src/io/mod.rs
```diff
@@ -500,6 +505,61 @@ pub trait AsyncWriteExt: AsyncWrite {
...
+    /// `bufs` will be as that depends on how many writes we needed to do. We
+    /// advice to see this function as taking ownership of `bufs` and don't use
+    /// the variable after the future returns. The underlying buffers, to which
+    /// `IoSlice` points (not the `IoSlice` itself), are unchanged and can be
+    /// reused.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// # futures::executor::block_on(async {
```
cramertj: Thanks for including this example!

futures-util/src/io/mod.rs
Thomasdezeeuw: The problem with writing all buffers is that is we need to deal with partially written `IoSlice`s. For which I wrote [`IoSlice::advance`](https://doc.rust-lang.org/std/io/struct.IoSlice.html#method.advance), which internally updates the slices, but needs mutable access. Basically `WriteAllVectored` needs ownership of `bufs`, to keep track of what bytes are already written. Also see the next section in the documentation.

In practice I don't think this will be a problem as `IoSlice` *slices*  are often created just before writing, e.g. see the example. But if you know of a better solution I'll love to hear it.

Thomasdezeeuw: Created https://github.com/rust-lang/rust/issues/70436.

Thomasdezeeuw:
@cramertj I've rebased this on master and changed the docs to match `io::Write::write_all_vectored` in the std lib (as merged in #70612).

Thomasdezeeuw:
An unrelated tests failed `select_on_non_unpin_size` on the CI. @cramertj should I just increase the expected size in test? Or is this being tracked elsewhere?

Nemo157:
Fixed in #2119, though that's still failing because of a nightly bug in the rustdoc check.

Thomasdezeeuw:
Now that #2119 is merged I've rebased this.

Thomasdezeeuw:
Friendly ping @cramertj, I don't want to have to rebase this again :)

cramertj:
Thanks for the ping, and for the changes! Sorry for the delay on my end.

PR #1743, opened 2019-07-17
Fix TryChain::is_terminated

### LINKED ISSUE #1740: Is TryChain::is_terminated correct?
Looking at https://github.com/rust-lang-nursery/futures-rs/blob/af434e1d495264dfd7f8b5094b69fbba490d423e/futures-util/src/try_future/try_chain.rs#L28-L33

Shouldn't those flipped? The `FusedFuture` docs say "Returns `true` if the underlying future should no longer be polled." Isn't the `Empty` state the completed state?

Thomasdezeeuw:
Closes #1740.

### DIFF
--- a/futures-util/src/try_future/try_chain.rs
+++ b/futures-util/src/try_future/try_chain.rs
@@ -27,8 +27,8 @@ impl<Fut1, Fut2, Data> TryChain<Fut1, Fut2, Data>
 
     pub(crate) fn is_terminated(&self) -> bool {
         match self {
-            TryChain::First(..) | TryChain::Second(_) => true,
-            TryChain::Empty => false,
+            TryChain::First(..) | TryChain::Second(_) => false,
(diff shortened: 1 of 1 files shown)


futures-util/src/try_future/try_chain.rs
```diff
@@ -53,7 +53,12 @@ impl<Fut1, Fut2, Data> TryChain<Fut1, Fut2, Data>
                 }
                 TryChain::Second(fut2) => {
                     // Poll the second future
-                    return unsafe { Pin::new_unchecked(fut2) }.try_poll(cx)
+                    return unsafe { Pin::new_unchecked(fut2) }
+                        .try_poll(cx)
+                        .map(|res| {
+                            *this = TryChain::Empty; // Drop fut2.
```
taiki-e: 👍

PR #1744, opened 2019-07-17
Add must_use attributes on Future and Stream (futures 0.1)

Speedy37:
I just found out that when using futures 0.1, a function returning `impl Future<...>` is not tagged as must_use. It took me some times to figure out which future I forgot to poll.

The only downside of this pull request I see is: this requires bumping the minimal rustc version to 1.32.

### DIFF
--- a/src/future/mod.rs
+++ b/src/future/mod.rs
@@ -174,6 +174,7 @@ use {Poll, stream};
 /// More information about combinators can be found [on tokio.rs].
 ///
 /// [on tokio.rs]: https://tokio.rs/docs/going-deeper-futures/futures-mechanics/
+#[must_use = "futures do nothing unless polled"]
 pub trait Future {
     /// The type of value that this future will resolved with if it is
     /// successful.
--- a/src/stream/mod.rs
+++ b/src/stream/mod.rs
@@ -180,6 +180,7 @@ if_std! {
 /// please feel welcome to comment on [the issue][being considered]!
 ///
 /// [being considered]: https://github.com/rust-lang-nursery/futures-rs/issues/206
+#[must_use = "streams do nothing unless polled"]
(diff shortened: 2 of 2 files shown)


seanmonstar:
> this requires bumping the minimal rustc version to 1.32

My initial feeling is that this is a blocker.

taiki-e:
If I remember correctly, `#[must_use]` on trait is no-op in the version before it is supported (`#[must_use]` on function was warning or hard error). So we should not need to increase minimal supported rustc version for this (I don't want to increase minimal supported rustc version greatly without increasing minor version).

If `#[must_use]` on trait gets warnings or errors with older rustc, I think it is better to use the [build script](https://doc.rust-lang.org/cargo/reference/build-scripts.html).

Speedy37:
Looks like you are right:

```
futures-rs> cargo +1.15.0 build
warning: unused manifest key: badges.appveyor.repository
warning: unused manifest key: badges.travis-ci.repository
warning: unused manifest key: package.categories
   Compiling futures v0.1.28 (file:///C:/Development/GitHub/futures-rs)
warning: use of deprecated item: recommended to use `Notify` instead, #[warn(deprecated)] on by default
   --> src\task_impl\std/mod.rs:107:13
    |
107 |     Old(Arc<Unpark>),
    |             ^^^^^^

    Finished debug [unoptimized + debuginfo] target(s) in 2.72 secs
```

I removed the version bump in travis.yml

dtolnay:
Ping @alexcrichton @cramertj -- would it be possible to get an 0.1 release with this?

cramertj:
Yup, I'll do that right now.

cramertj:
Futures 0.1.29 has been published with this change.

PR #1746, opened 2019-07-18
Fix Unpin not implemented for Pending

tomaka:
Not much to say.
`Pending` is often used as a dummy, and it unfortunately can't be passed to functions that require `Unpin`.


tomaka:
Related: https://github.com/rust-lang/rust/pull/62786

PR #1748, opened 2019-07-18
Change SendError to contain Option<T>

### LINKED ISSUE #1745: `mpsc::Sender::send` no longer returns the value to the caller on a failed send
With the standard library's `mpsc::Sender` and with the futures-0.1 `mpsc::Sender` you can try to send a value then get the value back if the receiver has hung up. futures-0.3 `mpsc::Sender` instead gives you back a `SendError` from which you can't extract the value you tried to send. There is a `try_send` method, but this is non-futures-aware and so useless to me.

Can we get this functionality back? Ownership and linearity are important in Rust. Values shouldn't just disappear into the void unless I want them to.

taiki-e:
Closes #1745

Also implements `SendError` <-> `TrySendError` conversion via `From`/`TryFrom` traits.

### DIFF
--- a/futures-channel/src/mpsc/mod.rs
+++ b/futures-channel/src/mpsc/mod.rs
@@ -82,12 +82,13 @@ use futures_core::stream::{FusedStream, Stream};
 use futures_core::task::{Context, Poll, Waker};
 use futures_core::task::__internal::AtomicWaker;
 use std::any::Any;
+use std::convert::TryFrom;
 use std::error::Error;
 use std::fmt;
 use std::pin::Pin;
-use std::sync::{Arc, Mutex};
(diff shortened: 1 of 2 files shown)


seanmonstar:
If I remember correctly, it used to be like this, and it was removed. I'll see if I can find it in the history.

seanmonstar:
Here it is: https://github.com/rust-lang-nursery/futures-rs/commit/a54b817f78b99a3ce74c587a57c3b5caae45b07f

taiki-e:
> Here it is: [a54b817](https://github.com/rust-lang-nursery/futures-rs/commit/a54b817f78b99a3ce74c587a57c3b5caae45b07f)

Oh, that looks reasonable to me.

PR #1750, opened 2019-07-19
Add ThreadPool::spawn_ok and spawn_obj_ok

tomaka:
Considering that `spawn` and `spawn_obj` never fail on the `ThreadPool`, it is desirable to add alternative methods that indeed cannot fail to spawn tasks.

I added this `_ok` suffix, otherwise that would be a breaking change. People's code would suddenly use the inherent method instead of the trait method, and likely get a compilation error because you can't call `unwrap()` on `()`.

Now that `SpawnError` unfortunately no longer contains the `Future` that failed to spawn, it is no longer possible to properly handle tasks spawn failures.


cramertj:
Thanks!

PR #1754, opened 2019-07-20
Remove #[allow(clippy::wrong_self_convention)]

taiki-e:


taiki-e:
Oops, today's nightly doesn't have clippy (this pr should work because the [other `get_pin_mut` methods are fine](https://github.com/rust-lang-nursery/futures-rs/blob/a705910a8d844f6d56527ab9d774be09eb3008d5/futures-util/src/try_future/flatten_stream_sink.rs#L61)) 
I will rebase when nightly has clippy.

taiki-e:
Rebased (Clippy is available today.)

taiki-e:
Seems #1880 resolved this issue.

Issue #1669, opened 2019-06-10
`&self` vs. `&mut self` for spawners


stbuehler:
Hi,

I'd like to reopen this discussion (#767 before for executors).

@aturon wrote in https://github.com/rust-lang-nursery/futures-rs/issues/767#issuecomment-365998767
> I think I'm convinced we should go with `&mut self`. Originally I had been thinking that with `&self` we could allow users of executors to decide on the memory management strategy (by applying an `Rc`/`Arc` themselves), but in practice I think these are usually baked into the executor anyway.

Exactly! The strategy is already baked in (`Spawn`/`LocalSpawn` implementations usually implement `Clone` too!) - so there is really no need for `&mut self`.

It would be way safer though to store a `&self` reference in TLS for the "current spawner".

Nemo157:
Do you mean to change both `Spawn` and `LocalSpawn` to take `&self`, or just `Spawn`?

It seems like there could definitely be a performance benefit from having a `LocalSpawn + !Clone+  !Send` implementation that can just directly access the current executors state to spawn a new task. It's much more limiting since you can only spawn new tasks from a single location, but sometimes that's all that's needed.

---

When I was playing around with global spawners I ended up using an internal trait for this

```rust
trait SharedSpawn {
    fn spawn_obj(&self, future: FutureObj<'static, ()>) -> Result<(), SpawnError>;

    fn status(&self) -> Result<(), SpawnError>;
}

impl<Sp> SharedSpawn for Sp where for<'a> &'a Sp: Spawn {
    fn spawn_obj(mut self: &Self, future: FutureObj<'static, ()>) -> Result<(), SpawnError> {
        Spawn::spawn_obj(&mut self, future)
    }

    fn status(&self) -> Result<(), SpawnError> {
        Spawn::status(&self)
    }
}
```

that allowed the global setter function to take in a single instance of the spawner and access that via an `RwLock` (since it's almost certainly going to just be set once at application startup).

```rust
fn set_global_spawner<Sp: Send + Sync + 'static>(spawner: Sp) where for<'a> &'a Sp: Spawn
```

The nice thing of this sort of pattern is it allows for both unique and shared spawners to exist behind the same basic trait, it's up to the implementor to implement it for shared references if they don't require unique access.

stbuehler:
I mean both `Spawn` and `LocalSpawn`.

[...]

PR #1763, opened 2019-07-28
Add SharedSpawn.

qnighy:
As a solution to #1669. It's sometimes convenient to have something like `Arc<dyn Spawn>` but `Spawn` isn't useful in this case because it accepts `&mut self`. I propose to have `SharedSpawn`/`Spawn` pair which work like the well-known `Fn`/`FnMut` pair.

### DIFF
--- a/futures-core/src/task/mod.rs
+++ b/futures-core/src/task/mod.rs
@@ -4,7 +4,7 @@
 mod poll;
 
 mod spawn;
-pub use self::spawn::{Spawn, LocalSpawn, SpawnError};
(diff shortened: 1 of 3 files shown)


stbuehler:
As pointed out in #1669 I'd prefer changing `Spawn` and `LocalSpawn` over adding more traits, given that I think no one actually needs the `&mut self` variants.

taiki-e (commented):
Thanks for the PR and sorry for the delay reviewing.
I'm going to close this as #1669 was addressed in #1950.

Issue #969, opened 2018-04-14
select_all does O(n) copies


stuhood:
Due to using `Vec::remove`, a `select_all` loop that drains a collection might do O(n) copies of the `Vec`.

As a strawman, `select_all_unordered` could use `Vec::swap_remove` and remove the index value from the API?

Nemo157:
Technically `SelectAll` does not guarantee any order on the returned vector, so changing it to `swap_remove` would be allowed...

It would be great to have examples of current `select_all` usage in the wild, depending on how it is actually used in practice it might be better to change it and introduce an alternative API for the use case where keeping the order consistent is necessary.

cramertj:
I'd just change it to use a `slab` or something instead of a `Vec` so that removal isn't so expensive.

cramertj:
Ah, can't change to `slab` because it yields a `Vec`. `swap_remove` it is!

PR #1764, opened 2019-07-28
Optimize future::select_all by not preserving order

udoprog:
This change makes use of `Vec::swap_remove` to shuffle less elements around in the backing `Vec` for `SelectAll`.

Related: #969 - which seems to have been missed when porting


cramertj:
Thanks!

PR #1766, opened 2019-07-28
Remote completion

stbuehler:
Allow polling remote handles (and oneshot channels) for "completion", i.e. when the result is available (`poll_ready` is used for "ready to send", so I used `poll_complete` - open to other suggestions), and to extract results without async context.

Add `RemoteCompletionHandle<'_>` to poll only for completion without extracting the result: can be passed to trait (object) methods as the underlying result type is erased (would be useful in https://github.com/rustasync/runtime/pull/73).

### DIFF
--- a/futures-channel/src/oneshot.rs
+++ b/futures-channel/src/oneshot.rs
@@ -297,6 +297,37 @@ impl<T> Inner<T> {
         }
     }
 
+    fn poll_complete(&self, cx: &mut Context<'_>) -> Poll<()> {
+        // Check to see if some data has arrived. If it hasn't then we need to
+        // block our task.
+        //
+        // Note that the acquisition of the `rx_task` lock might fail below, but
+        // the only situation where this can happen is during `Sender::drop`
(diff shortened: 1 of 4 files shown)


futures-util/src/future/remote_handle.rs
```diff
@@ -36,17 +44,87 @@ impl<T> RemoteHandle<T> {
...
+        Some(unwrap_result(self.rx.try_recv().transpose()?))
+    }
+
+    /// Extract final result
+    ///
+    /// # Panics
+    ///
+    /// Panics if underlying future is still running in addition to the
+    /// ways it would panic when polling `RemoteHandle`.
+    pub fn recv(mut self) -> T {
```
Nemo157: Is there some name that might better indicate this panics if still running, `unwrap_recv` maybe? (I feel like there's a good chance users might assume this does a blocking receive).

stbuehler: Hm. Or I could simply remove the method - `try_recv` should be good enough after all.

PR #1773, opened 2019-08-03
Don't require a full flush before pushing next item in SendAll

### LINKED ISSUE #1760: SendAll should not flush its Sink when it receives Pending from the Stream
[This `poll_flush`](https://github.com/rust-lang-nursery/futures-rs/blob/db4cad36ecbf59868c195738333f6d841b342fa2/futures-util/src/sink/send_all.rs#L90) should be replaced with `poll_ready` instead as all that `SendAll` is interested in is that it can send another item, at the moment it may get a late wakeup if the sink is able to take another item but hasn't _fully_ flushed itself.

I think this is just a consequence of there not being separate `poll_ready` and `start_send` methods when `SendAll` [was initially changed](https://github.com/rust-lang-nursery/futures-rs/blame/669980ba9215e6b1d50a730a3663629135bf2d06/futures-util/src/sink/send_all.rs) to do this loop and nobody realising it could be updated since then.

(Noticed when looking at https://github.com/matthunz/futures-codec/issues/13)

Nemo157:
Fixes #1760

### DIFF
--- a/futures-util/src/sink/send_all.rs
+++ b/futures-util/src/sink/send_all.rs
@@ -87,7 +87,7 @@ where
                     return Poll::Ready(Ok(()))
                 }
                 Poll::Pending => {
-                    ready!(Pin::new(&mut this.sink).poll_flush(cx))?;
+                    ready!(Pin::new(&mut this.sink).poll_ready(cx))?;
                     return Poll::Pending
                 }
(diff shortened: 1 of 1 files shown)


cramertj:
Good catch, thanks!

PR #1772, opened 2019-08-02
Fix warnings on the latest nightly

taiki-e:
This fixes warnings due to new clippy lints added by https://github.com/rust-lang/rust-clippy/pull/3766.

PR #1775, opened 2019-08-04
ci: Use the latest toolchain with clippy available if clippy is unavailable on the latest nightly

taiki-e:
Currently, the check is done only if there is clippy in the latest nighty. This PR change it to use the latest clippy available at that time.

Refs: https://github.com/rust-lang/rustup-components-history#the-web-part


taiki-e:
Build failures should be fixed in #1772.

PR #1778, opened 2019-08-05
document AtomicWaker memory ordering (#942)

stbuehler:
This is certainly no proof, but I tried to help with the part "someone gets pencil and paper out" :)

I don't think not documenting the way we think it should work helps anyone, so at least this documents the expected outcome, whether it is actually true or not.

### DIFF
--- a/futures-core/src/task/__internal/atomic_waker.rs
+++ b/futures-core/src/task/__internal/atomic_waker.rs
@@ -21,9 +21,71 @@ use crate::task::Waker;
 /// A single `AtomicWaker` may be reused for any number of calls to `register` or
 /// `wake`.
 ///
-/// `AtomicWaker` does not provide any memory ordering guarantees, as such the
-/// user should use caution and use other synchronization primitives to guard
-/// the result of the underlying computation.
+/// # Memory ordering
+///
+/// Calling `register` "acquires" all memory "released" by calls to `wake`
+/// before the call to `register`.  Later calls to `wake` will wake the
+/// registered waker (on contention this wake might be triggered in `register`).
+///
+/// For concurrent calls to `register` (should be avoided) the ordering is only
+/// guaranteed for the winning call.
+///
+/// # Examples
+///
+/// Here is a simple example providing a `Flag` that can be signalled manually
+/// when it is ready.
+///
+/// ```
+/// use futures::future::Future;
+/// use futures::task::{Context, Poll, AtomicWaker};
+/// use std::sync::Arc;
+/// use std::sync::atomic::AtomicBool;
+/// use std::sync::atomic::Ordering::Relaxed;
+/// use std::pin::Pin;
+///
+/// struct Inner {
+///     waker: AtomicWaker,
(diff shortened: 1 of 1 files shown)


futures-core/src/task/__internal/atomic_waker.rs
```diff
@@ -214,35 +283,40 @@ impl AtomicWaker {
...
+                            // wake() releasing self.state, all is fine.
                         }
                     }
                 }
             }
             WAKING => {
-                // Currently in the process of waking the task, i.e.,
-                // `wake` is currently being called on the old task handle.
-                // So, we call wake on the new waker
-                waker.wake_by_ref();
```
seanmonstar: With this change, if you called register the same time another thread was waking, the new waker would never get the wakeup.

stbuehler: That is certainly true.  But: if the other thread was waking just a little bit earlier, the new waker wouldn't get the wakeup either - and this is fine, because it doesn't have to: we just acquired a state that was relased by the other thread "waking", so the memory is synced as promised.

stbuehler: Damn. Those wakes are actually needed in case two waking threads ("producers") come along at the same time, and the first produced item has been consumed (without a call to `register`, which can happen). `register` will only sync with the first producer, and the second item might not be visible yet. I'll try to come up with some descriptions of those scenarious...

As far as I can see now:
- the first wake could probably be removed if the registration wasn't undone; all wakes during a registration should be properly synced, only the additional wakes after the registration are not synced (and don't find a registered waker right now).
- the second wake is probably not enough (strictly speaking), as Acquire/Release doesn't establish a transitive sync across threads, but another round through the event loop might simply take enough time.

stbuehler:
So I came up with 3 interesting scenarios, and the first two fail when I remove the explicit wakes.

I think for `A` could try some spinning before yielding, and for `B` we could simply leave the registration in place, but I guess for now I just document why the wakes are needed.

### A. Spurious wake preventing registration

```
wake();                             // spurious wake
    |           pop() -> Pending;
    #           register();         // state locked by wake, registration fails
    |           pop() -> Pending;   // doesn't see push(2)
    |   push(2);
    ?   wake();                     // either registration still locked or empty
```

### B. Spurious wake triggering undo of registration

```
                pop() -> Pending;
                register();
wake();             #               // spurious wake
                    |               // makes unlock in register fail, having it unregister the waker
                pop() -> Pending;   // doesn't see push(2)
        push(2);
        wake();                     // no waker to call
```

### C. Multiple wakes during registration

```
                pop() -> Pending;
                register();
wake();             #               // spurious wake
                    |               // makes unlock in register fail, having it unregister the waker
        push(2);    |
        wake();     |               // registration locked, but syncs on unlock (swap(AcqRel))
                    |
                pop() -> 2;
```

cramertj:
This looks like a great improvement, thanks! Sorry for taking so long to get this merged.

PR #1779, opened 2019-08-06
Add Future/Stream bounds to FusedFuture/FusedStream

taiki-e:
Just as `ExactSizeIterator` and `FusedIterator` have `Iterator` bounds, I think it's preferable for consistency to add `Future`/`Stream` bounds to `FusedFuture`/`FusedStream` as well.


cramertj:
Thanks!

PR #1780, opened 2019-08-06
Change StreamExt::boxed to return BoxStream and add StreamExt::boxed_local

### LINKED ISSUE #1767: `StreamExt::boxed` should return Pin<Box<dyn Stream>>, not Pin<Box<Self>>
This would match the behavior of `FutureExt::boxed`.

taiki-e:
Closes #1767


cramertj:
Nice, thanks!

PR #1784, opened 2019-08-07
Rename ArcWake to Wake

taiki-e:
Since both methods take Arc as an argument, it should not be difficult to understand that this trait is based on Arc. Also, I think this should be near to [the API we want to add to std/alloc in the future](https://github.com/rust-lang/rust/issues/59725#issuecomment-480594110).

### DIFF
--- a/futures-executor/src/local_pool.rs
+++ b/futures-executor/src/local_pool.rs
@@ -2,7 +2,7 @@ use crate::enter;
 use futures_core::future::{Future, FutureObj, LocalFutureObj};
 use futures_core::stream::{Stream};
 use futures_core::task::{Context, Poll, Spawn, LocalSpawn, SpawnError};
-use futures_util::task::{waker_ref, ArcWake};
+use futures_util::task::{waker_ref, Wake};
 use futures_util::stream::FuturesUnordered;
 use futures_util::stream::StreamExt;
 use futures_util::pin_mut;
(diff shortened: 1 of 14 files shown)


cramertj:
I have no other particularly strong feelings about this aside from the fact that this makes the trait sound more canonical, rather than like a convenient helper. cc @seanmonstar @carllerche who may have opinions about this.

LucioFranco:
Is making the ArcWake the Wake the right choice if down the line executors that do not need the atomic ref count?

carllerche:
I don't have any opinion. `ArcWake` is descriptive. It also is probably the strategy that should be used unless there is a reason not to.

cramertj:
> Is making the ArcWake the Wake the right choice if down the line executors that do not need the atomic ref count?

@LucioFranco Given that the `Wake` trait is a convenience which needn't be used by all executors, and that many executors (most?) will, I think, need the refcount, it makes sense to me to make this ergonomic. However, I don't particularly care either way here and would be happy to defer to someone with a stronger opinion :)

LucioFranco:
Yeah, that makes sense, though I think the descriptiveness is actually nice and clear imho. There are a lot of "wake" terms flying around and I think keeping it `ArcWake` just makes more sense. In the end we use this only in a few spots so im not too worried. Just thought I'd leave my opinion :)

PR #1788, opened 2019-08-09
Add an AsyncReadExt::take adapter.

jebrosen:
I wrote this adapter because it is used relatively frequently in `rocket_contrib`, examples, and tests, but I think it deserves a place in `futures-rs`.

Code is based on the synchronous equivalent in std - how/where should I attribute this?

Not ready:
* [x] docs
* [x] tests
* [x] manual `impl Unpin`?
* [ ] attribution to std if necessary

### DIFF
--- a/futures-util/src/io/mod.rs
+++ b/futures-util/src/io/mod.rs
@@ -75,6 +75,9 @@ pub use self::seek::Seek;
 mod split;
 pub use self::split::{ReadHalf, WriteHalf};
 
+mod take;
+pub use self::take::Take;
+
 mod window;
 pub use self::window::Window;
 
@@ -312,6 +315,33 @@ pub trait AsyncReadExt: AsyncRead {
         where Self: AsyncWrite + Sized,
     {
         split::split(self)
+    }
+
+    /// Creates an AsyncRead adapter which will read at most `limit` bytes
+    /// from the underlying reader.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// #![feature(async_await)]
+    /// # futures::executor::block_on(async {
+    /// use futures::io::AsyncReadExt;
+    /// use std::io::Cursor;
+    ///
+    /// let reader = Cursor::new(&b"12345678"[..]);
+    /// let mut buffer = [0; 5];
+    ///
+    /// let mut take = reader.take(4);
+    /// let n = take.read(&mut buffer).await?;
+    ///
+    /// assert_eq!(n, 4);
+    /// assert_eq!(&buffer, b"1234\0");
+    /// # Ok::<(), Box<dyn std::error::Error>>(()) }).unwrap();
+    /// ```
+    fn take(self, limit: u64) -> Take<Self>
+        where Self: Sized + Unpin
+    {
+        Take::new(self, limit)
     }
 
     /// Wraps an [`AsyncRead`] in a compatibility wrapper that allows it to be
--- a/futures-util/src/io/take.rs
+++ b/futures-util/src/io/take.rs
@@ -78,6 +78,84 @@ impl<R: AsyncRead + Unpin> Take<R> {
     pub fn set_limit(&mut self, limit: u64) {
         self.limit = limit
     }
+
+    /// Consumes the `Take`, returning the wrapped reader.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// #![feature(async_await)]
+    /// # futures::executor::block_on(async {
+    /// use futures::io::AsyncReadExt;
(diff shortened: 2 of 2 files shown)


futures-util/src/io/mod.rs
```diff
@@ -314,6 +317,13 @@ pub trait AsyncReadExt: AsyncRead {
         split::split(self)
     }
 
+    /// Creates an adaptor which will read at most `limit` bytes.
```
Nemo157: It would be nice to have an example as a quick smoketest of the functionality.

futures-util/src/io/take.rs
```diff
@@ -0,0 +1,35 @@
...
+use std::pin::Pin;
+
+/// Future for the [`take`](super::AsyncReadExt::take) method.
+#[derive(Debug)]
+#[must_use = "futures do nothing unless you `.await` or poll them"]
+pub struct Take<R: Unpin> {
+    inner: R,
+    limit: u64,
+}
+
```
jebrosen: Does this need a manual `Unpin` impl like some of the other adapters?

futures-util/src/io/mod.rs
```diff
@@ -314,6 +317,33 @@ pub trait AsyncReadExt: AsyncRead {
...
+    /// let mut buffer = [0; 5];
+    ///
+    /// let mut take = reader.take(4);
+    /// let n = take.read(&mut buffer).await?;
+    ///
+    /// assert_eq!(n, 4);
+    /// assert_eq!(&buffer, b"1234\0");
+    /// # Ok::<(), Box<dyn std::error::Error>>(()) }).unwrap();
+    /// ```
+    fn take(self, limit: u64) -> Take<Self>
```
cramertj: Did you consider having this take `&mut self` instead of `self`-by-value?

jebrosen: My (possibly wrong!) assumption is that `Take` is most often used as a "cap" and there's no interest in the rest of the underlying stream. I believe it would still be possible to write `(&mut reader).take(5)`, which would allow for the reuse of `reader` after dropping the `Take` instance. I can add a test for that if it's desired.

cramertj: Yep, that is possible-- both positions seem reasonable to me, so I'm happy to defer to your judgement here.

jebrosen: This test passes with the current API:

```rust
let mut reader = Cursor::new(&b"1234567"[..]);
let mut buffer = [0; 5];
                                               
{
    let mut take = (&mut reader).take(4);
    let n = take.read(&mut buffer).await?;
                                               
    assert_eq!(n, 4);
    assert_eq!(&buffer, b"1234\0");
}
                                               
let n = reader.read(&mut buffer).await?;
assert_eq!(n, 3);
assert_eq!(&buffer, b"5674\0");
```

I did try out a `Take` that uses `&mut R` instead of `R`, but I quickly realized that it requires the caller of `take()` to keep the original reader around -- which is much less flexible for consumers.

futures-util/src/io/take.rs
taiki-e: As u64 [always implements Unpin](https://doc.rust-lang.org/nightly/std/primitive.u64.html#impl-Unpin), manual Unpin implementation should not be necessary here, but it can also be added for clarity:
```rust
impl<R: Unpin> Unpin for Take<R> {}
```

futures-util/src/io/mod.rs
taiki-e: (Personally, I like this to have [the same signature (take `self`)](https://doc.rust-lang.org/nightly/std/io/trait.Read.html#method.take) as std.)

futures-util/src/io/take.rs
```diff
@@ -0,0 +1,35 @@
...
+impl<R: AsyncRead + Unpin> Take<R> {
```
taiki-e: We can probably add some methods [such as `limit`, `set_limit`](https://doc.rust-lang.org/nightly/std/io/struct.Take.html#methods). This can happen in this PR or follow-up work.

jebrosen: Those are trivial additions, I'll go ahead and add them now.

jebrosen: Take also has into_inner, get_ref, and get_mut - I can add those as well.

taiki-e: > Take also has into_inner, get_ref, and get_mut - I can add those as well.

I think they are useful.

jebrosen:
I believe this is ready for another review.

cramertj:
LGTM, thanks!

PR #1789, opened 2019-08-09
State 0.3.0-alpha.18

cramertj:
r? @taiki-e @Nemo157 @seanmonstar

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,7 +21,7 @@
 </p>
 
 <p align="center">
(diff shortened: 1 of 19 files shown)


cramertj:
cc @taiki-e looks like `buffer_unordered::works` is flaking :(

will investigate

cramertj:
In any case, it's not new to this release-- only the test is new, so I'm going to go ahead and merge this.

cramertj:
Opened https://github.com/rust-lang-nursery/futures-rs/issues/1790 to track the flake.

Issue #1790, opened 2019-08-09
Fix flake in buffer_unordered


cramertj:
seen on this change: https://github.com/rust-lang-nursery/futures-rs/pull/1789
in this job: https://travis-ci.com/rust-lang-nursery/futures-rs/jobs/224053387

taiki-e:
I tested it locally but could not reproduce it (Windows and Linux(WSL-Ubuntu)).
As far as I look at history ([pr](https://travis-ci.com/rust-lang-nursery/futures-rs/pull_requests) and [master](https://travis-ci.com/rust-lang-nursery/futures-rs/builds)), this problem seems not to occur in Linux, so I think this is an issue specific to macOS or a Travis issue.

Recent Travis's macOS build is very unstable regardless of this test (https://github.com/rust-lang-nursery/futures-rs/pull/1787#issuecomment-519628394), so for now, I'm guessing this is a Travis issue.

taiki-e:
(The OS looks irrelevant, but I have never been able to reproduce it locally.)

kentfredric:
I've been reproducing this on linux in futures `0.1.29`, the problem is its very ephemeral and prone to randomly working 99% of the time.

I can't make it do this on nightly rust, but I can do it on current stable rust.

Attaching gdb once it deadlocks gives me:
<details>
   <summary><code>gdb thread apply all bt full</code></summary>

```
Thread 4 (Thread 0x7f9d6d721700 (LWP 9458)):
#0  futex_wait_cancelable (private=0, expected=0, futex_word=0x7f9d64000cac) at ../sysdeps/unix/sysv/linux/futex-internal.h:88
        __ret = -512
        oldtype = 0
        err = <optimized out>
        oldtype = <optimized out>
        err = <optimized out>
        __ret = <optimized out>
        resultvar = <optimized out>
        __arg4 = <optimized out>
        __arg3 = <optimized out>
        __arg2 = <optimized out>
        __arg1 = <optimized out>
        _a4 = <optimized out>
        _a3 = <optimized out>
        _a2 = <optimized out>
        _a1 = <optimized out>
#1  __pthread_cond_wait_common (abstime=0x0, mutex=0x7f9d64000b20, cond=0x7f9d64000c80) at pthread_cond_wait.c:502
        spin = 0
[...]

PR #1791, opened 2019-08-10
Temporarily ignore buffer_unordered::works test on macOS

taiki-e:
Prevent other PRs from being blocked by #1790 until #1790 is resolved.

cc @cramertj


taiki-e:
😕 
```text
info: caused by: failed to make network request
```

PR #1792, opened 2019-08-10
Fix date of 0.3.0-alpha.18 release

killercup:
I know this crate is called futures but this still confused me


paulzhang5511:
#1794

cramertj:
Thanks for the fix!

PR #1797, opened 2019-08-12
Fix use-after-free on panic during ArcWake::wake_by_ref

stbuehler:
Wrap temporary `Arc<>`s in `ManuallyDrop` early instead of calling `forget()` later: that way even during unwinding for panics it doesn't drop the refcount it doesn't actually own.

Also it means `wake_by_ref` doesn't need an unwind section anymore.

Same thing in `increase_refcount` (although `Arc::clone` should only abort, not unwind).

### DIFF
--- a/futures-util/src/task/waker.rs
+++ b/futures-util/src/task/waker.rs
@@ -22,12 +22,10 @@ where
 // code here. We should guard against this by aborting.
 
 unsafe fn increase_refcount<T: ArcWake>(data: *const ()) {
-    // Retain Arc by creating a copy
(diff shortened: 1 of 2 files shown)


cramertj:
Nice! Would you mind adding a test for this?

stbuehler:
Added a test (if you prefer it in a separate commit before/after the fix I can split it of course).

The test fails for me without the fix applied:
```
error: process didn't exit successfully: `[...]/futures/target/debug/deps/arc_wake-e1af0914f1200023` (signal: 11, SIGSEGV: invalid memory reference)
```

cramertj:
Wonderful, thank you!

PR #1798, opened 2019-08-13
Update proc-macro2, syn, and quote to 1.0

taiki-e:


taiki-e:
```text
error[E0432]: unresolved import `AttributeArgs`
  --> /home/travis/.cargo/registry/src/github.com-1ecc6299db9ec823/syn-0.15.14/src/parse_macro_input.rs:85:5
   |
85 | use AttributeArgs;
   |     ^^^^^^^^^^^^^ no `AttributeArgs` in the root
```
[Perhaps proc-macro-hack...](https://github.com/dtolnay/proc-macro-hack/blob/f02410c217d027f4b1cc1dd807f6899daea8fb19/Cargo.toml#L17)

cramertj:
Thanks! :)

PR #1803, opened 2019-08-14
ThreadPoolBuilder::pool_size accepts NonZeroUsize

vorot93:
Let's put more invariants into the type system, shall we?

### DIFF
--- a/futures-executor/src/thread_pool.rs
+++ b/futures-executor/src/thread_pool.rs
@@ -5,6 +5,7 @@ use futures_core::task::{Context, Poll, Spawn, SpawnError};
 use futures_util::future::FutureExt;
 use futures_util::task::{ArcWake, waker_ref};
 use std::io;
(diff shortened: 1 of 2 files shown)


vorot93:
Please do remember that your average Joe programmer may assume zero to be an 'auto' (num of CPU threads) and not an invalid value and get the panic. To avoid stepping on this mine, he must read the documentation carefully. This PR moves this semi-unwritten contract into the type system which will lead to less unwanted panics among downstream users.

cramertj:
Thanks for the PR-- I appreciate efforts to express more invariants with the typesystem in order to prevent runtime failures. However, I'm going to close this PR because I think it's a pretty significant ergonomic hit for the majority of users, and I don't think it's going to benefit the typical usage. Users who accidentally misuse the function and provide `0` should immediately get an appropriate panic and be able to easily fix the error.

PR #1806, opened 2019-08-15
Use ManuallyDrop for WakerRef to use same vtable as normal Waker

stbuehler:
This makes `Waker::will_wake` work between `task::waker_ref` and `Wakers` cloned from it (or created by `task::waker`).

As `WakerRef` no longer stores an actual waker (as in owning a refcount), `WakerRef::new` takes now a reference instead of ownership. (Theoretically we could have an inner flag/enum to track ownership vs no ownership, but the struct name clearly suggests it shouldn't own anything.)

Replaces `waker_vtable!` with a `waker_vtable<W>()` function, hopefully making the returned vtable pointer (more) stable.

This also removes the safety guard "wake_unreachable" - but after all it should have been unreachable in the first place.

### DIFF
--- a/futures-util/src/compat/compat03as01.rs
+++ b/futures-util/src/compat/compat03as01.rs
@@ -201,9 +201,9 @@ impl Current {
 
         let ptr = current_to_ptr(self);
         let vtable = &RawWakerVTable::new(clone, wake, wake, drop);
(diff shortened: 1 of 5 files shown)


cramertj:
This is a great improvement, thanks!

PR #1816, opened 2019-08-20
Stabilize async-await feature

taiki-e:
rust-lang/rust#63209 has been merged. Thanks to everyone who has been working on this!

(Description of PR: Currently, this feature requires the nightly feature active to explicitly opt-in to unstable features, but this is no longer needed because rust-lang/rust#63209 has been merged.)

### DIFF
--- a/futures-util/src/async_await/join_mod.rs
+++ b/futures-util/src/async_await/join_mod.rs
@@ -19,7 +19,6 @@ macro_rules! document_join_macro {
         /// # Examples
         ///
         /// ```
-        /// #![feature(async_await)]
         /// # futures::executor::block_on(async {
         /// use futures::{join, future};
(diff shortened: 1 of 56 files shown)


futures-util/src/async_await/join_mod.rs
```diff
@@ -19,8 +19,7 @@ macro_rules! document_join_macro {
         /// # Examples
         ///
         /// ```
-        /// #![feature(async_await)]
-        /// # futures::executor::block_on(async {
+            /// # futures::executor::block_on(async {
```
killercup: accidental indent here and below?

taiki-e: Good catch, thanks.

PR #1820, opened 2019-08-21
Remove Unpin requirements from get_pin_mut and AsyncRead::take

taiki-e:


futures-util/src/io/chain.rs
```diff
@@ -55,6 +50,23 @@ where
...
     }
+
+    /// Gets pinned mutable references to the underlying readers in this `Chain`.
+    ///
+    /// Care should be taken to avoid modifying the internal I/O state of the
+    /// underlying readers as doing so may corrupt the internal state of this
+    /// `Chain`.
+    pub fn get_pin_mut(self: Pin<&mut Self>) -> (Pin<&mut T>, Pin<&mut U>) {
+        unsafe {
+            let Self { first, second, .. } = self.get_unchecked_mut();
```
taiki-e: ([pin-project](https://crates.io/crates/pin-project/0.4.0-alpha.2) allows this to be handled in safe code.)

PR #1826, opened 2019-08-24
Add AsyncRead::poll_read_buf and AsyncWrite::poll_write_buf

### LINKED ISSUE #1818: Add AsyncRead::poll_read_buf and AsyncWrite::poll_write_buf
Add the methods mentioned in https://github.com/rust-lang-nursery/futures-rs/pull/1688.
> Another change I'm thinking about is to add `poll_{read,write}_buf` like tokio's same name methods ([`poll_read_buf`](https://github.com/tokio-rs/tokio/blob/ed4d4a5353d07d2428072965ea23c9a6eba5d87d/tokio-io/src/async_read.rs#L91-L115), [`poll_write_buf`](https://github.com/tokio-rs/tokio/blob/ed4d4a5353d07d2428072965ea23c9a6eba5d87d/tokio-io/src/async_write.rs#L128-L143)). This makes `tokio::io::Async*` and `futures::io::Async*` almost identical, so that tokio probably can use the futures's io traits. (Stabilizing this is probably blocked by bytes 0.5 release, but since `poll_{read,write}_buf` have default implementations, it could be made to work well by adding a feature gate. )

~~Probably blocked by bytes 0.5-alpha.~~

cc https://github.com/tokio-rs/tokio/issues/1297

taiki-e:
Closes #1818

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -128,14 +128,23 @@ matrix:
             --no-default-features
             --features alloc
 
-    - name: cargo check (futures-util)
+    - name: cargo check (sub crates)
       rust: nightly
       script:
         - cargo run --manifest-path ci/remove-dev-dependencies/Cargo.toml */Cargo.toml
 
+        # futures-io
+        # Check default-features, all-features
+        - cargo check --manifest-path futures-io/Cargo.toml
+        - cargo check --manifest-path futures-io/Cargo.toml --all-features
+        # Check each features
+        - cargo check --manifest-path futures-io/Cargo.toml --features bytes,unstable
+
+        # futures-util
+        # Check default-features, all-features
         - cargo check --manifest-path futures-util/Cargo.toml
         - cargo check --manifest-path futures-util/Cargo.toml --all-features
-
+        # Check each features
         - cargo check --manifest-path futures-util/Cargo.toml --features sink
         - cargo check --manifest-path futures-util/Cargo.toml --features io
         - cargo check --manifest-path futures-util/Cargo.toml --features channel
@@ -150,6 +159,7 @@ matrix:
         - cargo check --manifest-path futures-util/Cargo.toml --features sink,bilock,unstable
         - cargo check --manifest-path futures-util/Cargo.toml --features io,bilock,unstable
         - cargo check --manifest-path futures-util/Cargo.toml --features sink,io
+        - cargo check --manifest-path futures-util/Cargo.toml --features bytes,unstable
 
         - cargo check --manifest-path futures-util/Cargo.toml --no-default-features
         - cargo check --manifest-path futures-util/Cargo.toml --no-default-features --features sink
--- a/futures-io/Cargo.toml
+++ b/futures-io/Cargo.toml
@@ -18,7 +18,13 @@ name = "futures_io"
 default = ["std"]
 std = []
 
+# Unstable features
+# `bytes` feature is outside of the normal semver guarantees and require the
+# `unstable` feature as an explicit opt-in to unstable API.
+unstable = []
+
 [dependencies]
+bytes = { version = "0.4.7", optional = true }
 
 [dev-dependencies]
 futures-preview = { path = "../futures", version = "=0.3.0-alpha.18" }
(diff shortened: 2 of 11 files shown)


taiki-e:
I have no strong opinion about whether to block this with bytes 0.5-alpha.

taiki-e:
cc @carllerche @seanmonstar @LucioFranco

LucioFranco:
@yoshuawuyts thank you for spending the time to write this up. I think there are some really good ideas here!

> tokio-rs/bytes doesn't currently seem to be on a path towards moving into the stdlib. This means that whichever additions are made to stdlib, the exposure of bytes in the API cannot be part of it. 

I think this is very reasonable and makes sense. :+1:

> However I'm sympathetic to tokio expressing the intent to using futures-io directly rather than exposing a fork. 

Yes, this is something we really do want to achieve because I think this is and will become a large pain point throughout the ecosystem. The tough part is that we have certain requirements that we want to maintain and this has become hard once the io traits got moved out of tokio. Which is why we still have our own versions since we have not had the full opportunity to explore them yet in relation to new futures.

> Expose Ext traits for tokio-specific functionality

I think this approach would work fine for our use case and maybe the correct approach here. Though I'd like to hear what @taiki-e has to say.

> We have 29 days left until the next Rust beta which will include async/await without a feature flag. I think we should be using the coming weeks before 1.39 to focus on making futures-rs ready for a release.

Now circling back to this section. What are the plans around this? I don't think the ecosystem has had enough time to _really_ explore all the traits besides `std::future::Future` and `Stream`? Do we have a solid and guided path for stabilizing futures? I personally have not seen much around this and of course, I would like to help move this forward. Since this was used as a point to not include this PR I'm curious what other thoughts you might have around stabilizing traits? I don't think we will see io traits within std anytime soon since they are still kind of new in relation to other runtimes etc. I'm happy to move this to another issue but thought I might as well bring it up and maybe its something we can discuss in today's async wg meeting?


Again, thanks for the write-up and I do hope soon we can all live on the same traits 😄

LucioFranco:
So another thing related to the extension trait that I totally forgot was that we can no longer specialize implementations with that method. Meaning we can't override the poll fn. This is something we would like to do but I'm not sure how we can support that without this. 😞

cramertj:
@LucioFranco can you give more information about the kinds of overriding implementations you'd like to provide that the current system doesn't support? Maybe we can find another way to make those work nicely.

carllerche:
@cramertj one use case is that the caller can avoid going through the `IoSlice` path if the read / write implementations do not support it. This is useful when the caller has a complex chain of buffers.

cramertj:
>  it also allows the AsyncRead / AsyncWrite implementation to decide the size of the IoSlice

We discussed this before, and I still don't think i ever quite understood why this is something you'd want-- `Read` implementations don't get to decide the size of the buffer provided to them, so I wouldn't expect `AsyncRead` to do anything different.

cramertj:
I think that's the basic crux of my argument here-- I'm not opposed to adding this as an unstable optional feature, but I'm opposed to ever having functions on `AsyncRead` and `AsyncWrite` that people rely on which we wouldn't expect to see if/when these traits are moved into the standard library.

LucioFranco:
> I'm opposed to ever having functions on AsyncRead and AsyncWrite that people rely on which we wouldn't expect to see if/when these traits are moved into the standard library.

I think this is a perfectly valid thing and I actually agree! But this just leads me to think that these traits may not be ready for `std` yet? I'm happy to help think through this further and I should be attending the next foundation meetings.

So from that, I think it may make sense to close this or throw it behind unstable.

cramertj:
> I think this is a perfectly valid thing and I actually agree! But this just leads me to think that these traits may not be ready for std yet?

I'm not sure how that conclusion relates-- these traits seem fine to me, and correspond pretty precisely to `std::io::Read`. This PR would make them no longer match, which concerns me.

LucioFranco:
Sure, then I say we close this and we can explore, for now, in our own traits.

PR #1829, opened 2019-08-27
Use our own io::{Empy, Repeat, Sink}

taiki-e:
If both `futures::io` ext traits and `std::io` traits are imported,  method names may conflict, so use our own.


futures-util/src/io/empty.rs
```diff
@@ -0,0 +1,64 @@
+use futures_core::task::{Context, Poll};
+use futures_io::{AsyncBufRead, AsyncRead, Initializer};
+use std::fmt;
+use std::io;
+use std::pin::Pin;
+
+/// Stream for the [`empty()`] function.
+#[must_use = "streams do nothing unless polled"]
```
Nemo157: `s/stream/reader` for these two lines, and similar across the other files.

taiki-e: Fixed.

taiki-e:
btw, I'm considering adding our own `Cursor` in another PR. (It was a little big so I separated it from this PR)

PR #1837, opened 2019-08-29
Remove rand dependency by using our own PRNG

### LINKED ISSUE #1807: Cannot use async-await feature with no_std
Attempting to compile a crate depending on `futures` with the `async-await` feature active (for `join!` and `select!`) fails when the target doesn't have `std`.

With the following dependency:

```toml
[dependencies.futures-preview]
version = "0.3.0-alpha.18"
default-features = false
features = ["async-await", "nightly"]
```

compiling fails for `thumbv7m-none-eabi` with:

```console
> cargo build --target thumbv7m-none-eabi
[...]
error[E0463]: can't find crate for `std`
error: Could not compile `futures-core-preview`.
[...]
error[E0463]: can't find crate for `std`
error: Could not compile `lazy_static`.
```

`std` is currently necessary because of the randomness inherent in `select!`, so fixing this will likely overlap with #1805, maybe by providing some way to provide the source of randomness to `select!`.

taiki-e:
Closes #1807

Also, this is a complete fix for #1489 (#1709 partially resolved this).

Refs: 
* `<[_] as rand::SliceRandom>::shuffle`: https://github.com/rust-random/rand/blob/684aa8f8a6de2a077dc3be2bd779bf4d91cc5167/src/seq/mod.rs#L448-L454
* https://en.wikipedia.org/wiki/Xorshift#xorshift*
* https://github.com/rayon-rs/rayon/pull/571

cc @cramertj @Nemo157

### DIFF
--- a/futures-util/src/async_await/random.rs
+++ b/futures-util/src/async_await/random.rs
@@ -1,0 +1,54 @@
+use std::{
+    cell::Cell,
+    collections::hash_map::DefaultHasher,
+    hash::Hasher,
+    num::Wrapping,
+    sync::atomic::{AtomicUsize, Ordering},
+};
+
+// Based on [Fisher–Yates shuffle].
+//
+// [Fisher–Yates shuffle]: https://en.wikipedia.org/wiki/Fisher–Yates_shuffle
+#[doc(hidden)]
+pub fn shuffle<T>(slice: &mut [T]) {
+    for i in (1..slice.len()).rev() {
+        slice.swap(i, gen_index(i + 1));
+    }
+}
+
+/// Return a value from `0..n`.
+fn gen_index(n: usize) -> usize {
+    (random() % n as u64) as usize
+}
+
+/// Pseudorandom number generator based on [xorshift*].
+///
+/// [xorshift*]: https://en.wikipedia.org/wiki/Xorshift#xorshift*
+fn random() -> u64 {
+    thread_local! {
+        static RNG: Cell<Wrapping<u64>> = Cell::new(Wrapping(prng_seed()));
+    }
+
+    fn prng_seed() -> u64 {
+        static COUNTER: AtomicUsize = AtomicUsize::new(0);
+
+        // Any non-zero seed will do
+        let mut seed = 0;
+        while seed == 0 {
+            let mut hasher = DefaultHasher::new();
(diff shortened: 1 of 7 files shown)


futures-util/src/async_await/random.rs
```diff
@@ -0,0 +1,51 @@
+use std::{
+    cell::Cell,
+    collections::hash_map::DefaultHasher,
+    hash::Hasher,
+    num::Wrapping,
+    sync::atomic::{AtomicUsize, Ordering},
+};
+
+#[doc(hidden)]
+pub fn shuffle<T>(slice: &mut [T]) {
```
Nemo157: Maybe it would be worth linking this to https://en.m.wikipedia.org/wiki/Fisher–Yates_shuffle

futures-util/src/async_await/random.rs
```diff
@@ -0,0 +1,51 @@
...
+    fn prng_seed() -> u64 {
+        static COUNTER: AtomicUsize = AtomicUsize::new(0);
+
+        // Any non-zero seed will do
+        let mut seed = 0;
+        while seed == 0 {
+            let mut hasher = DefaultHasher::new();
+            hasher.write_usize(COUNTER.fetch_add(1, Ordering::Relaxed));
+            seed = hasher.finish();
+        }
```
Nemo157: Is there a reason to use an obfuscated deterministic seed here instead of just something like `return 42;`?

Nemo157: (Removing the hasher also looks like it would be enough to make this no-std compatible).

Nemo157: I guess this makes the seed different per-thread, but so would just incrementing a counter used directly as the seed.

taiki-e: There was no particular reason, but I was thinking of using a simpler `cfg(not(feature = "std"))` hasher in no-std.

cramertj:
Seems reasonable to me!

futures-util/src/async_await/random.rs
Nemo157: This doesn't seem all that necessary given `xorshift*`s decent statistical randomness, but it doesn't seem like it hurts too much.

PR #1841, opened 2019-08-30
Add private fields to public types for forward compatibility

taiki-e:


futures-channel/src/oneshot.rs
```diff
@@ -377,8 +377,16 @@ impl<T> Drop for Sender<T> {
 
 /// Error returned from a [`Receiver`](Receiver) when the corresponding
 /// [`Sender`](Sender) is dropped.
-#[derive(Clone, Copy, PartialEq, Eq, Debug)]
```
cramertj: I actually like this one as-is because it allows you to easily write it as a pattern-- e.g.

```rust
match recv.await {
    Ok(x) => ...,
    Err(Canceled) => ...
}
```

I can't really imagine what else we'd want to add to it in the future.

futures-util/src/future/abortable.rs
```diff
@@ -121,16 +122,24 @@ pub fn abortable<Fut>(future: Fut) -> (Abortable<Fut>, AbortHandle)
 }
 
 /// Indicator that the `Abortable` future was aborted.
-#[derive(Copy, Clone, Debug, Eq, PartialEq)]
```
cramertj: Same comment as with `Canceled`-- this is nice for matching on, and I can't forsee wanting to add anything here.

futures-channel/src/oneshot.rs
taiki-e: Okay, these are in util/channel, so we can change them if we need them.

seanmonstar: FWIW, I tend to write it like:

```rust
match rx.await {
    Err(_canceled) => (),
}
```

futures/tests/ready_queue.rs
```diff
@@ -55,13 +55,19 @@ fn resolving_errors() {
 
         drop(tx2);
 
-        assert_eq!(Poll::Ready(Some(Err(oneshot::Canceled))), queue.poll_next_unpin(cx));
```
taiki-e: I think the bad thing about this PR is that we cannot do this.

taiki-e: However, actually, this is not just a problem for this PR. I also encountered this in #1785.

futures-channel/src/oneshot.rs
cramertj: @seanmonstar That doesn't actually assert that cancelled is what happened there, though-- you could be silencing a real error by accident.

PR #1842, opened 2019-08-30
Add stream::from_fn

### LINKED ISSUE #1609: replace `unfold` with `from_fn`
a la https://doc.rust-lang.org/std/iter/fn.from_fn.html

taiki-e:
Closes #1609

### DIFF
--- a/futures-util/src/stream/from_fn.rs
+++ b/futures-util/src/stream/from_fn.rs
@@ -1,0 +1,98 @@
+use core::fmt;
+use core::pin::Pin;
+use futures_core::future::Future;
+use futures_core::stream::Stream;
+use futures_core::task::{Context, Poll};
+use pin_utils::{unsafe_pinned, unsafe_unpinned};
+
+/// Creates a new stream where each iteration calls the provided closure
+/// `F: FnMut() -> Option<T>`.
+///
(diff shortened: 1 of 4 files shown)


futures-util/src/stream/from_fn.rs
```diff
@@ -0,0 +1,98 @@
...
+/// use futures::stream::{self, StreamExt};
+///
+/// let mut count = 0;
+/// let stream = stream::from_fn(|| {
+///     // Increment our count. This is why we started at zero.
+///     count += 1;
+///
+///     // Check to see if we've finished counting or not.
+///     if count < 6 {
+///         future::ready(Some(count))
```
taiki-e: Unfortunately, probably the equivalent of this cannot be written by async/await at this time (from_fn is not very compatible with async / await at this time.).

cramertj:
As I mentioned in #1609, I think that this actually just isn't very useful yet until we get proper support for async closures that can mutate their captures. Is there a place where you wanted to use this and found it better than the existing `unfold` function? I'd rather not add it if it's usually the wrong choice.

taiki-e:
I agreed. As far as I know, there is nothing better than unfold at this time. We should focus on functionality, not appearance.

PR #1844, opened 2019-08-30
Add manually Completable future.

EkardNT:
Adds a new Completable future which must be remotely completed using a CompletionHandle. My use case for adding this is to write an asynchronous long-polling server where some code outside of the scope of the long-poll request needs the ability to control when the long poll request completes. The Output type of the Completable future is Option<T>, with the None case representing the possibility of all handles being dropped without completion ever being manually triggered.

Example:
```rust
    let (handle, future) = futures::future::completable();
    // This thread is an example of how some other process can complete the future via
    // its handle.
    let thread = std::thread::spawn(|| {
        handle.complete(123);
    });

    assert_eq!(future.await, Some(123));
    thread.join().unwrap();
});
```

I have a couple of known issues I'd like feedback on:
- Completable is a fused iterator whose behavior is to return Poll::Ready(None) after the first Future::poll call that returned Poll::Ready(Some(...)). The alternative would be for it to continuously return Poll::Pending instead. I'm not sure if there's a convention that would make Poll::Pending better. I don't have a strong preference either way, and it would be easy to update if needed.
- Currently the future is only available when the std library is available because it depends on Mutex and Arc. If we used a no_std mutex like [spinlock](https://github.com/mvdnes/spin-rs) then we could make the Completable future available on `alloc` targets. But that would bringing in additional dependencies.


seanmonstar:
What is the difference between this proposed type and something like `futures_channel::oneshot`?

EkardNT:
Thanks for pointing out oneshot. Looks very similar, with some minor differences around error reporting and cloning that are not important. I'm going to proceed with using the `oneshot` approach for now and reopen this only if I find a use case that oneshot doesn't handle.

PR #1845, opened 2019-08-30
Make AsyncRead::initializer unstable API

### LINKED ISSUE #1761: Remove AsyncRead::initializer
The [tracking issue](https://github.com/rust-lang/rust/issues/42788) for `std::io::Read::initializer` makes it sound like that API is definitely going to be changed a lot, it would be good to remove the similar API here until the `std` API is closer to its final form.

taiki-e:
Closes #1761

r? @cramertj

### DIFF
--- a/futures-io/src/lib.rs
+++ b/futures-io/src/lib.rs
@@ -8,6 +8,8 @@
 //! All items of this library are only available when the `std` feature of this
 //! library is activated, and it is activated by default.
 
+#![cfg_attr(feature = "read_initializer", feature(read_initializer))]
+
 #![cfg_attr(not(feature = "std"), no_std)]
 
 #![warn(missing_docs, missing_debug_implementations, rust_2018_idioms, unreachable_pub)]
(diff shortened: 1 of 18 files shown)


futures-io/src/lib.rs
```diff
@@ -42,9 +45,11 @@ mod if_std {
 
     /// A type used to conditionally initialize buffers passed to `AsyncRead`
     /// methods, modeled after `std`.
+    #[cfg(feature = "read_initializer")]
     #[derive(Debug)]
     pub struct Initializer(bool);
```
taiki-e: Is it better to re-export the std ones instead of our own ones? (The stability of this feature depends on it.)

taiki-e: Done in https://github.com/rust-lang-nursery/futures-rs/pull/1845/commits/57686fd4726070f756a90dfba2c0c007846598ff

futures-io/src/lib.rs
```diff
@@ -390,12 +361,11 @@ mod if_std {
         }
     }
 
-    /// `unsafe` because the `io::Read` type must not access the buffer
-    /// before reading data into it.
-    macro_rules! unsafe_delegate_async_read_to_stdio {
+    macro_rules! delegate_async_read_to_stdio {
```
taiki-e: This is now safe.

PR #1846, opened 2019-08-30
invert return type of split for consistency

baloo:
`futures::io::AsyncReadExt::split` defines return as:
```
 (ReadHalf<Self>, WriteHalf<Self>)
```

This commit fix inconsistency of `futures::stream::StreamExt::split`

### DIFF
--- a/futures-util/src/stream/mod.rs
+++ b/futures-util/src/stream/mod.rs
@@ -1090,7 +1090,7 @@ pub trait StreamExt: Stream {
         cfg(all(target_has_atomic = "cas", target_has_atomic = "ptr"))
     )]
     #[cfg(feature = "alloc")]
-    fn split<Item>(self) -> (SplitSink<Self, Item>, SplitStream<Self>)
+    fn split<Item>(self) -> (SplitStream<Self>, SplitSink<Self, Item>)
         where Self: Sink<Item> + Sized
     {
         split::split(self)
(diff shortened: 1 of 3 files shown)


taiki-e:
This also matches the order (stream -> sink) in the document.
https://github.com/rust-lang-nursery/futures-rs/blob/90c83b8faca107e6c4db63d10b0d4f2ea36d6628/futures-util/src/stream/mod.rs#L1078-L1079

Nemo157:
The other thing I could imagine this being consistent with is `mpsc::channel` (both in `futures` and `std`) which returns `(Sender, Receiver)` i.e. it is consistent with the original definition. Maybe it would make more sense to swap `AsynReadExt::split` and change the documentation to match the order? Unfortunately this inconsistency already exists in the 0.1 traits so changing either will be an annoying migration hazard.

baloo:
@Nemo157 not too sure I understand what you want me to do.

Nemo157:
I don't have any strong opinion on what the best call would be, just noticed this other potential inconsistency because I'm much more familiar with using the channels than split and wanted to make sure all sides were considered.

baloo:
@Nemo157 @taiki-e who should I contact to have this merged?

taiki-e:
r? @cramertj 

(I don't have right to merge)

cramertj:
I have basically no opinion on this-- does anyone else feel strongly?

baloo:
any update?

cramertj:
I'm personally inclined to leave as-is just to avoid migration pains unless someone feels strongly that the new order used here is obviously preferable.

cramertj:
Closing-- it doesn't seem like anyone feels particularly strongly about this, and it'd be nice to avoid unnecessary breaking changes.

baloo:
fair enough :)

PR #1849, opened 2019-08-31
Add executor::block_on_sink

taiki-e:
Related: https://github.com/rust-lang-nursery/futures-rs/pull/1507#issuecomment-479585935

### DIFF
--- a/futures-executor/Cargo.toml
+++ b/futures-executor/Cargo.toml
@@ -15,10 +15,12 @@ Executors for asynchronous tasks based on the futures-rs library.
 default = ["std"]
 std = ["futures-core/std", "futures-task/std", "futures-util/std"]
 thread-pool = ["std", "num_cpus"]
+sink = ["futures-sink", "futures-util/sink"]
 
 [dependencies]
 futures-core = { path = "../futures-core", version = "0.3.1", default-features = false }
 futures-task = { path = "../futures-task", version = "0.3.1", default-features = false }
+futures-sink = { path = "../futures-sink", version = "0.3.1", default-features = false, optional = true }
 futures-util = { path = "../futures-util", version = "0.3.1", default-features = false }
 num_cpus = { version = "1.8.0", optional = true }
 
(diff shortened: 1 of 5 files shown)


futures-executor/src/local_pool.rs
```diff
@@ -334,6 +337,78 @@ impl<S: Stream + Unpin> Iterator for BlockingStream<S> {
...
+    /// the function returns `Pending` then the current thread is blocked
+    /// until it is otherwise ready to accept the value.
+    ///
+    /// # Return value
+    ///
+    /// If `Ok(())` is returned then the `value` provided was successfully sent
+    /// along the sink, and if `Err(e)` is returned then an error occurred
+    /// which prevented the value from being sent.
+    pub fn send(&mut self, item: Item) -> Result<(), S::Error> {
+        LocalPool::new().run_until(self.sink.send(item))
```
Matthias247: Why not just `block_on`? That should be cheaper.

Matthias247: But more of a general thing: I'm not sure if those wrappers are pullling their weight. Just using `block_on(sink.send(item))` is not much more boilerplate, works for all `Future`s, and doesn't require the original async type to get moved.

cramertj:
Apologies for the delayed review-- I thought we already had this at one point, but perhaps it disappeared. I don't have any objections to adding it. @taiki-e do you want to update this PR and get it landed as part of 0.3?

futures-executor/src/local_pool.rs
taiki-e: > Why not just `block_on`? That should be cheaper.

I just used the same thing as block_on_stream...
https://github.com/rust-lang-nursery/futures-rs/blob/25d638d720bfa128ff4f934d1742bfedfed20c89/futures-executor/src/local_pool.rs#L334

futures-executor/Cargo.toml
```diff
@@ -18,9 +18,11 @@ name = "futures_executor"
 default = ["std"]
 std = ["futures-core-preview/std", "futures-util-preview/std"]
 threadpool = ["num_cpus"]
+sink = ["futures-sink-preview", "futures-util-preview/sink"]
```
taiki-e: Add a feature flag (same as -channel and -util).

taiki-e:
I have no motivation to continue this. Just use `block_on(sink.send(item))`, as Matthias said.

PR #1851, opened 2019-08-31
Improve location of Option wrappers in Map/Chain

danieldg:
If the closure for Map or data for Chain is zero-size, this will avoid
expanding the size of the resulting Future as long as rustc can find a
usable hole in the original future.


taiki-e (commented):
Thanks for the PR and sorry for the delay reviewing.
I'm going to close this as `Chain` was removed by #2128.

PR #1853, opened 2019-09-01
Add Stream::size_hint

### LINKED ISSUE #170: Consider adding `size_hint` to `Stream`
This would be helpful for `Stream::collect`

Also, I often create empty streams to represent no body when working w/ HTTP, it would be nice to just skip all the work if the body is empty.

taiki-e:
Closes #170

r? @cramertj 
cc @Nemo157 @seanmonstar @carllerche @LucioFranco @yoshuawuyts

### DIFF
--- a/futures-core/src/future/mod.rs
+++ b/futures-core/src/future/mod.rs
@@ -57,7 +57,7 @@ mod private_try_future {
 
 /// A convenience for futures that return `Result` values that includes
 /// a variety of adapters tailored to such futures.
-pub trait TryFuture: private_try_future::Sealed {
+pub trait TryFuture: Future + private_try_future::Sealed {
     /// The type of successful values yielded by this future
     type Ok;
 
--- a/futures-core/src/stream.rs
+++ b/futures-core/src/stream.rs
@@ -59,6 +59,37 @@ pub trait Stream {
         self: Pin<&mut Self>,
         cx: &mut Context<'_>,
     ) -> Poll<Option<Self::Item>>;
+
+    /// Returns the bounds on the remaining length of the stream.
+    ///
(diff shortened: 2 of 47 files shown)


futures-core/src/stream.rs
```diff
@@ -126,7 +165,7 @@ mod private_try_stream {
 
 /// A convenience for streams that return `Result` values that includes
 /// a variety of adapters tailored to such futures.
-pub trait TryStream: private_try_stream::Sealed {
+pub trait TryStream: Stream + private_try_stream::Sealed {
```
taiki-e: Added this bound because it would be convenient if `stream.size_hint()` could be called on `St: TryStream`.

futures-core/src/future/mod.rs
```diff
@@ -57,7 +57,7 @@ mod private_try_future {
 
 /// A convenience for futures that return `Result` values that includes
 /// a variety of adapters tailored to such futures.
-pub trait TryFuture: private_try_future::Sealed {
+pub trait TryFuture: Future + private_try_future::Sealed {
```
taiki-e: Added for consistency with `TryStream`.

futures-core/src/stream.rs
taiki-e: oh...
```text
error[E0191]: the value of the associated type `Item` (from the trait `futures_core::stream::Stream`) must be specified
  --> futures/tests/object_safety.rs:20:30
   |
20 |     assert_is_object_safe::<&dyn TryStream<Ok = (), Error = ()>>();
   |                              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ associated type `Item` must be specified
```

cramertj:
Nice! Seems like a nice easy add.

PR #1855, opened 2019-09-01
Remove top-level re-exports

### LINKED ISSUE #1854: Remove top-level re-export of Poll?
https://github.com/rust-lang-nursery/futures-rs/blob/90c83b8faca107e6c4db63d10b0d4f2ea36d6628/futures/src/lib.rs#L63
https://github.com/rust-lang-nursery/futures-rs/blob/90c83b8faca107e6c4db63d10b0d4f2ea36d6628/futures-core/src/lib.rs#L30

Unlike 0.1, traits such as `Future` cannot be implemented with only items exported at the top-level (needs `Context`).

In addition, as Future can be created by async-await, I think the advantage of doing this would be very small.

Related: https://github.com/rust-lang-nursery/futures-rs/pull/1069#discussion_r200817126

cc @seanmonstar @Nemo157

taiki-e:
Closes #1854

### DIFF
--- a/futures-core/src/lib.rs
+++ b/futures-core/src/lib.rs
@@ -20,14 +20,13 @@ compile_error!("The `cfg-target-has-atomic` feature requires the `unstable` feat
 extern crate alloc;
 
 pub mod future;
-#[doc(hidden)] pub use self::future::{Future, FusedFuture, TryFuture};
 
 pub mod stream;
-#[doc(hidden)] pub use self::stream::{Stream, FusedStream, TryStream};
 
 #[macro_use]
 pub mod task;
 
+// Not public API.
 #[doc(hidden)]
 pub mod core_reexport {
     #[doc(hidden)]
--- a/futures-core/src/lib.rs
+++ b/futures-core/src/lib.rs
@@ -27,7 +27,6 @@ pub mod stream;
 
 #[macro_use]
 pub mod task;
-#[doc(hidden)] pub use self::task::Poll;
 
 #[doc(hidden)]
 pub mod core_reexport {
--- a/futures-test/src/lib.rs
+++ b/futures-test/src/lib.rs
@@ -1,4 +1,4 @@
-//! Utilities to make testing [`Future`s](futures_core::Future) easier
+//! Utilities to make testing [`Future`s](futures_core::future::Future) easier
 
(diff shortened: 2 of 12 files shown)


Nemo157:
What’s the feeling on removing all the `doc(hidden)` top-level reexports? It seems very odd to me to have undocumented exports for users to just accidentally stumble across.

taiki-e:
> What’s the feeling on removing all the `doc(hidden)` top-level reexports?

Personally I don't think I should depend on APIs that are not displayed in the documentation, so I would like to remove them, but unfortunately, [some of the current documentation depends on these reexports.](https://github.com/rust-lang-nursery/futures-rs/pull/1577#issue-275041960)

> It seems very odd to me to have undocumented exports for users to just accidentally stumble across.

I strongly agree.

taiki-e:
So, I added a commit to reduce some top-level re-exports and changed the title of PR.

taiki-e:
If we want to completely remove the remaining top-level reexports, you probably need to merge `try{future, stream}` module into `{future, stream}` module. (But this was what I was thinking of actually making a suggestion elsewhere.)
I will add a commit for this later.

taiki-e:
Nah, it's too big to change with this PR, so split it into another PR.

taiki-e:
Hmm, has intra-doc been improved? It seems that all top-level reexports can be removed.

cramertj:
I personally use these pretty often and find having to import from two different modules just to get `TryFutureExt` and `TryStreamExt` pretty annoying. If no one else is bothered by this, I don't feel strongly enough to block, but I do think it's a decent-sized ergonomic hurdle.

taiki-e:
@cramertj Okay, I separated the part not related to ext trait into #1858.

> I personally use these pretty often and find having to import from two different modules just to get `TryFutureExt` and `TryStreamExt` pretty annoying.

Can this be resolved by merging `try_{future, stream}` modules into `{future, stream}` modules.?

cramertj:
That would help, but IMO `use futures::{future::TryFutureExt, stream::TryStreamExt};` is still pretty noisy and the `future` and `stream` prefixes seem to stutter.

taiki-e:
I noticed that the current one is useful when using with macro.

```rust
use futures::{select, join, FutureExt, TryFutureExt};
```

Also, the removal of reexports of items other than traits that I originally wanted to do is included in #1858.
So, I'm going to close this PR.

PR #1858, opened 2019-09-03
Reduce top-level re-exports

### LINKED ISSUE #1854: Remove top-level re-export of Poll?

taiki-e:
Non-controversial part of #1855

Closes #1854

r? @cramertj

### DIFF
--- a/futures-core/src/lib.rs
+++ b/futures-core/src/lib.rs
@@ -27,8 +27,8 @@ pub mod stream;
 
 #[macro_use]
 pub mod task;
-#[doc(hidden)] pub use self::task::Poll;
 
+// Not public API.
 #[doc(hidden)]
 pub mod core_reexport {
     #[doc(hidden)]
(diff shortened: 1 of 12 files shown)


futures-core/src/lib.rs
```diff
@@ -20,15 +20,13 @@ compile_error!("The `cfg-target-has-atomic` feature requires the `unstable` feat
 extern crate alloc;
 
 pub mod future;
-#[doc(hidden)] pub use self::future::{Future, FusedFuture, TryFuture};
```
cramertj: Not sure if I agree that removing these is uncontroversial-- I use these imports frequently.

taiki-e: Reverted traits related changes.

PR #1860, opened 2019-09-05
0.1.29 release

cramertj:
r? @taiki-e @dtolnay


cramertj:
Thanks for taking a look @seanmonstar and @taiki-e!

PR #1862, opened 2019-09-08
Merge futures_util::try_{future, stream} into futures_util::{future, stream}

### LINKED ISSUE #1861: Merge futures_util::try_{future, stream} into futures_util::{future, stream}?
Currently, `futures-util` has a different module layout than `futures-core` and `futures`.

```rust
// futures
use futures::future::{Future, FutureExt, TryFuture, TryFutureExt};
// futures_util
use futures_util::{future::{Future, FutureExt}, try_future::{TryFuture, TryFutureExt}}
// futures_core
use futures::future::{Future, TryFuture};
```

It may be preferable to merge `try_{future, stream}` modules into `{future, stream}` modules and unify module layout with `futures-core` and `futures`.

```rust
use futures_util::future::{Future, FutureExt, TryFuture, TryFutureExt};
```

Related: https://github.com/rust-lang-nursery/futures-rs/pull/1855#issuecomment-527573231

taiki-e:
As ext traits are huge, they are split into individual files.

Closes #1861

r? @cramertj 
cc @Nemo157 @seanmonstar


taiki-e:
Hmm, I feel that there are too many files in `futures-util/future/` and `futures-util/stream/`.
It should probably split the adapter into directories for each ext traits.

taiki-e:
However, before doing that, I would like to hear maintainers' opinions as to whether we should go in this direction.

cramertj:
Sorry for taking so long to get back to you on this! I don't really have a strong opinion either way in terms of organization. If one approach is easier for you to work with, I'm fine with that. Otherwise, I find it sort of helpful to separate them to keep track of which combinators are coming from which extension traits.

taiki-e:
Thanks for the reply. I'll update this PR to separate them.

taiki-e:
@cramertj Updated

Issue #917, opened 2018-03-29
Format entire code base with cargo fmt


raviqqe:
It's better to format source files with `cargo fmt` in order to keep their style consistent.

[`rustfmt`](https://github.com/rust-lang-nursery/rustfmt) has multiple versions currently like `rustfmt-preview` or `rustfmt-nightly` and I don't know which we should use.

cramertj:
I'm not particularly interested in this change.

PR #1867, opened 2019-09-15
Run rustfmt over codebase

udoprog:
This is one of the few repos I have to disable automatic formatting for.

CC: #917

### DIFF
--- a/futures-channel/benches/sync_mpsc.rs
+++ b/futures-channel/benches/sync_mpsc.rs
@@ -7,8 +7,8 @@ use {
     futures::{
         channel::mpsc::{self, Sender, UnboundedSender},
         ready,
-        stream::{Stream, StreamExt},
         sink::Sink,
+        stream::{Stream, StreamExt},
         task::{Context, Poll},
     },
     futures_test::task::noop_context,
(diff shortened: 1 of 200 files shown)


taiki-e:
This kills a lot of the work I'm doing. Please wait for at least the crate to stabilize.

udoprog:
These failures seem to be unrelated to the formatting. I can reproduce them locally with `cargo test` on the current master:

```
failures:

---- join_size stdout ----
thread 'join_size' panicked at 'assertion failed: `(left == right)`
  left: `28`,
 right: `32`', futures\tests\async_await_macros.rs:177:5
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace.

---- try_join_size stdout ----
thread 'try_join_size' panicked at 'assertion failed: `(left == right)`
  left: `28`,
 right: `32`', futures\tests\async_await_macros.rs:193:5


failures:
    join_size
    try_join_size
```

udoprog:
> This kills a lot of the work I'm doing. Please wait for at least the crate to stabilize.

FWIW, in principle you should be able to rustfmt your branch as well and then apply your changes `git checkout my-branch .` over the formatted master and have the same result.

udoprog:
Note that the opposite is _not_ true, and I've already lost two patches in progress due to accidentally running `rustfmt` for one reason or another :cry:

taiki-e:
If I remember correctly, it should depend on the type of change.

taiki-e:
By the way, the current setting of rustfmt.toml is different from what we actually use. This is another reason I don't want this PR to be merged now (I personally want to remove rustfmt.toml).

taiki-e:
By the way, if conflicts with editor settings are a problem, I think adding `disable_all_formatting = true` to `rustfmt.toml` is also a good way.

cramertj:
I'd like to hold off on this for the time being.

PR #1868, opened 2019-09-15
Add FusedFuture implementations

taiki-e:
Related: #1831

r? @cramertj

### DIFF
--- a/futures-test/src/future/pending_once.rs
+++ b/futures-test/src/future/pending_once.rs
@@ -45,7 +45,7 @@ impl<Fut: Future> Future for PendingOnce<Fut> {
     }
 }
 
-impl<Fut: Future + FusedFuture> FusedFuture for PendingOnce<Fut> {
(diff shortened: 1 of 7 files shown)


futures-util/src/future/abortable.rs
```diff
@@ -155,14 +159,20 @@ impl<Fut> Future for Abortable<Fut> where Fut: Future {
...
             return Poll::Ready(Err(Aborted))
         }
 
         Poll::Pending
     }
 }
 
+impl<Fut> FusedFuture for Abortable<Fut> where Fut: FusedFuture {
+    fn is_terminated(&self) -> bool {
+        self.is_aborted() || self.future.is_terminated()
```
cramertj: I'm not sure if this is what we want-- the future may be aborted but not yet have yielded a value indicating that it has been aborted.

cramertj:
These all look good to me aside from the `Abortable` one, which I'm not sure about.

taiki-e:
Removed changes related to `Abortable`.

PR #1869, opened 2019-09-15
Update asserted size of join! and try_join!

taiki-e:


Matthias247:
Should we maybe update the assertions to `<= expected_size` to make this less susceptible to break?

cramertj:
@Matthias247 nope, I think it's helpful to know when these change and track it so we can report any regressions.

PR #1872, opened 2019-09-15
Support std::io::{Empty, Sink}

vorner:
Add async support for the „stub“ helpers of std::io.

Some motivation can be found here: https://vorner.github.io/2019/09/15/play-with-new-async.html#switching-to-tokio ‒ I find these helpers quite useful sometimes.


taiki-e:
I think you missed https://github.com/rust-lang-nursery/futures-rs/pull/1829.

vorner:
Sorry, I must have missed it, yes. I was checking by the version available on docs.rs.

PR #1873, opened 2019-09-15
Allow !Unpin Futures in select!

### LINKED ISSUE #1811: Make `futures::select!` not require it's arguments to be `Unpin`
I don't know if this restriction is due to the intended semantics of `futures::select!` or just due to its implementation. Either way, it's a huge PITA.

The consequence of requiring select's arguments to be `Unpin` is that you can't use `async fn`s directly in a call the `select!`, you have to pin them separately. This is usually a minor hurdle but sometimes it's a major hurdle.

I was recently in a situation where I had an `async fn` that takes a mutable reference:

```rust
async fn do_thing(foo: &mut Foo)
```

I wanted to call it using `select!`, but I also wanted to mutate its argument in a separate branch.

```rust
futures::select! {
    x = do_thing(&mut foo) => { ... },
    y = something_else() => {
        foo.mutate();
    },
}
```

This would be safe, since dropping the future returned by `do_thing(&mut foo)` should release the borrow. `select!` normally allows these kinds of cross-borrows between branches. However the `Unpin` requirement means this doesn't work and I had to pin the call to `do_thing`.

```rust
let do_thing_future = do_thing(&mut foo);
pin_mut!(do_thing_future);
futures::select! {
    x = do_thing_future => { ... },
    y = something_else() => {
        foo.mutate();
    },
}
```

Now the code is broken because the `&mut foo` borrow outlives the entire `select!` block. I can't see any obvious way around this. I tried re-implementing `do_thing` in terms of future combinators but that ended up creating a `Future` type that required dropping and so would also cause the borrow to outlive the `select!` block. In the end, the only way I could find to make this work, was to implement `do_thing` as a function which returns a struct which implements `Future` directly. Having to write `poll` implementations is error-prone and very cumbersome so it sucks that I couldn't find a way to avoid it.

Is the `Unpin` requirement strictly necessary? If it is, is there any way we could change the semantics of `select!` to make it unnecessary (or perhaps introduce two `select!` macros, one which requires `Unpin` and one which is somehow less featureful but doesn't require `Unpin`)?

Matthias247:
This adds support for directly `select!`ing on !Unpin Futures as long
as those Futures are returned by expressions by stack-pinning the
Futures.

For the common case where the futures are the result of an `async fn`
call this will increase the ergonomics of the macro a lot, since no
explicit binding and stack pinning is required.

Partially resolves #1811 (it does not affect the borrow-checking
requirements)

### DIFF
--- a/futures-select-macro/src/lib.rs
+++ b/futures-select-macro/src/lib.rs
@@ -168,13 +168,25 @@ pub fn select(input: TokenStream) -> TokenStream {
         .zip(variant_names.iter())
         .map(|(expr, variant_name)| {
             match expr {
-                // Don't bind futures that are already a path.
-                // This prevents creating redundant stack space
-                // for them.
-                syn::Expr::Path(path) => path,
+                syn::Expr::Path(path) => {
+                    // Don't bind futures that are already a path.
+                    // This prevents creating redundant stack space
+                    // for them.
+                    path
+                },
                 _ => {
+                    // Bind and pin the resulting Future on the stack. This is
+                    // necessary to support direct select! calls on !Unpin
+                    // Futures.
+                    // Safety: This is safe since the lifetime of the Future
+                    // is totally constraint to the lifetime of the select!
+                    // expression, and the Future can't get moved inside it
+                    // (it is shadowed).
                     future_let_bindings.push(quote! {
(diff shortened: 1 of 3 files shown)


Matthias247:
This has already been merged as part of #1874

PR #1874, opened 2019-09-16
Improve support for mutable borrows in select! blocks

### LINKED ISSUE #1811: Make `futures::select!` not require it's arguments to be `Unpin`

Matthias247:
Improve support for mutable borrows in select! blocks  …
As reported in #1811 mutable accesses were not allowed inside select!
blocks if a Future inside the same expression borrowed it. This is not
necessarily required, since the actual polling of the Futures occurs
before the result blocks are evaluated.

This change allows borrowing a variable inside a Future and referring to
the same variable inside other select branches. The change introduces
additional scoping into the select macro which separates the future poll
phase from executing the result blocks phase.

Fixes #1811

### DIFF
--- a/futures-select-macro/src/lib.rs
+++ b/futures-select-macro/src/lib.rs
@@ -173,21 +173,28 @@ pub fn select(input: TokenStream) -> TokenStream {
                     // This prevents creating redundant stack space
                     // for them.
                     // Passing Futures by path requires those Futures to implement Unpin.
+                    // We check for this condition here in order to be able to
+                    // safely use Pin::new_unchecked(&mut #path) later on.
+                    future_let_bindings.push(quote! {
+                        #futures_crate::async_await::assert_fused_future(&mut #path);
+                        #futures_crate::async_await::assert_unpin(&mut #path);
+                    });
                     path
                 },
                 _ => {
                     // Bind and pin the resulting Future on the stack. This is
                     // necessary to support direct select! calls on !Unpin
-                    // Futures.
+                    // Futures. The Future is not explicitly pinned here with
+                    // a Pin call, but assumed as pinned. The actual Pin is
+                    // created inside the poll() function below to defer the
+                    // creation of the temporary pointer, which would otherwise
+                    // increase the size of the generated Future.
                     // Safety: This is safe since the lifetime of the Future
                     // is totally constraint to the lifetime of the select!
                     // expression, and the Future can't get moved inside it
                     // (it is shadowed).
                     future_let_bindings.push(quote! {
                         let mut #variant_name = #expr;
-                        let mut #variant_name = unsafe {
-                            ::core::pin::Pin::new_unchecked(&mut #variant_name)
-                        };
                     });
                     parse_quote! { #variant_name }
                 }
@@ -199,8 +206,19 @@ pub fn select(input: TokenStream) -> TokenStream {
     // to use for polling that individual future. These will then be put in an array.
     let poll_functions = bound_future_names.iter().zip(variant_names.iter())
         .map(|(bound_future_name, variant_name)| {
+            // Below we lazily create the Pin on the Future below.
+            // This is done in order to avoid allocating memory in the generator
+            // for the Pin variable.
+            // Safety: This is safe because one of the following condition applies:
+            // 1. The Future is passed by the caller by name, and we assert that
+            //    it implements Unpin.
+            // 2. The Future is created in scope of the select! function and will
+            //    not be moved for the duration of it. It is thereby stack-pinned
             quote! {
                 let mut #variant_name = |__cx: &mut #futures_crate::task::Context<'_>| {
+                    let mut #bound_future_name = unsafe {
+                        ::core::pin::Pin::new_unchecked(&mut #bound_future_name)
+                    };
                     if #futures_crate::future::FusedFuture::is_terminated(&#bound_future_name) {
                         None
                     } else {
--- a/futures-select-macro/src/lib.rs
+++ b/futures-select-macro/src/lib.rs
@@ -172,6 +172,7 @@ pub fn select(input: TokenStream) -> TokenStream {
                     // Don't bind futures that are already a path.
                     // This prevents creating redundant stack space
                     // for them.
+                    // Passing Futures by path requires those Futures to implement Unpin.
                     path
                 },
                 _ => {
--- a/futures-select-macro/src/lib.rs
+++ b/futures-select-macro/src/lib.rs
@@ -246,21 +246,32 @@ pub fn select(input: TokenStream) -> TokenStream {
         #complete_branch
     };
 
-    let await_and_select = if let Some(default_expr) = parsed.default {
+    let await_select_fut = if parsed.default.is_some() {
+        // For select! with default this returns the Poll result
         quote! {
-            if let #futures_crate::task::Poll::Ready(x) =
-                __poll_fn(&mut #futures_crate::task::Context::from_waker(
-                    #futures_crate::task::noop_waker_ref()
-                ))
-            {
-                match x { #branches }
-            } else {
-                #default_expr
-            };
+            __poll_fn(&mut #futures_crate::task::Context::from_waker(
+                #futures_crate::task::noop_waker_ref()
+            ))
         }
     } else {
         quote! {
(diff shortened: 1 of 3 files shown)


Matthias247:
This builds on top of #1873 , which is also included in this change. If #1873 is separately merged I can rebase this one.

This also introduces some tests for using `select!` as an expression, which didn't seem to work before. It now works, but I'm actually not sure whether it was in that state before intentionally.

Matthias247:
What's your take on this @cramertj , @Nemo157 ?

futures-select-macro/src/lib.rs
```diff
@@ -168,13 +168,25 @@ pub fn select(input: TokenStream) -> TokenStream {
         .zip(variant_names.iter())
         .map(|(expr, variant_name)| {
             match expr {
-                // Don't bind futures that are already a path.
-                // This prevents creating redundant stack space
-                // for them.
-                syn::Expr::Path(path) => path,
+                syn::Expr::Path(path) => {
+                    // Don't bind futures that are already a path.
```
cramertj: Can you also add a comment here explaining that these futures must still implement `Unpin`?

cramertj: Additionally, extending the `select!` doc comment to explain this would be super useful.

futures-util/src/async_await/select_mod.rs
```diff
@@ -8,8 +8,15 @@ macro_rules! document_select_macro {
     ($item:item) => {
         /// Polls multiple futures and streams simultaneously, executing the branch
         /// for the future that finishes first. If multiple futures are ready,
-        /// one will be pseudo-randomly selected at runtime. Futures passed to
-        /// `select!` must be `Unpin` and implement `FusedFuture`.
+        /// one will be pseudo-randomly selected at runtime. Futures directly
+        /// passed to `select!` must be `Unpin` and implement `FusedFuture`.
+        ///
+        /// If an expression which yields a `Future` is passed to `select!`
+        /// (e.g. an `async fn` call) instead of a `Future` directly the `Unpin`
```
cramertj: rather than "a `Future` directly"` maybe say "by name" or something similar? Also, and example would probably help a lot here.

futures-select-macro/src/lib.rs
```diff
@@ -168,13 +168,25 @@ pub fn select(input: TokenStream) -> TokenStream {
...
+                    // Bind and pin the resulting Future on the stack. This is
+                    // necessary to support direct select! calls on !Unpin
+                    // Futures.
+                    // Safety: This is safe since the lifetime of the Future
+                    // is totally constraint to the lifetime of the select!
+                    // expression, and the Future can't get moved inside it
+                    // (it is shadowed).
                     future_let_bindings.push(quote! {
                         let mut #variant_name = #expr;
+                        let mut #variant_name = unsafe {
```
cramertj: Rather than creating an actual mutable reference here, which will cost extra stack space and indirection, what if we had the branch above (for direct bindings) add something like `future_let_bindings.push(quote! { #futures_crate::assert_type_behind_mut_ref_is_unpin(&mut path)})`. Then later we can use `Pin::new_unchecked` in each call to `poll`, since we already know each value being used must either be `Unpin` or be pinned.

futures-select-macro/src/lib.rs
```diff
@@ -246,58 +246,74 @@ pub fn select(input: TokenStream) -> TokenStream {
         #complete_branch
     };
 
-    let await_and_select = if let Some(default_expr) = parsed.default {
+    let await_select_fut = if parsed.default.is_some() {
```
cramertj: Nice! This is actually how `select!` used to work, but then when I reorganized to introduce randomization I dropped support for this. Thanks for adding it back.

cramertj:
> What's your take on this @cramertj , @Nemo157 ?

Thanks for the ping, and apologies for the delayed review. I was initially kind of skeptical about doing this, but I actually like how this turned out quite a bit-- AFAICT once my comments are addressed there's no regression to existing usecases, and there's a big ergonomic bump for the `select!`-once case.

I can imagine, though, that it'd be confusing to pull a future out into a named variable and have that cause your program to stop compiling, so I think this needs some more documentation and examples to make that behavior clear.

Thanks so much for the PR!

futures-select-macro/src/lib.rs
Matthias247: Done

futures-select-macro/src/lib.rs
Matthias247: Interesting idea. But I'm not sure whether it's really worth it. The pin creation here is just a pointer, and I would assume the compiler to observe where this is really used and it can also just put it into registers. I won't expect this to use any extra stack. But maybe we can confirm it.

The main downside of this proposal for me would be that we split the spread out the `unsafe` code over more places: We would introduce a reliance of the `poll()` call on the setup having verified all preconditions.

futures-util/src/async_await/select_mod.rs
Matthias247: I added examples for using `select!` with expressions as well as with identifiers. Also added some other clarifications.

futures-select-macro/src/lib.rs
cramertj: It will not do that-- since it's being held across an `await`, it will most definitely use not just stack space but increase the size of the generated future.

Matthias247: I implemented this now in a branch at https://github.com/Matthias247/futures-rs/compare/fix_select_mutability...Matthias247:fix_select_mutability_2?expand=1

Is that what you had been thinking about? I can't see it making any difference in the generated future size. It is 16 bytes for both approaches in the test where I added an assertion. If that's true then it doesn't matter which way we go. Then just let me know if I should add the commit too, or if we merge as it is.

If I'm missing something let me know too.

cramertj: It isn't changing size there because there's a `default` branch, which means that `async` block contains no `await`s and its size is only determined by the pointers to the variables it captures + the generator state index. If you remove the default branch, the size is 48 with that change, and 64 without it.

Matthias247: Oh ok. I unfortunately missed that part. I will add that commit then.

cramertj: Thanks! <3

cramertj: Thanks! <3

Matthias247: updated

cramertj:
Thanks again!

Issue #1834, opened 2019-08-29
Impl Sink for With should buffer the items


hosyvietanh:
I recently faced an issue when using `Sink` to send websocket messages.
I used `sink.with` in order to modify the msg before actually sending it. The items are not buffered which caused some messages being lost.
https://github.com/rust-lang-nursery/futures-rs/blob/e9cd539cc977fda701f6f4679ee435fa66986ed8/futures-util/src/sink/with.rs#L171

I'm not sure if I understand the purpose of `with` correctly, but I think it should buffer the items when calling `start_send` instead of replacing the old ones.

Nemo157:
There is definitely a bug because is `With::poll` not calling `self.sink.poll_ready` before it attempts to send the buffered item, but if the underlying sink isn’t ready to accept it then that should result in an error rather than missing items, so there may be something else I’m missing.

Nemo157:
Actually, I guess `With` itself doesn’t guarantee an error if you start sending before it’s ready, so maybe the underlying sink could be similar. Unfortunately I don’t see any way for `With` to create an error if `start_send` is called while it has a buffered item...

cramertj:
I think #1880 is all we're planning to do on this-- let me know if more is wanted here.

PR #1880, opened 2019-09-21
With: Ensure inner sink is ready before sending item

tmiasko:
Previously the `With` implementation never consulted the inner sink
`poll_ready` before initiating a send operation.

After changes the `With` combinator is considered ready when processing
of previous item has been completed and the inner sink is ready.

Additionally remove buffered state since it isn't actually used.

Issue: #1834

### DIFF
--- a/futures-util/src/sink/with.rs
+++ b/futures-util/src/sink/with.rs
@@ -1,6 +1,5 @@
 use core::fmt;
 use core::marker::PhantomData;
-use core::mem;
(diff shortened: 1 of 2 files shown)


cramertj:
Nice, thanks for the PR! Can you add a test to ensure that this behavior doesn't regress?

tmiasko:
Added a regression test & rebased.

cramertj:
Thanks!

PR #1881, opened 2019-09-22
Adds unbonded select

CJP10:
In a project I am working on I need the stream returned from select to close if either of the input streams are closed.

I figured this would be useful to someone else, so here I am.

I'm not sure about the name `unbonded` but I'm totally open to suggestions.


414owen:
@CJP10 This would be very useful for me! I'd love to see this merged. Surprised nobody else has needed this.

Looks like this needs to be updated to work with `SelectWithStrategy`. Not sure what the ideal semantics should be exactly, but at a brief glance, it looks like maybe `PollNext` could be generalized to have a `Nothing` option, to prevent future polling?

@taiki-e, if this code is updated and polished, would you be open to accepting the feature, and merging?

> I'm not sure about the name unbonded but I'm totally open to suggestions.

How about `select_short_circuit`, or `select_until_either_terminates`?

---

As for my use case, I have a stream of data that's being written to a tcp socket, with long waits between elements, and I have a stream representing a `heartbeat` to keep the connection alive. I don't want to keep sending heartbeats once the data stream terminates.

CJP10:
Similar to my use case. I'm open to refreshing the PR to be mergeable. It's be open for a while so might have to start from scratch.

414owen:
I've drafted something here: https://github.com/414owen/futures-rs/pull/1
It builds on https://github.com/rust-lang/futures-rs/pull/2583, so any reviews on that would be appreciated too!

Issue #1884, opened 2019-09-25
Release `0.3.0-alpha.19`


bmwill:
The last release of the futures-preview crate happened back in the beginning of August. Since then the `async_await` feature has been stabilized on nightly as well as in the 1.39.0 beta and it would be nice to be able to build this crate using the recently released beta compiler. So it would be great to do another release that includes #1816.

cc @cramertj @taiki-e

cramertj:
Sounds good! I'll put one together now.

PR #1885, opened 2019-09-26
Stage 0.3.0-alpha.19

cramertj:
r? @taiki-e @Nemo157 @seanmonstar 
Fix https://github.com/rust-lang-nursery/futures-rs/issues/1884


taiki-e:
(The build failure is https://github.com/rust-lang/cargo/issues/7427, so it is not related to this PR.)

Issue #1805, opened 2019-08-15
Provide alternative precedences for select!


gavrie:
The `select!` macro currently selects a random future in case multiple futures are ready.
It would be useful to be able to specify the precedence, for example choosing the first future like the existing behavior of `Select`.

This could be implemented by providing a parameter or with a separate wrapper macro such as `select_first!`.

Of course it's possible to work around this by not using the `select!` macro or by calling it multiple times, but it seems useful to have.

seanmonstar:
There is `select` and `selectUnbiased` in Kotlin, where the default doesn't use random, and a user opts in explicitly for randomness.

This differs from the current implementation, which I believe chooses to use randomness implicitly following Golang's lead.

(I think I'm in favor of random being opt-in, personally.)

Matthias247:
I would like to like to see it too! One benefit of that would be also that the `unbiased` version would work in nostd, due to not having a dependency on a random generator.

Regarding what the default should be, I am undecided:
- I understand Go opting for randomizing as a default since it avoids starvation/deadlocks if the same path is always ready. The `unbiased` method requires users to understand that.
- However it's often helpful being able to be able to specify priorities. E.g. a result always having a higher priority then a timeout. Or a value from one channnel (e.g. the control channel) being more important than from another one (the data channel).  I think some people even expect the order to reflect priorities when they write their code. The non randomized version will also perform slightly better.

In the current state maybe adding a `select_biased/select_with_priority` macro might be the favorable option for purely the reason that it won't break existing code.

mzabaluev:
Would it make sense to have a macro for equal priority round robin polling? Like `stream::Select`, but for a small static set of streams which may be optimized better than `SelectAll`: a `u32` value is enough to hold the round robin index and closed state flags for up to 27 streams.

mzabaluev:
> better than `SelectAll`

It would also allow constituents of different types and do not require `Unpin`.

PR #1891, opened 2019-09-28
Allow async-await macros to be used without std

arcnmx:
`select!()` is the only macro that requires std due to implementation details, so allow everything else through.

### DIFF
--- a/futures-util/src/async_await/mod.rs
+++ b/futures-util/src/async_await/mod.rs
@@ -31,8 +31,10 @@ mod select_mod;
 #[cfg(feature = "async-await-macro")]
 pub use self::select_mod::*;
 
+#[cfg(feature = "std")]
 #[cfg(feature = "async-await-macro")]
 mod random;
+#[cfg(feature = "std")]
 #[cfg(feature = "async-await-macro")]
 pub use self::random::*;
 
--- a/futures-util/src/async_await/select_mod.rs
+++ b/futures-util/src/async_await/select_mod.rs
@@ -306,6 +306,7 @@ macro_rules! document_select_macro {
 }
 
 document_select_macro! {
+    #[cfg(feature = "std")]
     #[proc_macro_hack(support_nested)]
     pub use futures_macro::select;
 
--- a/futures-util/src/lib.rs
+++ b/futures-util/src/lib.rs
@@ -39,12 +39,10 @@ pub use futures_core::ready;
 pub use pin_utils::pin_mut;
 
 // Not public API.
-#[cfg(feature = "std")]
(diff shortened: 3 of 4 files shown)


arcnmx:
I wanted to also address `random` on no-std but it seemed a bit too intrusive of a change for me to make all the design decisions...
- armv6m only has `AtomicUsize::{load, store}` which means reducing the xorshift state width depending on target pointer size
- Also means being okay with the state potentially repeating in the cases where calls overlap during the load+store?
- I think I actually want a non-randomized `select!()` anyway that polls in priority order so would need to reimplement select anyway.

So left that out for now, though I am tempted to leave the macro itself in so it could still be used if you override the random fn using a facade.

taiki-e:
(FYI: There is a discussion in https://github.com/rust-lang-nursery/futures-rs/issues/1805 about adding a select that doesn't use randomness. and seems implementation is not difficult: https://github.com/taiki-e/futures-rs/commit/2691aa2c9b5ad19e5dcbecb3655dae4db46d976f)

arcnmx:
Yeah that would be really nice, is there anything blocking that from becoming a PR? just docs? or name, default, syntax bikeshed? uncertainty around whether it should be provided?

taiki-e:
> is there anything blocking that from becoming a PR? just docs? or name, default, syntax bikeshed? uncertainty around whether it should be provided?

I didn't open the PR because the document was incomplete, but in practice, we will need to decide which is the default.

arcnmx:
Yeah, I can see the argument made for keeping unbiased as the default, though don't have any personal opinions there as long as the ability to choose can be made available.

Also, [I threw up what I had for no-std random](https://github.com/arcnmx/futures-rs/commit/random-nostd) and could PR in case it's something that could be accepted?

martell:
@arcnmx firstly thanks for looking at this,
Is there anything stopping this from being merged?
I'm looking forward to using async await without std upstream.

arcnmx:
Rebased onto master, so `select_biased!` is now an option on no-std with these changes. `select!` still requires #1912 but the biased alternative being available makes it a bit less of a blocker now.

arcnmx:
@taiki-e you've approved this previously, any chance it could be considered for merging? I've been experimentally making use of these changes for quite a while now and it'd be nice to have these available upstream - it seems like a benign change, and `select_biased` has made it even more appealing now.

(cc @Nemo157 as this is somewhat your domain as well)

taiki-e:
I'd like to merge this. cc @cramertj

cramertj:
This looks fine to me, thanks for the ping! I'll get a release out shortly.

PR #1895, opened 2019-10-01
Use docs.rs as default documentation

### LINKED ISSUE #1105: docs.rs fails for 0.3.0-alpha.1 and up
https://docs.rs/crate/futures-preview/0.3.0-alpha.1/builds/109321

How can we solve this?

taiki-e:
[docs-rs uses the latest nightly since September 30th](https://blog.rust-lang.org/2019/09/18/upcoming-docsrs-changes.html), so I think the problem has been resolved.

cc #1637

Closes #1105

cc @cramertj @Nemo157

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,7 +21,7 @@
 </p>
 
 <p align="center">
-  <a href="https://rust-lang-nursery.github.io/futures-api-docs/0.3.0-alpha.19/futures/">
+  <a href="https://docs.rs/futures-preview/">
     Documentation
   </a> | <a href="https://rust-lang-nursery.github.io/futures-rs/">
     Website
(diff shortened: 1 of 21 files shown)


cramertj:
It looks like some of the URLs no longer include an explicit version number, while others do. Is that intentional?

taiki-e:
Yes, this is intentional. IIUC, only [html_root_url](https://rust-lang-nursery.github.io/api-guidelines/documentation.html#crate-sets-html_root_url-attribute-c-html-root) requires an explicit version number and is not required elsewhere.

cramertj:
Do you think there's any benefit to continuing to specify an explicit version so that the published package points to the correct version of the docs, or do you think it doesn't matter?

taiki-e:
In pre-release versions, there are breaking changes from version to version, so it would probably make sense to link to the explicit version of the document. Other than that, cargo basically chooses the latest version, and the changes are backward compatible, so I don't think there is much merit to link to the explicit version, but it still makes sense when considering the increase in major (minor) versions.

taiki-e:
@cramertj Update to use explicit version numbers.

cramertj:
Thanks! Looks like travis failure is a new ICE, so this should be good to merge.

PR #1896, opened 2019-10-01
Remove unnecessary trait bounds from the definition of combinators

taiki-e:


cramertj:
note: one potential downside is that this makes it a breaking change to re-add these bounds in the future if the internals of the type were to change in such a way that these bounds were needed.

PR #1897, opened 2019-10-01
Target as symlink

mathstuf:


.gitignore
```diff
@@ -1,7 +1,7 @@
-target/
+/target
```
cramertj: What is the goal of this change?

mathstuf: It applies the change I made to Cargo's gitignore template in rust-lang/cargo#4944.

cramertj: What was the motivation for removing the leading slash?

mathstuf: Removing? I added the leading one. I can remove it again if wanted. It anchors the ignore to the root of the repo. Not sure why I moved it here. I think I've normally left that alone with these gitignore changes.

cramertj: Given that we have `target` inside the nested crates in the workspace as well, I think it makes sense to omit the leading slash.

cramertj:
Thanks!

Issue #1898, opened 2019-10-02
Documentation not available for alpha.19


elpiel:
Currently in [crates.io](https://crates.io/crates/futures-preview/0.3.0-alpha.19) the link to the documentation: https://rust-lang-nursery.github.io/futures-api-docs/0.3.0-alpha.19/futures is non existent.

Nemo157:
Looks like the auth token used to publish them is expired/revoked/doesn't have access: https://travis-ci.com/rust-lang-nursery/futures-rs/jobs/239482255#L1679

Nemo157:
Since #1895 stops using these published docs I think it's best to delete this automation and just manually publish the `alpha.19` docs so that the current link works, I can do that shortly.

Nemo157:
Should be fixed now.

PR #1899, opened 2019-10-02
Remove publishing of docs on tag

Nemo157:
As mentioned in #1898 this is currently broken because the auth token is no longer valid, with the changes in #1895 there's no need to really publish these docs here so instead of fixing the token just remove the job.


cramertj:
Works for me!

PR #1908, opened 2019-10-09
Rename oneshot:Sender::poll_cancel to poll_canceled

seanmonstar:
In most other places, `poll_action` means that 'action' is happening
asynchronously. The name `poll_cancel` implies that the call will cancel
the oneshot, but really we just wish to poll if it *has been* canceled.


cramertj:
Makes sense to me, thanks!

Issue #1893, opened 2019-09-30
Prepare for 0.3.0 release in conjunction with Rust 1.39


cramertj:
This is intended to serve as a tracking issue for things that need to be resolved before the 0.3.0 release which will be published alongside Rust 1.39. This release will remove the `-preview` suffixes from all the current crates and release them under `futures` and `futures-*`.

Matthias247:
I think @LucioFranco already raised the question in context of the release on zulip. But I would also be interested to know what the concrete idea for stabilization of the individual parts of this crate is. I think there contents with a variety of maturity and usefulness, and having a plan on what is intended to be long-term maintained and what not sounds great!

The answer could obviously be: **everything**. But that could it make it potentially harder to evolve the library than having a smaller content.

Here is my personal wishlist for stabilization:

### Must haves

The `join!` and `select!` macros as well as all their dependencies (which is `FusedFuture` at the moment). Used them in all programs, whether it's std or not.

### Nice to have

- All the extension traits on Futures, and e.g. thinks like `join_all`. I think the use-cases for those things got a lot smaller with the introduction of async_await. However those things seem fairly stable and useable for lots of people, so they are good to have.
- Synchronization primitives and channels. I think when this project initially started they were all awesome and a must-have to write any real world projects with futures. But now it seems like have at least 4 crates which implement the same things (futures-rs, tokio, async-std, futures-intrusive) - the first 3 probably in a very similar fashion. With each additional variant the motivation to long-term maintain a stable crate with those features seems lower. However since most applications will need something like a oneshot future and since the API surface is well fairly well understood it seems reasonable to provide those.

### Not sure / lower-priority

- futures-io. Not sure if the ecosystem has already converged on traits for this. Afaik Tokio has their own versions. GATs might make it at some point of time possible to model things with futures instead of direct `_poll()` methods. And I personally would also require the traits to be no-std compatible (error type).
- Stream/Sink. Also not sure whether we should see those stable as is, or if they won't be superseded medium-term by other things once GATs and other features become available.
- Executors. Great to get something out of the box when just taking `futures-rs` as a dependency. However I think most applications will in the end use executors of different libraries that are also integrated with IO reactors or represent the GUI mainloop.

taiki-e:
I think it is necessary to decide the crate name of futures-test (see https://github.com/rust-lang-nursery/futures-rs/issues/1169#issuecomment-416854334).

(There are many other things, but I will comment later.)

[...]

PR #1910, opened 2019-10-11
Make ThreadPool optional

### LINKED ISSUE #1724: Feature flag for thread pool executor
With the futures API having landed in libcore in Rust 1.36 and async/.await possibly landing in 1.38, the futures crate will need to eventually transition from alpha status into something more stable. Not all its parts have the same degree of maturity, though. In order to make the crate stable, the stable parts need to be separated from the unstable parts. This is obviously a long process. As one further small step towards this, I'd like to suggest to make the thread pool optional.

AFAIK the thread pool executor in futures-rs was always intended to serve more as an example of how a multi-threaded executor could work, not as an actual production-grade implementation that people should use. There are other implementations like @withoutboats's [juliex](https://github.com/withoutboats/juliex) and probably others that I've just not read about. Am I up-to-date with this? Is this still the purpose of the thread pool executor? If its purpose is still mainly to serve an example implementation, then I think we should make this clear by assigning it a feature flag and making it a non-default feature.

taiki-e:
This makes `ThreadPool` optional (disabled by default). 
cc #1893 

Closes #1724

r? @cramertj 
cc @seanmonstar

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,10 +15,12 @@ matrix:
     - name: cargo build (minimum required version)
       rust: 1.36.0
       script:
-        # default features & compat feature
         - cargo run --manifest-path ci/remove-dev-dependencies/Cargo.toml */Cargo.toml
+        # Check default-features
         - cargo build --all
+        # Check compat & threadpool features
         - cargo build --manifest-path futures/Cargo.toml --features io-compat
+        - cargo build --manifest-path futures/Cargo.toml --features threadpool
 
     # This is the minimum Rust version supported by `async-await` feature.
     # When updating this, the reminder to update the minimum required version of `async-await` feature in README.md.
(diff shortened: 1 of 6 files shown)


Matthias247:
This seems to have broken unit-tests with default features for me:

`cargo test` yields:

```
   Compiling futures-preview v0.3.0-alpha.19 (C:\Users\matth\Code\rust\futures-rs\futures)
error[E0432]: unresolved import `futures::executor::ThreadPool`
 --> futures\tests\eventual.rs:2:5
  |
2 | use futures::executor::ThreadPool;
  |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ no `ThreadPool` in `executor`

error[E0433]: failed to resolve: could not find `ThreadPool` in `executor`
  --> futures\tests\mutex.rs:39:39
   |
39 |     let mut pool = futures::executor::ThreadPool::builder()
   |                                       ^^^^^^^^^^ could not find `ThreadPool` in `executor`
``` 

Is there something missing, that it gets enabled for test builds again?

taiki-e:
We only test with `--all-features`, so it is recommended to pass `--all-features` flag when running tests locally.
https://github.com/rust-lang-nursery/futures-rs/blob/99d5ebed1288fe61e21151c86b2223e8872070c0/.travis.yml#L200-L201

PR #1916, opened 2019-10-18
feat: Allow oneshots to be reused

Marwes:
Probably not the best API but I could find this useful in https://github.com/mitsuhiko/redis-rs where I submit queries to a single task which later submits the response via a oneshot channel. With this the oneshot could be reused on the next query which improves overall performance by ~5-10% in one benchmark.

### DIFF
--- a/futures-channel/src/oneshot.rs
+++ b/futures-channel/src/oneshot.rs
@@ -104,9 +104,7 @@ pub fn channel<T>() -> (Sender<T>, Receiver<T>) {
     let receiver = Receiver {
         inner: inner.clone(),
     };
-    let sender = Sender {
-        inner,
-    };
+    let sender = Sender { inner };
     (sender, receiver)
 }
 
@@ -122,7 +120,7 @@ impl<T> Inner<T> {
 
     fn send(&self, t: T) -> Result<(), T> {
         if self.complete.load(SeqCst) {
-            return Err(t)
+            return Err(t);
         }
 
         // Note that this lock acquisition may fail if the receiver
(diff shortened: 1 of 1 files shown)


seanmonstar:
I'd personally find this counter-intuitive, since the name "oneshot" implies to me the channel cannot be reused.

Marwes:
Unless there is a better/more efficient way to implement `oneshot` which makes reusing impossible to implement I don't see why it being called `oneshot` has any weight on whether this should be added or not. If anything, it would imply that `oneshot` is a poor name for it.

Since the method takes `&mut` it isn't like any other part of the program can observe the `oneshot` being reused so the not-oneshotness is entirely contained to where `reset` is called.

Nemo157:
One implementation issue is that this appears prone to race conditions, you have no way to know when the sender has been dropped. If you just assume that once you have received a value the sender is dropped and attempt to reset the receiver you may get a panic, if the other side has left the sender sitting around while it does some operation after sending the value.

Marwes:
True. This is only supposed to be called after the `Receiver` has received a value and while the sender has likely dropped at that point (since `send` takes ownership and drops the `Sender`) it is not a guarantee.

Matthias247:
Why not just go for a bounded channel of size 1 instead? Having a resettable channel prevents some optimizations and might just give someone a hard time retaining the functionality if the implementation is ever changed.

There are now a ton of async channel implementations available (here, in tokio, futures-intrusive, async-std). I'm pretty sure one of them has an bounded channel implementation which works well for that use-case.

PR #1922, opened 2019-10-28
Add immutable iterators for FuturesUnordered

### LINKED ISSUE #1921: Add immutable iterators to FuturesUnordered
It's currently only possible to iterate over `FuturesUnordered` mutably. Immutable iterators would be nice for inspecting state, especially in types which wrap `FuturesUnordered` that don't want mutability in their public API when it is not strictly required.

rjsberry:
Adds `iter_pin_ref` and `iter` to `FuturesUnordered`. This is pretty much just a port of the mutable iterator implementations.

Closes #1921


cramertj:
Thanks!

PR #1924, opened 2019-10-30
Ignore clippy errors from debug_assert_with_mut_call

cramertj:


cramertj:
@seanmonstar thanks for the quick review!

Issue #1828, opened 2019-08-26
CopyBufInto::poll does not flush until the input stream is completely drained


najamelan:
@Nemo157 @talchas I saw #1771, but this issue is slightly different, specifically about flushing behavior.

It doesn't seem correct to me. I'm using it for an echo server over `AsyncRead`/`AsyncWrite`, because those have a small buffer of 8k, which hurts performance. Maybe that's not a correct use case, but then what should I use? `AsyncReadExt::copy_into` shows exactly the same behavior. Documentation for neither methods say anything about flushing. They just say that the future will only resolve once the reader has hit EOF.

```rust
BufReader::with_capacity( 512_000, reader ).copy_buf_into( &mut writer ).await
```

When `poll_fill_buf` pulls in some data, it calls `poll_write` on the writer. Since this is in a loop, we try `poll_fill_buf` again. Ok, there is no more data to read, so it returns pending and `CopyBufInto::poll` just returns, not flushing the writer. This means that the data will just sit in the writer's buffer and never gets send out over the network until the connection from the reader gets closed, or the buffer of the writer is full.

I recon that after writing, if there is no more data to read, we should flush the writer, since for networking code there is no telling when the next incoming message will be.

Nemo157:
Part of the `AsyncWrite::poll_write` contract is it must attempt to flush data out to accept the incoming write.

Nemo157:
Oh, actually I misunderstood, so what you're describing is a situation in which the reader will send some data, then wait a bit, then send more, and in the meantime you expect that initial chunk of data to make it to the writer?

This behaviour would depend on the writer you're using, I would expect any `UdpSocket` or `TcpStream` to already satisfy this since the underlying kernel IO will buffer the data and set a timeout to send it. If you are wrapping that in a `BufWriter` then it won't as it just buffers until an explicit flush is required.

Sounds like what you need is something like [`LineWriter`](https://doc.rust-lang.org/nightly/std/io/struct.LineWriter.html) that will inject flushes in based on the data being sent, or something which buffers and injects flushes based on timeout from last write. As far as I know neither of these abstractions exist yet.

najamelan:
[...]

PR #1926, opened 2019-10-31
Document AsyncWrite

najamelan:
When implementing `AsyncWrite` I found it unclear what behavior is expected from a type that implements it. The main question is when to flush the underlying object.

This PR is intended to settle on that issue and then document it to avoid confusion for further users. It follows up from #1828. Some further discussion is required to come to a conclusion before documenting.

## Explanation

Imagine we implement `AsyncWrite` on some type that does some internal buffering. Eg. this type will have data hanging in transit unless we call `poll_flush` on it. 

The questions occurred when using `AsyncReadExt::copy_buf_into` to create an echo server that just forwards all incoming data back out. The future returned by this method will only flush the `AsyncWrite` when the stream has completely ended. Until then, it will only call `poll_write`. So if some data comes in, then there is a pause, this data will now just sit in the writer without getting flushed out. 

The question is thus, should implementers sometimes flush in `poll_write` or not? If so, which context to pass to `poll_flush`? If any bytes have been written, `poll_write` needs to return `Poll::Ready(n)`, so a Pending from the `poll_flush` cannot be bubbled up unless no bytes where written...

### DIFF
--- a/futures-io/src/lib.rs
+++ b/futures-io/src/lib.rs
@@ -133,7 +133,7 @@ mod if_std {
     /// This trait is analogous to the `std::io::Write` trait, but integrates
     /// with the asynchronous task system. In particular, the `poll_write`
     /// method, unlike `Write::write`, will automatically queue the current task
-    /// for wakeup and return if data is not yet available, rather than blocking
+    /// for wakeup and return if the writer cannot take more data, rather than blocking
     /// the calling thread.
     pub trait AsyncWrite {
         /// Attempt to write bytes from `buf` into the object.
--- a/futures-io/src/lib.rs
+++ b/futures-io/src/lib.rs
@@ -208,6 +208,8 @@ mod if_std {
         /// `Interrupted`.  Implementations must convert `WouldBlock` into
         /// `Poll::Pending` and either internally retry or convert
         /// `Interrupted` into another error kind.
+        ///
+        /// It only makes sense to do anything here if you actually buffer data.
         fn poll_flush(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Result<()>>;
 
         /// Attempt to close the object.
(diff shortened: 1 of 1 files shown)


futures-io/src/lib.rs
```diff
@@ -148,6 +148,9 @@ mod if_std {
         /// `Interrupted`.  Implementations must convert `WouldBlock` into
         /// `Poll::Pending` and either internally retry or convert
         /// `Interrupted` into another error kind.
+        ///
+        /// `poll_write` must try to make progress by flushing the underlying object before returning
```
cramertj: This bit isn't quite right-- it mustn't flush unless flushing is what is necessary for the underlying IO object to become writable once more.

najamelan: @cramertj sorry to let this linger. I tried asking some clarification, but then the review feature of github kicked in and the comment wasn't actually posted. So trying to resolve this, I could make it say something like:
```rust
/// `poll_write` must try to make progress by flushing the underlying object if that is the only way the underlying object can become writable again.
```
But I feel that is rather confusing. Does that mean it can return pending without trying to flush? Does returning Pending not mean that it is not writable right now and needs to be flushed? What does client code do if `poll_write` returns pending? Do they manually call flush or do they just keep calling `poll_write`? For how long?

najamelan: sorry, that last part doesn't make sense. Obviously if poll_write returns pending, client code waits to be woken up by it. Still not sure what the best formulation is.

cramertj: The example that you wrote in the codeblock above sounds reasonable to me!

futures-io/src/lib.rs
```diff
@@ -205,6 +208,10 @@ mod if_std {
         /// `Interrupted`.  Implementations must convert `WouldBlock` into
         /// `Poll::Pending` and either internally retry or convert
         /// `Interrupted` into another error kind.
+        ///
+        /// It only makes sense to do anything here if you actually buffer data. Otherwise,
```
cramertj: I'm not sure this paragraph is helpful beyond perhaps the first sentence.

najamelan: ok, I shortened it.

cramertj:
Thanks!

PR #1927, opened 2019-10-31
Merge futures-select-macro and futures-join-macro

taiki-e:
They have exactly the same dependencies and are often used together. Management will be easier by merging.

r? @cramertj


cramertj:
Works for me, thanks for the PR!

PR #1931, opened 2019-11-01
Replace u64 with usize

yjhmelody:
I find that `take` and `skip` in std's iterator are using usize rather than u64.
```rust
fn skip(self, n: usize) -> Skip<Self>
where
    Self: Sized, 
```
Is there needed to alter them?


cramertj:
Looks reasonable to me!

PR #1933, opened 2019-11-01
improve comments and relax one cmpxch in AtomicWaker

rkuhn:
While reading AtomicWaker I stumbled over some comments that were pointing slightly off path, so I’m offering potential improvements in case my understanding of the code is correct.

Caveat emptor: I have only reasoned about the code, not tested it. If such code-based discussion is not common in the Rust community then please let me know where I should raise this point instead. I’m looking forward to learning from your responses!

### DIFF
--- a/futures-core/src/task/__internal/atomic_waker.rs
+++ b/futures-core/src/task/__internal/atomic_waker.rs
@@ -268,15 +268,18 @@ impl AtomicWaker {
                     // Release the lock. If the state transitioned to include
                     // the `WAKING` bit, this means that at least one wake has
                     // been called concurrently.
-                    //
+                    // 
                     // Start by assuming that the state is `REGISTERING` as this
-                    // is what we jut set it to.
+                    // is what we just set it to. If this holds, we know that no
+                    // other writes were performed in the meantime, so there is
+                    // nothing to acquire, but release the update. In case of
+                    // concurrent wakers, we need to acquire their releases.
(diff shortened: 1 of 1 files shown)


futures-core/src/task/__internal/atomic_waker.rs
```diff
@@ -268,15 +268,18 @@ impl AtomicWaker {
...
+                    // 
                     // Start by assuming that the state is `REGISTERING` as this
-                    // is what we jut set it to.
+                    // is what we just set it to. If this holds, we know that no
+                    // other writes were performed in the meantime, so there is
+                    // nothing to acquire, but release the update. In case of
+                    // concurrent wakers, we need to acquire their releases.
                     let res = self.state.compare_exchange(
-                        REGISTERING, WAITING, AcqRel, Acquire);
+                        REGISTERING, WAITING, Release, Acquire);
```
cramertj: Errors with `a failure ordering can't be stronger than a success ordering`

rkuhn: Ah, yes, makes sense, will update.

rkuhn:
Thanks for the review @cramertj! This code is so well documented that I think it makes sense to add the now remaining comment clarifications — it can serve as a tutorial for how to do wait-free synchronization in Rust :-)

cramertj:
Thanks so much!

PR #1936, opened 2019-11-03
Align io::Cursor implementations with std

### LINKED ISSUE #1510: impl AsyncWrite for Cursor<Vec<u8>> doesn't extend vector as needed

Nemo157:
Instead of providing a generic `AsyncWrite` implementation delegate the specific implementations of `Write`.

Fixes #1510

### DIFF
--- a/futures-io/src/lib.rs
+++ b/futures-io/src/lib.rs
@@ -26,7 +26,6 @@ compile_error!("The `read_initializer` feature requires the `unstable` feature a
 
 #[cfg(feature = "std")]
 mod if_std {
-    use std::cmp;
     use std::io;
     use std::ops::DerefMut;
     use std::pin::Pin;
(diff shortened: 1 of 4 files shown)


cramertj:
Thanks!

PR #1943, opened 2019-11-04
Use our own io::Cursor

taiki-e:
Based on #1936

Context: #1829

> If both `futures::io` ext traits and `std::io` traits are imported, method names may conflict, so use our own.

cc @Nemo157

### DIFF
--- a/futures-util/src/io/cursor.rs
+++ b/futures-util/src/io/cursor.rs
@@ -1,0 +1,238 @@
+use futures_core::task::{Context, Poll};
+#[cfg(feature = "read_initializer")]
+use futures_io::Initializer;
+use futures_io::{AsyncBufRead, AsyncRead, AsyncSeek, AsyncWrite, IoSlice, IoSliceMut, SeekFrom};
+use std::io;
+use std::pin::Pin;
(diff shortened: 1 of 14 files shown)


futures-util/src/io/cursor.rs
```diff
@@ -0,0 +1,238 @@
...
+/// are commonly used as a buffer, like `Cursor<`[`Vec`]`<u8>>` and
+/// `Cursor<`[`&[u8]`][bytes]`>`.
+///
+/// [`AsyncSeek`]: trait.AsyncSeek.html
+/// [`AsyncRead`]: trait.AsyncRead.html
+/// [`AsyncWrite`]: trait.AsyncWrite.html
+/// [bytes]: https://doc.rust-lang.org/std/primitive.slice.html
+#[derive(Clone, Debug, Default)]
+pub struct Cursor<T> {
+    inner: io::Cursor<T>,
```
taiki-e: As there is not enough time, it will be implemented as a newtype, but it can be optimized later.

cramertj:
Works for me, thanks!

Issue #1817, opened 2019-08-20
`SinkExt::send_all` should work with `TryStream`?


tedsta:
I have a `Sink<MyMsg, Error=MyError>` and I'd like to `send_all` it a stream, `Stream<Item=Result<MyMsg, MyError>>`.

`StreamExt::forward` almost works for me, but I need to keep using the `Sink` after.

Perhaps a `SinkExt::try_send_all` with `where St: TryStream<Ok = Item, Error = Error>`? I can send a PR if you guys are open to it.

cramertj:
Agreed-- `TryStream` seems sensible to support here.

PR #1946, opened 2019-11-05
Use TryStream in Sink::send_all

cramertj:
Closes https://github.com/rust-lang-nursery/futures-rs/issues/1817

### DIFF
--- a/futures-util/src/sink/mod.rs
+++ b/futures-util/src/sink/mod.rs
@@ -7,7 +7,7 @@
 //! library is activated, and it is activated by default.
 
 use futures_core::future::Future;
-use futures_core::stream::Stream;
+use futures_core::stream::{Stream, TryStream};
 use crate::future::Either;
 
 #[cfg(feature = "compat")]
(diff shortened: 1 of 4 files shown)


futures-util/src/sink/mod.rs
```diff
@@ -223,7 +223,7 @@ pub trait SinkExt<Item>: Sink<Item> {
         &'a mut self,
         stream: &'a mut St
     ) -> SendAll<'a, Self, St>
-        where St: Stream<Item = Item> + Unpin,
+        where &'a mut St: TryStream<Ok = Item, Error = Self::Error> + Unpin,
```
taiki-e: I was thinking about using `TryStream + Stream` bounds like [Stream::forward](https://github.com/rust-lang-nursery/futures-rs/blob/0c165a30954c98f281311ae7ce9be26491ca2932/futures-util/src/stream/stream/forward.rs#L25) - but I don't have any strong opinion on this.

PR #1948, opened 2019-11-05
Change copy_into/copy_buf_into to free functions for consistency with the standard library

taiki-e:


stjepang:
The signature of `std::io::copy` is:

```rust
pub fn copy<R: ?Sized, W: ?Sized>(reader: &mut R, writer: &mut W) -> Result<u64>
where
    R: Read,
    W: Write;
```

Is there a reason why in this PR we take ownership of `reader`, i.e. take `R` instead of `&mut R`?

Nemo157:
It was required to make delegating the implementation to `copy_buf_into` possible. It also more closely follows the API guideline [C-RW-VALUE](https://rust-lang.github.io/api-guidelines/interoperability.html#generic-readerwriter-functions-take-r-read-and-w-write-by-value-c-rw-value):

> Generic reader/writer functions take R: Read and W: Write by value

Given that `copy` exhausts the provided reader there's no reason to allow access to it afterwards, and users can still borrow it anyway if they want.

EDIT: [Here's the previous discussion when the API was changed](https://github.com/rust-lang-nursery/futures-rs/pull/1674#discussion_r293933381).

taiki-e:
The reason is explained in [this comment](https://github.com/rust-lang-nursery/futures-rs/pull/1674#discussion_r294042914) in #1674.

> we want to wrap the reader in a `BufReader` to allow the reuse, but then we don't have anywhere with ownership of that `BufReader` when we pass the reference into `CopyBufInto::new` (unless we make `CopyInto` a self-referential struct and only create the inner `CopyBufInto` when we're first polled, but I'd prefer to not deal with manually implemented self-referential structs).
> 

(Also, this is [why tokio is considering changing the definition of AsyncBufRead](https://github.com/tokio-rs/tokio/pull/1687#issuecomment-546488702))

EDIT: I didn't notice that @Nemo157 commented.

stjepang:
@Nemo157 If API guidelines suggest taking `Read` and `Write` arguments by value, why not take `writer` here by value, too (`W` instead of `&mut W`)?

I ran into similar issues before: https://github.com/async-rs/async-std/issues/365

Personally, I'd prefer if we always took `Read` and `Write` arguments by value (like API guidelines suggest), even in the standard library. We could even make this change in standard library's APIs without breaking them, AFAICT.

cramertj:
@stjepang unlike `reader`, `writer` is not exhausted by this operation.

PR #1954, opened 2019-11-05
Stage 0.3.0 release

cramertj:
r? @taiki-e @seanmonstar @Nemo157

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,24 @@
+# 0.3.0 - 2019-11-5
+* Stable release along with stable async/await!
+* Added async/await to default features (#1953)
+* Changed `Spawn` trait and `FuturesUnordered::push` to take `&self` (#1950)
+* Moved `Spawn` and `FutureObj` out of `futures-core` and into `futures-task (#1925)
+* Changed case convention for feature names (#1937)
+* Added `executor` feature (#1949)
+* Moved `copy_into`/`copy_buf_into` (#1948)
+* Changed `SinkExt::send_all` to accept a `TryStream` (#1946)
+* Removed `ThreadPool::run` (#1944)
+* Changed to use our own definition of `io::Cursor` (#1943)
+* Removed `BufReader::poll_seek_relative` (#1938)
+* Changed `skip` to take a `usize` rather than `u64` (#1931)
+* Removed `Stream` impl for `VecDeque` (#1930)
+* Renamed `Peekable::peek` to `poll_peek` (#1928)
+* Added immutable iterators for `FuturesUnordered` (#1922)
+* Made `ThreadPool` optional (#1910)
+* Renamed `oneshot::Sender::poll_cancel` to `poll_canceled` (#1908)
+* Added some missing `Clone` implementations
+* Documentation fixes
+
 # 0.3.0-alpha.19 - 2019-9-25
 * Stabilized the `async-await` feature (#1816)
 * Made `async-await` feature no longer require `std` feature (#1815)
@@ -216,7 +237,7 @@
 * `FuturesUnordered` optimization: Since the context stores a `&LocalWaker` reference, it was possible to avoid cloning the `Arc` of the waker
 * Futures-rs now uses Clippy
 * We now use in-band lifetimes
-* The `join!` and `select!` macros are now exposed by the `futures-preview` crate
+* The `join!` and `select!` macros are now exposed by the `futures` crate
 * The project logo was added to the `README.md`
 * `sink::MapErr::get_pinned_mut` is now called `get_pin_mut`
 * We now use the unstable `use_extern_macros` feature for macro reexports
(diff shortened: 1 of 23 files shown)


yoshuawuyts:
Running a test now on `async-std` to check if anything is broken. Will update this comment to report back.

cramertj:
@yoshuawuyts Thanks for helping review! <3

yoshuawuyts:
https://github.com/async-rs/async-std/pull/460 passed on CI, which means that as far as `async-std` is concerned this patch should be good to go! @cramertj thanks so much for the work you've put in to get this release out!

taiki-e:
There is a problem trying to update in tokio.
commit: https://github.com/tokio-rs/tokio/commit/fa71a584a601417331e3336fa22b9465e8d16c1d

```
error: implementation of `core::future::future::Future` is not general enough
   --> examples/connect.rs:31:5
    |
31  |       tokio::spawn(async move {
    |       ^^^^^^^^^^^^ implementation of `core::future::future::Future` is not general enough
    |
    = note: `futures_util::sink::send_all::SendAll<'0, tokio_util::codec::framed_write::FramedWrite<tokio::io::stdout::Stdout, codec::Bytes>, futures_util::stream::stream::map::Map<futures_util::stream::stream::filter_map::FilterMap<tokio_util::codec::framed_read::FramedRead<tokio::net::tcp::split::ReadHalf<'1>, codec::Bytes>, futures_util::future::ready::Ready<std::option::Option<std::vec::Vec<u8>>>, [closure@examples/connect.rs:99:25: 105:14]>, fn(std::vec::Vec<u8>) -> std::result::Result<std::vec::Vec<u8>, std::io::Error> {std::result::Result::<std::vec::Vec<u8>, std::io::Error>::Ok}>>` must implement `core::future::future::Future`, for any two lifetimes `'0` and `'1`...
    = note: ...but `core::future::future::Future` is actually implemented for the type `futures_util::sink::send_all::SendAll<'_, tokio_util::codec::framed_write::FramedWrite<tokio::io::stdout::Stdout, codec::Bytes>, futures_util::stream::stream::map::Map<futures_util::stream::stream::filter_map::FilterMap<tokio_util::codec::framed_read::FramedRead<tokio::net::tcp::split::ReadHalf<'_>, codec::Bytes>, futures_util::future::ready::Ready<std::option::Option<std::vec::Vec<u8>>>, [closure@examples/connect.rs:99:25: 105:14]>, fn(std::vec::Vec<u8>) -> std::result::Result<std::vec::Vec<u8>, std::io::Error> {std::result::Result::<std::vec::Vec<u8>, std::io::Error>::Ok}>>`
```

I will investigate later, but it may be https://github.com/rust-lang-nursery/futures-rs/pull/1946#discussion_r342560520.

taiki-e:
@cramertj  I've confirmed that the above problem can be fixed once #1955 is merged.

Nemo157:
+1 to removing the `=` requirements. It might pay to have another CI job/steps that verify the internal minimum versions are correct, but that seems like it might be complicated enough that it's not worth it (I believe that properly testing it would require a bunch of `Cargo.toml` rewriting to add and remove `path` dependencies as needed to test against the released versions).

cramertj:
I've landed #1955, rebased this change, and amended the versions to remove `=`. PTAL.

LucioFranco:
@taiki-e thank you for checking! <3

I say lets ship! Thank you for all your hard work @cramertj and others!

Nemo157:
CI failure is unrelated, the job itself is wrong and needs fixing.

theduke:
Out of curiosity, why is the `executor` feature on by default?

Seems like that would make sense as opt-in.

cramertj:
@theduke It is exposed as its own crate, `futures-executor`, which is reexported through the main facade crate. I have no strong opinion about whether or not it be enabled by default.

theduke:
Yeah I saw that. 

It only introduces one additional dependency ( `num_cpus`) but it seems to me like there is no big reason to tie it to the main facade at all and just provide it as the separate crate. It's certainly nice to have a simple stand alone executor available, but it would probably be a rather nieche thing to use, eg in small tests.

cramertj:
@theduke The primary value I see is convenience of having things like `block_on` available, which makes it possible to write small standalone tests and example code using only the `futures` crate.

seanmonstar:
It does seem useful for `futures` to be able to provide a `block_on` function out-of-the-box, even if a user probably shouldn't use it in most cases. 🤷‍♂

theduke:
I can see that, and looking at how little code it is (< 1k lines) I'm not terribly concerned anymore.

shepmaster:
As an arbitrary small datapoint, SNAFU compiles and tests pass against this branch.

Matthias247:
@cramertj Thanks for coordinating this! And even more so for going through the backlog and closing all long-term outstanding issues!

@seanmonstar 
> It does seem useful for `futures` to be able to provide a `block_on` function out-of-the-box, even if a user probably shouldn't use it in most cases. 🤷‍♂

There was some discussion on including `block_on` as a minimal executor into std - since it's currently quite hard to use async without any other dependency (you can't even run things in play.rust-lang without boilerplate). I created https://github.com/rust-lang/rust/pull/65875 to implement this - but I don't whether and when we will see it getting there.

PR #1956, opened 2019-11-06
Add StreamExt::buffer_unordered_adaptable

llebout:
It might be desirable to have adaptative parallelism for the
processing of buffered futures.

An example use case would be to cope with the ever changing
network conditions when sending requests to a remote host.
One might collect metrics about the performance of the
data transfer and adapt the parallelism parameters as
necessary to maximize throughput.


leo-lb:
Finally, CI is successful. **thumbv6m-none-eabi** does not have atomic instructions? how does it even multi core?

ebkalderon:
Nice work on this pull request! Exactly what I was looking for in a personal project. In regards to your question, according to https://github.com/rust-lang/rust/issues/54250, it seems that the platform simply does not support certain atomic operations in hardware.

taiki-e (changes requested):

I'm okay with adding such a feature, but I would like to have tests and docs for the following cases:

- `max` is `0` when `buffer_unordered_adaptable` is called.
- `max` was not `0` when `buffer_unordered_adaptable` is called, but changed to `0`.

(both are related to https://github.com/rust-lang/futures-rs/pull/2429 and https://github.com/rust-lang/futures-rs/issues/2348)

PR #1959, opened 2019-11-06
fix: [Local]SpawnExt should take &self as their base traits

Marwes:
Seems like an oversight in https://github.com/rust-lang-nursery/futures-rs/pull/1950

Technically a

BREAKING CHANGE


cramertj:
Yuck, I'll get this out ASAP. Thanks!

PR #1961, opened 2019-11-07
Iteration policy API for eager polling loops

### LINKED ISSUE #1957: ForEach, Fold, and similar stream combinators can run saturated without returning from poll
The combinators that poll a stream for exhaustion in a loop have a problem that's already been raised in #869: if the upstream consecutively returns `Ready` for a long time, the loop never breaks and the combinator's `poll` never returns for that long, starving other pending operations in the task from being polled.

To illustrate how this can be a problem for other code, consider this simple adapter for making futures cancellable:

```rust
use futures::channel::oneshot::{self, Canceled};
use futures::prelude::*;
use std::pin::Pin;
use std::task::{Context, Poll};

use futures::ready;
use pin_utils::unsafe_pinned;

struct CancelHandle(oneshot::Sender<()>);

#[derive(Debug)]
struct AlreadyDropped;

impl CancelHandle {
    fn cancel(self) -> Result<(), AlreadyDropped> {
        self.0.send(()).map_err(|()| AlreadyDropped)
    }
}

struct Cancelable<F> {
    op: F,
    stop_rx: oneshot::Receiver<()>,
}

impl<F> Cancelable<F> {
    unsafe_pinned!(op: F);
    unsafe_pinned!(stop_rx: oneshot::Receiver<()>);
}

impl<F: Unpin> Unpin for Cancelable<F> {}

impl<F> Future for Cancelable<F>
where
    F: Future,
{
    type Output = Result<F::Output, Canceled>;

    fn poll(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
        match self.as_mut().stop_rx().poll(cx) {
            Poll::Pending => {
                let output = ready!(self.as_mut().op().poll(cx));
                Ok(output).into()
            }
            Poll::Ready(_res) => Err(Canceled).into(),
        }
    }
}

fn make_cancelable<F>(op: F) -> (Cancelable<F>, CancelHandle) {
    let (stop_tx, stop_rx) = oneshot::channel();
    let fut = Cancelable { op, stop_rx };
    let handle = CancelHandle(stop_tx);
    (fut, handle)
}
```

It looks rather useful and intuitive, but this contrived example hangs with a busy-looping thread rather than canceling the task:

```rust
fn main() {
    let mut a = 0;
    let (fut, stop_handle) = make_cancelable(
        stream::repeat(1).for_each(move |n| {
            a += n;
            future::ready(())
        })
    );
    let mut executor = ThreadPool::new().unwrap();
    let res_handle = executor.spawn_with_handle(fut).unwrap();
    thread::sleep(Duration::from_millis(1));
    stop_handle.cancel().unwrap();
    let res = executor.run(res_handle);
    assert!(res.is_err());
}
```

In non-contrived usage with real streams, too, a `ForEach` with an always-ready processing closure will delay cancellation for as long as the stream yields items.

mzabaluev:
Work to fix #1957 and provide a new macro for responsible polling.

### DIFF
--- a/futures-core/src/iteration.rs
+++ b/futures-core/src/iteration.rs
@@ -26,7 +26,7 @@ pub trait LoopPolicy {
 /// an asynchronous source is pending or the poll resolves with an output
 /// value).
 #[derive(Debug)]
-pub struct Unlimited {}
+pub struct Unlimited;
 
 impl LoopPolicy for Unlimited {
     type State = ();
--- a/futures-core/src/iteration.rs
+++ b/futures-core/src/iteration.rs
@@ -1,6 +1,5 @@
 //! Utilities for implementing iterative polling in a cooperative way.
 
-use core::mem;
 use core::num::NonZeroU32;
 
 /// Iteration policy checker for eager polling loops.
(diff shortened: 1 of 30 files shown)


mzabaluev:
This is a simple all-dynamic solution that augments futures and streams which may be affected by saturation of their poll function with a counter-based relief valve.

A more elaborate, generic extension is also possible that will allow the users to opt out of the iteration counter statically, which will be backward compatible with these additions (except the `poll_loop!` macro that would need to make use of a protocol trait to be future proof). But that may be overkilling it. If I get around to do it, I will submit the generic solution in another PR.

mzabaluev:
I'm not sure about landing the `iteration` API and the `poll_loop!` macro in `futures-core`. These are implementation helpers, not essential definitions, and so they could be moved to `futures-util` or even to their own dedicated crate if deemed necessary.

taiki-e (commented):
Sorry for the late review.

I would not prefer to provide abstractions or APIs like this on `futures` side for now. I also would not prefer to use the way of increasing the size of the combinators like this approach.
I'm not clear what the medium-term solution to this problem is (In the short-term, we will need to use an approach like #2333), but I think it will be [resolved by std in the long-term](https://github.com/rust-lang/rust/pull/74335#issuecomment-742775411).

So I'm going to close this PR. Thanks anyway @mzabaluev!

PR #1962, opened 2019-11-07
Implemented hash_receiver method on mpsc

### LINKED ISSUE #1177: Allow mpsc senders to be stored in hash sets/maps
I have a use case where I need to store the senders in a hash set, but that is impossible to do directly right now because senders don't implement `Eq + Hash`. I wrote [`same`](https://crates.io/crates/same) crate, so I can workaround it using `Arc`, but it'd be better if I could skip that layer of indirection by putting the senders in the map directly (instead of putting them into `Arc` first).

I'll happily send a PR, but first I want to know which way to take:

* Implement `Eq`, `PartialEq` and `Hash` for senders, on the pointers. Does it make sense though? What "this sender is equal to that one" should mean? Would it be clear from user point of view?
* Implement `Same` and `RefHash` traits from my crate behind a feature flag - this may be more clear to the reader what is being compared/hashed.
* Expose methods `has_same_receiver()` and `channel_hash<H>()` and let the users use them - the methods clearly state what they do and avoid depending on `same`, but increase the burden on users.
* Do nothing - let the users use `Arc<Sender>`. This is inefficient.

What do you think?

Kixunil:
This method allows the sender to obtain the hash of the receiver
(actually `Option<pointer to shared structure>`). This is useful if one
wants to put the senders into `HashSet` or `HashMap`, using a newtype
pattern.

Closes #1177

### DIFF
--- a/futures-channel/src/mpsc/mod.rs
+++ b/futures-channel/src/mpsc/mod.rs
@@ -552,6 +552,13 @@ impl<T> SenderInner<T> {
         Arc::ptr_eq(&self.inner, &other.inner)
     }
 
+    /// Returns pointer to the Arc containing sender
+    ///
+    /// The returned pointer is not referenced and should be only used for hashing!
+    fn ptr(&self) -> *const Inner<T> {
+        &*self.inner
+    }
+
     /// Returns whether this channel is closed without needing a context.
     fn is_closed(&self) -> bool {
         !decode_state(self.inner.state.load(SeqCst)).is_open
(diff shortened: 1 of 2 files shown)


Kixunil:
Hmm, GitHub tells me that the branch is out-of-date, should I rebase?

Kixunil:
The tests failed due to CI bug. :(

cramertj:
This seems fine to me, thanks!

Kixunil:
Thank you too for merging! :)

Do you have any estimate when a new version containing the change could be released? Possibly even some way to subscribe to its publication?

Nemo157:
@Kixunil you can watch releases on the repo under the watch button in the top right.

Kixunil:
Thanks!

PR #1963, opened 2019-11-07
Update and fix CI config

taiki-e:
* Use Rust 1.39
* Fix `--features` flag


taiki-e:
Hmm... looks like failure is due to https://github.com/rust-lang-nursery/futures-rs/issues/1790

PR #1967, opened 2019-11-09
Added FusedStream impl to Buffered

jean-airoldie:
Note that the `buffered` method seem to be [missing tests](https://github.com/rust-lang-nursery/futures-rs/tree/master/futures/tests).

### DIFF
--- a/futures-util/src/stream/stream/buffered.rs
+++ b/futures-util/src/stream/stream/buffered.rs
@@ -1,4 +1,4 @@
-use crate::stream::{Fuse, FuturesOrdered, StreamExt};
+use crate::stream::{Fuse, FuturesOrdered, StreamExt, FusedStream};
 use futures_core::future::Future;
 use futures_core::stream::Stream;
 use futures_core::task::{Context, Poll};
@@ -18,6 +18,7 @@ where
     stream: Fuse<St>,
     in_progress_queue: FuturesOrdered<St::Item>,
     max: usize,
+    is_terminated: bool,
 }
 
 impl<St> Unpin for Buffered<St>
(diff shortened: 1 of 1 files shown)


futures-util/src/stream/stream/buffered.rs
```diff
@@ -40,6 +40,16 @@ where
     }
 }
 
+impl<St> FusedStream for Buffered<St>
+where
+    St: FusedStream,
+    St::Item: Future,
+{
+    fn is_terminated(&self) -> bool {
+        self.stream.is_terminated() && self.in_progress_queue.is_empty()
```
cramertj: I don't think this is quite right-- the stream might have terminated previously, and then the queue yielded a final element, but has not yet yielded `None`. In this case, the `Stream` still has to yield `None`, but this will return `true`

jean-airoldie: Why does a stream have to yield a `None` before terminating?

cramertj: it's part of the contract of `FusedStream` that `Stream`s which are not "stalled" should return `None` before being considered "terminated", similar to how `FusedIterator`s must return `None`.

jean-airoldie: Kk i'll fix that.

But what is the rational behind this contract? Why would you want to pretend that a stream is not yet terminated until its polled with None?

cramertj: The primary purpose of `FusedStream` is for use in `select!` to prevent polling a stream after it has yielded `None`, which violates the contract of the `Stream`. Users of `select!` would still generally expect to see a `None` value appear at some point and be able to handle the event of the stream becoming terminated, rather than it just stopping without ever yielding `None`.

jean-airoldie: Forgot about that PR, should be fixed I think.

taiki-e:
(Looks like the new `FusedStream` impl is the same as `.buffered().fuse()`.)

cramertj:
Mhm, I don't think we should add `FusedStream` impls to types like this if it means storing an extra boolean, since that's equivalent to wrapping with `.fuse()`.

jean-airoldie:
Fair enough, I agree with this reasoning. I'll close the PR.

taiki-e:
As `FuturesOrdered` is now implements `FusedStream` (#2205), I think this can be implemented by using `self.stream.is_terminated() && self.in_progress_queue.is_terminated()`.

PR #1968, opened 2019-11-09
readme: consistently require rust 1.39

FauxFaux:
The header of the README claims a different required Rust version to the body; update it, and its link.


taiki-e:
(I would like to mention supporting 1.36 when disabled async-await feature in readme, but it is fine even if it is another PR.)

PR #1970, opened 2019-11-10
Enable all stable features on the playground

Nemo157:
Prompted by [this u.rl.o thread](https://users.rust-lang.org/t/error-unresolved-import-futures-threadpool/34469), based on docs [in the playground source](https://github.com/integer32llc/rust-playground/blob/f68374b402b5dedbcea417086af29deb315e3031/top-crates/src/main.rs#L183-L192).

### DIFF
--- a/futures/Cargo.toml
+++ b/futures/Cargo.toml
@@ -56,3 +56,6 @@ read-initializer = ["futures-io/read-initializer", "futures-util/read-initialize
 
 [package.metadata.docs.rs]
 all-features = true
+
+[package.metadata.playground]
+features = ["std", "async-await", "compat", "io-compat", "executor", "thread-pool"]


Nemo157:
I assume that the playground builds all crates together in a single build so activating the features from the facade will affect all other crates and there's no point in marking them individually.

futures/Cargo.toml
```diff
@@ -56,3 +56,6 @@ read-initializer = ["futures-io/read-initializer", "futures-util/read-initialize
 
 [package.metadata.docs.rs]
 all-features = true
+
+[package.metadata.playground]
+all-features = false
```
taiki-e: The title says 'Enable', but I think '= false' means disable 'all-features'.

taiki-e:
also, some features depend on unstable features and require nightly rustc, so I think it is good to include only stabilized features.

Nemo157:
@taiki-e thanks, that's a good point. I've listed out just the stable features instead.

futures/Cargo.toml
```diff
@@ -56,3 +56,6 @@ read-initializer = ["futures-io/read-initializer", "futures-util/read-initialize
 
 [package.metadata.docs.rs]
 all-features = true
+
+[package.metadata.playground]
+features = ["std", "async-await", "compat", "io-compat", "executor", "threadpool"]
```
taiki-e:
```suggestion
features = ["std", "async-await", "compat", "io-compat", "executor", "thread-pool"]
```

I renamed 'threadpool' feature to 'thread-pool' in #1937: https://github.com/rust-lang-nursery/futures-rs/pull/1937/files#diff-41c03002d828b5290c4f39adc8d88060R45

PR #1976, opened 2019-11-15
Add select_biased! macro

### LINKED ISSUE #1805: Provide alternative precedences for select!

taiki-e:
Closes #1805

r? @cramertj


cramertj:
This looks great!

PR #1977, opened 2019-11-16
Add stream::try_unfold

### LINKED ISSUE #1952: Introduce stream::try_unfold
... restoring the behavior of futures 0.1's `stream::unfold` function. Without this, separate state tracking (like [here][]) is required in many situations.

[here]: https://github.com/ruma/ruma-client/blob/master/src/lib.rs#L321-L368

jplatte:
fixes #1952


futures-util/src/stream/try_stream/try_unfold.rs
```diff
@@ -0,0 +1,144 @@
...
+    unsafe_pinned!(fut: Option<Fut>);
+}
+
+impl<T, F, Fut, Item> FusedStream for TryUnfold<T, F, Fut>
+where
+    F: FnMut(T) -> Fut,
+    Fut: TryFuture<Ok = Option<(Item, T)>>,
+{
+    fn is_terminated(&self) -> bool {
+        self.state.is_none() && self.fut.is_none()
```
cramertj: This will yield `true` even before `None` has been returned in the case where the future errors, which is incorrect.

cramertj:
LGTM modulo the `FusedStream` impl

futures-util/src/stream/try_stream/try_unfold.rs
jplatte: Why is that incorrect? The docs for `FusedStream` say that `is_terminated` returns true if the stream should no longer be polled, which is usually the case after `None` is returned but may also be the case without that. When we know that all future calls to `poll_next` will return `None`, is it not reasonable to return `true` from `.is_terminated()`?

cramertj: No, `is_terminated()` should only return `true` once `None` has been yielded.

jplatte: Okay :man_shrugging:

jplatte:
Added a field `terminated: bool` that tracks whether the stream has terminated to implement the desired `FusedStream` semantics and rebased.

futures-util/src/stream/try_stream/try_unfold.rs
```diff
@@ -0,0 +1,149 @@
...
+    }
+}
+
+/// Stream for the [`try_unfold`] function.
+#[must_use = "streams do nothing unless polled"]
+pub struct TryUnfold<T, F, Fut> {
+    f: F,
+    state: Option<T>,
+    fut: Option<Fut>,
+    terminated: bool,
```
cramertj: Tracking `terminated` internally here doesn't make sense since users can use the `.fuse()` wrapper to provide this behavior. If you're looking to build termination into the combinator, one more optimized way to do this could be to use a three-state enum to store the future with a variant for terminated. However, it'd also be fine to just remove the `FusedFuture` impl.

jplatte: Okay, I'll just remove it then.

jplatte: Done

cramertj:
Thanks!

PR #1979, opened 2019-11-17
Move examples directory into the root

taiki-e:
This should make examples more discoverable.


cramertj:
LGTM! will merge once rebased

taiki-e:
Rebased

PR #1982, opened 2019-11-18
Remove panic when create

yjhmelody:
Since `self.pool_size` is private, the only way to set it is `pool_size` methed and the most time just using `new`. I think it's better to check and panic in `pool_size` method.


cramertj:
Works for me, thanks!

PR #1985, opened 2019-11-21
Remove note about async-await feature not active by default

mbrobbel:
The removed notes are no longer valid now that the `async-await` feature is enabled by default (https://github.com/rust-lang-nursery/futures-rs/pull/1953).


Nemo157:
I think this should still mention the fact that it's feature gated for `default-features = false` users. Maybe just remove the `_not_` and leave the rest of the sentence there?

mbrobbel:
> I think this should still mention the fact that it's feature gated for `default-features = false` users. Maybe just remove the `_not_` and leave the rest of the sentence there?

Good point. I'll update.

PR #1998, opened 2019-12-07
Use first nonempty buffer in vectored I/O

tmiasko:
Previous implementation would use the first buffer even when it was
empty and other nonempty buffers are provided. This could result in
`Ok(0)` return value which has otherwise a special meaning. The new
implementation uses the first nonempty buffer instead to avoid the
issue.

Previous implementation would return early if no buffers are provided.
The new implementation delegates operation using an empty buffer. This
gives an implementation an ability to decide how an empty read / write
operation should behave.

This change aligns the implementation with synchronous version of those
traits.


cramertj:
Thanks!

PR #2000, opened 2019-12-09
Fix CI failures

### LINKED ISSUE #1992: Unreachable code generated by try_join! with never error
This is currently causing CI failures because one of the tests inadvertently triggers it.

Given a crate with:

```rust
async fn foo() -> Result<(), !> {
    Ok(())
}

pub async fn bar() {
    let _: Result<((), ()), !> = futures::try_join!(foo(), foo());
}
```

compiling it produces two unreachable call warnings pointing at the `try_join!`:

```
warning: unreachable call
 --> src/lib.rs:6:34
  |
6 |     let _: Result<((), ()), !> = futures::try_join!(foo(), foo());
  |                                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  |                                  |
  |                                  unreachable call
  |                                  any code following this expression is unreachable
  |
  = note: `#[warn(unreachable_code)]` on by default
  = note: this warning originates in a macro outside of the current crate (in Nightly builds, run with -Z external-macro-backtrace for more info)
```

(`-Z external-macro-backtrace` doesn't give anything useful).

Nemo157:
Remove `-Z install-upgrade` for `cargo-hack`, this is now the default since rust-lang/cargo#7560.

Add `allow(unreachable_code)` for usage of `try_join` with the never type, fixes #1992.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -129,7 +129,7 @@ matrix:
     - name: cargo check (features)
       rust: nightly
       install:
-        - cargo install -Z install-upgrade cargo-hack
+        - cargo install cargo-hack
       script:
         # Check each specified feature works properly
         # * `--each-feature` - run for each feature which includes --no-default-features and default features of package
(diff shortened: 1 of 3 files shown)


Nemo157:
Although, given the latest conclusions in https://github.com/rust-lang/rust/issues/66757 that test will start failing on a future nightly. It can probably use `core::convert::Infallible` which will not actually test what it intends to test once the never type is reverted, but as soon as it is stabilised again the test will be correct and let us know if there's been a regression, and the fix works fine either way.

PR #2001, opened 2019-12-10
Used Self for uniform style

PirateDragon:
just as you see!


laizy:
😄, if you want your **uniform** style, update the `new` function for all types first?

PR #2002, opened 2019-12-11
Added ChunksLazy to Stream

### LINKED ISSUE #1919: Stream combinator to produce non-greedy chunks
The `Chunks` combinator does not yield a next chunk until it is full or the stream ends.

Sometimes, however, it's useful to yield the accumulated items for processing as soon as the upstream polls `Pending`. This might mean that new items won't be available in a while, or that a read buffer from which the items were decoded has been consumed and another OS read is in order, so it would make sense for the application to offer backpressure before attempting to consume that in turn.

fisherdarling:
This implements a "lazy" iterator over the chunks of a stream. The only difference between `ChunksLazy` and `Chunks` is that `ChunksLazy` yields the local buffer if the underlying stream returns `Poll::Pending`, even if there are no elements in the buffer.

This closes #1919


mzabaluev:
"Lazy" may be a misleading term; we have `future::lazy` which means something completely different.
I suggest renaming it to `chunks_eager`, as in eager to yield the buffer as soon as no items are immediately forthcoming.

mzabaluev:
> even if there are no elements in the buffer.

I don't think there is value in returning an empty chunk; in this case, `poll_next` should return `Pending`.

mzabaluev:
This has been made obsolete by #2123.

PR #2016, opened 2019-12-28
Clarify dropping behavior of RemoteHandle

najamelan:
It's quite surprising if you expect it to be dropped immediately, especially because it might just be Pending on something that no longer generates wakeups. I have had to select on a oneshot to avoid this problem.

### DIFF
--- a/futures-util/src/future/future/remote_handle.rs
+++ b/futures-util/src/future/future/remote_handle.rs
@@ -20,7 +20,8 @@ use {
 };
 
 /// The handle to a remote future returned by
-/// [`remote_handle`](crate::future::FutureExt::remote_handle).
+/// [`remote_handle`](crate::future::FutureExt::remote_handle). If you drop this,
+/// the remote future will be dropped __when it next get's polled__.
 #[must_use = "futures do nothing unless you `.await` or poll them"]
 #[derive(Debug)]
 pub struct RemoteHandle<T> {


Nemo157:
Dropping the receiver (in the `RemoteHandle`) _should_ cause a wakeup on the sender side (in the `Remote`) leading to it being dropped as soon as possible (and is literally using a oneshot for this purpose already).

najamelan:
Hmm, yes, I now see the poll impl of `Remote`:
```rust
    fn poll(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<()> {
        if let Poll::Ready(_) = self.as_mut().tx().as_mut().unwrap().poll_canceled(cx)
```
passing the `Context` to be waken up. For some reason this didn't work for me. I will investigate some more.

najamelan:
Ok, the result of today's debugging. I couldn't reproduce in a self contained example. 

In the original context, the `Remote` future did indeed get woken up as intended, but the inner future and it's data was not being dropped.

When debugging step by step I ended up in code from juliex, that I was still using somewhere in testing code. When I removed that from the equation by switching to LocalPool, the problem went away. I'm just in the process of changing how I pass executors around, so it's not very convenient right now to dig into it deeper. I will close this for now, in a few days I should be able to test it more thoroughly.

Even if it does work by dropping immediately, I think it would be good that that be mentioned in the documentation. I will file another PR when I'm confident where this issue came from.

najamelan:
@Nemo157 I just wonder, if this works correctly, why does `Abortable` exist when `RemoteHandle` provides the same functionality? Shouldn't we stick to one type?

Nemo157:
`Abortable` doesn’t capture the output and move it into the handle, it’s only for remote cancellation.

najamelan:
I can no longer reproduce, so assuming it was something juliex specific. 

> Abortable doesn’t capture the output and move it into the handle, it’s only for remote cancellation.
Do you mean they both exist because performance is different? I just meant if the functionality of `RemoteHandle` is a superset of `Abortable` then maybe it's better to have just one type, as in less is more.

PR #2021, opened 2020-01-03
`Peek` future for `Peekable` stream

### LINKED ISSUE #1871: Should provide function returning future for Peekable::peek
The current `Peekable::peek` signature doesn't seem right to me when related to other methods:

```rust
pub fn peek(
    self: Pin<&mut Peekable<St>>, 
    cx: &mut Context
) -> Poll<Option<&<St as Stream>::Item>>
```

I think this should be renamed to `poll_peek`, and another function added that returns a `Future` resolving to the peeked item:

```rust
pub fn peek(self: Pin<&mut Self>) -> impl Future<Output = Option<&<St as Stream>::Item>>;
```

Preferably this should also take something like `&mut self where Self: Unpin` to really avoid the issue from the below linked thread, unfortunately making a really useful signature like that would probably require defining an extra extension trait so that a `Pin<impl DerefMut<Target = Peekable<_>>>` would also be a `Peekable`.

(Related to [this u.rl.o thread](https://users.rust-lang.org/t/unable-to-use-futures-peek-method/32573/2))

dingxiangfei2009:
Fix #1871 

This pull request is an attempt to provide a sensible peeking behaviour for `Peekable` streams. Instead of exposing wires that resembles `poll` from `Future`, this construction wraps this exposure into a proper `Future`, with pinned borrow of the underlying `Peekable`. It works by swapping out the mutable borrow, attempting a poll, then either swapping the mutable borrow back in if `Pending`, or producing the `Poll` directly.

By the construction, `peek()` can be reinstated and the documentation becomes accurate again.

### DIFF
--- a/futures-util/src/stream/stream/peek.rs
+++ b/futures-util/src/stream/stream/peek.rs
@@ -1,10 +1,13 @@
-use crate::stream::{StreamExt, Fuse};
+use crate::future::Either;
+use crate::stream::{Fuse, StreamExt};
 use core::pin::Pin;
+use futures_core::future::{FusedFuture, Future};
 use futures_core::stream::{FusedStream, Stream};
 use futures_core::task::{Context, Poll};
 #[cfg(feature = "sink")]
 use futures_sink::Sink;
 use pin_utils::{unsafe_pinned, unsafe_unpinned};
+use std::fmt;
 
 /// A `Stream` that implements a `peek` method.
 ///
@@ -27,7 +30,7 @@ impl<St: Stream> Peekable<St> {
     pub(super) fn new(stream: St) -> Peekable<St> {
         Peekable {
             stream: stream.fuse(),
-            peeked: None
+            peeked: None,
         }
     }
 
@@ -67,21 +70,26 @@ impl<St: Stream> Peekable<St> {
     ///
     /// This method polls the underlying stream and return either a reference
     /// to the next item if the stream is ready or passes through any errors.
-    pub fn poll_peek(
+    pub fn peek(self: Pin<&mut Self>) -> Peek<'_, St> {
+        Peek { inner: Some(self) }
+    }
+
+    fn poll_peek(
         mut self: Pin<&mut Self>,
         cx: &mut Context<'_>,
-    ) -> Poll<Option<&St::Item>> {
+    ) -> Either<Pin<&mut Self>, Poll<Option<&St::Item>>> {
         if self.peeked.is_some() {
             let this: &Self = self.into_ref().get_ref();
-            return Poll::Ready(this.peeked.as_ref())
+            return Either::Right(Poll::Ready(this.peeked.as_ref()));
         }
-        match ready!(self.as_mut().stream().poll_next(cx)) {
-            None => Poll::Ready(None),
-            Some(item) => {
+        match self.as_mut().stream().poll_next(cx) {
+            Poll::Ready(None) => Either::Right(Poll::Ready(None)),
+            Poll::Ready(Some(item)) => {
                 *self.as_mut().peeked() = Some(item);
                 let this: &Self = self.into_ref().get_ref();
-                Poll::Ready(this.peeked.as_ref())
+                Either::Right(Poll::Ready(this.peeked.as_ref()))
             }
+            _ => Either::Left(self),
         }
     }
 }
@@ -95,12 +103,9 @@ impl<St: Stream> FusedStream for Peekable<St> {
 impl<S: Stream> Stream for Peekable<S> {
     type Item = S::Item;
 
-    fn poll_next(
-        mut self: Pin<&mut Self>,
-        cx: &mut Context<'_>,
-    ) -> Poll<Option<Self::Item>> {
+    fn poll_next(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
         if let Some(item) = self.as_mut().peeked().take() {
-            return Poll::Ready(Some(item))
+            return Poll::Ready(Some(item));
         }
         self.as_mut().stream().poll_next(cx)
     }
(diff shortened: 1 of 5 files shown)


Nemo157:
Changing `poll_peek` is a breaking change. It should be possible to implement the future while leaving `poll_peek` alone (just something like `self.inner.as_mut().poll_peek(...)`).

dingxiangfei2009:
@Nemo157 Is it possible? `self: Pin<&mut Peekable<_>>` cannot be recovered in case "poll_peek" returns `Poll::Pending`.

dingxiangfei2009:
Because of this, `poll_peek` at this current form is hardly of any use in and outside `async` block. In the current type signature, we cannot construct it in an `FnMut` and use it in `crate::future::poll_fn`, and we cannot construct concrete `Future` types that eventually resolves to a borrow of the "head" item... Unless it is intended that any `Future` polling on items from `Peekable` resolves immediately even if the underlying `Stream` is not ready, but this behaviour feels like a big breakage from the rest of the `Future`s constructible from this crate.

Nemo157:
`Pin::as_mut` reborrows the `Pin<&mut _>` for the duration of the `poll_peek` call.

dingxiangfei2009:
@Nemo157 Sorry, backtracking a little here.

I just realised that it is still not possible to use the original `poll_peek`. The main difficulty is because `Peek` needs return a borrow of bits of `Peekable`.

Say we use the original `poll_peek`. I am playing with the following implementation of `Future`:
```rust
impl<'a, St> Future for Peek<'a, St>
where
    St: Stream,
{
    type Output = Option<&'a St::Item>;
    fn poll(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
        if let Some(mut peekable) = self.inner.take() {
            match peekable.as_mut().poll_peek(cx) {
                Poll::Pending => {
                    self.inner = Some(peekable);
                    Poll::Pending
                }
                Poll::Ready(poll) => Poll::Ready(poll),
            }
        } else {
            Poll::Pending
        }
    }
}
```

Now we start to see lifetime issues, because we are borrowing from the `peekable` before `as_mut`.
```
error[E0505]: cannot move out of `peekable` because it is borrowed
   --> futures-util/src/stream/stream/peek.rs:165:39
    |
156 | impl<'a, St> Future for Peek<'a, St>
    |      -- lifetime `'a` defined here
...
163 |             match peekable.as_mut().poll_peek(cx) {
    |                   -------- borrow of `peekable` occurs here
164 |                 Poll::Pending => {
165 |                     self.inner = Some(peekable);
    |                                       ^^^^^^^^ move out of `peekable` occurs here
...
168 |                 Poll::Ready(poll) => Poll::Ready(poll),
    |                                      ----------------- returning this value requires that `peekable` is borrowed for `'a`

error[E0515]: cannot return value referencing local variable `peekable`
   --> futures-util/src/stream/stream/peek.rs:168:38
    |
163 |             match peekable.as_mut().poll_peek(cx) {
    |                   -------- `peekable` is borrowed here
...
168 |                 Poll::Ready(poll) => Poll::Ready(poll),
    |                                      ^^^^^^^^^^^^^^^^^ returns a value referencing data owned by the current function

error: aborting due to 2 previous errors
```

dingxiangfei2009:
There is a bit of tricky borrowing happening here when we stick to the old `poll_peek`. `Pin::as_mut` takes a mutable borrow, and in this case the first `peekable` is then mutably borrowed. The return value of `poll_peek` serves as a witness of this borrow, too, because it is a (potential) borrow of bits of `Peekable`. However, this mutable borrow only spans across `Peek::poll`, meaning that we cannot return this borrow of the head item, nor can we shelf the `Peekable` up for another `poll`.

Returning the mutable borrow back from `poll_peek`, however, will enable us to do both of them.

dingxiangfei2009:
@Nemo157 I have been playing around with `Pin::as_mut` with no success. Maybe I am missing something, so it would be great if you have suggestions or hints. :pray: :smile:

Nemo157:
Yeah, this is actually a really nasty intersection of lifetimes (similar to `poll_fill_buf` which I just realised we also don't have a `Future` returning function for). I've tried a couple of things but the only way I found was by accessing the `Peekable` internals directly, which is basically equivalent to your implementation.

For backwards compatibility I think you should rename the internal method something like `fn do_poll_peek`, then provide the existing `pub fn poll_peek` which calls it and just discards the `Left` value.

futures-util/src/stream/stream/peek.rs
```diff
@@ -120,9 +125,51 @@ impl<S: Stream> Stream for Peekable<S> {
...
+        if let Some(peekable) = self.inner.take() {
+            match peekable.poll_peek(cx) {
+                Either::Left(peekable) => {
+                    self.inner = Some(peekable);
+                    Poll::Pending
+                }
+                Either::Right(poll) => poll,
+            }
+        } else {
+            Poll::Pending
```
Nemo157: I think this should be `panic!("Peek polled after completion")`, most futures panic after completion to (hopefully) make debugging extra polling easier.

futures-util/src/stream/stream/peek.rs
```diff
@@ -67,21 +70,26 @@ impl<St: Stream> Peekable<St> {
...
-    pub fn poll_peek(
+    pub fn peek(self: Pin<&mut Self>) -> Peek<'_, St> {
+        Peek { inner: Some(self) }
+    }
+
+    fn poll_peek(
         mut self: Pin<&mut Self>,
         cx: &mut Context<'_>,
-    ) -> Poll<Option<&St::Item>> {
+    ) -> Either<Pin<&mut Self>, Poll<Option<&St::Item>>> {
```
Nemo157: Given that this never returns `Either::Right(Poll::Pending)` I think changing it to `Either<Pin<&mut Self>, Option<&St::Item>>` would make it a little more readable.

futures-util/src/stream/stream/peek.rs
```diff
@@ -120,9 +125,51 @@ impl<S: Stream> Stream for Peekable<S> {
...
+
+/// Future for the [`peek()`] function from [`Peekable`]
+#[must_use = "futures do nothing unless polled"]
+pub struct Peek<'a, St: Stream> {
+    inner: Option<Pin<&'a mut Peekable<St>>>,
+}
+
+impl<St: Stream> Unpin for Peek<'_, St> {}
+
+impl<St: Stream> fmt::Debug for Peek<'_, St> {
```
Nemo157: If you add `where St: fmt::Debug + St::Item: fmt::Debug` then you can add the `inner` field to the debug output.

futures-util/src/stream/stream/peek.rs
Nemo157: (Maybe with a comment that `Left` is equivalent to `Pending` and `Right` is equivalent to `Ready`).

dingxiangfei2009: Ah, I see.

cramertj:
Nice! Can you add a test for this?

dingxiangfei2009:
@cramertj Sure. A simple is added as an illustration of its function.

cramertj:
LGTM modulo remaining CI failures, which look legit:
- [doc issues](https://travis-ci.com/rust-lang/futures-rs/jobs/273400914)
- [clippy](https://travis-ci.com/rust-lang/futures-rs/jobs/273400909) (this one isn't your fault-- feel free to either fix or ignore)

cramertj:
Thanks!

PR #2023, opened 2020-01-04
Fix clippy issues in tests and add to CI

Nemo157:
There were some "errors" detected with explicitly dropping `Copy` values (mostly I think to discard unchecked `Result`s), the rest is just minor stylistic cleanups that doesn't matter much, but still makes some tests a little more readable.

### DIFF
--- a/futures-channel/tests/mpsc.rs
+++ b/futures-channel/tests/mpsc.rs
@@ -116,8 +116,6 @@ fn recv_close_gets_none() {
             Poll::Ready(Err(e)) => assert!(e.is_disconnected()),
         };
 
-        drop(&tx);
-
         Poll::Ready(())
     }));
 }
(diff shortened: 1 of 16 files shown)


futures-channel/tests/mpsc.rs
```diff
@@ -116,8 +116,6 @@ fn recv_close_gets_none() {
             Poll::Ready(Err(e)) => assert!(e.is_disconnected()),
         };
 
-        drop(&tx);
```
cramertj: Dropping a reference seems suspicious-- was this ever supposed to do something interesting?

cramertj:
Looks great, thanks!

futures-channel/tests/mpsc.rs
Nemo157: It was changed from a by-value drop to a by-ref drop in https://github.com/rust-lang/futures-rs/commit/fff7e7b09a60c73e303d6e757d20aeb0a6951d3d#diff-4198821522533ce7112d24aa37e3926dL132-R132, but even before that it seems to have been a useless drop just before the closure would have dropped it anyway.

PR #2028, opened 2020-01-10
Replace _priv field by tuple field

olegnn:


seanmonstar:
Why make this change? These types are purposefully _not_ tuple structs to prevent people from innocently matching them as such (like `Err(TryRecvError(..))`.

olegnn:
@seanmonstar `(())` works exactly the same way as `{ _priv: () }` (defines one private field to prevent pattern matching etc) except that it's more convenient to use.

taiki-e:
`_priv` field is not actually used and more descriptive than the unnamed field, so I prefer not to change this.

cramertj:
I agree with other comments. I think the old syntax was clearer about the intent of the field. Thanks for helping look for opportunities to clean up the codebase, though!

PR #2032, opened 2020-01-13
JoinAllFirstError: A combinator for an iterator of Result Futures

Nokel81:
This function and future pair is very similar to `join_all` but works only on `std::future::Future`'s whose `type Output = Result<T, E>`.

The return value of this future is `Result<Vec<T>, E>` where the possible `Err(e)` is the first error that is encountered during the polling.

### DIFF
--- a/futures-util/src/future/join_all.rs
+++ b/futures-util/src/future/join_all.rs
@@ -1,14 +1,14 @@
 //! Definition of the `JoinAll` combinator, waiting for all of a list of futures
 //! to finish.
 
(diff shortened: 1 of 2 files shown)


Pauan:
This already exists: https://docs.rs/futures/0.3.1/futures/future/fn.try_join_all.html

Nokel81:
I am sorry for the noise then

Pauan:
No problem, there's nothing wrong with trying to help!

PR #2033, opened 2020-01-16
Implement AsyncRead for BufWriter and vice versa

tomaka:
This adds two impls:

- `impl<R: AsyncWrite> AsyncWrite for BufReader<R>`
- `impl<W: AsyncRead> AsyncRead for BufWriter<W>`

The motivation is that you often want to do something like: `BufReader::new(BufWriter::new(socket))`.


tomaka:
CI failure looks unrelated to my PR.

PR #2038, opened 2020-01-19
Fix clippy warning that breaks CI

najamelan:
This fixes clippy warning https://rust-lang.github.io/rust-clippy/master/index.html#mem_replace_with_default

All tests passing.

### DIFF
--- a/futures-util/src/lib.rs
+++ b/futures-util/src/lib.rs
@@ -10,6 +10,11 @@
 #![cfg_attr(test, warn(single_use_lifetimes))]
 #![warn(clippy::all)]
 
+// The solution for this lint is not available on 1.39 which is the current minimum supported version.
+// Can be removed as of minimum supported 1.40 or if https://github.com/rust-lang/rust-clippy/issues/3941
+// get's implemented.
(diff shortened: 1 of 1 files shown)


najamelan:
Ha, it turns out this requires rustc 1.40. Don't know if we want to wait with this, but in the mean time it breaks CI on other PR, so if we don't want to bump, we probably need to silence the clippy warning for now.

Next stable comes out around January 30th, so if we want 2 versions back, the time this get's out to crates.io it might be fine to merge it.

Nemo157:
The stated policy is just 1.39+ (though it doesn’t say whether changing this is intended to be a breaking change), so I’d recommend just adding an `allow` for the lint.

najamelan:
Unfortunately rust-lang/rust-clippy/issues/3941 isn't implemented. I will silence it then.

Nemo157:
I don't know what's going on with CI, seems Travis added the same build under two different ids so one is stuck at queued. I tried restarting the build in case that would fix it somehow, but doesn't appear to have. @cramertj can you force merge since this is actually passing?

PR #2039, opened 2020-01-19
Add impl Spawn and LocalSpawn for Arc and Rc.

najamelan:
Since #1950 (0.3.0) the Spawn and LocalSpawn only require a shared reference
for spawning operations.

This adds blanket impls for `Arc<Sp: ?Sized + Spawn>` and `Rc<Sp: ?Sized + Spawn>`.
It does the same for LocalSpawn.

This allows client code to pass an Arc<Exec> when library code takes
parameters as `&dyn Spawn` or `impl Spawn`.

So far there were blanket impls for `&`, `&mut` and `Box`.

### DIFF
--- a/futures-task/src/spawn.rs
+++ b/futures-task/src/spawn.rs
@@ -126,7 +126,7 @@ impl<Sp: ?Sized + LocalSpawn> LocalSpawn for &mut Sp {
 #[cfg(feature = "alloc")]
 mod if_alloc {
     use super::*;
-    use alloc::boxed::Box;
+    use alloc::{ boxed::Box, rc::Rc };
 
     impl<Sp: ?Sized + Spawn> Spawn for Box<Sp> {
         fn spawn_obj(&self, future: FutureObj<'static, ()>) -> Result<(), SpawnError> {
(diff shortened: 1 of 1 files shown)


najamelan:
I don't know if something like this would make sense for `Pin`, but it's
not included here. I could add it if that makes sense.

najamelan:
This seems to fail on `thumbv6m-none-eabi`. Move the `Arc` impl under `std` feature?

taiki-e:
> Move the `Arc` impl under `std` feature?

Could you use [`cfg_target_has_atomic` macro](https://github.com/rust-lang/futures-rs/blob/0.3.1/futures-task/src/lib.rs#L32-L) for `Arc` impl?

najamelan:
Limited Arc impl under `cfg_target_has_atomic` as suggested.

cramertj:
Thanks!

najamelan:
Awesome. Any idea when a version 0.3.2 might be released? Seems quite some work has happened since November.

cramertj:
I'll release one next week, thanks for asking!

PR #2044, opened 2020-01-20
Add StreamExt::scan

olegnn:
This PR adds combinator similar to https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.scan. 
It may be useful for some cases when it's needed to cut stream or transform one stream into another using some state-based computations.

### DIFF
--- a/futures-util/src/stream/stream/mod.rs
+++ b/futures-util/src/stream/stream/mod.rs
@@ -559,8 +559,8 @@ pub trait StreamExt: Stream {
     /// let stream = stream::iter(1..=10);
     ///
     /// let stream = stream.scan(0, |state, x| {
-    ///         *state += x;
-    ///         future::ready(if *state < 10 { Some(x) } else { None })
+    ///     *state += x;
+    ///     future::ready(if *state < 10 { Some(x) } else { None })
     /// });
     ///
     /// assert_eq!(vec![1, 2, 3], stream.collect::<Vec<_>>().await);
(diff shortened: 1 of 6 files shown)


futures/tests_disabled/stream.rs
```diff
@@ -158,6 +158,12 @@ fn take_while() {
                 Ok(vec![1, 2]));
 }
 
+#[test]
+fn scan() {
```
taiki-e: Could you move this test into `futures/tests/stream.rs` (`futures/tests_disabled/stream.rs` is not tested)

futures-util/src/stream/stream/mod.rs
```diff
@@ -540,6 +544,37 @@ pub trait StreamExt: Stream {
...
+    ///     *state += x;
+    ///     future::ready(if *state < 10 { Some(x) } else { None })
+    /// });
+    ///
+    /// assert_eq!(vec![1, 2, 3], stream.collect::<Vec<_>>().await);
+    /// # });
+    /// ```
+    fn scan<S, B, Fut, F>(self, initial_state: S, f: F) -> Scan<Self, S, Fut, F>
+    where
+        F: FnMut(&mut S, Self::Item) -> Fut,
```
cramertj: There isn't a way to fix this at the moment, sadly, but this is subtly the wrong method signature for this function, because it doesn't allow `Fut` to hold onto `&mut S`. There's no way to fix this right now due to limitations of HRTBs / GATs, so I don't mind accepting this PR now, but note that this makes this API dramatically less useful than it could be otherwise.

cramertj:
Thanks for the PR! If you have time to send a PR to add documentation about the restriction I listed above, that'd be perfect, but no worries if not!

PR #2046, opened 2020-01-22
`StreamExt::scan`: specified lifetime requirements

olegnn:
As mentioned in https://github.com/rust-lang/futures-rs/pull/2044#discussion_r369249321, current `scan` realisation doesn't allow to capture `&mut` state in future. This PR adds adds lifetime requirements and allows to use `&mut` state in async functions and blocks. @cramertj is this solution ok?

### DIFF
--- a/futures-channel/Cargo.toml
+++ b/futures-channel/Cargo.toml
@@ -30,3 +30,6 @@ futures-sink = { path = "../futures-sink", version = "0.3.1", default-features =
 [dev-dependencies]
 futures = { path = "../futures", version = "0.3.1", default-features = true }
 futures-test = { path = "../futures-test", version = "0.3.1", default-features = true }
+
+[package.metadata.docs.rs]
+all-features = true
--- a/futures-core/Cargo.toml
+++ b/futures-core/Cargo.toml
@@ -26,3 +26,6 @@ cfg-target-has-atomic = []
 
 [dev-dependencies]
 futures = { path = "../futures", version = "0.3.1" }
+
+[package.metadata.docs.rs]
+all-features = true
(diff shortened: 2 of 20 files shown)


cramertj:
Unfortunately this is unsound-- it's giving the closure access to the mutable reference for the *entire* lifetime `'a`, while it can only soundly access the mutable reference until the next time the closure is called with said mutable reference, or the structure is destroyed. This becomes clearer if you think about that if the state type (`S`) has no lifetime parameters, you can choose `'a` to be `'static`. Now the closure is being called with a `&'static mut S`, which is clearly nonsense.

cramertj:
The real type of the closure you'd want is something like `for<'a> F: FnMut<&'a mut S, Self::Item, Output: Future<Output = Option<B>>>`.

olegnn:
@cramertj but future returned by provided function is stored internally in `Scan` structure and can only be executed from within poll function when stream will be polled again. If incomplete stream will be dropped, this future will be also. Even if future will have `static` lifetime, I still don’t see an issue because it’s internal.

olegnn:
I firstly tried higher rank trait bound of course, but it doesn’t work properly yet :(

olegnn:
Way to “hack” which I see is to spawn a thread from function with mutable reference or assign it to static items/other long-lived variables in scope. But I need to check if it compiles.

cramertj:
> . Even if future will have static lifetime, I still don’t see an issue because it’s internal.

The reference it gets will have a `'static` lifetime, so it can save that reference somewhere outside of the call, or pass it to a function like `spawn` which lets it continue to be used mutably forever.

> I firstly tried higher rank trait bound of course, but it doesn’t work properly yet :(

Yup, unfortunately there are a bunch of bugs around HRTB that make this not work yet. See https://github.com/rust-lang/futures-rs/pull/1023 for my own struggle to make this work. Maybe we can bug @nikomatsakis or someone else to take a look at fixing https://github.com/rust-lang/rust/issues/51004 :)

PR #2049, opened 2020-01-23
Avoid starvation from FuturesUnordered::poll_next

### LINKED ISSUE #2047: FuturesUnordered can block the executor
The implementation of `FuturesUnordered::poll_next` never yields as long as at least one future inside of it is ready. This is problematic if it manages a sufficient number of futures that the time it takes to poll all the ready futures is larger than the time it takes between when any given future it manages becomes ready. `FuturesUnordered::poll_next` will then enter what is effectively an infinite loop, holding up the executor and starving other tasks in the system.

For a simple example of this, consider a `FuturesUnordered` that manages futures that run on retry timers. If polling all the ready futures takes longer than the retry time of any one future, then by the time it finishes polling all the futures, the first future it polled will be ready again since its retry timer expired.

The easiest way to fix this is to place a limit on how many calls to `Future::poll` a given call to `FuturesUnordered::poll_next` can make. If the limit is exceeded, yield by calling `cx.waker().wake_by_ref()` and return `Poll::Pending`.

It is not inherently an issue to yield `Some` many times in a row, so only counting the calls inside of `poll_next` should be sufficient. It _does_ place the burden on the caller to implement this kind of limiting themselves too though, so there is an argument for doing a "forced yield" every `N` calls to `Future::poll`, resetting the counter whenever we return `Poll::Pending` (but not when we return `Poll::Ready(Some(_))`). This would help other `Future` implementors that _wrap_ `FuturesUnordered`, since they will get the protection from accidentally blocking the executor by no yielding for too long "for free".

jonhoo:
This fixes #2047.

### DIFF
--- a/futures-util/src/stream/futures_unordered/mod.rs
+++ b/futures-util/src/stream/futures_unordered/mod.rs
@@ -39,6 +39,23 @@ use self::ready_to_run_queue::{ReadyToRunQueue, Dequeue};
 /// without running out of ram.
 const TERMINATED_SENTINEL_LENGTH: usize = usize::max_value();
 
+/// Constant used for a `FuturesUnordered` to determine how many times it is
+/// allowed to poll underlying futures without yielding.
+///
+/// A single call to `poll_next` may potentially do a lot of work before
+/// yielding. This happens in particular if the underlying futures are awoken
+/// frequently but continue to return `Pending`. This is problematic if other
+/// tasks are waiting on the executor, since they do not get to run. This value
+/// caps the number of calls to `poll` on underlying futures a single call to
+/// `poll_next` is allowed to make.
+///
+/// The value itself is chosen somewhat arbitrarily. It needs to be high enough
+/// that amortize wakeup and scheduling costs, but low enough that we do not
+/// starve other tasks for long.
+///
+/// See also https://github.com/rust-lang/futures-rs/issues/2047.
+const YIELD_EVERY: usize = 32;
+
 /// A set of futures which may complete in any order.
 ///
 /// This structure is optimized to manage a large number of futures.
@@ -313,6 +330,10 @@ impl<Fut: Future> Stream for FuturesUnordered<Fut> {
     fn poll_next(mut self: Pin<&mut Self>, cx: &mut Context<'_>)
         -> Poll<Option<Self::Item>>
     {
+        // Keep track of how many child futures we have polled,
+        // in case we want to forcibly yield.
+        let mut polled = 0;
+
         // Ensure `parent` is correctly set.
         self.ready_to_run_queue.waker.register(cx.waker());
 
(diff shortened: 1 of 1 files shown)


cramertj:
I'm unsure about this. Overall, I think we need some better way to address this issue globally, as this behavior is also an issue with every other kind of Future/Stream/asynchronous function. If a task has work remaining that can be done, it will never yield to the executor. Making every single looping combinator track a maximum poll count seems like a possibility, but it seems like the kind of practice we'd want some discipline around, rather than picking arbitrary integers which will stack exponentially (with this PR, the limit of leaf-future polls for a `FuturesUnordered` containing futures which poll a `FuturesUnordered` is now 32^2=1024, and another layer would make it 32768). Perhaps @carllerche has thoughts?

jonhoo:
Yup, I agree it's not ideal, which is why I also proposed https://github.com/tokio-rs/tokio/pull/2160 as a more "complete" fix. I actually argued the exact same point about exponentials in the tokio Discord a few hours ago as a reason why we ultimately want a more global mechanism :p I _do_ think we need a solution in the short run too though, as global preemption is probably something that will take longer to settle, and this is certainly an issue _right now_ (for example, I am running into it in [Noria](https://github.com/mit-pdos/noria) as we speak).

carllerche:
I agree w/ @cramertj. IMO this should be tracked at the runtime task level and leaf resource operations should be the ones to impact the poll budget.

cramertj:
Merging-- see discussion ongoing in https://github.com/rust-lang/futures-rs/issues/2047

jonhoo:
@cramertj What's the procedure for doing a patch release with this change?

cramertj:
@jonhoo I'll put one out this week! Are you under time pressure? If so, I'll can try and do one tomorrow (the 28th).

jonhoo:
Nope, no rush on my end. I'm using it in https://github.com/mit-pdos/noria, but we're fine using patches in `Cargo.toml` to git for the time being :)

jonhoo:
@cramertj Any update on a release for this? Would be nice to get rid of the patches.

cramertj:
Yup, I'm doing one today!

jsancio:
Thanks for the feature @jonhoo. How are executor expected to use this feature? When checking for events (e.g. epoll) executors need to know if they should:

1. Block forever until an event triggers because there aren't any task in the ready queue.
2. Check for events without blocking because there are tasks in the ready queue.

`FuturesUnordered` doesn't expose this information. In other words, now `FutureUnordered::poll_next` can return `Poll::Pending` even if there are tasks in the ready queue.

jonhoo:
@jsancio I'm not entirely sure I follow? Whenever `FuturesUnordered` does this "forced yield", it also calls `.wake()` on its waker to ensure that it gets poll again. As long as the executor respects calls to `wake` during `poll`, it should be good. It shouldn't go to sleep when waiting for events, since it has at least one ready task (the `FuturesUnordered`). Does that answer your question?

jsancio:
Thanks @jonhoo. Yeah, it makes sense. I missed the part that `wake()` was getting called even it if it was returning `Poll::Pending`.

PR #2051, opened 2020-01-24
Remove impl Sync for FuturesUnordered

### LINKED ISSUE #2050: FuturesUnordered has an incorrect Sync impl
`FuturesUnordered` switched to using `Cell` for interior mutability of the `len` and `head_all` fields in 47f3ccb0f7243a02a2258d5cf0942e892f3ac858, but it still currently retains a `Sync` implementation. Since there isn't any synchronization performed when these fields are modified when calling the `push` method, the `Sync` implementation is invalid and should be dropped.

It looks like this was mentioned in #1669, but was overlooked in the actual commit.

okready:
FuturesUnordered was previously modified to use interior mutability for
the len and head_all fields using Cell, but its Sync implementation was
left intact. This implementation is no longer valid, so it has been
removed.

Fixes #2050.

### DIFF
--- a/futures-util/src/stream/futures_unordered/mod.rs
+++ b/futures-util/src/stream/futures_unordered/mod.rs
@@ -67,7 +67,6 @@ pub struct FuturesUnordered<Fut> {
 }
 
 unsafe impl<Fut: Send> Send for FuturesUnordered<Fut> {}
-unsafe impl<Fut: Sync> Sync for FuturesUnordered<Fut> {}
 impl<Fut> Unpin for FuturesUnordered<Fut> {}
 
 impl Spawn for FuturesUnordered<FutureObj<'_, ()>> {


seanmonstar:
I suspect code already depends on this bound, so it may be better to add the synchronization in instead of removing this bound, otherwise people's code will stop compiling in a patch update.

okready:
Good point. I realize simply wrapping the internals in a `Mutex` isn't ideal, but `FuturesUnordered::push` is a bit non-trivial, so it seems like a reasonable choice for now without performing any major rewrites, and it shouldn't slow down polling or other `&mut self` methods since they can simply call `get_mut` on the mutex. Any objections?

okready:
Disregard that, I didn't realize `Mutex` is only in `std`.

Pauan:
[`AtomicUsize`](https://doc.rust-lang.org/nightly/core/sync/atomic/struct.AtomicUsize.html) is available in `core` (and it should be used anyways, because it's faster than `Mutex`).

okready:
Fixing `push` to work concurrently using atomics seems straightforward (`len` and `head_all` can be briefly out-of-sync without things breaking). There's also the interaction between `push` and `iter` though. `iter` simply grabs the current `head_all` and `len` values, but they might not match up if `head_all` and `len` aren't synchronized.

We could just ignore the `len` field when creating an `IterPinRef` and either compute the length manually when creating the iterator or lazily when calling `size_hint`. It could be costly for large `FuturesUnordered` lists, but it would avoid having to do any additional synchronization between the two fields.

PR #2055, opened 2020-01-29
Allow in-place unwrapping of future::Ready

tyranron:
This PR adds `into_inner` method for `future::Ready`.

## Motivation

Due to lack of supporting `async fn` or `impl Trait` in traits at the moment, there are a lot of situations we are made to introduce redundant boxing just for "things to work".

```rust
impl<T> SomeTrait for AnotherType<T> 
where 
    T: SomeTrait<Future = future::Ready<u8>> ,
{
    type Future = LocalBoxFuture<'static, u32>;

    fn do_some(&self) -> Self::Future {
        self.0.do_some().map(|v| u32::from(v) + 32).boxed_local()
    }
}
``` 

But if we know the concrete `Future`'s type being `future::Ready`, we could have been able to avoid this redundant boxing if we have direct access to the inner immediately ready value:
```rust
impl<T> SomeTrait for AnotherType<T> 
where 
    T: SomeTrait<Future = future::Ready<u8>> ,
{
    type Future = future::Ready<u32>;

    fn do_some(&self) -> Self::Future {
        future::ready(self.0.do_some().into_inner().into() + 32)
    }
}
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+# 0.3.2 - 2020-??-??
+* Added `into_inner` method for `future::Ready` (#2055)
+
 # 0.3.1 - 2019-11-7
 * Fix signature of `LocalSpawn` trait (breaking change -- see #1959)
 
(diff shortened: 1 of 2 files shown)


cramertj:
Looks reasonable to me!

PR #2056, opened 2020-01-29
Add MappedMutexGuard API

adamlesinski:
MappedMutexGuard is used to map some locked data to a narrower
view of that data, while maintaining the lock in a new RAII wrapper.

This mirrors the same API in parking_lot.


cramertj:
Looks great, thanks!

PR #2058, opened 2020-01-31
Add TryFutureExt::map_ok_or_else method

Xaeroxe:
Similar to the recently stabilized `Result::map_or_else` method, this adds a method to the `TryFutureExt` trait that fulfills a similar purpose.

### DIFF
--- a/futures-util/src/future/try_future/map_ok_or_else.rs
+++ b/futures-util/src/future/try_future/map_ok_or_else.rs
@@ -1,0 +1,59 @@
+use core::pin::Pin;
+use futures_core::future::{FusedFuture, Future, TryFuture};
+use futures_core::task::{Context, Poll};
+use pin_utils::{unsafe_pinned, unsafe_unpinned};
+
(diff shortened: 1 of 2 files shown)


futures-util/src/future/try_future/map_ok_or_else.rs
```diff
@@ -0,0 +1,59 @@
...
+
+impl<Fut: Unpin, F, E> Unpin for MapOkOrElse<Fut, F, E> {}
+
+impl<Fut, F, E, T> FusedFuture for MapOkOrElse<Fut, F, E>
+    where Fut: TryFuture,
+          F: FnOnce(Fut::Ok) -> T,
+          E: FnOnce(Fut::Error) -> T,
+{
+    fn is_terminated(&self) -> bool {
+        self.f.is_none()
```
cramertj: This should be `self.f.is_none() || self.e.is_none()`

Xaeroxe: Good catch, fixed!

cramertj:
Thanks!

PR #2064, opened 2020-02-06
Add Drop impl to futures::channel::mpsc::UnboundedReceiver

adamlesinski:
Pre-0.3.2, clients could expect to drop an UnboundedReceiver and observe
that incoming messages are dropped and the channel closed.

This change re-introduces this behavior for UnboundedReceiver and adds
tests to prevent regressions.

Note that a drop impl already exists for BoundedReceiver.

### DIFF
--- a/futures-channel/src/mpsc/mod.rs
+++ b/futures-channel/src/mpsc/mod.rs
@@ -1211,6 +1211,18 @@ impl<T> Stream for UnboundedReceiver<T> {
     }
 }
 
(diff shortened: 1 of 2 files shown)


adamlesinski:
CI seems to be broken from a previous commit. The cargo clippy error is unrelated to this change.

cramertj:
Seems good to me to continue to match the behavior of BoundedReceiver.

PR #2067, opened 2020-02-07
Misc docs fixes

olegnn:
- Specified impossibility to capture item's references into futures (https://github.com/rust-lang/futures-rs/pull/2044#discussion_r369249321)
- Corrected order of `split` to (`Sink`, `Stream`) in docs
- Other fixes

### DIFF
--- a/futures-util/src/io/mod.rs
+++ b/futures-util/src/io/mod.rs
@@ -312,7 +312,7 @@ pub trait AsyncReadExt: AsyncRead {
     /// use futures::io::{self, AsyncReadExt, Cursor};
     ///
     /// // Note that for `Cursor` the read and write halves share a single
-    /// // seek position. This may or may not be `true` for other types that
+    /// // seek position. This may or may not be true for other types that
     /// // implement both `AsyncRead` and `AsyncWrite`.
     ///
     /// let reader = Cursor::new([1, 2, 3, 4]);
--- a/futures-util/src/io/mod.rs
+++ b/futures-util/src/io/mod.rs
@@ -312,7 +312,7 @@ pub trait AsyncReadExt: AsyncRead {
     /// use futures::io::{self, AsyncReadExt, Cursor};
     ///
     /// // Note that for `Cursor` the read and write halves share a single
-    /// // seek position. This may or may not be true for other types that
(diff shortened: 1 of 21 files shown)


futures-util/src/io/mod.rs
```diff
@@ -312,7 +312,7 @@ pub trait AsyncReadExt: AsyncRead {
     /// use futures::io::{self, AsyncReadExt, Cursor};
     ///
     /// // Note that for `Cursor` the read and write halves share a single
-    /// // seek position. This may or may not be true for other types that
+    /// // seek position. This may or may not be `true` for other types that
```
cramertj: This isn't referring to `true` the value in the language, but the truth of a statement (whether read and write halves share a single seek position).

futures-util/src/stream/stream/mod.rs
```diff
@@ -315,7 +315,9 @@ pub trait StreamExt: Stream {
     ///
     /// Note that this function consumes the stream passed into it and returns a
     /// wrapped version of it, similar to the existing `filter` methods in the
-    /// standard library.
+    /// standard library. Because of current HRTB limitations, reference to item
```
cramertj: I don't think this should be added-- I think it will confuse users more than help them.

futures-util/src/stream/stream/mod.rs
```diff
@@ -544,10 +546,14 @@ pub trait StreamExt: Stream {
...
 
-    /// Combinator similar to [`StreamExt::fold`] that holds internal state and produces a new stream.
+    /// Combinator similar to [`StreamExt::fold`] that holds internal state 
+    /// and produces a new stream.
     ///
-    /// Accepts initial state and closure which will be applied to each element of the stream until provided closure
-    /// returns `None`. Once `None` is returned, stream will be terminated.
+    /// Accepts initial state and closure which will be applied to each element
+    /// of the stream until provided closure returns `None`. Once `None` is
+    /// returned, stream will be terminated. Because of current HRTB limitations,
```
cramertj: Same here-- I don't think this should be added.

cramertj: and same elsewhere

cramertj:
Thanks!

PR #2068, opened 2020-02-07
Add `StreamExt::flat_map`

olegnn:
This combinator acts like https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.flat_map but for `Stream`s.

### DIFF
--- a/futures-util/src/stream/stream/flat_map.rs
+++ b/futures-util/src/stream/stream/flat_map.rs
@@ -1,0 +1,165 @@
+use super::Map;
+use core::fmt;
+use core::pin::Pin;
+use futures_core::stream::{FusedStream, Stream};
+use futures_core::task::{Context, Poll};
+#[cfg(feature = "sink")]
+use futures_sink::Sink;
+use pin_utils::unsafe_pinned;
+
+/// Stream for the [`flat_map`](super::StreamExt::flat_map) method.
+#[must_use = "streams do nothing unless polled"]
+pub struct FlatMap<St, U, F> {
+    stream: Map<St, F>,
+    inner_stream: Option<U>,
+}
+
+impl<St: Unpin, U, F> Unpin for FlatMap<St, U, F> {}
+
+impl<St, U, F> fmt::Debug for FlatMap<St, U, F>
+where
+    St: fmt::Debug,
+    U: fmt::Debug,
+{
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_struct("FlatMap")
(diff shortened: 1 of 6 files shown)


futures-util/src/stream/stream/flat_map.rs
```diff
@@ -0,0 +1,166 @@
...
+use pin_utils::unsafe_pinned;
+
+/// Stream for the [`flat_map`](super::StreamExt::flat_map) method.
+#[must_use = "streams do nothing unless polled"]
+pub struct FlatMap<St, U, F> {
+    stream: Map<St, F>,
+    inner_stream: Option<U>,
+}
+
+impl<St: Unpin, U, F> Unpin for FlatMap<St, U, F> {}
```
cramertj: This impl needs to also be conditional on `U: Unpin`

futures-util/src/stream/stream/mod.rs
```diff
@@ -544,6 +548,40 @@ pub trait StreamExt: Stream {
         Flatten::new(self)
     }
 
+    /// Maps a stream like [`StreamExt::map`] but flattens nested `Stream`s.
+    ///
+    /// [`StreamExt::map`] is very useful, but if it produces a `Stream` instead,
```
cramertj: I'd rephrase this to avoid talking about "extra layers"`-- perhaps refer to `.map(...).flatten()`? Also, consider mentioning that this combinator will yield all items from the first stream before coming back and pulling out another stream, so it's not suitable for handling values from multiple streams concurrently.

olegnn: @cramertj do we need some iterator which acts like `flat_map` but pulls streams concurrently?

cramertj: Absolutely-- I want it all the time, and opened https://github.com/rust-lang/futures-rs/issues/1781 for this. Unfortunately it's somewhat complicated to implement efficiently, and would require something similar to `FuturesUnordered` for streams.

futures-util/src/stream/stream/mod.rs
```diff
@@ -551,11 +551,11 @@ pub trait StreamExt: Stream {
     /// Maps a stream like [`StreamExt::map`] but flattens nested `Stream`s.
     ///
     /// [`StreamExt::map`] is very useful, but if it produces a `Stream` instead,
-    /// there's an extra layer of indirection. `flat_map()` will remove this extra
-    /// layer on its own.
+    /// you should chain cominators like `.map(f).flatten()` while this combinator
```
cramertj: > you should

"you would have to"?

PR #2078, opened 2020-02-12
Add thread-pool feature for doc tests.

najamelan:
The doc tests for futures-util require ThreadPool which is now behind a feature flag.


cramertj:
Thanks!

PR #2079, opened 2020-02-12
Fix/remote handle unwind safety

### LINKED ISSUE #2077: Is the use of catch_unwind in remote_handle correct?
@Nemo157 Would you mind clarifying the reasoning behind the use of `catch_unwind` here?

Im looking into [this code](https://github.com/rust-lang/futures-rs/blob/771513cea996bac0c8c4017be687e289c966f504/futures-util/src/future/future/remote_handle.rs#L100-L114):

```rust
pub(super) fn remote_handle<Fut: Future>(future: Fut) -> (Remote<Fut>, RemoteHandle<Fut::Output>) {
    let (tx, rx) = oneshot::channel();
    let keep_running = Arc::new(AtomicBool::new(false));

    // AssertUnwindSafe is used here because `Send + 'static` is basically
    // an alias for an implementation of the `UnwindSafe` trait but we can't
    // express that in the standard library right now.
    let wrapped = Remote {
        future: AssertUnwindSafe(future).catch_unwind(),
        tx: Some(tx),
        keep_running: keep_running.clone(),
    };

    (wrapped, RemoteHandle { rx, keep_running })
}
```

I can see a few potential problems:
1. The comment suggests that `Send + 'static` is `UnwindSafe`, however we don't require `Send + 'static` bounds on the future we assert unwind safe.
2. `Send + 'static != UnwindSafe`. There has been a lot of discussion on this around the [RFC](https://github.com/rust-lang/rfcs/blob/master/text/1236-stabilize-catch-panic.md) for `catch_unwind`. Thus it seems that even if the future were `Send + 'static`, we assert that it is `UnwindSafe` when it actually might not be, which means users still need to be aware of unwind safety (not an easy concept), and the documentation doesn't mention any of this.
3. So far, apart from here, the only place I have been able to find `catch_unwind` is in `tokio::task::harness`. It seems executors from the futures lib, async-std, wasm-bindgen-futures all don't catch panics. This makes me wonder, why here?
4. There is a `Send + 'static` bound on the `Future::Output` of the future. I wonder why it is there, since even though right now `remote_handle` can be used for `!Send` futures, it still requires the output to be `Send`, seemingly artificially limiting it's use cases. I reason that if a panic happens in the future, the output will not be produced and thus cannot be observed by other code, so it probably has no effect on unwind safety. I'm working on a lib that introduces a new `JoinHandle` type that wraps the one from tokio, the one from async-std and that I want to also work with the executors from the futures lib, in order to enable more executor agnostic code. However if I were to roll a custom `remote_handle` to overcome this restriction, I really need to understand the intricacies of unwind safety first, hence this question. 

For the record, the use of `AssertUnwindSafe` in tokio also seems incorrect to me, but at least they require `Send + 'static` on the future (which whacks some of the moles but not all).

najamelan:
closes #2077.

Please see detailed explanation in the commit messages.

This PR contains BREAKING CHANGES and version bumps are not included.

### DIFF
--- a/futures-util/src/future/future/remote_handle.rs
+++ b/futures-util/src/future/future/remote_handle.rs
@@ -9,7 +9,7 @@ use {
     std::{
         any::Any,
         fmt,
-        panic::{self, AssertUnwindSafe},
+        panic::{AssertUnwindSafe},
         pin::Pin,
         sync::{
             Arc,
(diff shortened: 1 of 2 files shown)


najamelan:
Having second thoughts. The main undesired property of tearing down the thread in which the panic happens is that threadpools will lose their threads. So I suppose that's the common use case, so I close the PR for now.

PR #2083, opened 2020-02-17
Basic `StreamExt::{flatten_unordered, flat_map_unordered}` impls

### LINKED ISSUE #1781: `Stream::flatten_unordered` would be nice

### LINKED ISSUE #2378: StreamExt::flat_map_unordered
(Inspired by the twitter thread at https://twitter.com/RReverser/status/1374544426622726145)

Flattening a stream of streams into a single stream provides an opportunity for concurrency if you don't care about the output order of the final stream. An efficient solution is difficult to realize with the built-in combinators.

- StreamExt::flat_map doesn't offer any concurrency
- stream::select_all requires collecting the stream of streams into an iterable.
- Using StreamAll directly can be done with StreamExt::collect, but it still requires collecting all streams from the stream of streams first before processing any of the output items.

What is really desired in this case is a combinator that introduces concurrency both within the set of streams yielded by the stream of streams and with the stream of streams itself. Something like (pseudocode)

```
    let stream_of_streams = ...;
    let accumulated_streams = SelectAll::new();
    loop {
        match select(stream_of_streams.next(), accumulated_streams.next()).await {
           StreamOfStreamsBranch(new_stream) => accumulated_streams.push(new_stream);
           AccumulatedStreamsBranch(item_to_pass_along) => break item_to_pass_along,
        }
    }
```

olegnn:
Closes #1781
Closes #2378

Basic implementations of `flatten_unordered` and `flat_map_unordered` using `FuturesUnordered` in combination with custom poll state and wakers under the hood. I didn't measure performance but I think it's fast enough. It differs from `flatten` and `flat_map` because it polls stream and inner streams concurrently.

Edited: I added benchmarks, which can be compared with `futures_unordered`. `flatten_unordered` with a stream of 10000 oneshots each of which produces an inner stream with one item taking approx about x1.5 time of the original `futures_unordered` benchmark with 10000 oneshots.

New update: after changing the logic to make a stream being polled in a loop instead of once per `poll_next`,  `flatten_unordered` takes about 1 - 1.1x time of the original  `futures_unordered` benchmark.


futures-util/src/stream/stream/flatten_unordered.rs
```diff
@@ -0,0 +1,388 @@
...
+
+        let item = if let Some(stream) = stream.as_mut().as_pin_mut() {
+            ready!(stream.poll_next(cx))
+        } else {
+            None
+        };
+
+        Poll::Ready(
+            item.map(|item| {
+                (item, PollStreamFut::new(unsafe { stream.get_unchecked_mut().take() }))
```
taiki-e, line 185: This `take` moves stream, so I think this is unsound if `St: !Unpin`.

futures-util/src/stream/stream/flatten_unordered.rs
```diff
@@ -0,0 +1,388 @@
...
+        // Only call waker if stream isn't being polled because of safety reasons.
+        // Waker will be called at the end of polling if state was changed.
+        if poll_state_value & POLLING == NONE {
+            if let Some(inner_waker) =
+                unsafe { self_arc.inner_waker.get().as_ref().cloned().flatten() }
+            {
+                // First, stop waking to allow polling stream
+                self_arc.stop_waking();
+                // Wake inner waker
+                inner_waker.wake();
```
taiki-e, line 142: If waker panics, the state will be fixed to POLLING and it makes FlattenUnordered unusable, right?

taiki-e (commented):
Sorry for the very late review!

futures-util/src/stream/stream/flatten_unordered.rs
olegnn: If waker panics, everything will go well as `self_arc.stop_waking()` was called prior to calling a waker, see https://github.com/rust-lang/futures-rs/pull/2083/commits/9bc4e4eedcc6e98218346d688cac2d2a3a4962b7

However, after any `poll_next`'s panic state was corrupted, so I added some state bombs to prevent this behavior.

futures-util/src/stream/stream/flatten_unordered.rs
```diff
@@ -0,0 +1,431 @@
...
+            }
+        }
+    }
+}
+
+pin_project! {
+    /// Future which contains optional stream. If it's `Some`, it will attempt
+    /// to call `poll_next` on it, returning `Some((item, next_item_fut))` in
+    /// case of `Poll::Ready(Some(...))` or `None` in case of `Poll::Ready(None)`.
+    /// If `poll_next` will return `Poll::Pending`, it will be forwared to
```
taiki-e, line 195:
```suggestion
    /// If `poll_next` will return `Poll::Pending`, it will be forward to
```

futures-util/src/stream/stream/mod.rs
```diff
@@ -199,6 +199,30 @@ mod buffered;
...
+
+#[cfg(not(futures_no_atomic_cas))]
+#[cfg(feature = "alloc")]
+delegate_all!(
+    /// Stream for the [`flatten_unordered`](StreamExt::flatten_unordered) method.
+    FlattenUnordered<St>(
+        flatten_unordered::FlattenUnordered<St, St::Item>
+    ): Debug + Sink + Stream + FusedStream + AccessInner[St, (.)] + New[|x: St, limit: Option<usize>| flatten_unordered::FlattenUnordered::new(x, limit)]
+    where St: Stream, St::Item: Stream, St::Item: Unpin
+);
```
taiki-e, line 214: Could you remove this wrapper type?

futures/tests/stream.rs
```diff
@@ -66,6 +69,257 @@ fn scan() {
...
+            .flat_map_unordered(1, |s| s.filter(|v| futures::future::ready(v % 2 == 0)))
+            .collect::<Vec<_>>()
+            .await;
+
+        fm_unordered.sort_unstable();
+
+        assert_eq!(fm_unordered, vec![0, 0, 2, 2, 4, 6, 8, 10]);
+    });
+
+    // wake up immmediately
```
taiki-e, line 173:
```suggestion
    // wake up immediately
```

taiki-e (commented):
Looks great!

futures/tests/stream.rs
```diff
@@ -66,6 +69,269 @@ fn scan() {
...
+
+    // stream panics
+    {
+        let st = once(async { once(async { panic!("Polled") }).boxed() }.boxed()).chain(
+            Interchanger { polled: false, base: 0, wake_immediately: true }
+                .then(|val| async move { val.boxed() }.boxed())
+                .take(10),
+        );
+
+        let stream = Arc::new(Mutex::new(st.boxed().flat_map_unordered(10, |s| s.map(identity))));
```
taiki-e, line 308: Is it possible to avoid boxing here? (That is the cause of Miri ICE)

If it is difficult to avoid boxing, you can avoid Miri ICE by skipping this test when cfg(miri) is enabled.

taiki-e (approved):
Thanks!

taiki-e:
Published in [0.3.20](https://github.com/rust-lang/futures-rs/releases/tag/0.3.20).

PR #2085, opened 2020-02-21
util: Relax bounds for FuturesUnordered

jonhoo:
There is no need for the `FuturesUnordered` constructors to require that
the provided `Fut` type implements `Future`. It needlessly requires
callers to also propagate that bound onto their constructors, all the
way up the chain. This patch removes the bounds except where necessary,
specifically for the `impl Stream`.


cramertj:
Somewhat sad that this is the best way forward given the eventual plan for implied bounds, but I suppose it's the standard way of doing things for now.

PR #2087, opened 2020-02-25
Specify units for stack_size

NikVolf:


cramertj:
Thanks for the clarification!

PR #2092, opened 2020-03-04
Add a `Cancellation` utility to `oneshot::Sender`

rphmeier:
This is a simple wrapper around the `poll_canceled` call, which is not `async/await` friendly. The primary motivation of this is to make writing code like this easier: 

```rust
let long_running_future = ...;
let mut my_sender = ...;

if let Either::Left(val) = future::select(
    long_running_future, 
    my_sender.cancellation(),
).await {
    let _ = my_sender.send(val);
} else {
    // the listener has dropped, so we can pre-emptively drop the long-running future.
}
```

Before this API, one either has to include this utility in their own codebase, or end up writing code which waits for the `long_running_future` to complete before attempting to send. This may be bad, because there are certain futures that may be incompletable but will never time out on their own.

### DIFF
--- a/futures-channel/src/oneshot.rs
+++ b/futures-channel/src/oneshot.rs
@@ -358,6 +358,15 @@ impl<T> Sender<T> {
         self.inner.poll_canceled(cx)
     }
 
+    /// Creates a future that resolves when this `Sender`'s corresponding
+    /// [`Receiver`](Receiver) half has hung up.
+    ///
+    /// This is a utility wrapping [`poll_canceled`](Sender::poll_canceled)
+    /// to expose a `Future`. 
+    pub fn await_canceled(&mut self) -> AwaitCanceled<'_, T> {
+        AwaitCanceled { inner: self }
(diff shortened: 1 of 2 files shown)


rphmeier:
Please let me know if there is more information on the documentation that you would like me to include. And go ahead and bikeshed names :P

rphmeier:
The CI failure seems unrelated, in `futures-util`.

cramertj:
> Please let me know if there is more information on the documentation that you would like me to include. And go ahead and bikeshed names :P

This looks great to me! WRT naming-- what would you think about just `canceled()`? `.await_cancelled().await` seems a little funky to me, but I don't feel too strongly.

cramertj:
Or maybe `.cancellation().await`?

rphmeier:
I like `cancellation` in both the plain and `select` contexts. Will push a renaming commit in a second.

cramertj:
Thanks!

Issue #776, opened 2018-02-18
Stream::take_until()/skip_until() a future is ready


manuels:
I would like to propose two new functions to add to `Stream`: `take_until(self, future: F)` and `skip_until(self, future: F)` where `F: Future<(), Stream::Error>`.

`skip_until()` would just be the 'reverse' of `take_until()`, so let's concentrate on `take_until()` for now:
`take_until()` should yield items of the stream until a side condition is true, i.e. the future is ready.
Note that this sounds a bit like the `take_while(self, pred: FnMut(&Self::Item) -> R)` function, that already exists, but `take_while()` *waits for new items* and then runs a predicate function on them, but `take_until()` should wait for the future asynchronously and stop the stream *even if no new item is available*.

Note that I am not really sure about the naming of the functions. There are probably better names.

troplin:
This would be nice especially for signals like ctrl-c (tokio-signal)

jonhoo:
Another use-case for this is to enable graceful shutdown of long-running futures. For example, consider a `tokio::net::TcpListener`'s `Incoming` `Stream`. You could then have a oneshot channel that indicates shutdown (maybe wrapped in a `Shared`), and write:
```rust
listener.incoming().take_until(shutdown_rx).for_each(...)
```

jonhoo:
For those in need of this, you can use my [`stream-cancel`](https://docs.rs/stream-cancel/) crate :)

PR #2096, opened 2020-03-05
util: Add take_until() in StreamExt

vojtechkral:
This is implementation similar to what was suggested in #776 and what is in the `stream-cancel` crate.

The `take_until()` utility is IMO especially useful together with tokio's new `Notify` thing (as well as any other similar futures) to crate cancellable 'infinite' tasks easily.

I thought it would be useful to have this in futures directly. (But of course I understand that there's some maintenance/backcompat burden.)

What do you think?

---

The `take_until()` method stops a stream once a `Future` passed to `take_until()`
resolves. This is implemented in the `TakeUntil` stream combinator.
Also, the future passed to `take_until()` can be extracted again out of `TakeUntil`
with the `take_future()` function. In that case, the stream continues uninterrupted.
The stopping `Future`'s eventual result can be extracted with `take_result()`.

Docs and tests included.

### DIFF
--- a/futures-util/src/stream/mod.rs
+++ b/futures-util/src/stream/mod.rs
@@ -15,7 +15,7 @@ mod stream;
 pub use self::stream::{
     Chain, Collect, Concat, Enumerate, Filter, FilterMap, FlatMap, Flatten, Fold, ForEach, Fuse,
     Inspect, Map, Next, Peek, Peekable, Scan, SelectNextSome, Skip, SkipWhile, StreamExt,
-    StreamFuture, Take, TakeWhile, Then, Zip,
+    StreamFuture, Take, TakeWhile, TakeUntil, Then, Zip,
 };
 
 #[cfg(feature = "std")]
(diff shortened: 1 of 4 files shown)


vojtechkral:
Not sure what the meaning of the CI failure is...

cramertj:
The CI failure is unrelated to your change, sorry.

cramertj:
This looks really good to me! What would you think about letting the future return an `Option<S::Item>` so that, if wanted, it could return data at the end of the stream? (possibly indicating why the stream ended early, such as on an error in the separate future's task)

vojtechkral:
@cramertj Thanks for having a look! If I understand right what you mean is that the `Option<S::Item>` returned by the futures would be yielded as the last element of the stream?

That would work, althoug the use would probably need to `map()` many potential futures to get the right return type...

Maybe instead the `TakeUntil` combinator could take a `Future` with any arbitrary `Output`, cache the result once the future resolves and provide method like `take_result(&mut self) -> Fut::Output` or similar using which the user could extract the return value once the stream is stopped. Do you think that's ok?

vojtechkral:
I've updated the PR with the `take_result()` idea above :)

cramertj:
Looks good to me! Nice idea :)

vojtechkral:
Cool, thanks for the review :)

PR #2098, opened 2020-03-13
futures: enable testing without --features thread-pool

kentfredric:
As is, running `cargo test` in the futures root directory
fails, as some of the tests are dependent on this functionality.

* <code>src/lib.rs</code>: The doc tests are a little hairy in `src/lib.rs` due to
default warnings about unused import,  combined with warnings
being `deny()`'d

* <code>tests/eventual.rs</code>: This whole test suite needs this functionality
so the test suite is turned into "no code to see here" without this
feature.

* <code>tests/mutex.rs</code>: Only one test relies on this feature, but a lot
of `cfg()` modifiers are employed to keep warnings about unused imports
to a minimum.

### DIFF
--- a/futures/src/lib.rs
+++ b/futures/src/lib.rs
@@ -25,11 +25,16 @@
 //! within macros and keywords such as async and await!.
 //!
 //! ```rust
+//! # #[cfg(features = "thread-pool")]
 //! # use futures::channel::mpsc;
+//! # #[cfg(features = "thread-pool")]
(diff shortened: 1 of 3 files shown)


kentfredric:
I may take a second pass at this approach and overhaul tests/ one test at a time to make sure they work under all --feature combinations.

`cargo hack` is such a useful tool here. :)

kentfredric:
The tests/ aspect of this is now superseded by #2101

PR #2100, opened 2020-03-17
full working example for oneshot

tworec:
A friend of mine was confused by the current example for oneshot channel.
I'm proposing more illustrative and *WORKING* version here ;)

Inspiration from: https://doc.rust-lang.org/std/sync/mpsc/fn.channel.html

### DIFF
--- a/futures-channel/src/oneshot.rs
+++ b/futures-channel/src/oneshot.rs
@@ -82,22 +82,22 @@ struct Inner<T> {
 ///
 /// ```
 /// use futures::channel::oneshot;
-/// use futures::future::FutureExt;
-/// use std::thread;
-///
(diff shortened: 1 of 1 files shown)


tworec:
notice that lines starting with
```
/// # 
```
are not visible on [docs.rs](https://docs.rs/futures/0.3.4/futures/channel/oneshot/fn.channel.html)

cramertj:
Nice, this looks like a good improvement. Would you mind also moving this test over to use async/await like our other examples?

tworec:
is this what you are asking for?

cramertj:
Normally we hide the `block_on` in our examples, but this works fine, thanks!

PR #2101, opened 2020-03-19
futures: rework tests/ to work under all --feature combinations

kentfredric:
This patch set extensively restructures the tests to: 

- Make import ownership of individual tests more apparent
- Simplify feature gating individual tests
- Simplify potential future test set fragmentation
- Execute all tests that can be expected to compile under a given configuration.

With this patch set, one can now do:

```bash
 cargo -q hack \
      --each-feature \
      --skip bilock \
      --skip cfg-target-has-atomic \
      --skip read-initializer \
      test -q --tests
```

```bash
 cargo -q hack \
      --features unstable \
      --each-feature \
      --skip unstable \
      test -q --tests
```

```bash
 cargo -q hack \
      --feature-powerset \
      --skip bilock \
      --skip cfg-target-has-atomic \
      --skip read-initializer \
      test -q --tests
```

```bash
 cargo -q hack \
      --features unstable \
      --feature-powerset \
      --skip unstable \
      test -q --tests
```

And experience no compilation failures, or compilation warnings.

( There's a bit of redundancy in my examples as `--feature-powerset` is a superset of `--each-feature`, but `--each-feature` is better for doing initial discovery of missing features, and a separation must be made for the 3 features that require "unstable" turned on )

Nothing in this PR addresses either the internal tests ( test --lib ) or the doc tests ( test --doc ) as they're both more contentious and difficult to do, and I'd rather see this PR merged first before embarking on shoring up those problems.

And it also may be worth adding these `cargo hack` invocations into the standard CI tests to ensure these features continue to be adequately tested.

Tips: Due to the approaches used for testing, this can cause `target/debug/` to become *incredibly* large, and prone to running you out of disk space if you're resource constrained.

I've personally taken to throwing
```
 CARGO_INCREMENTAL=0 RUSTFLAGS="-C debuginfo=0"
``` 

Into things, which cuts my `target/debug` dir down from `4.5GB` for the first `cargo hack` command, down to `2.6GB`.

Obviously this is a provisional PR and I can rebase like a champion, so any changes that would be required to make this merge welcome.

### DIFF
--- a/futures/tests/abortable.rs
+++ b/futures/tests/abortable.rs
@@ -1,11 +1,10 @@
-use futures::channel::oneshot;
-use futures::executor::block_on;
-use futures::future::{abortable, Aborted, FutureExt};
-use futures::task::{Context, Poll};
-use futures_test::task::new_count_waker;
-
+#[cfg(all(feature = "alloc", feature = "executor"))]
 #[test]
 fn abortable_works() {
+    use futures::channel::oneshot;
+    use futures::future::{abortable, Aborted};
+    use futures::executor::block_on;
+
     let (_tx, a_rx) = oneshot::channel::<()>();
     let (abortable_rx, abort_handle) = abortable(a_rx);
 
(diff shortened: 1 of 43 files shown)


cramertj:
This is awesome! Thank you so much for all this work!

PR #2103, opened 2020-03-19
future-utils: Avoid compile failures in tests w/o "std"

kentfredric:
This failure is not reproducable under the main futures workspace.

To replicate, one must first:
```
  future-utils/: cargo package
```
And then:
```
  cd "<prefix>/target/package/future-utils-0.3.4"
  cargo test --no-default-features
```
This patch remedies this issue by:

- Moving the imports required for this test into the test itself
- Gating the test so it only compiles when `--feature=std` (and thus,
  the `#[no_std]` in `lib.rs` doesn't activate)

This approach also future proofs the code somewhat and makes it
potentially easier to relocate to `tests/` one day.

Closes: https://github.com/rust-lang/futures-rs/issues/2102


cramertj:
Thanks!

PR #2104, opened 2020-03-22
reduce box allocation in bilock.

laizy:


cramertj:
Cute! Looks good to me :)

PR #2116, opened 2020-04-04
Refactor StreamExt#forward type bounds

xfix:
This makes the bounds easier to understand in documentation by removing bounds referring to each other.

Previously, the bounds in documentation appeared as such:

```rust
S: Sink<Self::Ok>,
Self: TryStream<Error = <S as Sink<Self::Ok>>::Error>,
```

This changes them to the following simpler to understand form.

```rust
S: Sink<Self::Ok, Error = Self::Error>,
Self: TryStream,
```


cramertj:
Nice!

PR #2117, opened 2020-04-06
Get rid of Chain as the building block for other combinators and remo…

Diggsey:
…ve some unsafe code

This is just the first step in a potential refactor that could remove a lot of safe and unsafe code, and I wanted to test the waters before going all out. As you can see the amount of code that can be deleted this way is high!

The basic ideas are:
- Switch from `pin-utils` to `pin-project` as it has a safe API and supports enums, which results in smaller combinator sizes in memory.
- Remove `Chain` as it is unnecessarily complicated. We can instead build up from primitive combinators like `Flatten` and `Map`.

There are a couple of open questions:
1) Is it OK to use raw type aliases as I have done with `Then`? Strictly speaking, it's a breaking change, however, it's exceptionally unlikely that it will actually cause any problems. If we don't want to do this, the fallback would be to have a macro to define a load of newtype wrappers that automatically implement the relevant traits.

2) Is it OK to get rid of the unused type parameter on  `Then`? This is a breaking change, although you wouldn't typically name this parameter during normal usage. If we don't want to get rid of it, then we'll need a newtype wrapper for at least `Then` and any other combinators with unused generic parameters.


Diggsey:
Oh, and this also fixes #2118 :)

Diggsey:
The test `select_on_non_unpin_size` is failing because I reduced the size of the future combinator, but I have no idea what it's actually trying to test by asserting on that size? Am I safe to just update the size?

futures-util/src/future/future/flatten.rs
```diff
@@ -1,56 +1,84 @@
...
+    type Output = <Fut::Output as Future>::Output;
 
-    pub(super) fn new(future: Fut) -> Flatten<Fut> {
-        Flatten {
-            state: Chain::new(future, ()),
-        }
+    fn poll(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
+        Poll::Ready(loop {
+            match self.as_mut().project() {
+                __InternalFlattenProjection::First(f) => {
```
Nemo157, line 42: This is an internal type of the generated code which shouldn't be used. You can use the `#[project]` attribute to allow matching on the projected variants.

Nemo157:
> The test `select_on_non_unpin_size` is failing because I reduced the size of the future combinator, but I have no idea what it's actually trying to test by asserting on that size? Am I safe to just update the size?

That is failing because of changes to the generator transform in `rustc`, I've opened https://github.com/rust-lang/futures-rs/pull/2119 to fix it.

futures-util/src/future/future/flatten.rs
Diggsey: Yeah, I'ved fixed that in the larger PR I have :)

Diggsey:
Superceded by #2128

PR #2119, opened 2020-04-07
Update expected size of `select_on_non_unpin_size`

Nemo157:
Probably comes from https://github.com/rust-lang/rust/pull/69837

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -145,7 +145,8 @@ matrix:
     - name: cargo doc
       rust: nightly
       script:
-        - RUSTDOCFLAGS=-Dwarnings cargo doc --workspace --no-deps --all-features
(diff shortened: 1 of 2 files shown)


Nemo157:
Doc build failure is https://github.com/rust-lang/rust/issues/70814

taiki-e:
> Doc build failure is likely [rust-lang/rust#70814](https://github.com/rust-lang/rust/issues/70814)

I would like to suggest adding  `-Aunused_braces` to `RUSTDOCFLAGS` until it is fixed.

https://github.com/rust-lang/futures-rs/blob/2f8943da9a0fcef208e86ae2f0f0152191629dcc/.travis.yml#L148

PR #2121, opened 2020-04-08
Code cleanup in shared.rs

Lucretiel:
- Replaced several uses of `if let` with a more idiomatic `match`
- added `will_wake` optimization
- cleaned up verbose logic in `record_waker`
- rustfmt


cramertj:
Nice, looks good!

PR #2122, opened 2020-04-10
Backport to 0.1: Avoid starvation from FuturesUnordered::poll_next

krallin:
This backports https://github.com/rust-lang/futures-rs/pull/2049/ to the
0.1 branch. Without this change, polling > 200 futures trough a
FuturesUnordered on a Tokio 0.2 executor results in a busy loop in
Tokio's cooperative scheduling module.

See for a repro of where this breaks:
https://github.com/tokio-rs/tokio/issues/2390

Tested by running the reproducer I submitted there. Without this change,
it hangs forever (spinning on CPU). With the change, it doesn't.


cramertj:
Thanks for the backport!

PR #2123, opened 2020-04-10
Introduce `ready_chunks` adaptor

glebpom:
This is the implementation of `ready_chunks` adapter. It is similar to `chunks`, but collects the data while the underlying stream returns `Poll::Ready`. Otherwise, the collected vector immediately returns so that the stream never blocks. The motivation behind it is that often we need to group multiple elements from the stream (specifically mpsc channels) and send them in a single system call. For example, it may be usable with calls like [`sendmmsg`](https://github.com/nix-rust/nix/pull/1208) or [`UDP GSO`](https://github.com/nix-rust/nix/pull/1209)

### DIFF
--- a/futures-util/src/stream/mod.rs
+++ b/futures-util/src/stream/mod.rs
@@ -24,6 +24,9 @@ pub use self::stream::CatchUnwind;
 #[cfg(feature = "alloc")]
 pub use self::stream::Chunks;
 
+#[cfg(feature = "alloc")]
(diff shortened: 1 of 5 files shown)


cramertj:
This looks great to me, thanks!

cramertj:
This looks great to me, thanks! I'd also be interested in a `read_next_ready_chunk(&mut buffer)` or similar that would read chunks into a preallocated vector to avoid the repeated heap allocations every time.

Issue #2061, opened 2020-02-04
error: expected item after attributes using Rust 1.41


galich:
Our app fails to compile because of the following error given by rustc 1.41.0 (5e1a79984 2020-01-27)

```
error: expected item after attributes
   --> C:\Users\serg\.cargo\registry\src\github.com-1ecc6299db9ec823\futures-0.3.1\src\lib.rs:529:19
    |
529 |     #[macro_export]
    |                   ^
```

Nemo157:
This appears to be an incompatibility introduced with `futures-util` 0.3.2, probably related to #1891. Easy fix is to update to 0.3.2 of `futures` as well or downgrade `futures-util` back to 0.3.2.

@cramertj worth yanking 0.3.2 (maybe just `-util`) and publishing a fix as 0.3.3? Seems like a difficult situation to get into unless you're pinning the version of `futures` but not pinning the underlying libraries.

Nemo157:
Tested with

```console
> cargo download -x futures==0.3.1
INFO: cargo-download v0.1.2
INFO: Crate `futures==0.3.1` downloaded successfully
INFO: Crate content extracted to ./futures-0.3.1/
> cd futures-0.3.1
> cargo +1.41.0 build
   Compiling proc-macro2 v1.0.8
   Compiling unicode-xid v0.2.0
   Compiling syn v1.0.14
   Compiling proc-macro-nested v0.1.3
   Compiling memchr v2.3.0
   Compiling futures-sink v0.3.2
   Compiling futures-core v0.3.2
   Compiling slab v0.4.2
   Compiling futures-io v0.3.2
   Compiling futures-task v0.3.2
   Compiling pin-utils v0.1.0-alpha.4
   Compiling futures-channel v0.3.2
   Compiling quote v1.0.2
   Compiling proc-macro-hack v0.5.11
   Compiling futures-macro v0.3.2
   Compiling futures-util v0.3.2
   Compiling futures-executor v0.3.2
   Compiling futures v0.3.1 (/private/var/folders/0p/5yvmrvhj5w3_vy1y8x7dvk7m0000gn/T/tmp.hJzVLqqR/futures-0.3.1)
error: expected item after attributes
   --> src/lib.rs:529:19
    |
529 |     #[macro_export]
    |                   ^

error: aborting due to previous error

error: could not compile `futures`.

To learn more, run the command again with --verbose.
> cargo update -p futures-util -p futures-executor --precise 0.3.1
[...]

PR #2124, opened 2020-04-10
Support reexport for procedural macros

### LINKED ISSUE #2099: Cannot use macros when crate is renamed
When the `futures` crate is renamed:
```toml
futures03 = { version = "0.3.1", package = "futures" }
```

Macros like `try_join` cannot find the crate.
Eg:
```rust
async fn test() {
    futures03::try_join!()
}
```

fails with
```
failed to resolve: could not find `futures` in `{{root}}`
```

This impedes migration from version 0.1 to version 0.3, since one of the two futures crates must be renamed.

taiki-e:
Create a hidden module that reexports all items in the crate and import it at the beginning of the wrapper macro. This allows supporting not only rename but also reexport.

```rust
#[doc(hidden)]
pub mod __reexport {
    #[doc(hidden)]
    pub use crate::*;
}

#[doc(hidden)]
#[proc_macro_hack(support_nested)]
pub use futures_macro::join_internal;

#[macro_export]
macro_rules! join {
    ($($tokens:tt)*) => {{
        // The code generated by proc-macro uses a path starting with `__futures_crate`.
        // Also, `use $crate;` in macro is not allowed (https://github.com/rust-lang/rust/issues/37390),
        // so create hidden module `__reexport`.
        use $crate::__reexport as __futures_crate;
        $crate::join_internal! {
            $( $tokens )*
        }
    }}
}
```

Note that proc-macro is renamed to `*_internal`.
IIRC, it is not expected to use futures-macro directly, so this is probably fine...
(Maybe using `=` requirements would help to avoid some problems like #2061.)

Closes #2099

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,6 +11,9 @@ members = [
   "futures-util",
   "futures-test",
 
+  "futures/tests/macro-tests",
+  "futures/tests/macro-reexport",
+
   "examples/functional",
   "examples/imperative",
 ]
--- a/futures-macro/src/join.rs
+++ b/futures-macro/src/join.rs
@@ -4,31 +4,16 @@ use proc_macro::TokenStream;
 use proc_macro2::{Span, TokenStream as TokenStream2};
 use quote::{format_ident, quote};
 use syn::parse::{Parse, ParseStream};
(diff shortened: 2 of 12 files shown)


cramertj:
Nice, thanks for fixing this!

PR #2139, opened 2020-04-25
impl AsyncRead/AsyncWrite for Arc<T>

stjepang:
In `async-std` it is common to share I/O types by storing them inside an `Arc`. It would be nice then for `Arc<T>` to also be an I/O type by implementing `AsyncRead` and `AsyncWrite`. In fact, this is exactly why the `io-arc` crate was invented: https://docs.rs/io-arc/1.0.0/io_arc/

I'm now working on a new runtime where this pattern is pervasive and am also using a variant of `IoArc`. It would be nice for `futures` to add this impl instead.

### DIFF
--- a/futures-io/src/lib.rs
+++ b/futures-io/src/lib.rs
@@ -348,6 +348,7 @@ mod if_std {
         deref_async_read!();
     }
 
+    #[allow(single_use_lifetimes)] // FIXME: https://github.com/rust-lang/rust/issues/69952
     impl<T: ?Sized + Unpin> AsyncRead for Arc<T>
     where
         for<'a> &'a T: AsyncRead,
@@ -450,6 +451,7 @@ mod if_std {
         deref_async_write!();
     }
 
+    #[allow(single_use_lifetimes)] // FIXME: https://github.com/rust-lang/rust/issues/69952
     impl<T: ?Sized + Unpin> AsyncWrite for Arc<T>
     where
         for<'a> &'a T: AsyncWrite,
--- a/futures-io/src/lib.rs
+++ b/futures-io/src/lib.rs
@@ -29,6 +29,7 @@ mod if_std {
     use std::io;
     use std::ops::DerefMut;
     use std::pin::Pin;
+    use std::sync::Arc;
     use std::task::{Context, Poll};
 
     // Re-export some types from `std::io` so that users don't have to deal
@@ -347,6 +348,28 @@ mod if_std {
         deref_async_read!();
     }
 
+    impl<T: ?Sized + Unpin> AsyncRead for Arc<T>
+    where
+        for<'a> &'a T: AsyncRead,
+    {
+        #[cfg(feature = "read-initializer")]
+        unsafe fn initializer(&self) -> Initializer {
+            (&**self).initializer()
+        }
+
+        fn poll_read(self: Pin<&mut Self>, cx: &mut Context<'_>, buf: &mut [u8])
+            -> Poll<Result<usize>>
+        {
+            Pin::new(&mut &**self).poll_read(cx, buf)
+        }
+
+        fn poll_read_vectored(self: Pin<&mut Self>, cx: &mut Context<'_>, bufs: &mut [IoSliceMut<'_>])
+            -> Poll<Result<usize>>
+        {
+            Pin::new(&mut &**self).poll_read_vectored(cx, bufs)
+        }
+    }
+
     impl<P> AsyncRead for Pin<P>
     where
         P: DerefMut + Unpin,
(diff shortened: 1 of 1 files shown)


futures-io/src/lib.rs
```diff
@@ -347,6 +348,28 @@ mod if_std {
...
+            -> Poll<Result<usize>>
+        {
+            Pin::new(&mut &**self).poll_read(cx, buf)
+        }
+
+        fn poll_read_vectored(self: Pin<&mut Self>, cx: &mut Context<'_>, bufs: &mut [IoSliceMut<'_>])
+            -> Poll<Result<usize>>
+        {
+            Pin::new(&mut &**self).poll_read_vectored(cx, bufs)
+        }
```
taiki-e, line 370:
```suggestion
        deref_async_read!();
```

futures-io/src/lib.rs
```diff
@@ -427,6 +450,31 @@ mod if_std {
...
+            Pin::new(&mut &**self).poll_write_vectored(cx, bufs)
+        }
+
+        fn poll_flush(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Result<()>> {
+            Pin::new(&mut &**self).poll_flush(cx)
+        }
+
+        fn poll_close(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Result<()>> {
+            Pin::new(&mut &**self).poll_close(cx)
+        }
```
taiki-e, line 475:
```suggestion
        deref_async_write!();
```

stjepang: This doesn't compile - note there is a slight difference between what this does and what `deref_async_write!()` does:

* `Pin::new(&mut &**self).poll_write(cx, buf)`
* `Pin::new(&mut **self).poll_write(cx, buf)`

taiki-e: Oops, you are right. Sorry for the wrong review!

Matthias247:
This seems broken for all `AsyncRead/AsyncWrite` implementations which are only able to store exactly one `Waker` - or one per direction - which are likely most of them. 

Unless you can add a constraint that guarantees that the underlying object is able to handle multiple concurrent IOs safely, I don't think this should be added.

stjepang:
@Matthias247 `AsyncRead`/`AsyncWrite` can store as many wakers as are needed, not just two.

Both [`async-std`](https://docs.rs/async-std) and [`smol`](https://docs.rs/smol) can safely handle any number of concurrent I/O handles.

Nemo157:
`where for<'a> &'a T: AsyncRead` at the very least _implies_ that it should support cross-task shared polling, checking Tokio it doesn't appear to provide shared impls, and checking async-std it does and it does appear to support multiple tasks.

My problem with these impls is just that it makes the chance of wrongly interleaved reads/writes higher, I assume in most cases the intention is to just support splitting into a read/write pair, which IMO would be better supported by inherent `split` methods.

If these are to be provided, should there be the same for `AsyncSeek` and `Stream` as well? `Stream` in particular seems like it avoids the issue I see with the IO impls since it provides atomic units of data instead of reading/writing some unknown amount at a time.

Matthias247:
> Both async-std and smol can safely handle any number of concurrent I/O operations.

Maybe. But letting users guess what is supported just sounds like a bad idea. If a certain implementation supports splitting/cloning, they can directly add methods that indicate that. I am not sure what the benefit of the generalization here is.

And some more background: I think `AsyncRead/Write/Stream/Sink` had been modelled after the `Future` type itself. The all take a `Context/Waker` as argument, and require a mutable pinned reference. `Future`s are commonly expected not to store multiple `Waker`s, and rather to replace the stored `Waker` on each poll.

stjepang:
> My problem with these impls is just that it makes the chance of wrongly interleaved reads/writes higher

Can you elaborate what is the problem here? How could operations be interleaved *wrongly* and under what circumstances?

To clarify, I *think* I understand what you're worried about - it's just that my opinion is that issues in question are so contrived and unrealistic they simply aren't a problem. Not once this came up in async-std's history. And the worst issue I can see is a memory leak. But then again, there are so many other non-contrived memory leaks that make this one not worth thinking about, in my opinion.

> If these are to be provided, should there be the same for `AsyncSeek` and `Stream` as well?

Yes - why not! :) Channels in `async-std` and `piper` are MPMC - their `Receiver`s implement `Stream`, but it would be trivial to also implement the trait for `&Receiver`.

> But letting users guess what is supported just sounds like a bad idea. If a certain implementation supports splitting/cloning, they can directly add methods that indicate that. I am not sure what the benefit of the generalization here is.

The point of the generalization is that we're forced to re-invent a fundamental type in `std` and use our own `Arc` because the `futures` crate doesn't have these impls yet.

> The all take a `Context/Waker` as argument, and require a mutable pinned reference. `Future`s are commonly expected not to store multiple `Waker`s, and rather to replace the stored `Waker` on each poll.

I fully agree! If `T: Future`, it can have only one `Waker`. But if `&T: Future`, you can have an infinite number of `&T`s, and therefore an infinite number of `Waker`s.

Nemo157:
By wrong interleaving I mean things like splinched reads

```rust
let x = spawn({
  let reader = reader.clone();
  async move {
    let mut buf = [0; 2];
    reader.read_exact(&mut buf).await?;
    Ok(buf)
  }
});
let y = spawn(async move {
  let mut buf = [0; 2];
  reader.read_exact(&mut buf).await?;
  Ok(buf)
});
let (x, y) = try_join(x, y).await?;
```

Depending on exact timings a data stream sent as `[0, 1, 2, 3]` could arrive as `x: [0, 1], y: [2, 3]` or `x: [0, 2], y: [1, 3]` or a handful of other results.

The same can happen with interleaved writes from concurrent tasks.

stjepang:
@Nemo157 Thanks for the example!

So, I see the issue, but this is also possible with I/O in the standard library, so why apply different standards here?

Nemo157:
I feel like the std shared IO impls were a mistake too, they might more directly match the underlying platform capabilities, but I've not seen the utility of them compared to the potential for issues (though TBH I've not seen any issues actually arise in real code).

I'm not really against adding these impls, if IO runtimes want to support shared access then having these will just make using those slightly easier, it's the IO runtime's choice to support them.

najamelan:
> This seems broken for all `AsyncRead/AsyncWrite` implementations which are only able to store exactly one `Waker`

Those types should not implement `AsyncRead/AsyncWrite` for `&T` then. They wouldn't be affected by this PR.

I think providing blanket impls where possible is a feature. It removes friction when designing API's and also when using them.

futures-io/src/lib.rs
```diff
@@ -347,6 +348,29 @@ mod if_std {
         deref_async_read!();
     }
 
+    #[allow(single_use_lifetimes)] // FIXME: https://github.com/rust-lang/rust/issues/69952
```
taiki-e, line 351: fwiw: This is https://github.com/rust-lang/rust/issues/55058, not 69952

taiki-e:
looks like there is no known AsyncRead/AsyncWrite implementation that actually supports any number of concurrent I/O operations (https://github.com/stjepang/async-io/pull/38#issuecomment-706919653), so I'll close this for now.

PR #2146, opened 2020-05-07
Implement Send + Sync for AssertUnmoved

Thomasdezeeuw:


futures-test/src/future/assert_unmoved.rs
```diff
@@ -17,6 +17,11 @@ pub struct AssertUnmoved<Fut> {
     _pinned: PhantomPinned,
 }
 
+// Safety: having a raw pointer in a struct makes it `!Send`, however the
+// pointer is never dereferenced so this is safe.
+unsafe impl<Fut: Sync + Send> Send for AssertUnmoved<Fut> {}
+unsafe impl<Fut: Sync + Send> Sync for AssertUnmoved<Fut> {}
```
Nemo157, line 23: These shouldn't each need both, right?

```suggestion
unsafe impl<Fut: Send> Send for AssertUnmoved<Fut> {}
unsafe impl<Fut: Sync> Sync for AssertUnmoved<Fut> {}
```

Thomasdezeeuw: I think you're right, but I just copied the implementation of the `Arc` implementation in std lib: https://doc.rust-lang.org/std/sync/struct.Arc.html#impl-Send. I'm ok with changing it, but I don't think it matter too much.

Nemo157: `Arc` is different since it allows creating cross-thread references from an owned instance (so can only be `Send` if `T: Sync`) and allows cross-thread cloning then `try_unwrap` to move a value (so can only be `Sync` if `T: Send`).

`AssertUnmoved` is fundamentally a newtype wrapper around `Fut`, so should use the normal inductive auto-trait definitions. It probably won't make much difference since so much async stuff is built around `+ Send + Sync` currently, but I have seen some stuff that is only `+ Send` occasionally, so it'd be good to support them.

Thomasdezeeuw: Ok, done.

Thomasdezeeuw:
The CI seems to be stuck, but it passed on Travis: https://travis-ci.com/github/rust-lang/futures-rs/builds/164448884.

PR #2148, opened 2020-05-08
Reduce pin related unsafe code

taiki-e:
This removes many of the remaining pin related unsafe code in futures-util and futures-test.

r? @cramertj @Nemo157

### DIFF
--- a/futures-test/src/future/assert_unmoved.rs
+++ b/futures-test/src/future/assert_unmoved.rs
@@ -1,7 +1,6 @@
 use futures_core::future::Future;
 use futures_core::task::{Context, Poll};
 use pin_project::{pin_project, pinned_drop};
-use std::marker::PhantomPinned;
 use std::pin::Pin;
 use std::ptr;
 use std::thread::panicking;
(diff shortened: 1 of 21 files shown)


futures-test/src/future/assert_unmoved.rs
```diff
@@ -9,11 +9,14 @@ use std::thread::panicking;
...
 /// method.
+#[pin_project(PinnedDrop)]
 #[derive(Debug, Clone)]
 #[must_use = "futures do nothing unless you `.await` or poll them"]
 pub struct AssertUnmoved<Fut> {
+    #[pin]
     future: Fut,
     this_ptr: *const AssertUnmoved<Fut>,
+    #[pin]
     _pinned: PhantomPinned,
```
taiki-e, line 20: Oh, I just added [an option for this in pin-project.](https://docs.rs/pin-project/0.4.14/pin_project/attr.pin_project.html#unpin)

futures-util/src/future/future/map.rs
```diff
@@ -56,34 +44,19 @@ impl<Fut, F, T> Future for Map<Fut, F>
     type Output = T;
 
     #[project]
+    #[project_replace]
```
taiki-e, line 47: It's not nice to need two wrapper attributes -- I'll fix it soon.

taiki-e: Fixed in pin-project 0.4.15

cramertj:
This looks great, thanks!

PR #2151, opened 2020-05-10
Fix documentation on futures::future::Map

nvanbenschoten:
The documentation was pointing at `futures::future::FutureExt::flatten` instead of at `futures::future::FutureExt::map`.


Nemo157:
Thanks!

PR #2154, opened 2020-05-13
Backport to 0.1: Shared must relinquish control to the executor if repolled

krallin:
This backports #2136 to Futures 0.1. There isn't much to add on top of
what's mentioned in #2136: the same bug exists in Futures 0.1, and it'll
fail in the same way when polled in recent versions of Tokio (#2418).

Backporting to 0.1 allows codebases that still have some Futures 0.1
code around to still use newer versions of Tokio.

### DIFF
--- a/src/future/shared.rs
+++ b/src/future/shared.rs
@@ -59,9 +59,8 @@ struct Notifier {
 
 const IDLE: usize = 0;
 const POLLING: usize = 1;
-const REPOLL: usize = 2;
-const COMPLETE: usize = 3;
-const POISONED: usize = 4;
+const COMPLETE: usize = 2;
+const POISONED: usize = 3;
 
 pub fn new<F: Future>(future: F) -> Shared<F> {
     Shared {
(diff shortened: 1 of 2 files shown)


krallin:
Any thoughts? :)

taiki-e:
#2122 hasn't been released yet, so it's nice to publish the next release of 0.1 after merging this.

(We need to decide how long we should maintain 0.1 branch, but this should be fine given that we have another backport already merged.)

krallin:
Could we get this merged and released? Having futures 0.1 being broken on newer versions of Tokio makes upgrading Tokio more painful than it should be :(

taiki-e:
Thanks, I'll prepare the next release.

krallin:
Thanks a lot @taiki-e !

krallin:
Hey @taiki-e , I'm sorry to be a bit of a bother here, but has this made it into a new release yet? Thanks :)

taiki-e:
Published 0.1.30, sorry for the delay.

krallin:
Thanks very much!

PR #2158, opened 2020-05-15
Add oneshot::Sender::is_connected_to method

canndrew:
I was gonna open an issue asking for this, but it was trivial to implement so I figured I'd just go straight to a PR.

This adds a method to `oneshot::Sender` to check whether it's connected to a given `Receiver`:

```rust
/// Tests to see whether this `Sender` is connected to the given `Receiver`. That is, whether
/// they were created by the same call to `channel`.
pub fn is_connected_to(&self, receiver: &Receiver<T>) -> bool {
    Arc::ptr_eq(&self.inner, &receiver.inner)
}
```

Does this seem like a good idea? It would be useful for me.

### DIFF
--- a/futures-channel/src/oneshot.rs
+++ b/futures-channel/src/oneshot.rs
@@ -377,6 +377,12 @@ impl<T> Sender<T> {
     pub fn is_canceled(&self) -> bool {
         self.inner.is_canceled()
     }
+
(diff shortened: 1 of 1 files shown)


taiki-e:
Thanks!

PR #2159, opened 2020-05-15
Add test utility that verifies an AsyncWrite is closed correctly

Nemo157:


Nemo157:
I used a slightly simplified version of this utility in https://github.com/Nemo157/async-compression/pull/92 to regression test a silly mistake.

futures-test/src/io/write/track_closed.rs
```diff
@@ -0,0 +1,110 @@
...
+    io::{Error, ErrorKind, IoSlice, Result},
+    pin::Pin,
+    task::{Context, Poll},
+};
+
+/// I/O wrapper that tracks whether it has been closed.
+///
+/// See the [`track_closed`] method for more details.
+///
+/// [`track_closed`]: super::AsyncWriteTestExt::limited_write
```
Nemo157, line 12:
```suggestion
/// [`track_closed`]: super::AsyncWriteTestExt::tracked_closed
```

Nemo157:
Yep, makes sense to have it working for sink too

PR #2160, opened 2020-05-17
Add doc feature flags.

jmjoy:
Is it a good idea to add some feature flags in `docs.rs`, just like `tokio` or `async-std`? For the users to read the docs friendly.
And the other `futures` crates also should have feature flags if it is a good idea.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -146,7 +146,7 @@ matrix:
       rust: nightly
       script:
         # TODO: Remove -Aunused_braces once https://github.com/rust-lang/rust/issues/70814 is fixed
-        - RUSTDOCFLAGS="-Dwarnings -Aunused_braces" cargo doc --workspace --no-deps --all-features
+        - RUSTDOCFLAGS="-Dwarnings -Aunused_braces --cfg docsrs" cargo doc --workspace --no-deps --all-features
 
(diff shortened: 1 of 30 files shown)


Nemo157:
:+1: I'm a big believer in showing `doc(cfg)` data on docs.rs.

My preferred setup is to not use a `feature` for this as it's not something anyone should be activating in their dependency list, instead in `Cargo.toml` setup

```toml
[package.metadata.docs.rs]
rustdoc-args = ["--cfg", "docsrs"]
```

then in code you can test this like

```rust
#![cfg_attr(docs, feature(doc_cfg))]
```

The `cargo doc` CI job should also have this added by putting `--cfg docsrs` into the `RUSTDOCFLAGS` at https://github.com/rust-lang/futures-rs/blob/8a95e9a84896fb3ce944801c12115b3784cf40ac/.travis.yml#L149

jmjoy:
@Nemo157 I have changed to using `cfg(docsrs)` and added some no default feature flags.

taiki-e (approved):
Thanks! This PR doesn't touch features that are enabled by default, such as the std feature, but I think it's ok because displaying doc-cfg with those features is quite noisy.

PR #2162, opened 2020-05-19
Fix description for Flatten

earthengine:
Fixes a documentation issue as evidented in the following snapshot from [docs.rs](https://docs.rs/futures/0.3.5/futures/stream/index.html):

image


taiki-e:
Hm, CI still doesn't work. This is a link fix, so I think it's okay to merge without running CI, but I'm not admin so I can't merge. 
cc @cramertj Could you merge this?

cramertj:
Done, sorry for all the delayed responses. I'm finally getting around to triaging all my hundreds of stale GH notifications :(

PR #2169, opened 2020-05-31
Add WeakShared.

khuey:
This makes implementing a cache with futures much easier. With the current architecture a cache that uses Shareds cannot honor the "drop as cancellation" feature of futures. And the cache is forced to either poll futures itself to ensure they're polled to completion or to leave half-polled futures dangling inside it (potentially consuming resources like sockets or database connections). WeakShared is to Shared as Weak is to Arc. If any copy of the underlying Shared exists that has not be dropped or polled to completion the WeakShared can be upgraded to it. This makes it possible to construct a cache that does not entrain futures, thus honoring the "drop as cancellation" feature.

### DIFF
--- a/futures-util/src/future/future/shared.rs
+++ b/futures-util/src/future/future/shared.rs
@@ -7,7 +7,7 @@ use std::fmt;
 use std::pin::Pin;
 use std::sync::atomic::AtomicUsize;
 use std::sync::atomic::Ordering::{Acquire, SeqCst};
-use std::sync::{Arc, Mutex};
+use std::sync::{Arc, Mutex, Weak};
 
 /// Future for the [`shared`](super::FutureExt::shared) method.
 #[must_use = "futures do nothing unless you `.await` or poll them"]
@@ -26,6 +26,9 @@ struct Notifier {
     wakers: Mutex<Option<Slab<Option<Waker>>>>,
 }
 
+/// A weak reference to a [`Shared`] that can be upgraded much like an `Arc`.
+pub struct WeakShared<Fut: Future>(Weak<Inner<Fut>>);
+
 // The future itself is polled behind the `Arc`, so it won't be moved
 // when `Shared` is moved.
 impl<Fut: Future> Unpin for Shared<Fut> {}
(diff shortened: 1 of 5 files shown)


khuey:
Ping?

khuey:
Looks like CI is finally happy.

futures-util/src/future/future/shared.rs
```diff
@@ -26,6 +26,9 @@ struct Notifier {
     wakers: Mutex<Option<Slab<Option<Waker>>>>,
 }
 
+/// A weak reference to a `Shared` than can be upgraded much like an `Arc`.
```
taiki-e, line 29:
```suggestion
/// A weak reference to a [`Shared`] than can be upgraded much like an [`Arc`].
///
/// [`Arc`]: std::sync::Arc
```

futures-util/src/future/future/shared.rs
```diff
@@ -107,6 +116,15 @@ where
         }
         None
     }
+
+    /// Returns [`Some`] containing a [`WeakShared`] that may later be upgraded back to this [`Shared`]
+    /// or [`None`] if it has already been polled to completion.
```
taiki-e, line 121:
```suggestion
    /// Creates a new [`WeakShared`] to this [`Shared`].
    ///
    /// Returns [`None`] if it has already been polled to completion.
```

futures-util/src/future/future/shared.rs
```diff
@@ -314,3 +332,16 @@ impl ArcWake for Notifier {
         }
     }
 }
+
+impl<Fut: Future> WeakShared<Fut>
+{
+    /// Attempts to upgrade this [`WeakShared`] into a [`Shared`]. Returns
+    /// [`None`] if all copies of the [`Shared`] have been dropped or polled
+    /// to completion, otherwise [`Some`]
```
taiki-e, line 340:
```suggestion
    /// Attempts to upgrade this [`WeakShared`] into a [`Shared`]. 
    ///
    /// Returns [`None`] if all clones of the [`Shared`] have been dropped or polled
    /// to completion.
```

futures/tests/shared.rs
```diff
@@ -143,6 +143,35 @@ fn peek() {
...
+    assert!(weak.upgrade().is_some());
+
+    assert_eq!(block_on(&mut shared2).unwrap(), 42);
+    // Now that all `Shared`s have been exhausted, we should not be able
+    // to get a new `WeakShared` or upgrade an existing one.
+    assert!(weak.upgrade().is_none());
+    assert!(shared2.downgrade().is_none());
+}
+
+#[cfg(all(feature = "alloc", feature = "executor"))] // channel:: + executor::
```
taiki-e, line 174:
```suggestion
```

taiki-e (commented):
Thanks for the PR and sorry for the delay reviewing. I have a few nits, but otherwise LGTM.

futures-util/src/future/future/shared.rs
khuey: I ultimately went with "for this [`Shared`]" here.

taiki-e (approved):

khuey:
Thanks.

PR #2176, opened 2020-06-05
Remove uses of pin_project::project attribute

taiki-e:
pin-project will deprecate the project attribute due to some unfixable
limitations.

Refs: https://github.com/taiki-e/pin-project/issues/225

*Although the deprecation of the project attribute has not been released yet, this PR is submitted in advance to avoid the CI from being broken by deprecated warnings.*

### DIFF
--- a/futures-util/src/future/future/map.rs
+++ b/futures-util/src/future/future/map.rs
@@ -6,7 +6,7 @@ use pin_project::pin_project;
 use crate::fns::FnOnce1;
 
 /// Internal Map future
-#[pin_project(Replace, project = MapProj, project_replace = MapProjOwn)]
+#[pin_project(project = MapProj, project_replace = MapProjOwn)]
 #[derive(Debug)]
 #[must_use = "futures do nothing unless you `.await` or poll them"]
 pub enum Map<Fut, F> {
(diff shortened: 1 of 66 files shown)


futures-util/src/future/future/map.rs
```diff
@@ -1,12 +1,12 @@
...
 use futures_core::future::{FusedFuture, Future};
 use futures_core::task::{Context, Poll};
-use pin_project::{pin_project, project};
+use pin_project::pin_project;
 
 use crate::fns::FnOnce1;
 
 /// Internal Map future
-#[pin_project(Replace)]
+#[pin_project(Replace, project = MapProj, project_replace = MapProjOwn)]
```
taiki-e, line 9: It is not ideal that we need to have both `Replace` and `project_replace = <name>` as `project_replace = <name>` implies that a `project_replace` method is generated. 
I'll fix it soon.

taiki-e: Fixed in pin-project 0.4.20

Nemo157:
Thanks! The number of PRs you've done for this deprecation is pretty epic 😁

PR #2177, opened 2020-06-11
Fix link in doc comment to `ready`

sollyucko:


futures-util/src/future/ready.rs
```diff
@@ -2,7 +2,7 @@ use core::pin::Pin;
 use futures_core::future::{FusedFuture, Future};
 use futures_core::task::{Context, Poll};
 
-/// Future for the [`ready`](ready()) function.
```
taiki-e, line 5: This is necessary to avoid the ambiguity between `future::ready` function and `ready!` macro.

```text
error: `ready` is both a function and a macro
 --> futures-util/src/future/ready.rs:5:21
  |
5 | /// Future for the [`ready`] function.
  |                     ^^^^^^^ ambiguous link
  |
  = note: `-D intra-doc-link-resolution-failure` implied by `-D warnings`
help: to link to the function, add parentheses
  |
5 | /// Future for the [`ready()`] function.
  |                     ^^^^^^^^^
help: to link to the macro, add an exclamation mark
  |
5 | /// Future for the [`ready!`] function.
  |                     ^^^^^^^^
```

https://travis-ci.com/github/rust-lang/futures-rs/jobs/347600517#L281-L312

sollyucko: Hmm, oops... For some reason, the link works properly for the docs.rs build of futures-util (https://docs.rs/futures-util/0.3.5/futures_util/future/struct.Ready.html links to https://docs.rs/futures-util/0.3.5/futures_util/future/fn.ready.html), but not for `futures` (https://docs.rs/futures/0.3.5/futures/future/struct.Ready.html links to https://docs.rs/futures/0.3.5/futures/future/ready()).

taiki-e: Hm, seems all of the links to free function are broken: https://docs.rs/futures/0.3.5/futures/future/index.html#structs

The workaround I know is to use `fn.ready.html`. (I'm not sure how to fix this with intra-doc-link. intra-doc-link still has many bugs...)

Nemo157: It would be very good to make a minimal repro of this and open an issue upstream. Testing locally it appears it's not fixed by https://github.com/rust-lang/rust/pull/73101.

Manishearth: I think this is because you're using ``[.`join`](join())``, and the markdown parser doesn't like that. Either use ``[`join()`]`` or use ``[`join`](function@join)``.

Manishearth: Nope, never mind

Manishearth: This is because of the funky reexport

https://github.com/rust-lang/rust/issues/73363

A quick fix would be to move that reexport to the bottom of the file, though since cross-crate stuff isn't working great anyway, might as well wait for https://github.com/rust-lang/rust/pull/73101

Manishearth: https://github.com/rust-lang/rust/pull/73365 should land very soon

taiki-e (commented):
I'm going to close this as the intra-doc link issues (rust-lang/rust#65983, https://github.com/rust-lang/rust/issues/73363) has been fixed and these links are now working correctly. Thanks anyway @sollyucko!

PR #2183, opened 2020-06-27
Fix some mpsc docs links

steven-joruk:
The `channel` function's link is broken here: https://docs.rs/futures/0.3.5/futures/channel/mpsc/struct.Sender.html

I've checked each link in `futures-channel/src/mpsc/mod.rs` and fixed as needed.


lights0123:
same in oneshot—this would be valid syntax if using `[]` with a nightly rustdoc (which docs.rs uses), but it isn't.

taiki-e (commented):
Thanks for the PR.
I'm going to close this as the intra-doc link issue (https://github.com/rust-lang/rust/issues/65983) has been fixed and these links are now working correctly.

Issue #1007, opened 2018-05-13
no_std support for a futures executor


dhoove:
This [comment ](https://github.com/rust-lang-nursery/futures-rs/issues/944#issuecomment-379144601) in #944 gave me some hope that it would soon be possible to use a single-threaded executor with no_std. I would like to be able to use async code for embedded applications. The general-purpose asynchronous CoAP library, [mentioned](https://github.com/rust-lang-nursery/futures-rs/issues/860#issue-304158699) in #860, also sounds very interesting for newer IoT devices. Wouldn't this also require a no_std executor?

TDHolmes:
This should be quite possible now that async fully works on embedded. [Here's a proof of concept](https://github.com/rust-embedded-community/async-on-embedded) from Ferrous Systems ([blog post](https://ferrous-systems.com/blog/async-on-embedded/) on it).

A number of things would have to change about the current single threaded executor to get this to work though. It uses `Arc/Rc` and spawns a thread, whereas for embedded reference counters aren't available in core and obviously you wouldn't spawn a thread for the executor.

Perhaps a simple, separate bare bones executor would be useful?

Nemo157:
The most basic possible executor requires 2 things that aren't available in `core`:

 1. Some form of shared ownership, this is `Arc` normally, but depending on other details could be something as simple as an `&'static _` (if you only want to support a single future in the lifetime of the application).
 2. Some way to do parking/unparking of the main thread, this is `Thread::{park, unpark}` normally, but some systems such as the Cortex-M series have dedicated assembly instructions that can be used for this (`WFE`/`SEV`).

Since these details are platform specific it _might_ be possible to have a generic executor that is capable of having concrete implementations plugged into it for them. But then the generic executor part becomes so trivial I don't see any reason to not just have platform specific executors.

TDHolmes:
> 1. Some form of shared ownership, this is `Arc` normally, but depending on other details could be something as simple as an `&'static _` (if you only want to support a single future in the lifetime of the application).

I think it's reasonable to require an allocator for a generic executor, so things like `Arc` and `Box` could be used in embedded ([alloc crate](https://doc.rust-lang.org/alloc/)).

> 2. Some way to do parking/unparking of the main thread, this is `Thread::{park, unpark}` normally, but some systems such as the Cortex-M series have dedicated assembly instructions that can be used for this (`WFE`/`SEV`).

[...]

PR #2185, opened 2020-06-29
Add a `no_std` compatible `LocalPool`

mkroening:
Related to #1007.

This is an effort to create a `LocalPool` which can be run on embedded devices without `std`.

While the proof of concept executor from Ferrous Systems relies on a private allocator ([rust-embedded-community/async-on-embedded/async-embedded/src/alloc.rs](https://github.com/rust-embedded-community/async-on-embedded/blob/17aa0f6ef1f1e3639acbce28093c6c30aacaf2c9/async-embedded/src/alloc.rs)), this approach requires a global allocator (like [alloc-cortex-m](https://github.com/rust-embedded/alloc-cortex-m)) and leverages existing abstractions like `FuturesUnordered`.

As we can't park/unpark the current thread without `std`, this approach is generic over the park/unpark functions. The user has to supply them, matching their environment, like this: `LocalPool::new(asm::wfe, asm::sev)`. Using empty functions for testing is possible as well.

This custom local pool is based on `LocalPool`. It was necessary to remove two occurences of `thread_local!`:

1. A waker to for the current thread was saved in thread local storage:
https://github.com/rust-lang/futures-rs/blob/19831dbecd13961fbd4243c1114663e24299c33d/futures-executor/src/local_pool.rs#L53-L58
This waker is now created by the pool, calling the provided function.

2. In `enter`:
https://github.com/rust-lang/futures-rs/blob/19831dbecd13961fbd4243c1114663e24299c33d/futures-executor/src/enter.rs#L4
This is used to avoid nested invoking of the executor. I did remove this and the two respective tests in my first attemt. Do you have an idea for a better solution? 

To view the changes from the `std`-based `LocalPool`, take a look at the second commit and following.
Note: I did not work on documentation and proper integration into `futures` jet. Also I am not sure about the naming of this executor.

Is this in scope of this project, or should I spin this off as another crate? What do you think?

### DIFF
--- a/futures-executor/Cargo.toml
+++ b/futures-executor/Cargo.toml
@@ -14,6 +14,7 @@ Executors for asynchronous tasks based on the futures-rs library.
 [features]
 default = ["std"]
 std = ["futures-core/std", "futures-task/std", "futures-util/std"]
+alloc = ["futures-core/alloc", "futures-task/alloc", "futures-util/alloc"]
 thread-pool = ["std", "num_cpus"]
 
 [dependencies]
--- a/futures-executor/src/custom_local_pool.rs
+++ b/futures-executor/src/custom_local_pool.rs
@@ -1,0 +1,402 @@
+use crate::enter;
+use futures_core::future::Future;
+use futures_core::stream::Stream;
+use futures_core::task::{Context, Poll};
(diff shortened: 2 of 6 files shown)


mkroening:
Closing this in favor of other, tailored `#[no_std]` executors.

PR #2186, opened 2020-07-06
Add type alias BoxFusedStream

mzabaluev:
Like `BoxStream`, this can be used for convenience.
Unlike `BoxStream`, the type is usable directly in `select!`.
Also add `LocalBoxFusedStream`.

### DIFF
--- a/futures-core/src/stream.rs
+++ b/futures-core/src/stream.rs
@@ -134,6 +134,15 @@ where
     }
 }
 
+/// An owned dynamically typed [`FusedStream`] for use in cases where you can't
+/// statically type your result or need to add some indirection.
+#[cfg(feature = "alloc")]
+pub type BoxFusedStream<'a, T> = Pin<alloc::boxed::Box<dyn FusedStream<Item = T> + Send + 'a>>;
+
+/// `BoxFusedStream`, but without the `Send` requirement.
+#[cfg(feature = "alloc")]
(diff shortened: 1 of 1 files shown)


taiki-e (commented):
Thanks for the PR. 
I would prefer not to expose more `(Local)Boxed*` type aliases. If we accept this PR, we will probably also need `(Local)BoxedFusedFuture` for consistency. And the same can happen if [new stream related traits](https://github.com/rust-lang/wg-async-foundations/blob/master/rfc-drafts/stream.md#other-traits) are added.

mzabaluev:
The case for these aliases is to facilitate direct use of boxed futures/streams in `select!`, using the fuse logic that is provided by the boxed future through dynamic dispatch.

I would not advocate proliferation of arbitrary boxed trait aliases for the sake of convenience alone.

taiki-e:
Sorry for the delayed reply.

I'm considering removing `Fused*` traits in the next major version (https://github.com/rust-lang/futures-rs/issues/2207) and it may be deprecated in the current major version at the earliest.
Also, it is difficult to implement `Fused*` traits correctly without specialization, so I personally do not recommend relying on those implementations.
So I recommended using `Fuse<BoxedFuture<..>>` or defining your own type alias, instead of improving convenience for features that will be deprecated.

I'm going to close this, but thanks @mzabaluev for the patch anyway.

PR #2181, opened 2020-06-19
Fix write-all-vectored feature

Thomasdezeeuw:
In Cargo.toml the feature is called "write-all-vectored" (with "-"), in
the code it used "write_all_vectored" (with "_"), this commit fixes the
code to use the correct name.

Also fixes the example.

PR #2187, opened 2020-07-06
Avoid writes without any data in write_all_vectored

tmiasko:
Previously, when non-empty sequence of empty IoSlices have been provided
to write_all_vectored, the buffers would be written as is with
write_vectored, and subsequently the return value of Ok(0) would be
misinterpreted as an error.

Avoid writes without any data by advancing the buffers first. This is
analogous to behaviour of write_all which makes no attempt to write if
there is no data.

Includes #2181.

### DIFF
--- a/futures-util/src/io/mod.rs
+++ b/futures-util/src/io/mod.rs
@@ -124,9 +124,9 @@ pub use self::write_vectored::WriteVectored;
 mod write_all;
 pub use self::write_all::WriteAll;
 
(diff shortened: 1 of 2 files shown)


Thomasdezeeuw:
Ping @taiki-e.

Thomasdezeeuw:
https://github.com/rust-lang/rust/pull/74088 is merged, can this be merged as well?

Thomasdezeeuw:
Since this has been reviewed, can it also be merged?

PR #2188, opened 2020-07-21
Remove mut from Sender::start_send()

garious:
I'm not expecting to see this PR merged - just looking to learn!

`std::sync::mpsc::Sender` and `async_channel::Sender` both have `send` methods that don't require `mut`, but `futures_channel::mpsc::Sender` does. Why exactly? I see the comment on `maybe_parked` says "this is an optimization to avoid having to lock the mutex most of the time". Making a "multi-producer" method `mut` to implement an optimization seems like a steep price. Would `AtomicBool` suffice?


garious:
Anyone able to comment here? Pretty nice:  let (~~mut~~ tx, rx) = mpsc::channel(0);

garious:
@taiki-e or @Nemo157 perhaps?

PR #2190, opened 2020-07-22
Mention FuturesOrdered as alternative to join_all

SuperFluffy:
The docs of `futures::future::join_all` say that results are collected “into a destination Vec<T> in the same order as they were provided”, but then go on to suggest `FuturesUnordered` as an alternative to it. `FuturesOrdered` should be mentioned as a proper replacement for keeping the order, while suggesting `FuturesUnordered` as an additional option.


taiki-e:
Thanks!

PR #2195, opened 2020-08-09
extend example for join!/try_join!

### LINKED ISSUE #2120: Expand futures::join example to show that it is variadic
Currently the example only shows a 2-tuple being joined, should make it more obvious that the macro is variadic.

JOE1994:
Hello :crab: , this PR makes the following change.
* Update examples for `join!`/`try_join!` to make explicit that these macros are variadic.

closes #2120 

Thank you for reviewing this PR :+1:

### DIFF
--- a/futures-util/src/async_await/join_mod.rs
+++ b/futures-util/src/async_await/join_mod.rs
@@ -22,8 +22,13 @@ macro_rules! document_join_macro {
         ///
         /// let a = async { 1 };
         /// let b = async { 2 };
-        ///
+        /// let c = async { 3 };
+        /// let d = async { 4 };
+        /// let e = async { 5 };
+        /// 
+        /// // `join!` is variadic
         /// assert_eq!(join!(a, b), (1, 2));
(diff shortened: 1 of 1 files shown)


JOE1994:
The initial commit failed in CI due to an issue external to this PR

taiki-e (commented):
I think it is easier to read if the definitions of variable a,b and variable c,d,e are separated.

Also, CI failure will be fixed in #2203

taiki-e:
Also, it may be good to add a comment explaining "variadic" after the comment "`join!`/`try_join!` is variadic". 
e.g., "`join!` is variadic, so you can pass any number of futures."

JOE1994:
> Also, it may be good to add a comment explaining "variadic" after the comment "`join!`/`try_join!` is variadic".
> e.g., "`join!` is variadic, so you can pass any number of futures."

Thank you for your feedback! I made commits to apply the suggested changes :smiley_cat:

taiki-e (approved):

taiki-e:
Thanks!

PR #2198, opened 2020-08-11
minor fixes to docs: 'futures_channel::oneshot'

### LINKED ISSUE #1487: The term "one-shot" is not defined in the documentation.
> A one-shot, futures-aware channel

> Creates a new futures-aware, one-shot channel.

It's unclear to me if this means single consumer single producer or if this means you can only send one message on the channel and not two or whether only one message can exist inside the channel at a time.

Seems like this would be a very useful piece of information to add to [the documentation](https://docs.rs/futures/0.1.25/futures/sync/oneshot/index.html).

JOE1994:
Hello :crab: , this PR makes the following changes to the docs for module `futures_channel::oneshot`.
* fix broken links to docs for `fn channel()`
* clarify that `oneshot` is spsc (single-producer, single-consumer)
* one grammar fix in docs for `Sender::poll_canceled`
* small change in docs for `Sender::send` to make it more concise

#### closes #1487 
  * The latest docs explains that `oneshot` is for sending a value once, so I only added clarification that `oneshot` is a spsc channel.

Thank you for reviewing this PR :cat:

### DIFF
--- a/futures-channel/src/oneshot.rs
+++ b/futures-channel/src/oneshot.rs
@@ -70,7 +70,7 @@ struct Inner<T> {
     tx_task: Lock<Option<Waker>>,
 }
 
-/// Creates a new one-shot channel for sending values across asynchronous tasks.
+/// Creates a new one-shot channel for sending a single value across asynchronous tasks.
 ///
 /// The channel works for a spsc (single-producer, single-consumer) scheme.
 ///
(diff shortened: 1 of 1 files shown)


futures-channel/src/oneshot.rs
```diff
@@ -70,6 +72,8 @@ struct Inner<T> {
 
 /// Creates a new one-shot channel for sending values across asynchronous tasks.
```
taiki-e, line 73:
```suggestion
/// Creates a new one-shot channel for sending a single value across asynchronous tasks.
```

PR #2205, opened 2020-08-31
Implement `FusedStream` for `FuturesOrdered`

### LINKED ISSUE #2192: Why is there no `FusedStream` implementation for `FuturesOrdered`?
While `FuturesUnordered` has a [`FusedStream` implementation](https://docs.rs/futures/0.3.5/futures/stream/struct.FuturesUnordered.html#impl-FusedStream), the `FuturesOrdered` type does not. Is there any reason for this missing implementation or is this just an oversight?

phil-opp:
This uses the same approach as the `FusedStream` implementation for `FuturesUnordered`.

Fixes #2192 

cc @taiki-e

### DIFF
--- a/futures-util/src/stream/futures_ordered.rs
+++ b/futures-util/src/stream/futures_ordered.rs
@@ -1,7 +1,7 @@
 use crate::stream::{FuturesUnordered, StreamExt};
 use futures_core::future::Future;
 use futures_core::stream::Stream;
-use futures_core::task::{Context, Poll};
+use futures_core::{FusedStream, task::{Context, Poll}};
 use pin_project::pin_project;
 use core::cmp::Ordering;
 use core::fmt::{self, Debug};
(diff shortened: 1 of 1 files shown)


taiki-e:
`FuturesUnordered` implements `FusedStream` with the same logic that this PR introduces, and `is_terminated` field is `true` only when `in_progress_queue.poll_next_unpin()` returns `None`, so I think `self.in_progress_queue.is_terminated() && self.queued_outputs.is_empty()` is enough for this.

phil-opp:
Thanks for the suggestion, I pushed a new commit with it.

taiki-e (approved):

phil-opp:
The clippy failure in src/mpsc/mod.rs seems unrelated to this PR.

taiki-e:
Yes, that will be fixed in #2203

taiki-e:
Thanks!

PR #2210, opened 2020-09-07
Make Flatten doc point to flatten method

brianhv:
The [doc for `Flatten`](https://docs.rs/futures/0.3.5/futures/stream/struct.Flatten.html) claims to be for the `inspect` method. I believe this is probably a copy/paste oversight. This change replaces the reference to `inspect` with one to `flatten`.


taiki-e (commented):
Thanks for the PR.
I'm going to close this as there is already a PR that solves this: #2162

PR #2212, opened 2020-09-14
Implement try_take_while.

bittrance:
Implementation of `try_take_while`, symmetric to `try_skip_while`. I found that I needed it because it works as a variant of `try_for_each` which allows termination of the stream, effectively working as an `break`-able fallible iteration over the stream.


taiki-e (approved):

PR #2216, opened 2020-09-22
futures: Require all stable features in tests

taiki-e:
In projects that have many feature flags and tests, it is hard to maintain the support of the feature flag combination in tests (due to complex cfgs, long test time, etc.), so it is probably better to fail tests if all stable features are not activated.

Related: #2101
Closes: #2098

cc @cramertj @Nemo157 @kentfredric

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -38,10 +38,16 @@ matrix:
     # When updating this, the reminder to update the minimum required version of `async-await` feature in README.md.
     - name: cargo build --features async-await (minimum required version)
       rust: 1.39.0
+      install:
+        # cargo does not support for --features/--no-default-features with workspace, so use cargo-hack instead.
+        # Refs: cargo#3620, cargo#4106, cargo#4463, cargo#4753, cargo#5015, cargo#5364, cargo#6195
+        - if ! cargo hack -V 2>/dev/null; then
+            cargo install cargo-hack;
+          fi
       script:
-        - cargo run --manifest-path ci/remove-dev-dependencies/Cargo.toml */Cargo.toml
         # async-await feature is activated by default.
-        - cargo build --workspace
+        - cargo hack build --workspace --no-dev-deps
+        - cargo build --tests --features default,thread-pool,io-compat --manifest-path futures/Cargo.toml
 
     - name: cargo +stable build
       rust: stable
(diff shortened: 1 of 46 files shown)


kentfredric:
A workaround for "the tests take a long time" might be to set up a "cron" driven CI test,  so they don't demand feature combination with PRs and soforth, but a reminder to keep on top of it still happens.

taiki-e:
Even if we don't do it in PR, someone has to do it to fix if cron fails...

kentfredric:
Yeah,  just the downside really is without testing feature combos, you have no evidence features can be arbitrarily combined in a dependent.

So whether or not you have tests that break,  you can still have code that breaks,  no?

seanmonstar (approved):

taiki-e:
Our CI already does [feature flag check](https://github.com/rust-lang/futures-rs/blob/1d93f5d3696fe545cdd9cc0e59a57b658cbf009a/.travis.yml#L129-L143), and, IIRC, all the current `futures`'s stable feature flags are additive (not only APIs but also implementations). So this PR shouldn't have a negative impact on the actually published crate.

kentfredric:
> Our CI already does [feature flag check](https://github.com/rust-lang/futures-rs/blob/1d93f5d3696fe545cdd9cc0e59a57b658cbf009a/.travis.yml#L129-L143), and, IIRC, all the current `futures`'s stable feature flags are additive (not only APIs but also implementations). So this PR shouldn't have a negative impact on the actually published crate.

That looks adequate for ensuring everything compiles, sure, but doesn't give many assurances of expected behaviour.

Just restricting this this way makes it harder to test things downstream in a comprehensive way. ( which is a relatively important thing to do, when you're a linux vendor )

Nemo157:
Part of Rusts feature design is that _features do not change behaviour_ they only expand the available API. As far as I can tell `futures` strictly follows this design principle, so that testing with all features active tests the exact same code paths that would be tested with some features inactive.

kentfredric:
Fair enough, and I hope you're right :). I've just tended to preferred to be able to prove things via evidence, as opposed to constructing a conjecture that it *should* be ok, for I find that, however reasonable we are when making such a conjecture, humans always possess the risk of missing something, which only an evidential test will expose. :)

PR #2217, opened 2020-09-22
Migrate CI to GitHub Actions

taiki-e:


seanmonstar (approved):

.github/workflows/ci.yml
```diff
@@ -0,0 +1,225 @@
...
+          - nightly
+        os:
+         - ubuntu-latest
+         - macos-latest
+         - windows-latest
+    runs-on: ${{ matrix.os }}
+    steps:
+      - uses: actions/checkout@v2
+      - name: Install Rust
+        run: rustup update ${{ matrix.rust }} && rustup default ${{ matrix.rust }}
```
taiki-e, line 33: Seems `rustup update ..` is does not actually need, as the recent rustup's `rustup default` automatically installs the toolchain if the toolchain is unavailable.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,225 @@
...
+
+  thumbv6m:
+    name: cargo build --target thumbv6m-none-eabi
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Install Rust
+        run: rustup update nightly && rustup default nightly
+      - run: rustup target add thumbv6m-none-eabi
+      - run: cargo run --manifest-path ci/remove-dev-dependencies/Cargo.toml */Cargo.toml
```
taiki-e, line 134: This is equivalent to `cargo hack --remove-dev-deps`. It may be okay to remove this script given that this script [takes as long to compile](https://github.com/rust-lang/futures-rs/runs/1287718190?check_suite_focus=true
) as [installing cargo-hack](https://github.com/rust-lang/futures-rs/runs/1287717540?check_suite_focus=true).

FYI @Nemo157

taiki-e:
It needs to change the "required statuses" of the branch protection settings to merge this (requires admin rights).

@cramertj: Could you change the branch protection settings? (or could you add me on admins of this repo?)

.github/workflows/ci.yml
taiki-e: Oh, no, this means if the default toolchain is stable and that is outdated, then the toolchain does not update. I'll revert...

taiki-e:
I'm going to merge this. Branch protection will be bypassed by leaving an empty Travis config. That's not a very good way, but I don't want to continue using Travis as it is very slow and not stable. (IIRC, it sometimes takes about an hour after the PR is opened before the build starts even if the queue is empty.)

PR #2221, opened 2020-09-30
0.1: Allow deprecated warnings

taiki-e:
They are warnings about deprecated items or syntax, but alternatives are not available in 0.1's msrv.

### DIFF
--- a/src/future/shared.rs
+++ b/src/future/shared.rs
@@ -288,6 +288,7 @@ impl<E> fmt::Display for SharedError<E>
 impl<E> error::Error for SharedError<E>
     where E: error::Error,
 {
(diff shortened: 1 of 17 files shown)


src/future/shared.rs
```diff
@@ -288,6 +288,7 @@ impl<E> fmt::Display for SharedError<E>
 impl<E> error::Error for SharedError<E>
     where E: error::Error,
 {
+    #[allow(deprecated)]
     fn description(&self) -> &str {
```
taiki-e, line 292: `Error::description` is a required method in rust 1.15: https://doc.rust-lang.org/1.15.0/std/error/trait.Error.html#tymethod.description

PR #2224, opened 2020-10-02
stream::ForEach can no longer block indefinitely

Lucretiel:
This PR removes the `loop` from `stream::ForEach`. Instead, the future
makes at most 1 attempt to poll its interior, and if it can make more
progress, it immediately calls `wake_by_ref()`, then returns Pending.
In this way, under a fair executor, it can't accidentally starve the
task queue.

### DIFF
--- a/futures-util/src/stream/stream/for_each.rs
+++ b/futures-util/src/stream/stream/for_each.rs
@@ -30,9 +30,10 @@ where
 }
 
 impl<St, Fut, F> ForEach<St, Fut, F>
-where St: Stream,
-      F: FnMut(St::Item) -> Fut,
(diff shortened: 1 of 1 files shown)


cramertj:
I'm skeptical of this approach-- see my the related discussion between myself, @jonhoo , and @carllerche in this thread: https://github.com/rust-lang/futures-rs/pull/2049#issuecomment-577780571

I also anticipate that this PR would have significant negative performance implications if we adopted this approach throughout the ecosystem: yielding back up to the top-level executor with every step seems like it would interact poorly with locality, and could be very inefficient depending on the top-level executor's queuing implementation.

PR #2228, opened 2020-10-13
Update pin-project to 1

taiki-e:


seanmonstar (approved):

taiki-e:
This breaks undocumented MSRV (1.36) of futures/futures-util/futures-executor/futures-test with `default-features = false`.
However, as said in https://github.com/taiki-e/pin-project/pull/292, this should probably be fine as that is undocumented and there is a possibility of switching to pin-project-lite that requires 1.37. (I'm working on enum support for pin-project-lite (https://github.com/taiki-e/pin-project-lite/pull/28), and once that is done, we can probably switch.)

Issue #826, opened 2018-03-02
Consider switchin stream::once to take T: IntoFuture


carllerche:
AFAIK, there is no way to convert a future to a stream of one value right now. `stream::once` takes a Result. It probably wouldn't be hard to update that to take `T: IntoFuture`.

cramertj:
Are you looking for [this method](https://rust-lang-nursery.github.io/futures-rs/futures/prelude/trait.FutureExt.html#method.into_stream)?

carllerche:
I guess so, it wasn't clear where to look for it. My first thought was `stream::foo(...)`

PR #2230, opened 2020-10-15
Remove outdated comment

travisbrown:
Small doc fix: this is no longer accurate since `once` was changed to take a future instead of an item in #885 / #826.


taiki-e (approved):

taiki-e:
Thanks!

PR #2233, opened 2020-10-15
Clean up private modules

taiki-e:
* Merge into one module per crate.
* Rename to __private. To prevent [someone from accidentally using it](https://github.com/rust-lang-nursery/failure/issues/342)...
* Shortened path to some private items.
* Use '=' requirement to futures-macro to avoid breakage due to private API changes. (This should be done in #2124, but I missed that.)

### DIFF
--- a/futures-util/src/async_await/mod.rs
+++ b/futures-util/src/async_await/mod.rs
@@ -3,37 +3,37 @@
 //! This module contains a number of functions and combinators for working
 //! with `async`/`await` code.
 
-use futures_core::future::Future;
-use futures_core::stream::Stream;
-
-#[doc(hidden)]
-pub use futures_core::future::FusedFuture;
-#[doc(hidden)]
-pub use futures_core::stream::FusedStream;
+use futures_core::future::{Future, FusedFuture};
+use futures_core::stream::{Stream, FusedStream};
 
 #[macro_use]
 mod poll;
+#[allow(unreachable_pub)] // https://github.com/rust-lang/rust/issues/64762
 pub use self::poll::*;
 
 #[macro_use]
 mod pending;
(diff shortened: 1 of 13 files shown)


futures-util/src/async_await/mod.rs
```diff
@@ -3,37 +3,37 @@
...
-#[doc(hidden)]
-pub use futures_core::future::FusedFuture;
-#[doc(hidden)]
-pub use futures_core::stream::FusedStream;
+use futures_core::future::{Future, FusedFuture};
+use futures_core::stream::{Stream, FusedStream};
 
 #[macro_use]
 mod poll;
+#[allow(unreachable_pub)] // https://github.com/rust-lang/rust/issues/57411
```
Nemo157, line 11: I don't think this is that issue, that seems to specifically be a bad interaction between a fix-suggestion and having a deny-lint active (deleting the `pub use` in the example there results in a working program).

It seems since we don't make `futures-util::async_await` a `pub mod` and instead `pub use async_await::*;` these _can_ all be `pub(crate)` and still work.

Nemo157: Oh, no, I didn't realise re-exports could further constrain the visibility of the re-exported item, rather than just having a visibility applied to the re-export itself. I think this is more likely https://github.com/rust-lang/rust/issues/64762

Nemo157 (approved):

futures-util/src/async_await/mod.rs
taiki-e: Thanks, it certainly seems more likely to be https://github.com/rust-lang/rust/issues/64762. 
(Actually, the code mentioned in [this comment](https://github.com/rust-lang/rust/issues/64762#issuecomment-567555428) looks almost the same as in this PR's case, except for the module itself warned.)

PR #2236, opened 2020-10-17
Don't double panic in futures-test::test::panic_waker::wake_panic

asomers:
This function can get called during Drop, so it's important not to panic
if the thread is already panicking. Doing so makes debugging panics much
harder.


taiki-e (approved):
Thanks!

PR #2244, opened 2020-10-30
Add Shared::ready method to create a shared future that is immediately ready.

afranchuk:
Importantly, this affects the behavior of peek _without_ requiring an async context. Otherwise within an async context one can simply clone and await a shared future to achieve the same effect (though perhaps less efficiently).

### DIFF
--- a/futures-util/src/future/future/shared.rs
+++ b/futures-util/src/future/future/shared.rs
@@ -95,6 +95,22 @@ impl<Fut: Future> Shared<Fut> {
             waker_key: NULL_WAKER_KEY,
         }
     }
+
+    /// Creates a [`Shared`] that is immediately ready with a value.
+    pub fn ready(output: Fut::Output) -> Shared<Fut> {
+        let inner = Inner {
+            future_or_output: UnsafeCell::new(FutureOrOutput::Output(output)),
+            notifier: Arc::new(Notifier {
+                state: AtomicUsize::new(COMPLETE),
+                wakers: Mutex::new(None)
+            })
+        };
+
+        Shared {
+            inner: Some(Arc::new(inner)),
+            waker_key: NULL_WAKER_KEY
+        }
+    }
 }
 
 impl<Fut> Shared<Fut>
--- a/futures/tests/shared.rs
+++ b/futures/tests/shared.rs
@@ -214,6 +214,20 @@ fn dont_do_unnecessary_clones_on_output() {
 }
 
 #[test]
+fn ready() {
+    use futures::executor::block_on;
+    use futures::future::{FutureExt, Shared};
+
+    let mut futs = Vec::new();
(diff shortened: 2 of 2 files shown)


taiki-e (commented):
I'm not sure why you want to add this as part of `Shared`. [`future::ready` implements `Clone`](https://docs.rs/futures/0.3.7/futures/future/struct.Ready.html#impl-Clone). I guess what you really need is a method that returns a reference to the value of `future::Ready`.

afranchuk:
Well I kind of mentioned it; this differs in allowing you to immediately peek the value. If you are storing Shared to be reused (much along the same lines as the recent commit that adds the Weak Shared references), this makes a difference in allowing you to peek without having ever awaited the value in an async context. I know it seems fairly niche, but I need it at the moment :) however I suppose I _could_ change my store to simply be some trait object that is Future, Clone, and has peek as well. But I don't really see a reason not to add this method.

taiki-e:
> If you are storing Shared to be reused (much along the same lines as the recent commit that adds the Weak Shared references), this makes a difference in allowing you to peek without having ever awaited the value in an async context.

I'm not sure how the reuse of `Shared` is related to `WeakShared`.
(`WeakShared` is a weak ptr to the `Shared` that downgraded. Replacing the original `Shared` with the new `Shared` does not update the value of cloned `Shared` and downgraded `WeakShared`.)

If you need the ability to keep updating shared values, [`tokio::sync::watch`](https://docs.rs/tokio/0.3.4/tokio/sync/watch/index.html) is probably the right solution.

afranchuk:
Sorry, what I meant was that I saw the commit for `WeakShared` mentions the use case of having a cache of stored `Shared`, and being able to create `ready` `Shared` instances in a cache is also useful (as that is what I need it for). So only very loosely related to `WeakShared`, more related to their example use case.

afranchuk:
But if this is a controversial addition, I was able to make a separate abstraction that achieved what was desired. I do think there's no harm in allowing people to directly create `Shared` in each of its possible states.

taiki-e:
@afranchuk Thanks for clarifying.
My main concern is that `Shared` is used as a future of shared value, not a future of shared future. I think `Shared` is not efficient enough for that purpose, and in most cases, [custom futures based on `ready`](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=34464cd875ff29cf5bc9709bb5cdd227) are more efficient and `tokio::sync::watch` is more flexible.
It's true that there are cases where it's easier to write with this feature like a test that this PR added, but I don't think they're more efficient than combining `MaybeDone` or `Either` with `Shared`.

taiki-e:
(If this has enough documentation on performance (like mentioned in the previous comment), we may accept this, but we recommend that you avoid using it if possible.)

PR #2245, opened 2020-10-31
Add TryStreamExt::try_buffered

khuey:
A copy of TryBufferUnordered, with the underlying FuturesUnordered swapped out for a FuturesOrdered.


taiki-e (changes requested):
Thanks for the PR. Please add doc examples or tests.

khuey:
Alright, how about that?

taiki-e (approved):
LGTM, thanks!

Issue #2249, opened 2020-10-31
try_filter_map causes stream to panic after an error is returned


Kouzukii:
When try_filter_map() is used on a Stream of Results and an Error is returned, the stream becomes unusable.
Calling .next() on it afterwards results in a panic.

Related discussion on the rust forums: https://users.rust-lang.org/t/take-first-non-error-result-out-of-a-stream-of-results-in-rust/50897

Playground to reproduce the bug: https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=82e01c4891d60df13a12528451d12225

taiki-e:
Good catch! Seems there are the same bugs in `try_skip_while` and `try_take_while`.

taiki-e:
Filed #2250 to fix this.

PR #2254, opened 2020-11-04
feat: Add sink::from_fn

Marwes:
Felt the need for this a couple times to mock a sink or inject something
that is sink-like, but doesn't implement sink itself.

### DIFF
--- a/futures-util/src/sink/from_fn.rs
+++ b/futures-util/src/sink/from_fn.rs
@@ -59,11 +59,13 @@ where
 
     fn poll_flush(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
         let mut this = self.project();
-        if let Some(future) = this.future.as_mut().as_pin_mut() {
-            ready!(future.poll(cx))?;
-        }
+        let result = if let Some(future) = this.future.as_mut().as_pin_mut() {
+            ready!(future.poll(cx))
+        } else {
+            Ok(())
+        };
         this.future.set(None);
-        Poll::Ready(Ok(()))
+        Poll::Ready(result)
     }
 
     fn poll_close(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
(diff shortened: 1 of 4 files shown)


Nemo157:
See previously https://github.com/rust-lang/futures-rs/pull/1493, actually capturing state inside the closure is painful.

Comparing the implementations I see two bugs here:

* `poll_ready` needs to flush, otherwise it will never return `Ready`.
* `poll_flush` needs to clear the future, otherwise flushing again after it's flushed will panic.

Marwes:
> See previously #1493, actually capturing state inside the closure is painful.

A mutex does the trick when that is necessary :shrug: . There are also cases where there isn't a need to capture state (say pushing to an external database.

futures-util/src/sink/from_fn.rs
```diff
@@ -0,0 +1,72 @@
...
+        this.future.set(Some(future));
+        Ok(())
+    }
+
+    fn poll_flush(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
+        let mut this = self.project();
+        if let Some(future) = this.future.as_mut().as_pin_mut() {
+            ready!(future.poll(cx))?;
+        }
+        this.future.set(None);
```
taiki-e, line 65: If the error is recoverable and the user calls `poll_{flush, ready, close}` again, this will result in a panic. (related: #2249)

```suggestion
            let res = ready!(future.poll(cx));
            this.future.set(None);
            res?;
        }
```

taiki-e:
I have no objection to adding a way to convert `impl Fn* -> impl Future` to `Sink`, as sink doesn't currently have an API to do it.

However, I'm not clear if this implementation is preferred. We tried to add something similar to stream (https://github.com/rust-lang/futures-rs/issues/1609 https://github.com/rust-lang/futures-rs/pull/1842), but we decided that it would be preferable to provide only [`unfold`](https://docs.rs/futures/0.3.8/futures/stream/fn.unfold.html) due to the limitations of async closure.

So, I'm starting to feel that adding an API like `unfold` that passes the state separately from the closure is better. @Marwes @Nemo157 Any thought?

Marwes:
https://github.com/rust-lang/futures-rs/pull/2268

taiki-e:
Closing this in favor of #2268.

Issue #2253, opened 2020-11-03
Not building on 1.47 anymore


spacekookie:
For some reason this has stopped building.

```
error: `proc-macro` crate types currently cannot export any items other than functions tagged with `#[proc_macro]`, `#[proc_macro_derive]`, or `#[proc_macro_attribute]`
  --> /home/.cargo/registry/src/github.com-1ecc6299db9ec823/futures-macro-0.3.7/src/lib.rs:31:1
   |
31 | pub fn join_internal(input: TokenStream) -> TokenStream {
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

error: `proc-macro` crate types currently cannot export any items other than functions tagged with `#[proc_macro]`, `#[proc_macro_derive]`, or `#[proc_macro_attribute]`
  --> /home/.cargo/registry/src/github.com-1ecc6299db9ec823/futures-macro-0.3.7/src/lib.rs:37:1
   |
37 | pub fn try_join_internal(input: TokenStream) -> TokenStream {
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

error: `proc-macro` crate types currently cannot export any items other than functions tagged with `#[proc_macro]`, `#[proc_macro_derive]`, or `#[proc_macro_attribute]`
  --> /home/.cargo/registry/src/github.com-1ecc6299db9ec823/futures-macro-0.3.7/src/lib.rs:43:1
   |
43 | pub fn select_internal(input: TokenStream) -> TokenStream {
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

error: `proc-macro` crate types currently cannot export any items other than functions tagged with `#[proc_macro]`, `#[proc_macro_derive]`, or `#[proc_macro_attribute]`
  --> /home/.cargo/registry/src/github.com-1ecc6299db9ec823/futures-macro-0.3.7/src/lib.rs:49:1
   |
49 | pub fn select_biased_internal(input: TokenStream) -> TokenStream {
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

```

taiki-e:
[...]

PR #2255, opened 2020-11-04
Release 0.3.8

taiki-e:
#2243 fixed some proc-macro-related issues (#2026, #2097, rust-lang/rust#78234). I don't think it fixes proc-macro issues on NixOS (https://github.com/rust-lang/futures-rs/issues/2242#issuecomment-718102805 and probably #2253), but a release that includes this should be useful.

Changes:
* Switch proc-macros to use native `#[proc_macro]` at Rust 1.45+ (#2243)
* Add `WeakShared` (#2169)
* Add `TryStreamExt::try_buffered` (#2245)
* Add `StreamExt::cycle` (#2252)
* Implemented `Clone` for `stream::{Empty, Pending, Repeat, Iter}` (#2248, #2252)
* Fix panic in some `TryStreamExt` combinators (#2250)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,11 @@
+# 0.3.8 - 2020-11-04
+* Switch proc-macros to use native `#[proc_macro]` at Rust 1.45+ (#2243)
+* Add `WeakShared` (#2169)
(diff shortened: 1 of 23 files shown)


cramertj:
LGTM!

Issue #2234, opened 2020-10-16
Missing StreamExt::unzip


LLFourn:
See [`core::iterator::unzip`]

[`core::iterator::unzip`]: https://doc.rust-lang.org/core/iter/trait.Iterator.html#method.unzip

binier:
Are you looking for this?
https://github.com/rust-lang/futures-rs/blob/7340d3d5d6fe8082a73069582b048ebaef6626b1/futures-util/src/stream/stream/zip.rs#L56-L62

Or are you talking about a single stream which returns a tuple and you want unzip that into two streams?

LLFourn:
>Or are you talking about a single stream which returns a tuple and you want unzip that into two streams?

Yes. This is what I wanted.

PR #2256, opened 2020-11-05
feat(futures-util/stream): implement stream.unzip adapter

binier:
The way this works is by putting the **base stream** in the `Arc` and sharing that `Arc` in `UnzipLeft` and `UnzipRight` streams.

Initially only `UnzipLeft` is polling the base stream, emits first tuple item as it's result and sends the other one through `mpsc::channel` to the `UnzipRight`. At the end of the stream, `UnzipLeft` sends `Poll::Ready(None)` through the channel to left the `UnzipRight` know stream is finished.

- In case `UnzipLeft` is dropped, first it drops `mpsc::Sender` and then wakes up the `UnzipRight` task, which is a way of telling to `UnzipRight` to take over polling the **base stream** (Since `try_recv` called by right stream will error with `TryRecvError::Disconnected` if sender is dropped).
- In case `UnzipRight` is dropped, `right_waker.upgrade()` will fail and return `None` since `right_waker` is a `Weak` type of `Arc`. If that happens, send simply won't happen.
- If both is dropped, **base stream** will drop too.

There is **unsafe** code when converting **base stream** from `Pin<&mut Arc<Stream>>` to `Pin<&mut Stream>` in order to poll it, but since we are accessing **base stream** only from one of the streams at a time, it's safe right?

closes: #2234

### DIFF
--- a/futures-util/src/stream/stream/unzip.rs
+++ b/futures-util/src/stream/stream/unzip.rs
@@ -1,0 +1,171 @@
+use crate::task::AtomicWaker;
+use alloc::sync::{Arc, Weak};
+use core::pin::Pin;
+use futures_core::stream::{FusedStream, Stream};
+use futures_core::task::{Context, Poll};
+use pin_project::{pin_project, pinned_drop};
+use std::sync::mpsc;
+
+/// SAFETY: safe because only one of two unzipped streams is guaranteed
+/// to be accessing underlying stream. This is guaranteed by mpsc. Right
+/// stream will access underlying stream only if Sender (or left stream)
+/// is dropped in which case try_recv returns disconnected error.
+unsafe fn poll_unzipped<S, T1, T2>(
+    stream: Pin<&mut Arc<S>>,
+    cx: &mut Context<'_>,
+) -> Poll<Option<S::Item>>
+where
+    S: Stream<Item = (T1, T2)>,
+{
+    stream
+        .map_unchecked_mut(|x| &mut *(Arc::as_ptr(x) as *mut S))
+        .poll_next(cx)
+}
+
+#[pin_project(PinnedDrop)]
+#[derive(Debug)]
+#[must_use = "streams do nothing unless polled"]
+pub struct UnzipLeft<S, T1, T2>
+where
+    S: Stream<Item = (T1, T2)>,
+{
+    #[pin]
+    stream: Arc<S>,
+    right_waker: Weak<AtomicWaker>,
+    right_queue: mpsc::Sender<Option<T2>>,
+}
+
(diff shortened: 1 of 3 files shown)


binier:
I want to add some tests though, but not sure where to add them, since tests for example for adaptor `zip` is in disabled tests...

futures-util/src/stream/stream/unzip.rs
```diff
@@ -0,0 +1,171 @@
...
+use core::pin::Pin;
+use futures_core::stream::{FusedStream, Stream};
+use futures_core::task::{Context, Poll};
+use pin_project::{pin_project, pinned_drop};
+use std::sync::mpsc;
+
+/// SAFETY: safe because only one of two unzipped streams is guaranteed
+/// to be accessing underlying stream. This is guaranteed by mpsc. Right
+/// stream will access underlying stream only if Sender (or left stream)
+/// is dropped in which case try_recv returns disconnected error.
```
taiki-e, line 12: I don't think this is correct. You can call `UnzipLeft(Right)::poll_next` and `UnzipRight(Left)::is_terminated` at the same time.

taiki-e:
Given that `Iterator::unzip` is like `.collect::<(Vec<_>, Vec<_>)>`, I would prefer to separate functionality like Shared Stream from Unzip.

> I want to add some tests though, but not sure where to add them, since tests for example for adaptor `zip` is in disabled tests...

I recommend adding to `futures/tests/stream.rs` or `futures/tests/stream_unzip.rs` (new file).

futures-util/src/stream/stream/unzip.rs
binier: Oh right! But I can simply maintain `is_done` boolean in `UnzipRight` and set it to `true` if I receive `Poll::Ready(None)` right?

**Or** is there a case when: `is_terminated()` returns true for underline stream but it still hasn't emitted `Poll::Ready(None)` and won't ever emit it?

If thats the case, would `UnzipLeft` be dropped if it `is_terminated`? So would it be sufficient if `is_terminated()` returned:
- `is_done` if left is **not** dropped.
- `stream.is_terminated()` if left **is** dropped, since it's now safe to call that method.

binier:
> Given that `Iterator::unzip` is like `.collect::<(Vec<_>, Vec<_>)>`, I would prefer to separate functionality like Shared Stream from Unzip.

Hm... If I understand this right, you mean that unzip should not separate streams, but I didn't quite get the alternative. What would the api look like for the unzip that you are talking about?

> I recommend adding to `futures/tests/stream.rs` or `futures/tests/stream_unzip.rs` (new file).

Thanks! Will add tests there.

PR #2262, opened 2020-11-13
Stream select macro

Xaeroxe:
Adds a new procedural macro to the futures crate that I've found some use for. `stream_select!` works similarly to the `select_all` function, but doesn't require the streams to all be the same type. Nor does it use a `Box<dyn Stream>` to accomplish that. ~~It's implemented in terms of the `select!` macro~~ , but represents a type rather than multiple branches of execution. In my use case I would feed the stream from `stream_select!` into a `select!` statement, as many of the streams I want to `select!` over should be treated identically in the `select!` statement.

Reviewers may ask "Why not just use the `select` function repeatedly?" And the answer to that is that recursive use of the `select` function results in an unfair stream, with priority being given to streams on the upper layers.

### DIFF
--- a/futures-macro/src/stream_select.rs
+++ b/futures-macro/src/stream_select.rs
@@ -4,8 +4,7 @@ use syn::{parse::Parser, punctuated::Punctuated, Expr, Index, Token};
 
 /// The `stream_select!` macro.
 pub(crate) fn stream_select(input: TokenStream) -> Result<TokenStream, syn::Error> {
-    let args = Punctuated::<Expr, Token![,]>::parse_terminated
-        .parse2(input)?;
+    let args = Punctuated::<Expr, Token![,]>::parse_terminated.parse2(input)?;
     let generic_idents = (0..args.len()).map(|i| format_ident!("_{}", i)).collect::<Vec<_>>();
     let field_idents = (0..args.len()).map(|i| format_ident!("__{}", i)).collect::<Vec<_>>();
     let field_idents_2 = (0..args.len()).map(|i| format_ident!("___{}", i)).collect::<Vec<_>>();
--- a/futures-macro/src/stream_select.rs
+++ b/futures-macro/src/stream_select.rs
@@ -1,19 +1,18 @@
-use proc_macro::TokenStream;
+use proc_macro2::TokenStream;
 use quote::{format_ident, quote, ToTokens};
 use syn::{parse::Parser, punctuated::Punctuated, Expr, Index, Token};
 
 /// The `stream_select!` macro.
-pub(crate) fn stream_select(input: TokenStream) -> TokenStream {
+pub(crate) fn stream_select(input: TokenStream) -> Result<TokenStream, syn::Error> {
     let args = Punctuated::<Expr, Token![,]>::parse_terminated
-        .parse(input)
-        .expect("macro expects a comma separated list of expressions");
+        .parse2(input)?;
     let generic_idents = (0..args.len()).map(|i| format_ident!("_{}", i)).collect::<Vec<_>>();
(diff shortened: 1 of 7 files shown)


taiki-e (changes requested):

I think it makes sense to add a feature like this (this is actually a stream version of [simple select](https://github.com/rust-lang/futures-rs/issues/2031)), but I don't want to add features that rely on `Fused{Future,Stream}` because [`Fused{Future,Stream}` will be removed in the next major version](https://github.com/rust-lang/futures-rs/issues/2207#issuecomment-782050833).

Xaeroxe:
@taiki-e I've removed the FusedStream dependency, this is now ready for review.

futures/tests/macro-tests/src/main.rs
```diff
@@ -65,4 +65,37 @@ fn main() {
             _ = b => unreachable!(),
         };
     });
+
+    // stream_select! macro
+    let _ = block_on(async {
+        let endless_ints = |i| stream::iter(vec![i].into_iter().cycle());
+
+        let mut endless_ones = futures04::stream_select!(endless_ints(1i32), stream::pending());
```
taiki-e, line 73: This module is only used to check things related to macro re-export. So, could you move the test to `futures/tests/async_await_macros.rs`?

futures-macro/src/stream_select.rs
```diff
@@ -0,0 +1,110 @@
+use proc_macro::TokenStream;
+use quote::{format_ident, quote, ToTokens};
+use syn::{parse::Parser, punctuated::Punctuated, Expr, Index, Token};
+
+/// The `stream_select!` macro.
+pub(crate) fn stream_select(input: TokenStream) -> TokenStream {
+    let args = Punctuated::<Expr, Token![,]>::parse_terminated
+        .parse(input)
+        .expect("macro expects a comma separated list of expressions");
```
taiki-e, line 9: I would like to avoid panic on proc-macro. Could you change stream_select function to return Result and handle it on the top-level stream_select_internal function? Like:

```rs
pub fn stream_select_internal(input: TokenStream) -> TokenStream {
    crate::stream_select::stream_select(input).unwrap_or_else(Error::into_compile_error)
}
```

taiki-e (changes requested):
impl LGTM. Sorry for the late review!

Xaeroxe:
@taiki-e ready for re-review, thank you!

futures/tests/async_await_macros.rs
```diff
@@ -308,6 +310,42 @@ fn select_on_mutable_borrowing_future_with_same_borrow_in_block_and_default() {
...
+
+        let mut endless_ones = stream_select!(endless_ints(1i32), stream::pending());
+        assert_eq!(endless_ones.next().await, Some(1));
+        assert_eq!(endless_ones.next().await, Some(1));
+
+        let mut finite_list = stream_select!(stream::iter(vec![1, 2, 3].into_iter()));
+        assert_eq!(finite_list.next().await, Some(1));
+        assert_eq!(finite_list.next().await, Some(2));
+        assert_eq!(finite_list.next().await, Some(3));
+        assert_eq!(finite_list.next().await, None);
```
taiki-e, line 328: If stream_select takes one stream, it is more inefficient than simply calling next on that stream, right?
If so, it would probably be preferable to make that case an error.

Xaeroxe: Reasonable, and corrected. This is ready for re-review!

taiki-e (approved):
LGTM. Thanks @Xaeroxe!

taiki-e:
Published in 0.3.17

PR #2263, opened 2020-11-14
feat(futures-util/stream): unzip operator

binier:
> Given that `Iterator::unzip` is like `.collect::<(Vec<_>, Vec<_>)>`, I would prefer to separate functionality like Shared Stream from Unzip.

_Originally posted by @taiki-e in https://github.com/rust-lang/futures-rs/issues/2256#issuecomment-723367469_

This version does exactly that.

closes: #2234


futures-util/src/stream/stream/unzip.rs
```diff
@@ -0,0 +1,66 @@
...
+    left: FromA,
+    right: FromB,
+}
+
+impl<St: Stream, FromA: Default, FromB: Default> Unzip<St, FromA, FromB> {
+    fn finish(self: Pin<&mut Self>) -> (FromA, FromB) {
+        let this = self.project();
+        (
+            mem::take(this.left),
+            mem::take(this.right),
```
taiki-e, line 24:
```suggestion
            mem::replace(this.left, Default::default()),
            mem::replace(this.right, Default::default()),
```
(`mem::take` requires Rust 1.40.)

taiki-e (commented):
Thanks!

taiki-e (approved):

PR #2268, opened 2020-11-23
feat: Add sink::unfold

Marwes:
A more general function than `from_fn` (#2254), as suggested in
https://github.com/rust-lang/futures-rs/pull/2254#issuecomment-731860678

### DIFF
--- a/futures-util/src/sink/unfold.rs
+++ b/futures-util/src/sink/unfold.rs
@@ -1,0 +1,83 @@
+use core::{future::Future, pin::Pin};
+use futures_core::task::{Context, Poll};
+use futures_sink::Sink;
+use pin_project::pin_project;
+
+/// Sink for the [`unfold`] function.
+#[pin_project]
+#[derive(Debug)]
+#[must_use = "sinks do nothing unless polled"]
+pub struct Unfold<T, F, R> {
+    state: Option<T>,
+    function: F,
+    #[pin]
+    future: Option<R>,
+}
+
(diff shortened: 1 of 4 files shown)


taiki-e (approved):
Thanks, this looks good to me. I think we'll deprecate this and add something like #2254 once async closures became working properly, but for now, this is more flexible and I prefer this than #2254. (AFAIK, no one is currently working on async closures, so I don't think async closures' problems will be fixed soon.)

cc @cramertj @Nemo157: Any opinion on adding this? (I tend to merge this)

futures-util/src/sink/unfold.rs
```diff
@@ -0,0 +1,83 @@
...
+
+/// Sink for the [`unfold`] function.
+#[pin_project]
+#[derive(Debug)]
+#[must_use = "sinks do nothing unless polled"]
+pub struct Unfold<T, F, R> {
+    state: Option<T>,
+    function: F,
+    #[pin]
+    future: Option<R>,
```
piegamesde, line 14: `state` and `future` are never present at the same time, so maybe the use of an enum could reduce the size of the struct?

taiki-e: I would accept a patch to reduce the size of `unfold`.

sdroege:
The way how this is implemented is slightly suboptimal for type inference, see https://github.com/rust-lang/futures-rs/pull/2311

PR #2273, opened 2020-11-28
futures-util: Migrate from pin-project to pin-project-lite

### LINKED ISSUE #2170: Replace pin-project with pin-project-lite
Almost any program doing async I/O uses `futures-utils`, which pulls in `syn` and `quote` dependencies, causing compilation times to be much higher than is needed.

We should consider replacing `pin-project` with `pin-project-line`, which is not using proc macros.

cc @taiki-e

taiki-e:
[`pin-project-lite`](https://github.com/taiki-e/pin-project-lite) 0.2 supports enum. However, unfortunately, some enums (`Either`, `(Try)MaybeDone`) that require pin-projection used in futures-util have tuple variants and are also public APIs, so, we cannot use pin-project-lite for these enums.

`futures-test` also depends on pin-project, but this PR doesn't touch that as that needs support for custom Drop implementation (https://github.com/taiki-e/pin-project-lite/pull/25) and `futures-test` is a crate intended to be used as a dev-dependency.

Closes #2170

### DIFF
--- a/futures-util/Cargo.toml
+++ b/futures-util/Cargo.toml
@@ -46,7 +46,7 @@ memchr = { version = "2.2", optional = true }
 futures_01 = { version = "0.1.25", optional = true, package = "futures" }
 tokio-io = { version = "0.1.9", optional = true }
 pin-utils = "0.1.0"
-pin-project = "1.0.1"
+pin-project-lite = "0.2"
 
 [dev-dependencies]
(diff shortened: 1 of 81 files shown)


taiki-e:
compile time with **no-default-features** + `std` feature (`cargo build --release -Ztimings --no-default-features --features std`)
before (master):



after (this pr):

cramertj (approved):

taiki-e:
@cramertj Thanks for the review!

PR #2277, opened 2020-12-01
ci: Use --feature-powerset --depth 2 instead of --each-feature

taiki-e:
This patch increases feature combinations to be checked (total of all public crates) from 69 to 210.

We have too many features to check the full powerset (1457!), so limit the max number of simultaneous feature flags by [`--depth` option](https://github.com/taiki-e/cargo-hack/pull/59). I think this should be sufficient in most cases as explained in https://github.com/taiki-e/cargo-hack/issues/58.

It took about 8-9 minutes to run in locally, but let's see how long it takes in CI... (If it's less than 10 minutes, it's probably okay to merge as windows CI takes about 8 minutes.)

FYI: @Nemo157 @jhpratt

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -192,7 +192,7 @@ jobs:
       - run: cargo bench --manifest-path futures-util/Cargo.toml --features=bilock,unstable
 
   features:
-    name: cargo hack check --each-feature
+    name: cargo hack check --feature-powerset
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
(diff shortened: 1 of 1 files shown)


taiki-e:
Ok, this [takes about 9 minutes in CI](https://github.com/rust-lang/futures-rs/pull/2277/checks?check_run_id=1478180939#step:5:1) (install 40s + check 8m). Given cargo-hack installation time can be reduced by [installing from GitHub release](https://github.com/taiki-e/cargo-hack/pull/91) (and [Travis does not return status for an hour](https://github.com/rust-lang/futures-rs/pull/2217#issuecomment-727688106)), this is probably okay.

PR #2279, opened 2020-12-07
Add stream::repeat_with() method

zhanghanyun:
If the element type of the stream you need does not implement Clone, or if you do not want to keep the repeated element in memory, you can instead use the repeat_with() function.


futures-util/src/stream/repeat.rs
```diff
@@ -52,3 +52,93 @@ impl<T> FusedStream for Repeat<T>
...
+    repeater: F,
+}
+
+impl<A, F: FnMut() -> A> Unpin for RepeatWith<F> {}
+
+impl<A, F: FnMut() -> A> Stream for RepeatWith<F> {
+    type Item = A;
+
+    fn poll_next(mut self: Pin<&mut Self>, _: &mut Context<'_>) -> Poll<Option<Self::Item>> {
+        Poll::Ready(Some((self.repeater)()))
```
taiki-e, line 73:
```suggestion
        Poll::Ready(Some((&mut self.repeater)()))
```

taiki-e (changes requested):
Thanks for the PR! This looks reasonable to me. Could you do the following?
- Rebase (to fix clippy fail)
- Apply my code suggestion (to fix build fail on old compilers)
- Move `replase_with` related code from `repeate.rs` to a new file `futures-util/src/stream/repeat_with.rs`. (futures-util usually has its own file for each combinator.)

futures-util/src/stream/repeat.rs
zhanghanyun: Thanks for your suggestion

taiki-e (approved):

PR #2283, opened 2020-12-08
perf: Pack the state and future of unfolds in the same memory

Marwes:
cc https://github.com/rust-lang/futures-rs/pull/2268#discussion_r532011280

### DIFF
--- a/futures-util/src/sink/unfold.rs
+++ b/futures-util/src/sink/unfold.rs
@@ -1,3 +1,4 @@
+use crate::unfold_state::UnfoldState;
 use core::{future::Future, pin::Pin};
 use futures_core::ready;
 use futures_core::task::{Context, Poll};
@@ -11,39 +12,7 @@ pin_project! {
     pub struct Unfold<T, F, R> {
         function: F,
         #[pin]
-        state: State<T, R>,
-    }
-}
-
-#[derive(Debug)]
-pub(crate) enum State<T, R> {
-    Value(T),
-    Future(/* #[pin] */ R),
-    Empty,
-}
-
-impl<T, R> State<T, R> {
-    pub(crate) fn project_future(self: Pin<&mut Self>) -> Option<Pin<&mut R>> {
-        // SAFETY Normal pin projection on the `Future` variant
-        unsafe {
-            match self.get_unchecked_mut() {
-                Self::Future(f) => Some(Pin::new_unchecked(f)),
-                _ => None,
-            }
-        }
-    }
-
-    pub(crate) fn take_value(self: Pin<&mut Self>) -> Option<T> {
-        // SAFETY We only move out of the `Value` variant which is not pinned
-        match *self {
-            Self::Value(_) => unsafe {
-                match std::mem::replace(self.get_unchecked_mut(), State::Empty) {
-                    State::Value(v) => Some(v),
-                    _ => std::hint::unreachable_unchecked(),
-                }
-            },
-            _ => None,
-        }
+        state: UnfoldState<T, R>,
     }
 }
 
@@ -69,7 +38,7 @@ impl<T, R> State<T, R> {
 pub fn unfold<T, F, R>(init: T, function: F) -> Unfold<T, F, R> {
     Unfold {
         function,
-        state: State::Value(init),
+        state: UnfoldState::Value(init),
     }
 }
 
@@ -92,7 +61,7 @@ where
                 panic!("start_send called without poll_ready being called first")
             }
         };
-        this.state.set(State::Future(future));
+        this.state.set(UnfoldState::Future(future));
         Ok(())
     }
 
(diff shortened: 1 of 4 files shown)


futures-util/src/sink/unfold.rs
```diff
@@ -55,23 +61,24 @@ where
 
     fn start_send(self: Pin<&mut Self>, item: Item) -> Result<(), Self::Error> {
         let mut this = self.project();
-        debug_assert!(this.future.is_none());
-        let future = (this.function)(this.state.take().unwrap(), item);
-        this.future.set(Some(future));
+        let future = match this.state.as_mut().project_replace(State::Empty) {
+            StateProjOwn::Value(value) => (this.function)(value, item),
+            StateProjOwn::Future(_) | StateProjOwn::Empty => unreachable!()
```
taiki-e, line 66:
```suggestion
            StateProjOwn::Future(_) => panic!("poll_ready not called first"),
            StateProjOwn::Empty => unreachable!()
```

nit: `StateProjOwn::Empty` is really unreachable unless closure panics and the user catch it, but `StateProjOwn::Future` can be reached if the user makes a mistake. A different panic message may be useful.

futures-util/src/stream/unfold.rs
```diff
@@ -74,44 +81,46 @@ where
...
 impl<T, F, Fut, Item> FusedStream for Unfold<T, F, Fut>
-    where F: FnMut(T) -> Fut,
-          Fut: Future<Output = Option<(Item, T)>>,
+where
+    F: FnMut(T) -> Fut,
+    Fut: Future<Output = Option<(Item, T)>>,
 {
     fn is_terminated(&self) -> bool {
-        self.state.is_none() && self.fut.is_none()
+        matches!(self.state, State::Empty)
```
taiki-e, line 94: `matches!` macro is not available in our MSRV (1.37). Could you replace it with `match` or `if let`?

futures-util/src/stream/unfold.rs
```diff
@@ -74,44 +81,46 @@ where
...
-    fn poll_next(
-        self: Pin<&mut Self>,
-        cx: &mut Context<'_>,
-    ) -> Poll<Option<Self::Item>> {
+    fn poll_next(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
         let mut this = self.project();
 
-        if let Some(state) = this.state.take() {
-            this.fut.set(Some((this.f)(state)));
+        if matches!(&*this.state, State::Value(_)) {
```
taiki-e, line 108: Same here

taiki-e (commented):
Thanks! This looks good to me aside from a few nits.

futures-util/src/sink/unfold.rs
```diff
@@ -55,27 +61,33 @@ where
...
+        Poll::Ready(
+            if let StateProj::Future(future) = this.state.as_mut().project() {
+                let result = match ready!(future.poll(cx)) {
+                    Ok(state) => {
+                        this.state.set(State::Value(state));
+                        Ok(())
+                    }
+                    Err(err) => Err(err),
+                };
+                result
```
taiki-e, line 86:
```suggestion
                match ready!(future.poll(cx)) {
                    Ok(state) => {
                        this.state.set(State::Value(state));
                        Ok(())
                    }
                    Err(err) => Err(err),
                }
```

Marwes:
Had to updated this to manual pin projections due to https://github.com/rust-lang/futures-rs/pull/2273 .

taiki-e (changes requested):
Thanks. The implementation LGTM, but seems CI is still failed. Could you fix it? (probably you need to replace `std` with `core` in `unfold_state.rs`)

In addition, there are many irrelevant format changes in `lib.rs`, could you remove them?

futures-util/src/unfold_state.rs
```diff
@@ -0,0 +1,34 @@
+use std::pin::Pin;
```
taiki-e, line 1:
```suggestion
use core::pin::Pin;
```

futures-util/src/unfold_state.rs
```diff
@@ -0,0 +1,34 @@
...
+                _ => None,
+            }
+        }
+    }
+
+    pub(crate) fn take_value(self: Pin<&mut Self>) -> Option<T> {
+        // SAFETY We only move out of the `Value` variant which is not pinned
+        match *self {
+            Self::Value(_) => unsafe {
+                match std::mem::replace(self.get_unchecked_mut(), UnfoldState::Empty) {
```
taiki-e, line 26:
```suggestion
                match core::mem::replace(self.get_unchecked_mut(), UnfoldState::Empty) {
```

futures-util/src/unfold_state.rs
```diff
@@ -0,0 +1,34 @@
...
+                    UnfoldState::Value(v) => Some(v),
+                    _ => std::hint::unreachable_unchecked(),
```
taiki-e, line 28:
```suggestion
                    _ => core::hint::unreachable_unchecked(),
```

futures-util/src/sink/unfold.rs
```diff
@@ -1,4 +1,4 @@
-use crate::unfold_state::UnfoldState;
+se crate::unfold_state::UnfoldState;
```
piegamesde, line 1: Typo?

taiki-e (approved):

PR #2290, opened 2020-12-13
Implement Clone on With combinator

akhramov:
Consider the following use-case of mpsc channels. There are two
categories of producers which produce items of two distinct types,
later unified in one using `With` combinator:

```
let (sender, receiver) = mspc::channel(100);

let download_status = sender.clone().with(|item| {
    futures::future::ok(Status::Download(item))
});

let unpack_status = sender.clone().with(|item| {
    futures::future::ok(Status::Unpack(item))
});
```

It would be convenient for `With` combinator to implement `Clone`,
since the underlying sink, `futures::channel::mpsc::Sender`,
implements it.

This change adds `#[derive(Clone)]` to `With` combinator

### DIFF
--- a/futures-util/src/sink/with.rs
+++ b/futures-util/src/sink/with.rs
@@ -11,7 +11,6 @@ use pin_project_lite::pin_project;
 pin_project! {
     /// Sink for the [`with`](super::SinkExt::with) method.
     #[must_use = "sinks do nothing unless polled"]
-    #[derive(Clone)]
     pub struct With<Si, Item, U, Fut, F> {
         #[pin]
         sink: Si,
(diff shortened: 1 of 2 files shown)


taiki-e (changes requested):
Thanks for the PR. derive is not the better way to do this, as that adds `U: Clone, Item: Clone` bounds. Could you use manual impl instead of deriving?

akhramov:
Thanks for the review. 

> Could you use manual impl instead of deriving?

Sure thing, please see the update. 

Also, I'm not familiar with the code style of the project, so please feel free to point out any slipups.

taiki-e (approved):
LGTM, thanks

PR #2293, opened 2020-12-22
futures_util::stream::SelectAll::push should use &self.

archseer:
FuturesUnordered::push bound is &self, so SelectAll should be able to use the same bound.

### DIFF
--- a/futures-util/src/stream/select_all.rs
+++ b/futures-util/src/stream/select_all.rs
@@ -60,7 +60,7 @@ impl<St: Stream + Unpin> SelectAll<St> {
     /// function will not call `poll` on the submitted stream. The caller must
     /// ensure that `SelectAll::poll` is called in order to receive task
     /// notifications.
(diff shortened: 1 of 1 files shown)


taiki-e (approved):
Thanks! As this is a breaking change, I'll merge this when master switches to 0.4-alpha.

archseer:
Is it a breaking change though? It's just loosening the bound from mut to ref. Should be similar to going from `FnMut` to `Fn`.

taiki-e:
`impl Fn` type is guaranteed to be `impl FnMut`, but `fn(&mut T, ..)` and `fn(&T, ..)` are different types.
Given that `SelectAll::push` has multiple arguments (self + item), I think this is very unlikely to cause actual breakage, but I tend to consider this as a breaking change. (Unfortunately, [RFC 1105](https://github.com/rust-lang/rfcs/blob/master/text/1105-api-evolution.md#signatures-in-functions) has no explicit mention of changes like this.)

Issue #2207, opened 2020-09-04
futures-core 1.0 release


taiki-e:
This is a proposal for the next major release of futures-core.
(I will open issues for other crates later. They are probably 0.4 instead of 1.0.)

## About

I propose to further reduce the APIs of futures-core and only provide what is really stable.
This makes futures-core a shim crate that provides future/task-related APIs of the standard library.
Eventually (maybe after `core::stream::Stream` stabilized), only crates that need compatibility with older compilers will depend on futures-core, and other crates will not need to depend on futures-core.

## APIs provided by this crate

The following two kinds of APIs are provided:
* APIs already stable in the standard library:
  * In versions where it is stable, re-export from the standard library 
  * In versions where it doesn't exist in the standard library or isn't stable yet, futures-core defines & provides the same (use build-script to detect it).
    This allows you to use items that are only available in the latest rustc, even if you are using an older version of rustc.
* APIs not yet stable in the standard library:
  * If we know what (definition of items) will be added to the standard library, we can define and provide the same until it stabilizes.
  * However, **basically do not add at this stage.**
    * `Stream` trait and `ready!` macro are probably the only exceptions to this.
  * Once stable, we can re-export with later versions (use build-script to detect it).

Other APIs are provided by individual `futures-*` crate.

## MSRV

APIs of this crate is really small, so it would be nice to be able to pin it to one of the following:

* Rust 1.36 that future/task-api stabilized.
* Rust 1.39 that async-await stabilized.

Every time the MSRV is increased, a new minor version is released.

## Changes

Here is a list of proposed actual changes:

  * `FusedFuture`, `FusedStream`.
    * Remove. See https://github.com/rust-lang/futures-rs/issues/1989 and https://github.com/rust-lang/futures-rs/issues/1894 for main reasons.
    * Uses of them in `select!` macro is replaced with the way suggested in https://github.com/rust-lang/futures-rs/issues/1989.
  * Type aliases (`BoxFuture`, `LocalBoxFuture`, `BoxStream`, `LocalBoxStream`)
    * Move to `futures-util`.
    * They are just type aliases of `Pin<Box>`ed future/stream.
  * Trait aliases (`TryFuture`, `TryStream`)
    * Move to `futures-util`. 
    * They are just trait aliases of future/stream that return `Result`, but the real trait aliases are not yet stable, so they have some limitations. See https://github.com/rust-lang/futures-rs/issues/2200 for one of the limitations.
  * `Stream` trait
[...]

PR #2294, opened 2020-12-22
Use autocfg instead of unstable feature(cfg_target_has_atomic)

taiki-e:
Instead of depending on unstable features of the compiler, use [`autocfg`](https://docs.rs/autocfg/1.0.1/autocfg/struct.AutoCfg.html#method.probe_expression) crate to detect if atomic CAS is available.

This completely removes the dependency on unstable features from crates other than futures and futures-util.
I'm planning to remove the dependency on unstable features from all crates, which is the first step in that plan.

~~`feature(cfg_target_has_atomic)` is propagated to the `core`/`alloc` imported in all dependencies if any of the dependencies graph crates are uses `feature(cfg_target_has_atomic)`, so it works if the user just enables the `feature(cfg_target_has_atomic)`.~~ EDIT: See https://github.com/rust-lang/futures-rs/pull/2294#issuecomment-749730512
(Previously you had to enable both `feature(cfg_target_has_atomic)` and `futures` crate's `cfg-target-has-atomic` feature.)

Unfortunately, this has some (potential) concerns about breakage:

- As said in https://github.com/rust-lang/futures-rs/issues/2207#issuecomment-689605831, there are concerns about compatibility with non-cargo build systems that do not run build-scripts. 
  If the build script is not called on these build systems, modules/types that require `cfg(target_has_atomic = "ptr")`, such as channels, will not compile. However, I think this can workaround by always treating `cfg(target_has_atomic = "ptr")` as true when the build script doesn't run. (This should behave exactly the same as current `futures` without `cfg-target-has-atomic` feature.)
  Also, I haven't actually tested it, but Bazel ([cargo-raze](https://github.com/google/cargo-raze)) and Buck([reindeer](https://github.com/facebookincubator/reindeer/tree/master/example)) seem to support build scripts, so I'm not too worried about this.
- Compile time will probably increase a little. `autocfg` is lightweight and widely used, and this is a much smaller increase compared to time reduced by #2273, so I don't think this is really a problem.
- **The biggest concern about this** is that modules/types that require `cfg(target_has_atomic = "ptr")` will not compile if `autocfg` does not work as expected. See https://github.com/rust-lang/rust/issues/78979, https://github.com/bluss/indexmap/issues/151, https://github.com/bluss/indexmap/issues/144, https://github.com/rust-lang/rust/issues/54216, https://github.com/rust-num/num-traits/issues/177, https://github.com/rust-num/num-complex/issues/83 for the actual case where similar issue occurred. 
  A known workaround for this seems to be to add a cargo feature that always treats `cfg(target_has_atomic = "ptr")` as true, like https://github.com/bluss/indexmap/pull/145, https://github.com/rust-num/num-traits/pull/185.

I haven't implemented workarounds yet, but I think there are workarounds for possible issues (basically).

Also, the approach I adopt in this PR is basically the same as the one I'm planning to use in the changes proposed in #2207.


Nemo157:
LGTM

> `feature(cfg_target_has_atomic)` is propagated to the `core`/`alloc` imported in all dependencies if any of the dependencies graph crates are uses `feature(cfg_target_has_atomic)`, so it works if the user just enables the `feature(cfg_target_has_atomic)`.

The user doesn't even need to enable this, for anyone not supporting targets like `thumbv6m` they can just assume that the target supports atomics, or they could use a similar `autocfg` script checking the existence of the `futures` APIs they need.

futures-task/build.rs
```diff
@@ -0,0 +1,22 @@
...
+
+// The rustc-cfg strings below are *not* public API. Please let us know by
+// opening a GitHub issue if your build environment requires some way to enable
+// these cfgs other than by executing our build script.
+fn main() {
+    let cfg = match AutoCfg::new() {
+        Ok(cfg) => cfg,
+        Err(e) => {
+            println!(
+                "cargo:warning=futures-core: unable to determine rustc version: {}",
```
taiki-e, line 11: Oh... this is not `futures-core`...

futures-channel/src/lib.rs
```diff
@@ -22,12 +20,9 @@
 
 #![doc(test(attr(deny(warnings), allow(dead_code, unused_assignments, unused_variables))))]
 
-#[cfg(all(feature = "cfg-target-has-atomic", not(feature = "unstable")))]
-compile_error!("The `cfg-target-has-atomic` feature requires the `unstable` feature as an explicit opt-in to unstable features");
-
 macro_rules! cfg_target_has_atomic {
     ($($item:item)*) => {$(
-        #[cfg_attr(feature = "cfg-target-has-atomic", cfg(target_has_atomic = "ptr"))]
+        #[cfg(not(no_atomic_cas))]
```
taiki-e, line 25: It's unfortunate that we need to use double negation, but AFAIK, this is the easiest way to implement a workaround for the first problem.

taiki-e:
> * A known workaround for this seems to be to add a cargo feature that always treats `cfg(target_has_atomic = "ptr")` as true, like [bluss/indexmap#145](https://github.com/bluss/indexmap/pull/145), [rust-num/num-traits#185](https://github.com/rust-num/num-traits/pull/185).

Another workaround for this is to pass a no-op expression (empty block or unit) to `autocfg` and treat `cfg(target_has_atomic = "ptr")` as true if it fails. This allows detecting cases that fail for unrelated reasons such as io.

cramertj:
Just seeing this now. I'm very familiar with the hassle that custom `build.rs` scripts like this can cause other build systems, and it'd be really nice if we didn't introduce another point of friction for those users. That said, those users are already accustomed to some friction, so it seems right to look more closely at the costs to those users vs. the benefits to `futures`.

It's easy for custom build systems to enable a feature flag, but running arbitrary other rust code (`autocfg`) at build time may or may not be simple. Is it possible to use autocfg to detect whether the feature should be used, and then go through normal feature flags to enable the behavior?

> This completely removes the dependency on unstable features from crates other than futures and futures-util.
I'm planning to remove the dependency on unstable features from all crates, which is the first step in that plan.

I'm not totally sure I understand the benefits of this change. These unstable feature usages are already gated on an `unstable` cargo feature flag right? That is, it isn't quite clear to me what the benefit is of removing unstable feature usages that aren't enabled without explicit opt-in.

taiki-e:
> Is it possible to use autocfg to detect whether the feature should be used, and then go through normal feature flags to enable the behavior?

Do you mean adding something like bluss/indexmap#145 and rust-num/num-traits#185? If so, it's definitely possible.

> These unstable feature usages are already gated on an `unstable` cargo feature flag right? That is, it isn't quite clear to me what the benefit is of removing unstable feature usages that aren't enabled without explicit opt-in.

I think it's nice to remove dependencies on unstable features, whether explicitly gated or not.

That said, I agree that there isn't much benefit from this patch, as the use still need to enable unstable features on the user side to actually use cfg_target_has_atomic.

PR #2306, opened 2021-01-07
perf: Avoid an Option in the `Map*` futures

Marwes:
Forces another `unsafe` block but shrinks the generated future a bit

### DIFF
--- a/futures-util/src/future/future/map.rs
+++ b/futures-util/src/future/future/map.rs
@@ -9,13 +9,14 @@ use crate::fns::FnOnce1;
 pin_project! {
     /// Internal Map future
     #[project = MapProj]
+    #[project_replace = MapProjReplace]
     #[derive(Debug)]
     #[must_use = "futures do nothing unless you `.await` or poll them"]
     pub enum Map<Fut, F> {
         Incomplete {
             #[pin]
             future: Fut,
-            f: Option<F>,
+            f: F,
         },
         Complete,
     }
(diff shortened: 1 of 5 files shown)


futures-util/src/future/future/map.rs
```diff
@@ -24,13 +24,26 @@ pin_project! {
...
     /// Creates a new Map.
     pub(crate) fn new(future: Fut, f: F) -> Self {
-        Self::Incomplete { future, f: Some(f) }
+        Self::Incomplete { future, f }
+    }
+
+    fn take_function(self: Pin<&mut Self>) -> F {
+        // SAFETY The `f` field is not pinned so we can move it out and return it
+        unsafe {
+            match core::mem::replace(self.get_unchecked_mut(), Self::Complete) {
```
taiki-e, line 33: This moves `future` field.  I think it is not sound, even if you don't access the field (it moves before the drop).
So I think you have to use [code like this](https://github.com/rust-lang/futures-rs/pull/2148/files#diff-1b645f83a773708969f122114e61b5c7e67898aa58c6d701e9d6f0337a3d746aL60-L86). (It's almost equivalent to what the pin-project's [`project_replace`](https://docs.rs/pin-project/1.0.3/pin_project/attr.pin_project.html#project_replace-method) method is [actually using](https://github.com/taiki-e/pin-project/blob/master/examples/project_replace-expanded.rs).)

Marwes: Yep, just realized that and browsed through what project_replace does. It is not as clear cut as I thought it was. Do you think `pin-project-lite` could/should support project_replace as well or is that out of scope? It would have helper in https://github.com/rust-lang/futures-rs/pull/2283 as well, and perhaps in other places in futures-rs also

taiki-e: I think it's theoretically can be implemented and it's okay to add it to pin-project-lite, but I'm not sure how easy it is to implement (currently, I don't have the time to investigate it).

futures-util/src/unfold_state.rs
```diff
@@ -1,32 +1,37 @@
...
-        match *self {
-            Self::Value(_) => unsafe {
-                match core::mem::replace(self.get_unchecked_mut(), UnfoldState::Empty) {
-                    UnfoldState::Value(v) => Some(v),
-                    _ => core::hint::unreachable_unchecked(),
-                }
+        match &*self {
+            UnfoldState::Value { .. } => match self.project_replace(UnfoldState::Empty) {
+                UnfoldStateProjReplace::Value { value } => Some(value),
+                _ => None,
```
taiki-e, line 34:
```suggestion
                _ => unreachable!(),
```

taiki-e (approved):
LGTM. Thanks!

PR #2322, opened 2021-01-15
Use assert_* functions for all future/stream/sink/io combinators

### LINKED ISSUE #2320: Add assertions for Future/Stream/Sink
Measures to prevent recurrence of #2311.

Refs: #2020

taiki-e:
Measures to prevent recurrence of #2311.

Refs: #2020
Closes #2320

### DIFF
--- a/futures-util/src/sink/mod.rs
+++ b/futures-util/src/sink/mod.rs
@@ -6,7 +6,7 @@
 //! - The [`SinkExt`] trait, which provides adapters for chaining and composing
 //!   sinks.
 
-use crate::future::Either;
+use crate::future::{assert_future, Either};
 use core::pin::Pin;
 use futures_core::future::Future;
 use futures_core::stream::{Stream, TryStream};
@@ -81,7 +81,7 @@ pub trait SinkExt<Item>: Sink<Item> {
         E: From<Self::Error>,
         Self: Sized,
     {
-        With::new(self, f)
+        assert_sink::<U, E, _>(With::new(self, f))
     }
 
     /// Composes a function *in front of* the sink.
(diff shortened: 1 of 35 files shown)


futures-util/src/sink/mod.rs
```diff
@@ -250,8 +251,11 @@ pub trait SinkExt<Item>: Sink<Item> {
     fn send_all<'a, St>(&'a mut self, stream: &'a mut St) -> SendAll<'a, Self, St>
     where
         St: TryStream<Ok = Item, Error = Self::Error> + Stream + Unpin + ?Sized,
+        // St: Stream<Item = Result<Item, Self::Error>> + Unpin + ?Sized,
         Self: Unpin,
     {
+        // TODO: type mismatch resolving `<St as Stream>::Item == std::result::Result<Item, <Self as futures_sink::Sink<Item>>::Error>`
+        // assert_future::<Result<(), Self::Error>, _>(SendAll::new(self, stream))
```
taiki-e, line 258: IIUC, this is not because the bounds are not satisfied, but because the compiler cannot resolve projection (assoc type).

futures-util/src/stream/stream/mod.rs
```diff
@@ -1334,7 +1336,10 @@ pub trait StreamExt: Stream {
     where
         S: Sink<Self::Ok, Error = Self::Error>,
         Self: TryStream + Sized,
+        // Self: TryStream + Sized + Stream<Item = Result<<Self as TryStream>::Ok, <Self as TryStream>::Error>>,
     {
+        // TODO: type mismatch resolving `<Self as futures_core::Stream>::Item == std::result::Result<<Self as futures_core::TryStream>::Ok, <Self as futures_core::TryStream>::Error>`
+        // assert_future::<Result<(), Self::Error>, _>(Forward::new(self, sink))
```
taiki-e, line 1342: Same here

futures-util/src/sink/mod.rs
taiki-e: Since `TryStream` is sealed, I think it's safe to replace the `TryStream` bounds with `Stream<Item = Result<..>>` bounds, but for now, I'll leave it as it is.

PR #2325, opened 2021-01-18
Implement Stream::size_hint() for Either

benkay86:
`futures::future::Either` should delegate `futures::stream::Stream::size_hint()` to its left/right branches.  This was done automatically by the either crate in futures 0.2, but the functionality was lost in futures 0.3 when we implemented our own `Either` type.


taiki-e (approved):
Thanks

PR #2332, opened 2021-02-06
Avoid once_cell in static wakers

KaiJewson:
`Waker`s are `#[repr(transparent)]` over their inner `RawWakers`. Therefore, instead of using a `once_cell` to lazily initialize a `Waker` with `Waker::from_raw`, we can instead simply store a static `RawWaker` and then use pointer casts to obtain a `Waker`. This also removes the `once_cell` dependency.


taiki-e (approved):
LGTM, thanks!

Issue #2053, opened 2020-01-27
Best practices for combinators with their own task pools (a la FuturesUnordered)


cramertj:
See https://github.com/rust-lang/futures-rs/issues/2047. When a combinator keeps its own task pool rather than deferring directly to the parent executor, cooperatively yielding (`cx.waker.wake_by_ref(); return Poll::Pending`) works "less well" because the task pool will remain active and thus be polled again. #2049 introduced a mitigation for this by limiting the number of times a `FuturesUnordered` is polled, but it'd be good to develop general guidance for these types of combinators, and perhaps a more principled solution for `FuturesUnordered`.

cc @jonhoo @carllerche

jonhoo:
I think the most appealing solution to this is for `FuturesUnordered` to also cooperatively yield using the same mechanism that the underlying tasks are. #2049 is sort of a blunt solution since it just blindly yields every so often, and that has some sad implications (such as the multiplicative property @cramertj points out in #2047). If a mechanism like https://github.com/tokio-rs/tokio/pull/2160 was available throughout the ecosystem, then `FuturesUnordered` to simply use `poll_cooperate` (in the lingo of that PR) each time it detects that it has looped for example. I _think_ the mechanism from that PR can be de-coupled from the executor by just providing hooks for the executor to bind into — see `tokio::league::ceded` — but it will certainly require some careful thought to be general enough.

cramertj:
Yes, if there was some sort of global mechanism for setting budget and decrementing / resetting where appropriate, I could see `FuturesUnordered` making use of that.

jonhoo:
One other option as briefly mentioned in https://github.com/rust-lang/futures-rs/issues/2047#issuecomment-577469813 is to make `FuturesUnordered` only ever iterate over its futures at most once. That is, if it ever detects that it is about to poll a given future for the second time in the same call to `poll_next`, then it returns `Poll::Pending`. How to do this efficiently given the current implementation is not entirely clear though.

udoprog:
> One other option as briefly mentioned in [#2047 (comment)](https://github.com/rust-lang/futures-rs/issues/2047#issuecomment-577469813) is to make `FuturesUnordered` only ever iterate over its futures at most once. That is, if it ever detects that it is about to poll a given future for the second time in the same call to `poll_next`, then it returns `Poll::Pending`. How to do this efficiently given the current implementation is not entirely clear though.

[...]

PR #2333, opened 2021-02-07
FuturesUnordered: Do not poll the same future twice per iteration

taiki-e:
Instead of using a constant, use the length of FuturesUnordered to ensures that each future is polled only once at most per iteration. This does not solve the fundamental problem, but should be able to improve the current situation somewhat.

This fixes at least one of the problems reported in https://github.com/tokio-rs/tokio/issues/3493. See that issue for details of the bug.

r? @cramertj @jonhoo
cc #2053

### DIFF
--- a/futures-util/src/stream/futures_unordered/mod.rs
+++ b/futures-util/src/stream/futures_unordered/mod.rs
@@ -30,22 +30,6 @@ use self::task::Task;
 mod ready_to_run_queue;
 use self::ready_to_run_queue::{ReadyToRunQueue, Dequeue};
 
-/// Constant used for a `FuturesUnordered` to determine how many times it is
-/// allowed to poll underlying futures without yielding.
-///
-/// A single call to `poll_next` may potentially do a lot of work before
-/// yielding. This happens in particular if the underlying futures are awoken
-/// frequently but continue to return `Pending`. This is problematic if other
-/// tasks are waiting on the executor, since they do not get to run. This value
-/// caps the number of calls to `poll` on underlying futures a single call to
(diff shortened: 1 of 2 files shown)


jonhoo:
I think my only concern here is whether the futures are guaranteed to not have their order changed during a single call to `FuturesUnordered::poll` If that were the case, then we might end up calling (worst case) a single future `.len()` times if it keeps being moved to the front of the poll list. This is just from memory, but I seem to recall there being a linked list in the implementation with an insert on `.wake()`, which is where this fear is coming from. If that's not the case, then +1 from me.

It doesn't solve the underlying issue, but does make things at least a bit more sane.

taiki-e:
> I think my only concern here is whether the futures are guaranteed to not have their order changed during a single call to `FuturesUnordered::poll`

Good point.
IIUC, `ReadyToRunQueue::dequeue` is the only function that can change the order of futures within the range (`0..self.len`) that acquired at the first of `poll_next`. And AFAIK this is only called in `Stream::poll_next` and `Drop::drop` that take mutable reference.

https://github.com/rust-lang/futures-rs/blob/ddbf5228660f5e06c4edd7bb6d57dfe429e5fb8a/futures-util/src/stream/futures_unordered/ready_to_run_queue.rs#L45-L49
https://github.com/rust-lang/futures-rs/blob/ddbf5228660f5e06c4edd7bb6d57dfe429e5fb8a/futures-util/src/stream/futures_unordered/mod.rs#L425-L427

So, my understanding is that the order doesn't change. But I might have missed something.

PR #2338, opened 2021-02-13
Format most crates with rustfmt

taiki-e:
This formats the following crates with rustfmt.

- futures
- futures-core
- futures-executor
- futures-io
- futures-macro
- futures-sink
- futures-task
- futures-test
- examples

futures-util and futures-channel conflicts with many of other PRs and will not be formatted in this PR.
futures-executor also have some PR, but I plan to close them (I'll describe in detail when closing them).

### DIFF
--- a/.rustfmt.toml
+++ b/.rustfmt.toml
@@ -1,0 +1,1 @@
+use_small_heuristics = "Max"
--- a/examples/functional/src/main.rs
+++ b/examples/functional/src/main.rs
@@ -30,9 +30,7 @@ fn main() {
         // responsible for transmission
         pool.spawn_ok(fut_tx_result);
 
-        let fut_values = rx
-            .map(|v| v * 2)
-            .collect();
+        let fut_values = rx.map(|v| v * 2).collect();
 
         // Use the executor provided to this async block to wait for the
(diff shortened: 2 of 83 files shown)


taiki-e:
By the way, one of the reasons for doing this is that it's painful to review a PR that contains a lot of irrelevant format changes like https://github.com/rust-lang/futures-rs/pull/2283

taiki-e:
> +850 −652

I don't like the rustfmt to inflate the code. Perhaps it makes sense to use `use_small_heuristics = "Max"` like [rust-lang/rust](https://github.com/rust-lang/rust/commit/ce39461ca75ae5cdc5e7798869d88666ad139646#diff-a947c9d2523f34e53b9cdcd6411174a3148f23ea03ad8a61c1aee7b3ea005440)

taiki-e:
> they don't really conflict with this PR (futures-channel)

Hmm, the reason futures-channel doesn't conflict is actually due to a bug where rustfmt can't format the code contained in the module inside the macro. I need to exclude futures-channel from the list.

taiki-e:
This blocks a lot of other work I want to do, so I'll close this for now.

PR #2345, opened 2021-02-14
small change in `select` example.

### LINKED ISSUE #2334: `select` example in docs
I am an async noob but the 'simple example' in [docs](https://docs.rs/futures/0.3.12/futures/future/fn.select.html) on `select` may be a bad example.
The example `select`s one of the futures:
```rust
let future1 = async { 1 };
let future2 = async { 2 };
```
The problem in this is that the computations are both `Poll::Ready` immediately (at least on my machine).
So the result of the `select` is always `Either::Left`, as I would expect from [the implementation](https://github.com/rust-lang/futures-rs/blob/master/futures-util/src/future/select.rs#L92).
Therefore, the `assert!` in the example is misleading IMHO.

I would suggest being explicit about one future finishing first.
The simplest example I came up with is [here](https://gist.github.com/tungli/f89599edd09ab192707c9df2b56a06c8).

tungli:
In the previous example one future always finished first (because of implementation details), possibly confusing users about what `select` does.

This commit resolves the issue by being explicit about one future finishing before the other.

Closes #2334

### DIFF
--- a/futures-util/src/future/select.rs
+++ b/futures-util/src/future/select.rs
@@ -40,7 +40,7 @@ impl<A: Unpin, B: Unpin> Unpin for Select<A, B> {}
 /// 
 /// // These two futures have different types even though their outputs have the same type.
 /// let future1 = async {
-///     let () = future::pending().await; // will never finish
+///     future::pending::<()>().await; // will never finish
 ///     1
 /// };
 /// let future2 = async { 
--- a/futures-util/src/future/select.rs
+++ b/futures-util/src/future/select.rs
@@ -52,10 +52,10 @@ impl<A: Unpin, B: Unpin> Unpin for Select<A, B> {}
 /// pin_mut!(future2);
 ///
 /// let value = match future::select(future1, future2).await {
-///     Either::Left((value1, _)) => value1, // `value1` is resolved from `future1`
-///     // `_` represents `future2`
+///     Either::Left((value1, _)) => value1,  // `value1` is resolved from `future1`
+///                                           // `_` represents `future2`
 ///     Either::Right((value2, _)) => value2, // `value2` is resolved from `future2`
-///     // `_` represents `future1`
+///                                           // `_` represents `future1`
 /// };
 ///
 /// assert!(value == 2);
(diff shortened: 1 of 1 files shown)


futures-util/src/future/select.rs
```diff
@@ -32,25 +32,33 @@ impl<A: Unpin, B: Unpin> Unpin for Select<A, B> {}
...
-/// let future2 = async { 2 };
+/// use futures::{
+///     pin_mut,
+///     future::Either,
+///     future::self,
+/// };
+/// 
+/// // These two futures have different types even though their outputs have the same type.
+/// let future1 = async {
+///     let () = future::pending().await; // will never finish
```
taiki-e, line 43:
```suggestion
///     future::pending::<()>().await; // will never finish
```

IIRC, clippy warns `let () =`

futures-util/src/future/select.rs
```diff
@@ -32,25 +32,33 @@ impl<A: Unpin, B: Unpin> Unpin for Select<A, B> {}
...
 /// pin_mut!(future1);
 /// pin_mut!(future2);
 ///
 /// let value = match future::select(future1, future2).await {
 ///     Either::Left((value1, _)) => value1, // `value1` is resolved from `future1`
-///                                          // `_` represents `future2`
+///     // `_` represents `future2`
 ///     Either::Right((value2, _)) => value2, // `value2` is resolved from `future2`
-///                                           // `_` represents `future1`
+///     // `_` represents `future1`
```
taiki-e, line 58: "`_` represents `future2`" is a description of `Either::Left((value1, _)) => value1` on the previous line, so it's probably easier to read if you format it like this:

```rust
///     // `value1` is resolved from `future1`
///     // `_` represents `future2`
///     Either::Left((value1, _)) => value1,
///     // `value2` is resolved from `future2`
///     // `_` represents `future1`
///     Either::Right((value2, _)) => value2,
```

taiki-e (commented):
Thanks! This looks good to me aside from a few nits.

futures-util/src/future/select.rs
tungli: yeah, I haven't noticed I changed the spacing accidentally. I edited it to match the original example to:
```rust
/// let value = match future::select(future1, future2).await {
///     Either::Left((value1, _)) => value1,  // `value1` is resolved from `future1`
///                                           // `_` represents `future2`
///     Either::Right((value2, _)) => value2, // `value2` is resolved from `future2`
///                                           // `_` represents `future1`
/// };
```

taiki-e (approved):

PR #2353, opened 2021-02-22
tests: cfg-ignore compat tests if there is no compat feature

nrc:
I realise that `cargo test` is meant to be run using `--all-features`, however, this is not obvious the first time one runs tests because the error about the missing `compat` module happens first, and the helpful message about  `--all-features` never gets output. With this change, we get far enough into compilation that you get to see the helpful error message, not the unhelpful ones. I believe this is an improvement to a contributors first experience (it would have saved me some time, for sure) and that outweighs the small noise of having the additional `cfg` lines.


taiki-e (approved):
That makes sense. Thanks!

PR #2355, opened 2021-02-22
Expose implementable Fn-trait API variants for futures-util

### LINKED ISSUE #2349: Consider exposing Fn*1-traits publicly
I would like to wrap a tokio receiver into a stream using `stream::unfold()` while giving a concrete type. As the used closure's type is not available (and Fn-traits are not implementable on stable), this cannot be done yet.

However, there are already the wrapper traits in [fns.rs](https://github.com/rust-lang/futures-rs/blob/master/futures-util/src/fns.rs). Exposing them publicly and using them in `stream::unfold()` (and possibly more places) would allow for concrete types. This is already done (internally) e.g. for [`FuturesExt::map_into()`](https://github.com/rust-lang/futures-rs/blob/4124111369d1f69a07183fccc7d97006e8846008/futures-util/src/future/future/mod.rs#L64).

As a downside the api would look a little more cluttered, when replacing trait bounds like `FnMut(A) -> O` by `FnMut1<A>`.

b3t33:
**Consider this a proof of concept, or sketch how to close #2349.**

Expand the existing `Fn(Once|Mut)?1`-traits for zero and two arguments. These and dual API using these instead of `Fn(..) ->` is exposed publicly gated by the `fntraits` feature in futures-util.

Modify all internal structures for which the enriched API is present (see modules of `futures_util::fns`) to use Fn-traits from `fns`.
Since existing public API might break for a drop-in replacement, expose a dual API. There where problems, at least for `inspect*`-variants, requiring type annotated closures (this would be a breaking change,
which is avoided by the dual API use).

Possibly the feature should be named differently to your liking, also I'm not sure if is better included under unstable features.

### DIFF
--- a/futures-util/Cargo.toml
+++ b/futures-util/Cargo.toml
@@ -22,6 +22,7 @@ io-compat = ["io", "compat", "tokio-io"]
 sink = ["futures-sink"]
 io = ["std", "futures-io", "memchr"]
 channel = ["std", "futures-channel"]
+fntraits = []
 
 # Unstable features
 # These features are outside of the normal semver guarantees and require the
--- a/futures-util/src/fns.rs
+++ b/futures-util/src/fns.rs
@@ -1,8 +1,127 @@
-use core::marker::PhantomData;
-use core::fmt::{self, Debug};
+#![cfg_attr(feature = "fntraits", doc = r#"
+Wrap `Fn`, `FnMut`, and `FnOnce` traits allowing for more concrete typed adapters.
+# This module contains
+## Fn-trait wrappers
+- [`FnOnce0`], [`FnMut0`], and [`Fn0`] traits mimicking `Fn*() -> Output`.
+- [`FnOnce1`], [`FnMut1`], and [`Fn1`] traits mimicking `Fn*(A) -> Output`.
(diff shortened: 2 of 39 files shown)


taiki-e:
I understand the use cases for this feature, but these traits are like hacks and I wasn't planning on using them in the long-term...

b3t33:
Indeed, this is quite hacky...at least all checks succeeded now. However, the refactoring was straight forward, even more when desugaring the standard Fn-traits `Fn(A) -> O` as `Fn<(A,), Output = O>`.

In the end, you as a maintainer are free to not merge this. Maybe the changes to `futures/tests/auto_traits.rs` are interesting in either case, as `Inspect_(Ok|Err)` relate to try-streams.

taiki-e (commented):
I would prefer not to expose these traits as we want to remove these hacks in the end. 

So I'm going to close this. Thanks anyway @b3t33!

rollo-b2c2:
Hey @taiki-e could you please reconsider this as this is a major pain in terms of developer quality of life.

FnTraits have been unstable since 2015 and this is still an issue 2 years since @b3t33 raised it.

What just ends up happening is that devs just copy the trait hierachy into their projects so they can write their own stream functions .

PR #2357, opened 2021-02-22
Mention StreamExt::next in documentation of the poll!(..) macro

### LINKED ISSUE #2356: Feature-Request: Add poll_next! macro, analogous to poll!.
For futures there is a convenient `poll!` macro that allows polling a future once inside of an async context. I couldn't find an equivalent for the `poll_next` method of streams. This would be really useful in my opinion.

This could be done somewhat like this:
```rust
use futures::{Stream, StreamExt};
use std::future::Future;
use std::pin::Pin;
use std::task::{Context, Poll};

fn poll_next<StreamType>(stream: StreamType) -> PollNextOnce<StreamType> {
    PollNextOnce {
        stream,
    }
}

struct PollNextOnce<StreamType> {
    stream: StreamType,
}

impl<StreamType> Future for PollNextOnce<StreamType>
where
    StreamType: Stream + Unpin,
{
    type Output = Poll<Option<StreamType::Item>>;

    fn poll(self: Pin<&mut Self>, context: &mut Context) -> Poll<Self::Output> {
        Poll::Ready(self.get_mut().stream.poll_next_unpin(context))
    }
}
```

FSMaxB-dooshop:
Closes #2356

### DIFF
--- a/futures-util/src/async_await/poll.rs
+++ b/futures-util/src/async_await/poll.rs
@@ -9,6 +9,10 @@ use futures_core::task::{Context, Poll};
 /// This macro is only usable inside of `async` functions, closures, and blocks.
 /// It is also gated behind the `async-await` feature of this library, which is
 /// activated by default.
+///
+/// If you need the result of polling a [`Stream`](crate::stream::Stream),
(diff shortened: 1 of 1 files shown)


taiki-e (approved):
Thanks!

PR #2374, opened 2021-03-15
Add TryStreamExt::try_collect

yotamofek:
First-time contribution, hope I did everything right!

So this PR adds a `TryCollect` combinator to `TryStream`s that discards all items from inner stream and returns just a total count of them. I'm not sure if it's generally useful enough to be added to the `futures` crate, but I found myself having to use my own implementation of this functionality a couple of times, so I'd thought I'd see if there's enough interest for this to be included.


taiki-e (commented):

When I saw the title, the first thing I imagined was a function like `.filter(|res| async move { res.is_ok() }).count()`, but this seems to be equivalent to `.try_fold(0, |count, _| async move { Ok(count + 1) })`, and can be written by using a single existing combinator (try_fold).

You can also use the following helper functions to avoid redundancy when using multiple times.

```rust
// for futures 0.3
// usage: `.try_fold(0, try_count::<_, Error>)`
async fn try_count<T, E>(count: usize, _: T) -> Result<usize, E> {
    Ok(count + 1)
}

// for futures 0.4
// usage: `.try_fold(0, try_count)`
async fn try_count<T, E>(count: usize, res: Result<T, E>) -> Result<usize, E> {
    res?;
    Ok(count + 1)
}
```

I think it can be easily written using an existing combinator, and also I feel there are not too many use cases. So I tend to prefer to postpone adding this to `StreamExt` for now.

PR #2377, opened 2021-03-22
Reexport Future, Stream, and Sink from the root of futures_util

coolreader18:
This would make it much easier to migrate from futures to futures-util: just a simple find and replace, as long as you're not using `futures::prelude::*`


taiki-e (commented):
Thanks! Could you do the same for Try* traits and io traits?

coolreader18:
Oh I *knew* there was something I was forgetting :)

taiki-e (approved):

PR #2397, opened 2021-04-17
Fastmutex

nathdobson:


futures-util/src/lock/mutex.rs
```diff
@@ -169,8 +185,11 @@ impl<T: ?Sized> Mutex<T> {
         let old_state = self.state.fetch_and(!IS_LOCKED, Ordering::AcqRel);
         if (old_state & HAS_WAITERS) != 0 {
             let mut waiters = self.waiters.lock().unwrap();
-            if let Some((_i, waiter)) = waiters.iter_mut().next() {
-                waiter.wake();
+            while let Some(wait_key) = waiters.queue.pop_front() {
+                if waiters.slab.contains(wait_key) {
```
hyyking, line 189: `Index<_>` is implemented using `get()` so the check could be simplified with:
```rust
if let Some(present) = waiters.slab.get(wait_key) { 
    present.wake();
    break;
}
```

PR #2400, opened 2021-04-20
Remove uses of unstable feature(cfg_target_has_atomic)

### LINKED ISSUE #2399: Feedback Request: "disable-cas-atomics" feature
Hey all!

I'm looking to use the `futures` crate on an embedded target with no CAS atomics (specifically `thumbv6m-none-eabi`), and currently I am required to use the following features which require nightly:

```toml
[dependencies.futures]
version = "0.3.14"
default-features = false
features = [ "cfg-target-has-atomic", "unstable" ]
```

Without these flags, I get the following errors:

```
error[E0599]: no method named `compare_exchange` found for struct `AtomicUsize` in the current scope
   --> /home/james/.cargo/registry/src/github.com-1ecc6299db9ec823/futures-core-0.3.14/src/task/__internal/atomic_waker.rs:264:14
    |
264 |             .compare_exchange(WAITING, REGISTERING, Acquire, Acquire)
    |              ^^^^^^^^^^^^^^^^ method not found in `AtomicUsize`

error[E0599]: no method named `compare_exchange` found for struct `AtomicUsize` in the current scope
   --> /home/james/.cargo/registry/src/github.com-1ecc6299db9ec823/futures-core-0.3.14/src/task/__internal/atomic_waker.rs:282:42
    |
282 |                     let res = self.state.compare_exchange(
    |                                          ^^^^^^^^^^^^^^^^ method not found in `AtomicUsize`

error[E0599]: no method named `swap` found for struct `AtomicUsize` in the current scope
   --> /home/james/.cargo/registry/src/github.com-1ecc6299db9ec823/futures-core-0.3.14/src/task/__internal/atomic_waker.rs:308:40
    |
308 | ...                   self.state.swap(WAITING, AcqRel);
    |                                  ^^^^ method not found in `AtomicUsize`

error[E0599]: no method named `fetch_or` found for struct `AtomicUsize` in the current scope
   --> /home/james/.cargo/registry/src/github.com-1ecc6299db9ec823/futures-core-0.3.14/src/task/__internal/atomic_waker.rs:375:26
    |
375 |         match self.state.fetch_or(WAKING, AcqRel) {
    |                          ^^^^^^^^ method not found in `AtomicUsize`

error[E0599]: no method named `fetch_and` found for struct `AtomicUsize` in the current scope
   --> /home/james/.cargo/registry/src/github.com-1ecc6299db9ec823/futures-core-0.3.14/src/task/__internal/atomic_waker.rs:381:28
    |
381 |                 self.state.fetch_and(!WAKING, Release);
    |                            ^^^^^^^^^ method not found in `AtomicUsize`

error: aborting due to 5 previous errors
```

This is due to the target atomic configuration flag currently being nightly-only. Unfortunately, this means that the library that I am working on would also be restricted to nightly-only, or I'd have to fork the `futures` crate, which would be a compatibility problem.

Would you be open to adding a `disable-cas-atomics` feature that ALWAYS disables these features (without checking the target cfg)? I believe this should be a straightfoward, but somewhat verbose PR, so I wanted check before working on it.

CC @dirbaio, who has also run into this for embedded async support.

Thanks!

taiki-e:
Instead of depending on unstable features of the compiler, use the TARGET environment variables provided by cargo for the build script and a list of targets that do not have atomic CAS.

This completely removes the dependency on unstable features from crates other than futures, futures-util, futures-io.

This way is inspired by the way [heapless](https://github.com/japaric/heapless/commit/44c66a74845eacd3f18eae5be8ae6d887f4f7f0b) and [defmt](https://github.com/knurling-rs/defmt/blob/963152f0fc530fca64ba4ff1492d9c4b7bf76062/build.rs#L42-L51) do, but this doesn't maintain the target list manually. (It's not really fully automated, but [it's very easy to update](https://github.com/taiki-e/futures-rs/blob/cfg_target_has_atomic-2/.github/workflows/ci.yml#L194).)

Closes #2399 
Replaces #2294

r? @Nemo157 @cramertj
cc @jamesmunns @Dirbaio


Dirbaio:
This is great! Thanks for working on it :) 

Removing `cfg-target-has-atomic` will break users building for [custom targets](https://docs.rust-embedded.org/embedonomicon/custom-target.html) that have no atomics, though.

taiki-e:
> Removing `cfg-target-has-atomic` will break users building for [custom targets](https://docs.rust-embedded.org/embedonomicon/custom-target.html) that have no atomics, though.

Hmm, you are right. 

That said, we want to reduce dependency on unstable features as much as possible... So perhaps the compromise is to make `cfg(no_atomic_cas)` an unstable public API so that atomics can be disabled via RUSTFLAGS:

```sh
RUSTFLAGS='--cfg futures_no_atomic_cas' cargo build
```

Dirbaio:
Why not simply keep `cfg-target-has-atomic` as-is? I'm not a fan of having to use RUSTFLAGS. The standard way to customize lib behavior is cargo features.

Aditionally, removing it` (or keeping it but making it do nothing) is semver-breaking: it makes custom targets stop compiling.

taiki-e:
> Why not simply keep `cfg-target-has-atomic` as-is? I'm not a fan of having to use RUSTFLAGS. The standard way to customize lib behavior is cargo features.

- The current design (enabling unstable features by feature flags) is that dependencies can enable unstable features even if the user doesn't enable them directly, and it is problematic.
  See also https://github.com/rayon-rs/rayon/issues/369, https://github.com/rayon-rs/rayon/issues/364, https://github.com/alexcrichton/proc-macro2#unstable-features, https://github.com/tokio-rs/tokio/pull/1774#issuecomment-554505014, etc.

  > Note that this must not only be done for your crate, but for any crate that depends on your crate. This infectious nature is intentional, as it serves as a reminder that you are outside of the normal semver guarantees. 

- `feature(cfg_target_has_atomic)` will never stable as far as I understand. Continuing to rely on such "forever unstable" features is a maintenance burden. (Actually, I think it will be removed in favor of cfg(accessible) in the future, not "forever unstable".)


> semver-breaking

No, cfg-target-has-atomic feature is unstable and stability is not guaranteed.

https://github.com/rust-lang/futures-rs/blob/c77cd3c9ba19b65db3d1931d2087032752fe9c7d/futures/Cargo.toml#L46-L48

Dirbaio:
I wasn't aware of this issue with `unstable` features and the shift to use RUSTFLAGS. Looks good to me then! :+1:

jamesmunns:
Hey @taiki-e, thank you for implementing this!

I have tested this change in my `cassette` repo (which prompted this), and I was able to replace my copy/paste exports with just exporting the futures dependency.

I can confirm this change works as expected for my project, at least :)

taiki-e:
@jamesmunns Thanks for testing this!

futures-channel/build.rs
```diff
@@ -0,0 +1,40 @@
...
+// - `futures_no_atomic_cas`
+//      Assume the target does not have atomic CAS (compare-and-swap).
+//      This is usually detected automatically by the build script, but you may
+//      need to enable it manually when building for custom targets or using
+//      non-cargo build systems that don't run the build script.
+//
+// With the exceptions mentioned above, the rustc-cfg strings below are
+// *not* public API. Please let us know by opening a GitHub issue if your build
+// environment requires some way to enable these cfgs other than by executing
+// our build script.
```
taiki-e, line 17: As said in https://github.com/rust-lang/futures-rs/pull/2400#issuecomment-823391327, exposes cfg(futures_no_atomic_cas) as unstable public API (outside of the normal semver guarantees).

taiki-e: Also, we probably need to set `#![cfg_attr(docsrs, doc(cfg_hide(futures_no_atomic_cas)))]` when https://github.com/rust-lang/rust/pull/79341 merged...

jyn514: That might be a while :sweat_smile:

PR #2404, opened 2021-04-30
Document flush behavior on error for `send_all` & `forward`

### LINKED ISSUE #2402: `SendAll` doesn't flush on encountering `Err` from `TryStream`
The implementation of `fn poll` in `Future` implementation for `SendAll` propagates `Err` case without flushing `Sink`. I think it can be expected from the implementation to flush the sink even in that case that is when the `Stream` errors. 

https://github.com/rust-lang/futures-rs/blob/c77cd3c9ba19b65db3d1931d2087032752fe9c7d/futures-util/src/sink/send_all.rs#L84

ktff:
Closes #2402


taiki-e (approved):
Thanks!

PR #2407, opened 2021-05-05
Use #[proc_macro] at Rust 1.45+

### LINKED ISSUE #2406: Removing `proc_macro_hack`
The macros using `proc_macro_hack` like `select!`, `try_join!` don't work with Rust Analyzer ( https://github.com/rust-analyzer/rust-analyzer/issues/6835 ).

I don't know what is required to remove `proc_macro_hack`, but given that [it's page](https://github.com/dtolnay/proc-macro-hack) says:

> Note: As of Rust 1.45 this crate is superseded by native support for #[proc_macro] in expression position. Only consider using this crate if you care about supporting compilers between 1.31 and 1.45.

It could be replaceable with `proc_macro`, which has better support.

taiki-e:
Closes #2406

### DIFF
--- a/futures-macro/Cargo.toml
+++ b/futures-macro/Cargo.toml
@@ -16,6 +16,9 @@ proc-macro = true
 
 [features]
 
+[build-dependencies]
+autocfg = "1"
+
 [dependencies]
 proc-macro2 = "1.0"
 proc-macro-hack = "0.5.19"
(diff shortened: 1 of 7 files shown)


taiki-e:
Seems to work as expected (`: i32` annotation in `let out`):

PR #2408, opened 2021-05-06
improve must_use comment on RemoteHandle

### LINKED ISSUE #2152: Improve must_use comment on RemoteHandle
https://github.com/rust-lang/futures-rs/blob/3bf5ac99e939e6ff70283937fa91e8d77e5f70e8/futures-util/src/future/future/remote_handle.rs#L38

This is burying the lede, it seems more important to mention the behaviour on drop, which will cause many more issues if the handle is not used

https://github.com/rust-lang/futures-rs/blob/3bf5ac99e939e6ff70283937fa91e8d77e5f70e8/futures-util/src/future/future/remote_handle.rs#L23-L24

ibraheemdev:
resolves #2152


taiki-e (approved):
Thanks

PR #2409, opened 2021-05-06
Async test function attribute

### LINKED ISSUE #1890: Add #[async_test] to futures-test crate?
Hi!

I'm interested what you're thinking about adding an `#[async_test]` proc-macro to the `futures-test` crate.

There were previous discussions about it at https://github.com/rust-lang-nursery/futures-rs/issues/1169#issuecomment-416301993 and there are also some implementations existing already.

- I have a kind of ugly (sry but I'm not good at proc-macros) but working implementation at [futures-await-test-preview](https://github.com/ngg/futures-await-test/tree/0.3) which uses `futures_executor::LocalPool::new().run_until()`, I'm not sure if this is better or worse then using `block_on`, for my use-cases it seems to work perfectly.
- There is one created for tokio: [tokio-async-await-test](https://docs.rs/tokio-async-await-test/0.1.0/tokio_async_await_test/)

Examples for intended use-case:
```rust
#[async_test]
async fn normal_test() {
    assert!(create_future().await == 4);
}

#[async_test]
#[should_panic]
async fn panic_test() {
    assert!(create_future().await == 5);
}

#[async_test]
async fn result_test() -> Result<(), ()> {
    assert!(create_result_future().await? == 4);
    Ok(())
}
```

Currently I use my implementation to write tests for a Rust wrapper around an asynchronous C++ library, which uses only `futures-0.3`, not `tokio`. I find it really useful, perhaps others would too even if it uses `futures-executor` only at first. Later it could be somehow extended to support other executors as well..

Would you be interested in adding something like this to `futures-test`?

ibraheemdev:
```rust
#[futures_test::test]
async fn my_test() {
     let fut = async { true };
     assert!(fut.await);
}
```

Resolves #1890

### DIFF
--- a/futures-test/src/lib.rs
+++ b/futures-test/src/lib.rs
@@ -54,7 +54,7 @@ mod track_closed;
 /// Enables an `async` test function. The generated future will be run to completion with
 /// [`futures_executor::block_on`](futures_executor::block_on).
 ///
-/// ```no_run
+/// ```
 /// #[futures_test::test]
 /// async fn my_test() {
 ///     let fut = async { true };
--- a/futures-test/src/lib.rs
+++ b/futures-test/src/lib.rs
@@ -16,6 +16,7 @@ compile_error!(
 #[cfg(feature = "std")]
 pub mod __private {
     pub use futures_core::{future, stream, task};
+    pub use futures_executor::block_on;
     pub use std::{
         option::Option::{None, Some},
         pin::Pin,
@@ -49,3 +50,16 @@ pub mod io;
 mod assert_unmoved;
 mod interleave_pending;
 mod track_closed;
+
+/// Enables an `async` test function. The generated future will be run to completion with
+/// [`futures_executor::block_on`](futures_executor::block_on).
+///
+/// ```no_run
+/// #[futures_test::test]
+/// async fn my_test() {
+///     let fut = async { true };
+///     assert!(fut.await);
+/// }
+/// ```
+#[cfg(feature = "std")]
+pub use futures_macro::test_internal as test;
--- a/futures-test/src/lib.rs
+++ b/futures-test/src/lib.rs
@@ -61,5 +61,17 @@ mod track_closed;
 ///     assert!(fut.await);
 /// }
 /// ```
+///
+/// This is equivalent to the following code:
+///
+/// ```
+/// #[test]
+/// fn my_test() {
+///     futures::executor::block_on(async move {
+///         let fut = async { true };
+///         assert!(fut.await);
+///     })
+/// }
+/// ```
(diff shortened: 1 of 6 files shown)


ibraheemdev:
This macro is actually not even testing specific. Maybe it should be called `block_on` and re-exported from `futures_util`/`futures` also?

futures-test/src/lib.rs
```diff
@@ -49,3 +50,16 @@ pub mod io;
...
+/// Enables an `async` test function. The generated future will be run to completion with
+/// [`futures_executor::block_on`](futures_executor::block_on).
+///
+/// ```
+/// #[futures_test::test]
+/// async fn my_test() {
+///     let fut = async { true };
+///     assert!(fut.await);
+/// }
+/// ```
```
taiki-e, line 63: rustdoc does the equivalent of cargo run, not cargo test, so doctests don't actually do anything in this example. Could you add tests to `futures/tests/`?

taiki-e (commented):
Thanks!


> Maybe it should be called `block_on` and re-exported from `futures_util`/`futures` also?

I tend to prefer to provide this attribute as part of futures-test until https://github.com/rust-lang/futures-rs/issues/2295 is done. (Also, block_on is provided by futures-executor that depend on futures-util, so it is difficult to provide this as part of futures-util.)

futures/tests/test_macro.rs
```diff
@@ -0,0 +1,18 @@
...
+        let fut = async { true };
+        assert!(fut.await);
+
+        let fut = async { false };
+        assert!(!fut.await);
+    }
+
+    #[futures_test::test]
+    #[should_panic]
+    async fn it_is_being_run() {
```
ibraheemdev, line 14: I don't know if there is a reliable way to test if the generate test actually gets run (has the `test` attribute).

taiki-e: This is good as-is for now.
[macrotest](https://crates.io/crates/macrotest) is a crate for that use case ([example](https://github.com/taiki-e/pin-project/tree/main/tests/expand)), but it's difficult to use with `#[futures_test::test]`, as it will show other code generated by `#[test]`.

futures-test/src/lib.rs
taiki-e, line 63:
```suggestion
/// ```
///
/// This is equivalent to the following code:
///
/// ```rust
/// #[test]
/// fn my_test() {
///     futures::executor::block_on(async move {
///         let fut = async { true };
///         assert!(fut.await);
///     })
/// }
/// ```
```

taiki-e (commented):
LGTM aside from a few nits.

taiki-e (approved):
Thanks again.

PR #2410, opened 2021-05-06
Abortable streams

### LINKED ISSUE #1424: Abortable Streams
Feature request: an analogue to `futures_util::future::Abortable` but for fusing Streams.

Also as a side note, wondering if it would make sense to provide an `AbortGuard` type/api on top of `AbortHandle` that aborts on drop? Previous discussions seemed to indicate that this behaviour shouldn't be the default, but it seems maybe reasonable to provide it in some way for the scenarios where it's appropriate or wanted?

ibraheemdev:
This PR makes `Abortable` able to work with both futures and streams. 

Is there a better place to put `Abortable` than `futures-util/src/abortable.rs`? Maybe a `shared/` folder?

Resolves #1424


futures-util/src/abortable.rs
```diff
@@ -0,0 +1,175 @@
...
+    /// # use futures::stream::{self, StreamExt};
+    ///
+    /// let (abort_handle, abort_registration) = AbortHandle::new_pair();
+    /// let mut stream = Abortable::stream(stream::iter(vec![1, 2, 3]), abort_registration);
+    /// abort_handle.abort();
+    /// assert_eq!(stream.next().await, None);
+    /// # });
+    /// ```
+    pub fn new(future: T, reg: AbortRegistration) -> Self {
+        Self { task: future, inner: reg.inner }
```
taiki-e, line 59:
```suggestion
    pub fn new(task: T, reg: AbortRegistration) -> Self {
        Self { task, inner: reg.inner }
```

taiki-e (commented):
Thanks! This looks good to me!

> Is there a better place to put `Abortable` than `futures-util/src/abortable.rs`? Maybe a `shared/` folder?

`shared/` is confusing with `future/shared`, so I think `futures-util/src/abortable.rs` is fine.

futures-util/src/abortable.rs
```diff
@@ -0,0 +1,175 @@
...
+    fn poll(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
+        self.try_poll(cx, |fut, cx| fut.poll(cx))
+    }
+}
+
+impl<St> Stream for Abortable<St>
+where
+    St: Stream,
+{
+    type Item = St::Item;
```
taiki-e, line 157: There may be use cases where it is preferable to separate between cases canceled by abort and others, but I think it is fine to return `St::Item` since we can add a method like is_aborted if such a feature is needed.

futures-util/src/abortable.rs
```diff
@@ -0,0 +1,175 @@
...
+    ///
+    /// Usage with streams:
+    ///
+    /// ```
+    /// # futures::executor::block_on(async {
+    /// # use futures::future::{Abortable, AbortHandle};
+    /// # use futures::stream::{self, StreamExt};
+    ///
+    /// let (abort_handle, abort_registration) = AbortHandle::new_pair();
+    /// let mut stream = Abortable::stream(stream::iter(vec![1, 2, 3]), abort_registration);
```
taiki-e, line 53: This function doesn't seem to exist, but I think it's not actually a bad idea to have both `Abortable::new` (require `T: Future`) and `Abortable::stream` (require `T: Stream`) because we can prevent the creation of Abortables that cannot be polled. -- However, it may feel redundant when we provide other abortables (e.g., abortable sink) in the future.

futures-util/src/abortable.rs
```diff
@@ -0,0 +1,180 @@
...
+        (Self { inner: inner.clone() }, AbortRegistration { inner })
+    }
+}
+
+// Inner type storing the waker to awaken and a bool indicating that it
+// should be aborted.
+#[derive(Debug)]
+struct AbortInner {
+    waker: AtomicWaker,
+    abort: AtomicBool,
```
taiki-e, line 100:
```suggestion
    aborted: AtomicBool,
```

futures-util/src/abortable.rs
```diff
@@ -0,0 +1,180 @@
...
+    /// # });
+    /// ```
+    pub fn new(task: T, reg: AbortRegistration) -> Self {
+        Self { task, inner: reg.inner }
+    }
+
+    /// Checks whether the task has been aborted. See [`AbortHandle::abort`] for details.
+    pub fn is_aborted(&self) -> bool {
+        self.inner.aborted.load(Ordering::Relaxed)
+    }
```
taiki-e, line 65: This is actually a method that returns true if it is notified that the task should be aborted (= `abort` was called), and I think there are some notes that are desirable to explain in the documentation.

- This will return true even if `abort` is called after the task has completed.
- This will return true even if the task still is running at the time.

taiki-e: Alternatively, we can omit `is_aborted` method until someone actually requests it.

ibraheemdev: We might as well keep it, because we can use it internally in `poll_next` instead of loading the flag manually. I'll update the docs.

ibraheemdev: Updated the docs, is `Ordering::Relaxed` enough for the user facing API?

taiki-e (commented):
Thanks. LGTM aside from a nit.

taiki-e (approved):

PR #2411, opened 2021-05-06
add either example

### LINKED ISSUE #2276: doc: Either could use an example
The docs on the Either enum are fairly unclear (as in: how to use it, what it's for).

An example would do wonders, say one that shows what you need to write if you want to construct a single future that is the merger of two futures chosen (conditionally) at-present:

~~~
if cond {
    Either::Left(async { ... })
} else {
    Either::Right(async { ... })
}
~~~

ibraheemdev:
Resolves #2276


taiki-e (approved):
Thanks!

PR #2413, opened 2021-05-06
clean up try(join) combinator

ibraheemdev:
Now that `Try/JoinX` was removed, the macros are just noise.

### DIFF
--- a/futures-util/src/future/join.rs
+++ b/futures-util/src/future/join.rs
@@ -1,81 +1,59 @@
-#![allow(non_snake_case)]
-
-use super::assert_future;
-use crate::future::{maybe_done, MaybeDone};
+use crate::future::{assert_future, maybe_done, MaybeDone};
 use core::fmt;
 use core::pin::Pin;
 use futures_core::future::{FusedFuture, Future};
 use futures_core::task::{Context, Poll};
 use pin_project_lite::pin_project;
 
-macro_rules! generate {
-    ($(
-        $(#[$doc:meta])*
-        ($Join:ident, <$($Fut:ident),*>),
-    )*) => ($(
-        pin_project! {
-            $(#[$doc])*
-            #[must_use = "futures do nothing unless you `.await` or poll them"]
-            pub struct $Join<$($Fut: Future),*> {
-                $(#[pin] $Fut: MaybeDone<$Fut>,)*
-            }
-        }
-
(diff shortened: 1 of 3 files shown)


futures-util/src/future/join.rs
```diff
@@ -1,81 +1,59 @@
...
+where
+    Fut1: Future + fmt::Debug,
+    Fut1::Output: fmt::Debug,
+    Fut2: Future + fmt::Debug,
+    Fut2::Output: fmt::Debug,
+{
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_struct("Join").field("fut1", &self.fut1).field("fut2", &self.fut2).finish()
+    }
+}
```
taiki-e, line 33: I'm not sure why the previous code didn't do it, but this can probably be replaced by `#[derive(Debug)]`.

futures-util/src/future/try_join.rs
```diff
@@ -1,97 +1,62 @@
...
+    Fut1::Ok: fmt::Debug,
+    Fut1::Error: fmt::Debug,
+    Fut2: TryFuture + fmt::Debug,
+    Fut2::Ok: fmt::Debug,
+    Fut2::Error: fmt::Debug,
+{
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_struct("TryJoin").field("fut1", &self.fut1).field("fut2", &self.fut2).finish()
+    }
+}
```
taiki-e, line 35: Same here.

taiki-e (commented):
Thanks!

futures-util/src/future/join.rs
ibraheemdev: The `Debug` macro can't handle associated type trait bounds:
```rust
error[E0277]: `<Fut1 as Future>::Output` doesn't implement `Debug`
  --> futures-util/src/future/join.rs:11:14
   |
11 |     #[derive(Debug)]
   |              ^^^^^ `<Fut1 as Future>::Output` cannot be formatted using `{:?}` because it doesn't implement `Debug`
```

taiki-e: Oh, sorry! I missed that problem.

taiki-e (approved):

PR #2414, opened 2021-05-06
Override Debug for oneshot::{Sender,Receiver}.

goffrie:
Prior to this diff, the Debug impl for oneshot channels printed something like:
```
Sender { inner: Inner { complete: false, data: Lock { locked: false, data: UnsafeCell }, rx_task: Lock { locked: false, data: UnsafeCell }, tx_task: Lock { locked: false, data: UnsafeCell } } }
```
which isn't very helpful. Instead, just print
`Sender { complete: false }` (or `true`, if the other end has dropped).

Note that the `T: Debug` bound is retained to allow for the possibility
of allowing slightly more interesting debug output in the future.

### DIFF
--- a/futures-channel/src/oneshot.rs
+++ b/futures-channel/src/oneshot.rs
@@ -16,7 +16,6 @@ use crate::lock::Lock;
 ///
 /// This is created by the [`channel`](channel) function.
 #[must_use = "futures do nothing unless you `.await` or poll them"]
-#[derive(Debug)]
 pub struct Receiver<T> {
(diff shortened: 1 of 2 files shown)


ibraheemdev:
Some users may want the detailed debug information. It may be particularly helpful for debugging deadlocks.

goffrie:
A user can't actually lock the internal locks themselves, so if there were a deadlock it would be a bug internal to futures-rs.

taiki-e (approved):
Thanks!

PR #2415, opened 2021-05-07
FuturesUnordered::clear

### LINKED ISSUE #2327: Consider adding FuturesUnordered::clear() and SelectAll::clear()
We're using `FuturesUnordered` to race multiple connection futures in order to implement the RFC 8305 Happy Eyeballs spec. Once we make a successful connection, we're supposed to close all the other pending connections in order to free up the resources. We are wrapping our `FuturesUnordered` in an `Option` to make sure we can drop any other pending connections as soon as possible.

However, it be simpler if we could just call a `FuturesUnordered::clear()` to throw away the futures. Likewise, I imagine a `SelectAll::clear()` could also be useful.

ibraheemdev:
resolves #2327


futures-util/src/stream/futures_unordered/mod.rs
```diff
@@ -576,19 +552,26 @@ impl<Fut> Debug for FuturesUnordered<Fut> {
...
+impl<Fut> FuturesUnordered<Fut> {
+    /// Clear all the futures associated with this task.
+    pub fn clear(&mut self) {
         unsafe {
             while !self.head_all.get_mut().is_null() {
                 let head = *self.head_all.get_mut();
                 let task = self.unlink(head);
                 self.release_task(task);
             }
         }
```
taiki-e, line 564: In the `drop`, we can skip resetting some fields because we know that the user will not access the FuturesUnordered again. However, in the `clear`, we will need to reset other fields as well (at least `self.is_terminated.store(false, Relaxed)` is needed).

futures/tests/stream_futures_unordered.rs
```diff
@@ -300,3 +300,14 @@ fn polled_only_once_at_most_per_iteration() {
...
+
+#[test]
+fn clear() {
+    let mut tasks = FuturesUnordered::from_iter(vec![future::ready(1), future::ready(2)]);
+
+    assert_eq!(block_on(tasks.next()), Some(1));
+    assert!(!tasks.is_empty());
+
+    tasks.clear();
+    assert!(tasks.is_empty());
```
taiki-e, line 312: Could you add a case for re-adding elements after cleared?

taiki-e (commented):
Thanks!

futures-util/src/stream/futures_unordered/mod.rs
ibraheemdev: should the `ready_to_run_queue` be reset as well?

ibraheemdev: and, can `self.is_terminated.store(false, Relaxed)` also be called in the `drop` impl, or is that extra overhead?

taiki-e: Yes

ibraheemdev: Do you mean `is_terminated.store(true)`, or `false`?

taiki-e: I meant `false`. That is the same behavior as `FuturesUnordered::new`:

https://github.com/rust-lang/futures-rs/blob/569106699d179ef4a3fbdf1417d7b772da095d28/futures-util/src/stream/futures_unordered/mod.rs#L133-L137

(I'm hoping that `FuturesUnordered::clear` will work as an efficient version of `mem::replace(futures_unordered, FuturesUnordered::new())`.)

ibraheemdev: Sounds good, updated.

futures-util/src/stream/futures_unordered/ready_to_run_queue.rs
```diff
@@ -85,25 +85,34 @@ impl<Fut> ReadyToRunQueue<Fut> {
...
+    //
+    // # Safety
+    //
+    // Note that each task has a strong reference count associated with it
+    // which is owned by the ready to run queue. All tasks MUST have had
+    // their futures dropped already by `FuturesUnordered::clear`. This
+    // method just pulls out tasks and drops their refcounts.
+    pub(crate) unsafe fn clear(&self) {
+        loop {
+            match self.dequeue() {
```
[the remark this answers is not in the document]
ibraheemdev: Then how does `FuturesUnorder::clear` call `self.ready_to_run_queue.clear()` through the `Arc`?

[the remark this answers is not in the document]
ibraheemdev: I think the best we can do is add a safety requirement of unique access. The current `self` still has weak references to itself through it's tasks, so `Arc::get_mut` will not work.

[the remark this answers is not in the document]
taiki-e: Oh, sorry. I forgot it is wrapped in `Arc`.

> I missed that requirement.

I think that safety requirement is guaranteed by `&mut self` of `FuturesUnorder::clear`, but it would be nice to have code comments on that.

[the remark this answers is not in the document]
taiki-e: > I think the best we can do is add a safety requirement of unique access.

Agreed.

ibraheemdev:
With this change `is_terminated` and `ready_to_run_queue` are modified on drop as well, should this be avoided?

taiki-e (commented):
LGTM. Could you resolve the merge conflicts?

taiki-e (approved):
Thanks!

PR #2416, opened 2021-05-07
impl Send + Sync for FuturesUnordered iterators

### LINKED ISSUE #1801: futures_unordered::IterMut does not implement Send
I'm trying to write code that looks like this:

```rust
async fn foo() {
    for future in futures_unordered.iter_mut() {
        do_something_async().await;
    }
}
```

This, unfortunately, makes the future returned by `foo` non-`Send` because the `IterMut` exists across an `await` boundary. Is it possible to fix this and have `IterMut` implement `Send`?

ibraheemdev:
Resolves #1801


taiki-e (approved):
Thanks!

PR #2417, opened 2021-05-07
add size_hint to mpsc receiver

ibraheemdev:
Return the max buffer size of `mpsc::Receiver` as the upper bound in `Stream::size_hint`.

### DIFF
--- a/futures-channel/src/mpsc/mod.rs
+++ b/futures-channel/src/mpsc/mod.rs
@@ -282,7 +282,7 @@ struct UnboundedInner<T> {
 
 #[derive(Debug)]
 struct BoundedInner<T> {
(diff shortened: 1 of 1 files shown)


taiki-e (approved):
Thanks!

Nemo157:
At first glance this looks incorrect, if another item is pushed onto the stream while it's being consumed then it will exceed the size hint. Only once all senders have been dropped could this be used as the size-hint (and in that case it should be possible to see how many items are in the queue exactly).

ibraheemdev:
I only just realized this was wrong after creating it... It is still possible to provide a size-hint in the case that all senders have been dropped, or the channel is closed.

PR #2421, opened 2021-05-07
Add rustfmt check to CI

taiki-e:
- Add rustfmt check to CI
- Format modules defined inside macros

### DIFF
--- a/tools/fmt.sh
+++ b/tools/fmt.sh
@@ -1,0 +1,26 @@
+#!/bin/bash
+
+# Format all rust code.
+#
+# Usage:
+#    ./tools/fmt.sh
+#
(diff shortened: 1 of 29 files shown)


tools/fmt.sh
```diff
@@ -0,0 +1,26 @@
...
+# Refs: https://github.com/rust-lang/rustfmt/issues/4078
+
+set -euo pipefail
+IFS=$'\n\t'
+
+cd "$(cd "$(dirname "${0}")" && pwd)"/..
+
+# shellcheck disable=SC2046
+if [[ -z "${CI:-}" ]]; then
+    (
```
taiki-e, line 18: I'm not sure why I added these extra parentheses...

PR #2423, opened 2021-05-08
add FuturesUnordered::into_iter, make iter_pin_ref public

### LINKED ISSUE #1795: Add FuturesUnordered into_iter method
I have a `FuturesUnordered` that I'd like to deconstruct and get the futures out of. Currently the only way to achieve this is to wrap the futures in some kind of `OptionFuture` type which allows me to call `FuturesUnordered::iter_mut`, steal the future, then drop the `FuturesUnordered` before it can be polled again. This is extra boilerplate and involves adding unwraps to my code. Could `FuturesUnordered` support this directly?

ibraheemdev:
- Remove `IterPinRef`. `FuturesUnordered::iter_pin_ref` is private for some reason, so it is impossible to create a `IterPinRef`. Of course, this means that no one used it. I don't see how iterating over a `Pin<&Fut>` is useful anyways. This is still a breaking change however, because the struct was public, albeit impossible to construct.
- Implement `IntoIterator` for `FuturesUnordered`, `&FuturesUnordered`, and `&mut FuturesUnordered`, and add an `IntoIter` iterator struct for the first.

Resolves #1795.

### DIFF
--- a/futures-util/src/stream/futures_unordered/iter.rs
+++ b/futures-util/src/stream/futures_unordered/iter.rs
@@ -18,12 +18,16 @@ pub struct IterMut<'a, Fut: Unpin>(pub(super) IterPinMut<'a, Fut>);
 
 #[derive(Debug)]
 /// Immutable iterator over all futures in the unordered set.
-pub struct Iter<'a, Fut: Unpin> {
+pub struct IterPinRef<'a, Fut> {
     pub(super) task: *const Task<Fut>,
     pub(super) len: usize,
     pub(super) pending_next_all: *mut Task<Fut>,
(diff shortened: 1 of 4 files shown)


taiki-e:
I think it's fine to make `iter_pin_ref` public. The API it adds is very small.

If we remove `IterPinRef` from public API, I would prefer to remove only the export of it, instead of remove `IterPinRef` itself. [`Pin<&Self>` itself has several use cases (though, the use cases are not related to future, in most cases)](https://github.com/taiki-e/pin-project/issues/92#issuecomment-532569295), we can easily expose it when someone requests it.

taiki-e (approved):
Thanks!

PR #2427, opened 2021-05-08
Change StreamExt::scan to pass state to closure by value

### LINKED ISSUE #2171: `StreamExt::scan` is barely useful due to lifetime issue
Continuing #2046, the lifetime issue is still a headache, that the async block basically cannot access the state when it's executed. Currently, I stick to `futures::stream::unfold` workaround to bind streams with states.. IMO, the future block of `scan` could own the state and returns to stream just like what `unfold` did . Perhaps we can make this change, or are there any other considerations for this design?

taiki-e:
Fixes #2171

```diff
      fn scan<S, B, Fut, F>(self, initial_state: S, f: F) -> Scan<Self, S, Fut, F>
      where
-         F: FnMut(&mut S, Self::Item) -> Fut,
-         Fut: Future<Output = Option<B>>,
+         F: FnMut(S, Self::Item) -> Fut,
+         Fut: Future<Output = Option<(S, B)>>,
          Self: Sized,
```

Related: https://github.com/rust-lang/futures-rs/issues/1609#issuecomment-493526043, https://github.com/rust-lang/futures-rs/pull/2254#issuecomment-731860678

### DIFF
--- a/futures-util/src/stream/stream/mod.rs
+++ b/futures-util/src/stream/stream/mod.rs
@@ -715,18 +715,25 @@ pub trait StreamExt: Stream {
     /// of the stream until provided closure returns `None`. Once `None` is
     /// returned, stream will be terminated.
     ///
+    /// Unlike [`Iterator::scan`], the closure takes the state by value instead of
+    /// mutable reference to avoid [the limitation of the async
+    /// block](https://github.com/rust-lang/futures-rs/issues/2171).
+    ///
     /// # Examples
     ///
     /// ```
     /// # futures::executor::block_on(async {
-    /// use futures::future;
     /// use futures::stream::{self, StreamExt};
     ///
     /// let stream = stream::iter(1..=10);
     ///
-    /// let stream = stream.scan(0, |state, x| {
-    ///     *state += x;
-    ///     future::ready(if *state < 10 { Some(x) } else { None })
+    /// let stream = stream.scan(0, |mut state, x| async move {
+    ///     state += x;
+    ///     if state < 10 {
+    ///         Some((state, x))
+    ///     } else {
+    ///         None
+    ///     }
     /// });
     ///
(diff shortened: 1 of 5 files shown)


taiki-e:
FYI @cramertj @olegnn

futures-util/src/stream/stream/mod.rs
```diff
@@ -715,27 +715,34 @@ pub trait StreamExt: Stream {
...
     ///
     /// assert_eq!(vec![1, 2, 3], stream.collect::<Vec<_>>().await);
     /// # });
     /// ```
     fn scan<S, B, Fut, F>(self, initial_state: S, f: F) -> Scan<Self, S, Fut, F>
     where
-        F: FnMut(&mut S, Self::Item) -> Fut,
-        Fut: Future<Output = Option<B>>,
+        F: FnMut(S, Self::Item) -> Fut,
+        Fut: Future<Output = Option<(S, B)>>,
```
taiki-e, line 745: >         Fut: Future<Output = Option<(S, B)>>,

Hmm, it seems this (state, item) is the reverse order of the tuples returned by unfold's closure (item, state). However, given the order of arguments of scan/forld's closure (state, item), I feel this is reasonable.

benluelo:
Are there plans to cut a new release containing this change? Considering it was over 2 years go, I'm quite surprised it's still unreleased: https://docs.rs/futures/latest/futures/stream/trait.StreamExt.html#method.scan

taiki-e:
This is a breaking change. Unfortunately, breaking releases of such widely used crates cannot be released often, as they have a significant impact on the ecosystem. This will be released once many (some?) other issues included in the milestone are resolved, so please consider helping to resolve the other issues included in the milestone.

michaelbeaumont:
Would it not be possible to simply add another combinator in the mean time?

PR #2428, opened 2021-05-09
expose iterators from SelectAll

ibraheemdev:
Closes #1964

Issue #1964, opened 2019-11-07
Expose .iter_mut() from SelectAll


yshui:
Currently there is no way to modify the set of streams in SelectAll, yet the underlying FuturesUnordered does have this capability.

I wonder if it's possible to expose the same thing in SelectAll as well?

ibraheemdev:
when #2423 is merged `IntoIter` should be exposed as well, so this issue should be kept open until then

PR #2431, opened 2021-05-10
Change SelectAll iterators to return stream instead of StreamFuture

taiki-e:
The conversion from StreamFuture to the underlying stream always succeeds because FuturesUnordered removes completed futures. So, I think it makes sense and more useful that SelectAll iterators return stream instead of StreamFuture.

This also removes iter_pin_ref and iter_pin_mut, because SelectAll's streams are always Unpin.

cc #1964, #2428
FYI @ibraheemdev


ibraheemdev:
ah nice, you newtyped the iterators, I was going to bring that up.

PR #2438, opened 2021-05-11
Implement try_chunks

### LINKED ISSUE #1947: try_chunks for futures 0.3
There is a [chunks](https://docs.rs/futures-preview/0.3.0-alpha.19/futures/stream/trait.StreamExt.html#method.chunks) method for `StreamExt`.

Could there be a `try_chunks` for `TryStreamExt`? The closure's domain would be `Vec<T>`, converting from `Result<T, E>`. 

One can produce the same result by performing the fold themselves, however it doesn't look and feel nice.

ivan770:
Closes #1947

### DIFF
--- a/futures-util/src/stream/try_stream/mod.rs
+++ b/futures-util/src/stream/try_stream/mod.rs
@@ -12,6 +12,8 @@ use crate::fns::{
 use crate::future::assert_future;
 use crate::stream::assert_stream;
 use crate::stream::{Inspect, Map};
+#[cfg(feature = "alloc")]
+use alloc::vec::Vec;
 use core::pin::Pin;
 use futures_core::{
     future::{Future, TryFuture},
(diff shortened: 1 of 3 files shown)


taiki-e (commented):
Thanks for the PR! I'm ok with adding such a feature, but I'm not sure if the behavior when an error occurs is good with this.

I feel it might be preferable to have our own error type that contains the error returned by the stream and the elements of the buffer at the time the error occurred.
What do you think?

ivan770:
> 
> 
> Thanks for the PR! I'm ok with adding such a feature, but I'm not sure if the behavior when an error occurs is good with this.
> 
> I feel it might be preferable to have our own error type that contains the error returned by the stream and the elements of the buffer at the time the error occurred.
> What do you think?

Yeah, I thought about adding a tuple of all elements before error and the error itself, but my approach would lead to conversion of `TryStream` to `Stream`. I think your approach with custom error type seems more suitable, as `TryStream` is preserved. I'll modify PR a bit later.

ivan770:
Sorry for lot's of intermediate commits and CI activation requests. I guess simple ID verification could be better for GitHub Actions 😅

futures-util/src/stream/try_stream/mod.rs
```diff
@@ -506,6 +506,7 @@ pub trait TryStreamExt: TryStream {
     where
         Self: Sized,
     {
+        #[allow(clippy::type_complexity)]
```
ivan770, line 509: I think it's ok to allow type complexity here, since it's still easy to read

taiki-e (approved):
Thanks!

taiki-e:
Published in 0.3.16.

PR #2445, opened 2021-05-15
adding loop_fn back to futures-rs, 0.3 intial implementation

### LINKED ISSUE #2444: adding loop_fn_back in to the future-rs
I have created a small library that isn't entirely completed but almost there to add loop fn back in, 
I wasn't entirely sure how I would alter the mod files to add it in but i think there 

the library is at 
[https://github.com/warlord500/loop_fn_future](https://github.com/warlord500/loop_fn_future)

warlord500:
so this an attempt to add loop_fn back in again right now there are a few issues one.

- [ ] add docs to struct without warnings!
- [ ] add at least one working test! 
- [ ]  change doc test
- [ ] fix atomic_cas_target.rs
this was done so i could get the code to build.

thanks to @ibraheemdev for helping me get rid of unsafe!!

Closes #2444


taiki-e (commented):
Thanks for the PR! As said in #2444, I would prefer not to add this unless there is a specific use case that requires this.

PR #2449, opened 2021-06-08
Fix typos

taiki-e:


seanmonstar (approved):

futures-executor/tests/local_pool.rs
```diff
@@ -404,7 +404,7 @@ fn tasks_are_scheduled_fairly() {
 }
 
 // Tests that the use of park/unpark in user-code has no
-// effect on the expected behaviour of the executor.
+// effect on the expected behavior of the executor.
```
taiki-e, line 407: "behaviour" is not actually wrong (it is British English), but we already use "behavior" (American English) in most places, so prefer to use "behavior".

ibraheemdev: Or we could use "behaviour" everywhere :wink:

PR #2450, opened 2021-06-08
Introduce stream::select_with_strategy

najamelan:
The current stream::select function only supports round robin. To change that
would be a breaking change, so for now I propose a `select_bias` which
takes a closure that defines the selection strategy.

- When a breaking change is planned, we could re-consider this, but it
  does have 4 generic parameters of which one is a closure, compared to
  the 2 streams in `select`. We could also reimplement `select` on top of this,
  we just need to deal with hiding the generics of the closure, but I reckon a 
  function pointer can be named and we don't need to capture anything....
- note the name select_bias. It's not consistent with `select_biased` used
  elsewhere in the lib. This can obviously be changed, but I feel the "ed"
  at the end does not add any semantic value but makes function names
  longer.
- There isn't really integration tests included right now, but the examples
  in the docs get tested by doc test.

### DIFF
--- a/futures-util/src/stream/select.rs
+++ b/futures-util/src/stream/select.rs
@@ -29,7 +29,7 @@ pin_project! {
 ///
 /// ```rust
 /// # futures::executor::block_on(async {
-/// use futures_util::stream::{ repeat, select, StreamExt };
+/// use futures::stream::{ repeat, select, StreamExt };
 ///
 /// let left = repeat(1);
 /// let right = repeat(2);
--- a/futures-util/src/stream/select.rs
+++ b/futures-util/src/stream/select.rs
@@ -1,5 +1,5 @@
 use super::assert_stream;
-use crate::stream::{Fuse, StreamExt};
+use crate::stream::{select_with_strategy, SelectWithStrategy, PollNext};
 use core::pin::Pin;
 use futures_core::stream::{FusedStream, Stream};
 use futures_core::task::{Context, Poll};
@@ -7,14 +7,10 @@ use pin_project_lite::pin_project;
 
 pin_project! {
     /// Stream for the [`select()`] function.
-    #[derive(Debug)]
     #[must_use = "streams do nothing unless polled"]
     pub struct Select<St1, St2> {
         #[pin]
-        stream1: Fuse<St1>,
-        #[pin]
-        stream2: Fuse<St2>,
-        flag: bool,
+        inner: SelectWithStrategy<St1, St2, fn(&mut PollNext)-> PollNext, PollNext>,
     }
 }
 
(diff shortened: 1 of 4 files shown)


futures-util/src/stream/select.rs
```diff
@@ -1,50 +1,67 @@
...
 /// streams have completed.
 ///
 /// Note that this function consumes both streams and returns a wrapped
 /// version of them.
+///
+/// ## Examples
+///
+/// ```rust
+/// # futures::executor::block_on(async {
+/// use futures_util::stream::{ repeat, select, StreamExt };
```
taiki-e, line 32: I would prefer to always refer `futures` in examples, not futures-util. (like existing examples)

futures-util/src/stream/select.rs
```diff
@@ -1,50 +1,67 @@
...
-use crate::stream::{Fuse, StreamExt};
+use crate::stream::{select_with_strategy, PollNext, SelectWithStrategy};
 use core::pin::Pin;
 use futures_core::stream::{FusedStream, Stream};
 use futures_core::task::{Context, Poll};
 use pin_project_lite::pin_project;
 
 pin_project! {
     /// Stream for the [`select()`] function.
-    #[derive(Debug)]
```
taiki-e, line 10: Why Debug impl is removed?

najamelan: ah, sorry. My bad. I intended to put it back but forgot. The reason is this:

```
error: implementation of `Debug` is not general enough
  --> futures-util/src/stream/select.rs:10:14
   |
10 |     #[derive(Debug)]
   |              ^^^^^ implementation of `Debug` is not general enough
   |
   = note: `Debug` would have to be implemented for the type `for<'r> fn(&'r mut PollNext) -> PollNext`
   = note: ...but `Debug` is actually implemented for the type `fn(&'0 mut PollNext) -> PollNext`, for some specific lifetime `'0`
```

I think we need a manual implementation, but I'm not sure what to put in it. I suppose the derive only implemented it when the generics implemented it. We can access inner through SelectWithStrategy to call their Debug I suppose.

najamelan: We can do this:
```rs
impl<St1, St2> fmt::Debug for Select<St1, St2>
where
    St1: fmt::Debug,
    St2: fmt::Debug,
{
    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
        write!( f, "Select over {:?}", self.inner.get_ref() )
    }
}
```

With the two repeat streams of the example it gives:
```
Select over (Repeat { item: 1 }, Repeat { item: 2 })
```
I should probably also add a Debug to `SelectWithStrategy`. What do you think of the above implementation?

taiki-e: Ah, I think we need to change the `Debug` impl of `SelectWithStrategy` not to print closure (like https://github.com/rust-lang/futures-rs/pull/1677 did). After that, we could re-add `#[derive(Debug)]` to `Select`.

najamelan: Ok, makes sense. The result for `Select` is:

`Select { inner: SelectWithStrategy { stream1: Fuse { stream: Repeat { item: 1 }, done: false }, stream2: Fuse { stream: Repeat { item: 2 }, done: false } } }`

futures-util/src/stream/select_with_strategy.rs
```diff
@@ -0,0 +1,228 @@
...
+impl<St1, St2, Clos, State> fmt::Debug for SelectWithStrategy<St1, St2, Clos, State>
+where
+    St1: fmt::Debug,
+    St2: fmt::Debug,
+{
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_struct("SelectWithStrategy")
+            .field("stream1", &self.stream1)
+            .field("stream2", &self.stream2)
+            .finish()
```
taiki-e, line 226: It probably makes sense to print the `state` field.

najamelan: I have added the state field. It does means implementing `Debug` depends on `State` implementing it.

taiki-e (approved):
LGTM. Thanks @najamelan!

futures-util/src/stream/select_with_strategy.rs
```diff
@@ -0,0 +1,230 @@
...
+pin_project! {
+    /// Stream for the [`select_with_strategy()`] function. See function docs for details.
+    #[must_use = "streams do nothing unless polled"]
+    pub struct SelectWithStrategy<St1, St2, Clos, State> {
+        #[pin]
+        stream1: Fuse<St1>,
+        #[pin]
+        stream2: Fuse<St2>,
+        state: State,
+        clos: Clos,
```
pickfire, line 46: Is this a typo for close?

najamelan: It's the generic representing the closure. So not a typo. Of course another name can be chosen. I don't really care about what it's called.

taiki-e:
Published in 0.3.16.

PR #2452, opened 2021-06-17
New poll_immediate functions to immediately return from a poll

### LINKED ISSUE #2257: Utility for polling a future from an async function
```rust
struct Check<'a, F: Future + ?Sized>(Pin<&'a mut F>);
impl<F: Future + ?Sized> Future for Check<'_, F> {
    type Output = Poll<F::Output>;
    fn poll(mut self: Pin<&mut Self>, cx: &mut Context) -> Poll<Self::Output> {
        Poll::Ready(self.0.as_mut().poll(cx))
    }
}
async fn checkup(tasks: Vec<Task>) {
    for mut task in tasks {
        if let Poll::Ready(val) = Check(task.fut.as_mut()).await {
            handle(val);
            cleanup(task);
        }
    }
}
```

I've seen this pattern come up in a few different projects now, and it seems generic enough for the `futures` crate. It could well be too niche though.

tylerhawkes:
I can open an issue as well if needed, but since I've already done the work I thought it would be just as easy to actually talk about the additions proposed in the pull request.

The reasons for these new functions are because I'm using a tokio::sync::mpsc::Receiver and sometimes I need to just wait until a value is ready and sometimes I need to check if one is ready and if not do some other processing. This can be accomplished by doing
```rust
    futures_util::select_biased! {
        msg = receiver.recv().await => {...}
        _ = futures_util::future::ready(()) => {...}
    }
```
but it seems like it would be nice to have a dedicated Future for this that works better with IDE's. 

I could probably use the `now_or_never()` function from `FuturesExt`, but that consumes the future and these new functions and futures allow for polling repeatedly (due to the Stream implementation) to see if they are done without blocking and they can also get the future back if it is `Unpin` which is useful if the futures are expensive to construct or if you want to just wait for them to be done at a later time.

Closes #2257

### DIFF
--- a/futures-util/src/future/poll_immediate.rs
+++ b/futures-util/src/future/poll_immediate.rs
@@ -1,0 +1,201 @@
+use super::assert_future;
+use crate::FutureExt;
+use core::pin::Pin;
+use futures_core::task::{Context, Poll};
+use futures_core::{FusedFuture, Future, Stream};
+
+/// Future for the [`poll_immediate`](poll_immediate()) function.
+#[derive(Debug, Clone)]
+#[must_use = "futures do nothing unless you `.await` or poll them"]
+pub struct PollImmediate<T>(Option<T>);
+
+impl<T, F> Future for PollImmediate<F>
+where
+    F: Future<Output = T>,
+{
+    type Output = Option<T>;
+
+    #[inline]
+    fn poll(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<T>> {
+        // # Safety
+        // This is the only time that this future will ever be polled.
+        let inner =
+            unsafe { self.get_unchecked_mut().0.take().expect("PollOnce polled after completion") };
+        crate::pin_mut!(inner);
+        match inner.poll(cx) {
+            Poll::Ready(t) => Poll::Ready(Some(t)),
+            Poll::Pending => Poll::Ready(None),
+        }
+    }
+}
+
+impl<T: Future> FusedFuture for PollImmediate<T> {
+    fn is_terminated(&self) -> bool {
+        self.0.is_none()
+    }
+}
+
+/// Creates a stream that can be polled repeatedly until the future is done
+/// ```
+/// # futures::executor::block_on(async {
+/// use futures::task::Poll;
+/// use futures::{StreamExt, future, pin_mut};
+/// use future::FusedFuture;
+///
+/// let f = async { 1_u32 };
+/// pin_mut!(f);
+/// let mut r = future::poll_immediate(f);
+/// assert_eq!(r.next().await, Some(Poll::Ready(1)));
+///
+/// let f = async {futures::pending!(); 42_u8};
+/// pin_mut!(f);
+/// let mut p = future::poll_immediate(f);
+/// assert_eq!(p.next().await, Some(Poll::Pending));
+/// assert!(!p.is_terminated());
+/// assert_eq!(p.next().await, Some(Poll::Ready(42)));
+/// assert!(p.is_terminated());
+/// assert_eq!(p.next().await, None);
+/// # });
+/// ```
+impl<T, F> Stream for PollImmediate<F>
+where
+    F: Future<Output = T>,
+{
+    type Item = Poll<T>;
+
+    fn poll_next(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
+        unsafe {
+            // # Safety
+            // We never move the inner value until it is done. We only get a reference to it.
+            let inner = &mut self.get_unchecked_mut().0;
+            let fut = match inner.as_mut() {
+                // inner is gone, so we can signal that the stream is closed.
+                None => return Poll::Ready(None),
+                Some(inner) => inner,
+            };
+            let fut = Pin::new_unchecked(fut);
+            Poll::Ready(Some(fut.poll(cx).map(|t| {
+                // # Safety
+                // The inner option value is done, so we need to drop it. We do it without moving it
+                // by using drop in place. We then write over the value without trying to drop it first
+                // This should uphold all the safety requirements of `Pin`
+                std::ptr::drop_in_place(inner);
+                std::ptr::write(inner, None);
+                t
+            })))
+        }
+    }
+}
+
+/// Creates a future that is immediately ready with an Option of a value.
+///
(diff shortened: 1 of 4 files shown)


futures-util/src/future/poll_immediate.rs
```diff
@@ -0,0 +1,201 @@
...
+            };
+            let fut = Pin::new_unchecked(fut);
+            Poll::Ready(Some(fut.poll(cx).map(|t| {
+                // # Safety
+                // The inner option value is done, so we need to drop it. We do it without moving it
+                // by using drop in place. We then write over the value without trying to drop it first
+                // This should uphold all the safety requirements of `Pin`
+                std::ptr::drop_in_place(inner);
+                std::ptr::write(inner, None);
+                t
```
taiki-e, line 84: this is unsound because `inner` may panic

futures-util/src/stream/poll_immediate.rs
```diff
@@ -0,0 +1,80 @@
...
+            let stream = Pin::new_unchecked(fut);
+            match stream.poll_next(cx) {
+                Poll::Ready(Some(t)) => Poll::Ready(Some(Poll::Ready(t))),
+                Poll::Ready(None) => {
+                    // # Safety
+                    // The inner stream is done, so we need to drop it. We do it without moving it
+                    // by using drop in place. We then write over the value without trying to drop it first
+                    // This should uphold all the safety requirements of `Pin`
+                    std::ptr::drop_in_place(inner);
+                    std::ptr::write(inner, None);
```
taiki-e, line 35: same here

futures-util/src/future/poll_immediate.rs
```diff
@@ -0,0 +1,201 @@
...
+    F: Future<Output = T>,
+{
+    type Output = Option<T>;
+
+    #[inline]
+    fn poll(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<T>> {
+        // # Safety
+        // This is the only time that this future will ever be polled.
+        let inner =
+            unsafe { self.get_unchecked_mut().0.take().expect("PollOnce polled after completion") };
```
taiki-e, line 23: this is unsound when `Future::poll` called after `fut.poll()` (in `Stream` impl) panicked.

taiki-e (changes requested):
Thanks for the PR. Almost all of unsafe code used in this PR is not correct. Please use pin-project-lite as existing code does.

futures-util/src/future/poll_immediate.rs
tylerhawkes: I found the `set()` function in the abortable code.

futures-util/src/stream/poll_immediate.rs
tylerhawkes: I found the set() function in the abortable code.

futures-util/src/future/poll_immediate.rs
tylerhawkes: I've removed all unsafe code, so hopefully this is now sound.

tylerhawkes:
@taiki-e I think all the checks should pass now.

taiki-e (commented):
Some futures expect it will be polled until complete, and such usage may cause problems. (see https://github.com/smol-rs/async-io/pull/37#discussion_r501784717)
Also, creating futures frequently and then dropping them may cause performance problems. (see https://github.com/rust-lang/futures-rs/issues/2173)

I am not strongly opposed to adding such features, but I think it would be preferable to mention these issues in the documentation.

futures-util/src/future/poll_immediate.rs
```diff
@@ -0,0 +1,198 @@
...
+/// }));
+/// match p.await {
+///     Ok(_) => unreachable!(),
+///     Err(e) => {
+///         assert_eq!(e.await, 42);
+///     }
+/// }
+/// # });
+/// ```
+pub fn poll_immediate_reuse<F: Future + Unpin>(f: F) -> PollImmediateReuse<F> {
```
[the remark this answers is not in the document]
tylerhawkes: That's fair. I'll remove it.

futures-util/src/stream/poll_immediate.rs
```diff
@@ -0,0 +1,77 @@
...
+}
+
+impl<S: Stream> super::FusedStream for PollImmediate<S> {
+    fn is_terminated(&self) -> bool {
+        self.stream.is_none()
+    }
+}
+
+/// Creates a new stream that never blocks when awaiting it. This is useful
+/// when immediacy is more important than waiting for the next item to be ready
```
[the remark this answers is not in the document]
tylerhawkes: I'll update the docs.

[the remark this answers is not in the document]
tylerhawkes: The docs have been updated. @taiki-e can you take another look?

futures-util/src/future/poll_immediate.rs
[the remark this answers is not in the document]
tylerhawkes: @taiki-e this function and the related struct have been removed.

futures-util/src/future/poll_immediate.rs
```diff
@@ -0,0 +1,126 @@
...
+
+/// Creates a future that is immediately ready with an Option of a value.
+/// Specifically this means that [poll](core::future::Future::poll()) always returns [Poll::Ready](core::task::Poll::Ready).
+///
+/// # Caution
+///
+/// Some futures expect to run until completion. If the future passed to this function isn't some sort of `&mut Future` then it will
+/// be dropped and can cause performance problems when creating and dropping futures continuously. In some cases this is fine like
+/// when asking for the [next()](crate::stream::StreamExt::next()) value of a stream and dropping the [Next](crate::stream::Next) future
+/// doesn't change any state in the stream.
```
taiki-e, line 97: It would be preferable to clarify in which cases the problem may occur. Something like:

```suggestion
/// When consuming the future by this function, note the following:
///
/// - This function does not guarantee that the future will run to completion, so it is generally incompatible with passing the non-cancellation-safe future by value.
/// - Even if the future is cancellation-safe, creating and dropping new futures frequently may lead to performance problems.
```

taiki-e (commented):
Thanks. LGTM aside from a nit.

taiki-e (approved):
Thanks!

taiki-e:
Published in 0.3.17

PR #2456, opened 2021-06-24
Remove an incorrect doctest comment

teor2345:
It looks like this comment was copy-pasted from the previous doc test.

But I'd like to double-check that with @najamelan before merging.


najamelan:
Yes, good catch it's a copy paste mistake indeed.

najamelan (approved):

taiki-e (approved):
Thanks.

Issue #2458, opened 2021-06-25
`StreamExt::All / StreamExt::Any`


barvirm:
Is there a reason why these methods aren't implemented?
Both of these can be implemented with fold, or Am'I missing something?

taiki-e:
I don't think there is any reason other than "no one has ever requested it before. 

I would accept a PR to add them.

barvirm:
[https://github.com/rust-lang/futures-rs/pull/2460](https://github.com/rust-lang/futures-rs/pull/2460)

PR #2460, opened 2021-06-26
StreamExt::All / StreamExt::Any

barvirm:
Primitive implementation of All and Any predicates over the stream. Most of the code is reused from Fold implementation.
I was thinking about reusing Fold inside All and Any, but we lose the possibility of an early exit.


barvirm:
#2458

taiki-e (approved):
Thanks!

taiki-e:
Published in 0.3.16.

PR #2469, opened 2021-07-27
Add `TryStreamExt::try_forward`, remove `TryStream` bound from `StreamExt::forward`

### LINKED ISSUE #2004: StreamExt::forward should be TryStreamExt::forward
The current implementation of `StreamExt::forward` should actually be in `TryStreamExt` as `try_forward`, I think, because it requires a fallible stream. `StreamExt` should then provide a `forward` method that does not require the sink to produce `Result`s.

In other words the fix that merged in #1946 doesn't much sense to me. The code that was there prior to #1946 should be resurrected as `StreamExt::forward`, as the code that is there now should be `TryStreamExt::try_forward`.

Or is there some edge case that I'm missing? Happy to submit a PR if what I described above seems reasonable.

/cc @cramertj

ibraheemdev:
Closes #2004


futures/tests/auto_traits.rs
```diff
@@ -1268,17 +1268,29 @@ pub mod stream {
...
+    assert_not_impl!(Forward<LocalStream, ()>: Send);
+    assert_impl!(Forward<SyncStream<()>, ()>: Sync);
+    assert_not_impl!(Forward<SyncStream, ()>: Sync);
+    assert_not_impl!(Forward<SyncStream<()>, *const ()>: Sync);
+    assert_not_impl!(Forward<LocalStream, ()>: Sync);
+    assert_impl!(Forward<UnpinStream, ()>: Unpin);
+    assert_not_impl!(Forward<UnpinStream, PhantomPinned>: Unpin);
+    assert_not_impl!(Forward<PinnedStream, ()>: Unpin);
+
+    assert_impl!(TryForward<SendTryStream<()>, ()>: Send);
```
ibraheemdev, line 1283: I'm not sure why this is failing and the previous assertion for `Forward` was not?
```
error[E0277]: `*const ()` cannot be sent between threads safely
    --> futures/tests/auto_traits.rs:1283:5
     |
1283 |     assert_impl!(TryForward<SendTryStream<()>, ()>: Send);
     |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
     |     |
     |     `*const ()` cannot be sent between threads safely
     |     required by this bound in `stream::_::{closure#0}::assert_impl_all`
     |
     = help: within `futures::stream::TryForward<Pin<Box<(dyn futures::Stream<Item = Result<(), *const ()>> + std::marker::Send + 'static)>>, ()>`, the trait `std::marker::Send` is not implemented for `*const ()`
     = note: required because it appears within the type `Result<(), *const ()>`
     = note: required because it appears within the type `Option<Result<(), *const ()>>`
     = note: required because it appears within the type `futures::stream::try_stream::try_forward::TryForward<Pin<Box<(dyn futures::Stream<Item = Result<(), *const ()>> + std::marker::Send + 'static)>>, (), Result<(), *const ()>>`
     = note: required because it appears within the type `futures::stream::TryForward<Pin<Box<(dyn futures::Stream<Item = Result<(), *const ()>> + std::marker::Send + 'static)>>, ()>`
     = note: this error originates in the macro `assert_impl` (in Nightly builds, run with -Z macro-backtrace for more info)

error[E0277]: `*const ()` cannot be shared between threads safely
    --> futures/tests/auto_traits.rs:1287:5
     |
1287 |     assert_impl!(TryForward<SyncTryStream<()>, ()>: Sync);
     |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
     |     |
     |     `*const ()` cannot be shared between threads safely
     |     required by this bound in `stream::_::{closure#0}::assert_impl_all`
     |
     = help: within `futures::stream::TryForward<Pin<Box<(dyn futures::Stream<Item = Result<(), *const ()>> + Sync + 'static)>>, ()>`, the trait `Sync` is not implemented for `*const ()`
     = note: required because it appears within the type `Result<(), *const ()>`
     = note: required because it appears within the type `Option<Result<(), *const ()>>`
     = note: required because it appears within the type `futures::stream::try_stream::try_forward::TryForward<Pin<Box<(dyn futures::Stream<Item = Result<(), *const ()>> + Sync + 'static)>>, (), Result<(), *const ()>>`
     = note: required because it appears within the type `futures::stream::TryForward<Pin<Box<(dyn futures::Stream<Item = Result<(), *const ()>> + Sync + 'static)>>, ()>`
     = note: this error originates in the macro `assert_impl` (in Nightly builds, run with -Z macro-backtrace for more info)
```

futures-util/src/stream/try_stream/mod.rs
```diff
@@ -76,6 +78,18 @@ mod try_filter;
...
 pub use self::try_filter::TryFilter;
 
+mod try_forward;
+
+#[cfg(feature = "sink")]
+delegate_all!(
+    /// Future for the [`try_forward`](super::TryStreamExt::try_forward) method.
+    #[cfg_attr(docsrs, doc(cfg(feature = "sink")))]
+    TryForward<St, Si>(
+        try_forward::TryForward<St, Si, St::Item>
```
yaymukund, line 88: This is the source of the compile error.
```suggestion
        try_forward::TryForward<St, Si, St::Ok>
```

yaymukund (changes requested):
I was able to get tests to pass by applying the inline suggested change and then fixing the failing test like:
```diff
diff --git futures-util/src/io/mod.rs futures-util/src/io/mod.rs
index b96223d1..00231fa1 100644
--- futures-util/src/io/mod.rs
+++ futures-util/src/io/mod.rs
@@ -570,13 +570,13 @@ pub trait AsyncWriteExt: AsyncWrite {
     /// ```
     /// # futures::executor::block_on(async {
     /// use futures::io::AsyncWriteExt;
-    /// use futures::stream::{self, StreamExt};
+    /// use futures::stream::{self, TryStreamExt};
     ///
     /// let stream = stream::iter(vec![Ok([1, 2, 3]), Ok([4, 5, 6])]);
     ///
     /// let mut writer = vec![];
     ///
-    /// stream.forward((&mut writer).into_sink()).await?;
+    /// stream.try_forward((&mut writer).into_sink()).await?;
     ///
     /// assert_eq!(writer, vec![1, 2, 3, 4, 5, 6]);
     /// # Ok::<(), Box<dyn std::error::Error>>(())
```

futures-util/src/stream/try_stream/mod.rs
ibraheemdev: Aha, I left that from the previous changes.

futures-util/src/stream/try_stream/mod.rs
```diff
@@ -76,6 +78,18 @@ mod try_filter;
 #[allow(unreachable_pub)] // https://github.com/rust-lang/rust/issues/57411
 pub use self::try_filter::TryFilter;
 
+mod try_forward;
```
yaymukund, line 81: I think this should fix the CI tests.
```suggestion
#[cfg(feature = "sink")]
mod try_forward;
```

yaymukund (approved):
Cool! I'm no maintainer or anything, but this looks good to me apart from packaging considerations (i.e. I think this is a breaking change).

futures-util/src/stream/try_stream/mod.rs
```diff
@@ -287,6 +302,27 @@ pub trait TryStreamExt: TryStream {
...
+    /// order to preserve access to the `Sink`. If the stream produces an error,
+    /// that error will be returned by this future without flushing/closing the sink.
+    #[cfg(feature = "sink")]
+    #[cfg_attr(docsrs, doc(cfg(feature = "sink")))]
+    fn try_forward<S>(self, sink: S) -> TryForward<Self, S>
+    where
+        S: Sink<Self::Ok, Error = Self::Error>,
+        Self: Sized,
+    {
+        assert_future::<Result<(), S::Error>, _>(TryForward::new(self, sink))
```
yaymukund, line 323:
```suggestion
        Self: Sized + Stream<Item = Result<Self::Ok, Self::Error>>,
    {
        assert_future::<Result<(), Self::Error>, _>(TryForward::new(self, sink))
```

You'll also need to `use crate::stream::Stream` at the top & fix a rebase conflict.

ibraheemdev:
Hmmm..:
```rust
 error[E0391]: cycle detected when computing the bounds for type parameter `Self`
   --> futures-util/src/stream/try_stream/mod.rs:322:44
    |
322 |         Self: Sized + Stream<Item = Result<Self::Ok, Self::Error>>,
    |                                            ^^^^^^^^
    |
    = note: ...which again requires computing the bounds for type parameter `Self`, completing the cycle
note: cycle used when processing `stream::try_stream::TryStreamExt::try_forward`
   --> futures-util/src/stream/try_stream/mod.rs:321:17
    |
321 |         S: Sink<Self::Ok, Error = Self::Error>,
    |                 ^^^^^^^^
```

The extra `Stream` bound is weird anyways, are you sure that's the correct fix?

taiki-e:
Maybe it can be avoided by removing the `Stream` bounds from the `TryStreamExt::try_forward` and always using the `TryStream` bounds instead of the `Stream` bounds in `try_forword.rs`.

ibraheemdev:
Aha, I think that's it.

taiki-e (approved):
Thanks!

PR #2471, opened 2021-08-05
Implement `Default` trait for `OptionFuture`

EFanZh:
This provides an easy way to create empty `OptionFuture`s.


taiki-e (approved):
Thanks!

PR #2479, opened 2021-08-20
Prefer doc(no_inline) to doc(hidden) for reexports

### LINKED ISSUE #2472: `doc(hidden)` reexports
`futures` and `futures-core` reexport a bunch of things with `#[doc(hidden)]`. rust-analyzer will in the future hide `doc(hidden)` items from completions and autoimport (it already does so if the item itself is hidden, it's just not handling hidden reexports yet), so if these reexport are actually intended to be used, they probably shouldn't be hidden; they might instead need `doc(no_inline)` or just no annotation.

CC https://github.com/rust-analyzer/rust-analyzer/issues/7718 (see the discussion at the end) and https://github.com/rust-analyzer/rust-analyzer/issues/9197.

nipunn1313:
Fixes #2472

### DIFF
--- a/futures-core/src/lib.rs
+++ b/futures-core/src/lib.rs
@@ -16,11 +16,11 @@
 extern crate alloc;
 
 pub mod future;
-#[doc(hidden)]
+#[doc(no_inline)]
 pub use self::future::{FusedFuture, Future, TryFuture};
 
 pub mod stream;
-#[doc(hidden)]
+#[doc(no_inline)]
 pub use self::stream::{FusedStream, Stream, TryStream};
 
 #[macro_use]
--- a/futures-util/src/lib.rs
+++ b/futures-util/src/lib.rs
@@ -301,18 +301,18 @@ macro_rules! delegate_all {
 }
 
 pub mod future;
-#[doc(hidden)]
+#[doc(no_inline)]
 pub use crate::future::{Future, FutureExt, TryFuture, TryFutureExt};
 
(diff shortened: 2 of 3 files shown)


nipunn1313:
The `cargo clippy` failure appears to be a regression introduced in nightly 2021-08-17. Afaict it's a false positive from clippy.

The failure of cargo +1.41.0 is interesting since it's clearly ok on some stables around it. It's also clearly a non-regression (claiming that things aren't documented there). What do you suggest we do here to make CI happy? Could conditionally ignore the warning with something like https://github.com/dtolnay/rustversion - or unconditionally ignore the warning.

taiki-e (changes requested):
Thanks! Could you change to use `no_inline` only on trait reexports? `*_internal` macros and `async_await` module are not public APIs.

taiki-e:
Could you do the same for the trait reexports in `futures`?

https://github.com/rust-lang/futures-rs/blob/fdb96f5d86bbb66e0908fbb9ad8002daca37a50d/futures/src/lib.rs#L105-L125

nipunn1313:
yep - thanks for catching! I missed that file in my audit! (I had used `vim -p $(rg "doc.*hidden")` which capped to an upper bound number of files by mistake). I double checked and confirmed I covered everything after hitting up futures/src/lib.rs

Thanks!

taiki-e (approved):
Thanks!

PR #2481, opened 2021-08-20
Add vim swapfiles to .gitignore

nipunn1313:


taiki-e (commented):
Thanks for the PR! For editor-specific settings, it is recommended to add to a global gitignore file (`~/.config/git/ignore`).
See also: https://docs.github.com/en/github/using-git/ignoring-files#configuring-ignored-files-for-all-repositories-on-your-computer

PR #2484, opened 2021-08-23
Correcting overly restrictive lifetimes in vectored IO

### LINKED ISSUE #2483: Lifetime issues around write_vectored et al
(I first spotted this while trying to use `write_all_vectored` within `AsyncWriteExt`, but the same applies elsewhere in that trait.)

The signature of fn `write_all_vectored` is, 
```
fn write_all_vectored<'a>(&'a mut self, bufs: &'a mut [IoSlice<'a>]) -> WriteAllVectored<'a, Self>
```

I don't understand this choice of lifetimes, particularly the `&'a Something<'a>` construction, which I have gathered is suggestive of a self-refential struct and should be avoided. When I tried to charge ahead and use the function anyway, I got very confusing lifetime errors about my structure being borrowed at the point where it's dropped. I asked at URLO, and they suggested to report this as a bug. 

To understand more, I tried re-implementing it while making all three lifetimes independent (ala [this](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=c4272b1a0a92225224c080cb5ccb400a)) but this appeared to work exactly as intended, which just left me further mystified. Does making all the lifetimes the same serve a purpose I've missed, or is this truly a bug?

If this is a bug and the playground snippet I've posted is the correct resolution, I'd be quite happy to raise a PR for it and corresponding changes to other parts of `AsyncWriteExt`, I just wanted to check that this was truly a mistake first.

Additionally, for `write_all_vectored` in particular, the documentation suggests that `bufs` is unusable and in an undefined state after the function completes. I similarly cannot see any reason why it was not consumed if that's the case.

ejmount:
The current input lifetimes for `write_vectored` and `write_all_vectored` in `AsyncWriteExt` and `read_vectored` in `AsyncReadExt` are unnecessarily restrictive and disallow otherwise reasonable uses. This change fixes #2483 by relaxing that restriction, making the lifetime of the incoming references independent of the inner lifetime that the `IoSlice`s refer to. (They could be relaxed further, but I can't see any practical difference it would make so I have tried to be consistent with the surrounding code)

Please let me know if you'd like any other changes made for this issue, or any other comments you have.

(Edit: Apologies for any log/notification noise the repeated CI runs generated, some unrelated tests wouldn't run cleanly on my local machine)

### DIFF
--- a/futures-util/src/io/write_all_vectored.rs
+++ b/futures-util/src/io/write_all_vectored.rs
@@ -193,6 +193,11 @@ mod tests {
 
     #[test]
     fn test_write_all_vectored_distinct_lifetimes() {
+        struct WrapVec<T>(Vec<T>);
+        impl<T> Drop for WrapVec<T> {
+            fn drop(&mut self) {}
+        }
+
         let waker = noop_waker();
         let mut cx = Context::from_waker(&waker);
         let mut dst = test_writer(2, 2);
@@ -201,8 +206,8 @@ mod tests {
             // Force the lifetimes of the underlying data and IOSlice to be unequal
             let data = vec![1, 2, 3, 4];
             {
-                let mut slices: Vec<_> = data.chunks(2).map(IoSlice::new).collect();
-                let mut future = dst.write_all_vectored(&mut *slices);
+                let mut slices = WrapVec(data.chunks(2).map(IoSlice::new).collect());
+                let mut future = dst.write_all_vectored(&mut *slices.0);
                 match Pin::new(&mut future).poll(&mut cx) {
                     Poll::Ready(Ok(())) => {}
                     other => panic!("unexpected result polling future: {:?}", other),
--- a/futures-util/src/io/write_all_vectored.rs
+++ b/futures-util/src/io/write_all_vectored.rs
@@ -190,4 +190,25 @@ mod tests {
             assert_eq!(&*dst.written, &*wanted);
         }
     }
+
+    #[test]
+    fn test_write_all_vectored_distinct_lifetimes() {
+        let waker = noop_waker();
+        let mut cx = Context::from_waker(&waker);
+        let mut dst = test_writer(2, 2);
+
+        {
+            // Force the lifetimes of the underlying data and IOSlice to be unequal
+            let data = vec![1, 2, 3, 4];
+            {
+                let mut slices: Vec<_> = data.chunks(2).map(IoSlice::new).collect();
+                let mut future = dst.write_all_vectored(&mut *slices);
+                match Pin::new(&mut future).poll(&mut cx) {
+                    Poll::Ready(Ok(())) => {}
+                    other => panic!("unexpected result polling future: {:?}", other),
+                }
+            }
+            assert_eq!(&*dst.written, &*data);
+        }
+    }
 }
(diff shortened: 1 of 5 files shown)


taiki-e:
Thanks! Could you add a test to prevent this change from being accidentally reverted in the future?

futures-util/src/io/write_all_vectored.rs
```diff
@@ -10,21 +10,21 @@ use std::pin::Pin;
...
-    pub(super) fn new(writer: &'a mut W, mut bufs: &'a mut [IoSlice<'a>]) -> Self {
+impl<'a, 'b, W: AsyncWrite + ?Sized + Unpin> WriteAllVectored<'a, 'b, W> {
+    pub(super) fn new(writer: &'a mut W, mut bufs: &'a mut [IoSlice<'b>]) -> Self {
         IoSlice::advance_slices(&mut bufs, 0);
         Self { writer, bufs }
     }
 }
 
-impl<W: AsyncWrite + ?Sized + Unpin> Future for WriteAllVectored<'_, W> {
+impl<W: AsyncWrite + ?Sized + Unpin> Future for WriteAllVectored<'_, '_, W> {
```
ibraheemdev, line 27: Isn't this `impl<'a, W> ... WriteAllVectored<'a, 'a, W> {}`?

ejmount: Would you be able to elaborate on why you think so? AIUI `Future` should be valid for any choice of lifetimes, and it doesn't need to refer to them by name.

ibraheemdev: Nvm, was mistaken.

ejmount:
I've added a corresponding test to the existing ones for `write_all_vectored`, but couldn't see any that use the corresponding interface with `read_` / `write_vectored` - the tests for those structures appear to directly call `poll_*` rather than the methods on `Async*Ext`,  and the `poll_*` implementations are already correct. Let me know if you think testing that "gap" in the interface is worth creating an entirely new harness for, though.

futures-util/src/io/write_all_vectored.rs
```diff
@@ -190,4 +190,25 @@ mod tests {
...
+        let waker = noop_waker();
+        let mut cx = Context::from_waker(&waker);
+        let mut dst = test_writer(2, 2);
+
+        {
+            // Force the lifetimes of the underlying data and IOSlice to be unequal
+            let data = vec![1, 2, 3, 4];
+            {
+                let mut slices: Vec<_> = data.chunks(2).map(IoSlice::new).collect();
+                let mut future = dst.write_all_vectored(&mut *slices);
```
taiki-e, line 205: Currently, this test seems to work fine without this patch.


```suggestion
        struct WrapVec<T>(Vec<T>);
        impl<T> Drop for WrapVec<T> {
            fn drop(&mut self) {}
        }

        let waker = noop_waker();
        let mut cx = Context::from_waker(&waker);
        let mut dst = test_writer(2, 2);

        {
            // Force the lifetimes of the underlying data and IOSlice to be unequal
            let data = vec![1, 2, 3, 4];
            {
                let mut slices = WrapVec(data.chunks(2).map(IoSlice::new).collect());
                let mut future = dst.write_all_vectored(&mut *slices.0);
```

ejmount: You are quite right it does, oops. I don't quite follow why your revision correctly fails in the original version where mine works, but I've accepted it just now anyway

taiki-e (approved):
Thanks!

PR #2489, opened 2021-08-28
Add BufReader::seek_relative

taiki-e:
Equivalents to std::io::BufReader::seek_relative that stabilized in Rust 1.53.

refs: https://doc.rust-lang.org/nightly/std/io/struct.BufReader.html#method.seek_relative

### DIFF
--- a/futures-util/src/io/buf_reader.rs
+++ b/futures-util/src/io/buf_reader.rs
@@ -1,4 +1,5 @@
 use super::DEFAULT_BUF_SIZE;
+use futures_core::future::Future;
 use futures_core::ready;
 use futures_core::task::{Context, Poll};
 #[cfg(feature = "read-initializer")]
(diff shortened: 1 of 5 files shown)


futures-util/src/io/buf_reader.rs
```diff
@@ -73,6 +73,39 @@ impl<R: AsyncRead> BufReader<R> {
...
+    /// This method does not return the location of the underlying reader, so the caller
+    /// must track this information themselves if it is required.
+    pub fn seek_relative(&mut self, offset: i64) -> SeeKRelative<'_, R>
+    where
+        R: Unpin,
+    {
+        SeeKRelative { inner: self, offset, first: true }
+    }
+
+    fn poll_seek_relative(
```
taiki-e, line 88:
```suggestion
    /// Attempts to seek relative to the current position. If the new position lies within the buffer,
    /// the buffer will not be flushed, allowing for more efficient seeks.
    /// This method does not return the location of the underlying reader, so the caller
    /// must track this information themselves if it is required.
    pub fn poll_seek_relative(
```

futures-util/src/io/buf_reader.rs
```diff
@@ -163,6 +196,9 @@ impl<R: AsyncRead + AsyncSeek> AsyncSeek for BufReader<R> {
     /// `.into_inner()` immediately after a seek yields the underlying reader
     /// at the same position.
     ///
+    /// To seek without discarding the internal buffer, use
+    /// [`BufReader::poll_seek_relative`](BufReader::poll_seek_relative).
```
taiki-e, line 200:
```suggestion
    /// To seek without discarding the internal buffer, use
    /// [`BufReader::seek_relative`](BufReader::seek_relative) or
    /// [`BufReader::poll_seek_relative`](BufReader::poll_seek_relative).
```

futures-util/src/io/buf_reader.rs
```diff
@@ -73,6 +74,43 @@ impl<R: AsyncRead> BufReader<R> {
     }
 }
 
+impl<R: AsyncRead + AsyncSeek> BufReader<R> {
+    /// Seeks relative to the current position. If the new position lies within the buffer,
+    /// the buffer will not be flushed, allowing for more efficient seeks.
+    /// This method does not return the location of the underlying reader, so the caller
+    /// must track this information themselves if it is required.
+    pub fn seek_relative(&mut self, offset: i64) -> SeeKRelative<'_, R>
```
taiki-e, line 82: hmm, should this be `self: Pin<&mut Self>`?

taiki-e: Oh, I cannot call seek_relative after creating BufReader if `R: !Unpin` and seek_relative needs `R: Unpin`.

PR #2493, opened 2021-09-02
Remove implicit clear in `ReadyToRunQueue::drop`

### LINKED ISSUE #2473: Safety risk with reusing ReadyToRunQueue
In Fuchsia, we performed an [audit of Futures 0.3.15](https://fuchsia-review.googlesource.com/c/fuchsia/+/567474), and we noticed that [ReadyToRunQueue::drop](https://github.com/rust-lang/futures-rs/blob/0c4d6a38f3929ff9e1a7c56f655173262197d736/futures-util/src/stream/futures_unordered/ready_to_run_queue.rs#L111) is not inherently safe, and instead relies on `FuturesUnordered::drop` to execute first.

As best as we can tell, this appears to be safe, but there's a risk that if `ReadyToRunQueue` is ever used elsewhere, it would be pretty easy to miss this safety constraint. Is there a way we could protect against this? At a minimum, I think it might be worth adding a `// Safety: ...` comment to the struct definition describing this requirement, but perhaps there's a way to refactor the code to avoid this risk all together.

cc @anp and @ibraheemdev

ibraheemdev:
Resolves #2473

### DIFF
--- a/futures-util/src/stream/futures_unordered/mod.rs
+++ b/futures-util/src/stream/futures_unordered/mod.rs
@@ -558,7 +558,7 @@ impl<Fut> FuturesUnordered<Fut> {
     pub fn clear(&mut self) {
         self.clear_head_all();
 
-        // we just cleared all the tasks, and we have &mut self, so this is safe.
+        // SAFETY: we just cleared all the tasks and we have &mut self
         unsafe { self.ready_to_run_queue.clear() };
 
         self.is_terminated.store(false, Relaxed);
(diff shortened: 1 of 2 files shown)


taiki-e:
Thanks! Could you also update safety docs on ReadyToRunQueue::clear? (FuturesUnordered::clear -> FuturesUnordered::clear_head_all)

https://github.com/rust-lang/futures-rs/blob/cc7e98e0fe96ceb5410dce7aa25a29e314aa949d/futures-util/src/stream/futures_unordered/ready_to_run_queue.rs#L95-L97

taiki-e (approved):

PR #2494, opened 2021-09-03
reexport `alloc::task::Wake`

ibraheemdev:
Per #2207.


ibraheemdev:
Should it be re-exported as `ArcWake` to avoid breakage?

taiki-e:
> Should it be re-exported as `ArcWake` to avoid breakage?

Since the default methods for Wake and ArcWake are different, this is breaking change anyway, and the re-export as ArcWake is not needed.

PR #2495, opened 2021-09-08
futures-util: add `StreamExt::count` method

petrosagg:
Works the same way as its `Iterator::count` counterpart


taiki-e (approved):
Thanks!

taiki-e:
Published in 0.3.19.

PR #2497, opened 2021-09-15
Introduce an abortable mechanism into futures_util::io::copy & copy_buf

kazuki0824:
# Summary
When transferring data from one place to another using the copy_buf function, you may want to abort the operation in the middle of the transfer. Since the transfer is divided into small increments by internal buffers and probably must not be done at a time, the interruption logic can be implemented by checking for the interruption request at every poll_fill_buf().

# This patch has

- a bunch of modification that introduces a new behaviour, similar to what futures::abortable::Abortable does, into CopyBuf<T>.
- changes of return values coming from copy_buf(...) & copy(...). 
- changes of visibility of some Abortable-related private members, such as AbortInner and its children.

# How to use
Unlike before, copy() & copy_buf() returns the combination of Future and AbortHandle. If you want to abort the operation, you should extract the tuple and trigger the inner AbortHandle.

# Backward compatibility
If you want to fit the existing code to this change, all you have to do is simply adding ".0" after every copy_buf() or copy().


taiki-e:
Thanks for the PR!

- copy is a very common function, and most of its use cases do not require this feature. If you want to add this feature, I would prefer to add it as a separate function.

- Is it possible to do this same thing by implementing AsyncRead/AsyncWrite for Abortable? If so, that would be more generic and preferable.

PR #2499, opened 2021-09-17
Fix Unusable Sink Implementation on Scan

drewkett:
The `Sink` implementation for `futures::stream::Scan` has generic bounds on it that require the stream and state types on `Scan` to be the same type, which naturally makes it not useful to use the implementation. This change adds an extra generic to the `Sink` implementation to separate out the type of the state argument in `Scan`.


taiki-e (approved):
Thanks!

taiki-e:
Published in 0.3.18.

PR #2501, opened 2021-09-24
add repeat_then function

andrewgazelka:
# Problem
Currently, there is no way (as far as I know) to create an infinite `Stream` which re-runs a future generator dependent on a mutable reference. The following PR creates a `repeat_then` function, which functions similar to the combination of `repeat` and `then`. 

# Solution
`repeat_then` functions differently, in that it allows using a mutable reference in the future generator. It does so by taking ownership of the to-be-referenced variable, and for each Future generated providing a mutable reference to `input` it can be built by. The invariants of the `RepeatThen` stream assure that only one mutable reference will be provided for each active `Future`, which is something Rust cannot verify by chaining a `repeat` and `then`.

# Use case
```rust
#[derive(Default)]
struct State {
    inner: i32
}
                                                                                             
impl State {
    async fn increment_and_return(&mut self) -> i32 {
        self.inner+=1;
        self.inner
    }
}
                                                                                             
{
    // consuming state
    let stream = stream::repeat_then(State::default(), |state| state.increment_and_return());
    assert_eq!(vec![1, 2, 3, 4, 5], stream.take(5).collect::<Vec<i32>>().await);
}
{
    // mut reference to state (can use after done with the stream)
    let mut state = State::default();
    let stream = stream::repeat_then(&mut state, |state| state.increment_and_return());
    assert_eq!(vec![1, 2, 3, 4, 5], stream.take(5).collect::<Vec<i32>>().await);
    assert_eq!(5, state.inner);
    assert_eq!(6, state.increment_and_return().await)
}                                                                               
});
```

# Todo
Although I have tried to verify the validity of the code, I am not sure it is completely sound. Can anyone help me out on this? Should I add more doctests?

### DIFF
--- a/futures-util/src/stream/mod.rs
+++ b/futures-util/src/stream/mod.rs
@@ -105,6 +105,9 @@ pub use self::select_with_strategy::{select_with_strategy, PollNext, SelectWithS
 mod unfold;
 pub use self::unfold::{unfold, Unfold};
 
+mod repeat_then;
+pub use self::repeat_then::{repeat_then, RepeatThen};
+
 #[cfg(not(futures_no_atomic_cas))]
 #[cfg(feature = "alloc")]
 mod futures_ordered;
--- a/futures-util/src/stream/repeat_then.rs
+++ b/futures-util/src/stream/repeat_then.rs
@@ -1,0 +1,132 @@
+use core::marker::PhantomData;
+use core::pin::Pin;
+
+use crate::stream::assert_stream;
+use crate::task::{Context, Poll};
+use crate::{Future, Stream};
+
+/// Stream for the [`repeat_then`] function.
+#[derive(Debug)]
+pub struct RepeatThen<'a, OUT, INPUT, FUT: Future<Output = OUT>, GEN: Fn(&'a mut INPUT) -> FUT> {
+    /// The generator function for a future
+    gen: GEN,
+
+    /// fut references input, so this [`RepeatThen`] is ![`Unpin`]. Since also could contain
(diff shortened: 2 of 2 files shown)


khuey:
Why can't you use `stream::unfold`?

e.g.
```rust
        #[derive(Default)]
        struct State {
            inner: i32
        }
                                                                                                     
        impl State {
            async fn increment_and_return(mut self) -> (i32, Self) {
                self.inner+=1;
                let i = self.inner;
                (i, self)
            }
            
            async fn increment_and_return_ref(&mut self) -> i32 {
                self.inner+=1;
                self.inner
            }
        }
                                                                                                     
        {
            // consuming state
            let stream = stream::unfold(State::default(), |state| async move { Some(state.increment_and_return().await) });
            assert_eq!(vec![1, 2, 3, 4, 5], stream.take(5).collect::<Vec<i32>>().await);
        }
        {
            // mut reference to state (can use after done with the stream)
            let mut state = State::default();
            let stream = stream::unfold(&mut state, |state| async move { Some((state.increment_and_return_ref().await, state)) });
            assert_eq!(vec![1, 2, 3, 4, 5], stream.take(5).collect::<Vec<i32>>().await);
            assert_eq!(5, state.inner);
            assert_eq!(6, state.increment_and_return_ref().await)
        }                                                                               
    });
```

andrewgazelka:
This probably works. Let me test it out.

andrewgazelka:
Yea it looks like it works in all the scenarios I tested. Thanks!

PR #2505, opened 2021-10-08
Make `futures::task::noop_waker_ref` available without `std`.

kpreid:
#2332 removed the `std` dependency and some `#[cfg(feature = "std")]`, but missed the one on the reexport from `futures_util`. This PR removes that cfg attribute, so `noop_waker_ref` should be equally available from all places it is reexported. I searched for other mentions of `noop_waker_ref` in this repository and did not find any other changes needed.


taiki-e:
Published in 0.3.18.

PR #2507, opened 2021-10-11
Create copy_buf_abortable, which enables to stop copying in the middle.

kazuki0824:
See #2497 .

# Summary
When transferring data from one place to another using the copy_buf function, you may want to abort the operation in the middle of the transfer, for example, by SIGINT. Since the transfer is divided into small increments by internal buffers and probably must not be done at a time, the interruption logic can be implemented by checking for the aborting request flag at every poll_fill_buf().

# This patch has

- a new func and struct that is like CopyBuf<T> and has a new behaviour, similar to what futures::abortable::Abortable does.
- changes of visibility of some Abortable-related private members, such as AbortInner and its children.

### DIFF
--- a/futures-util/src/io/copy_buf_abortable.rs
+++ b/futures-util/src/io/copy_buf_abortable.rs
@@ -42,7 +42,7 @@ use std::sync::Arc;
 ///         Err::<usize, Aborted>(a)
 ///     }
 /// }
-///
+/// #  }).unwrap();
 /// ```
 pub fn copy_buf_abortable<R, W>(
     reader: R,
--- a/futures-util/src/io/copy_buf_abortable.rs
+++ b/futures-util/src/io/copy_buf_abortable.rs
@@ -110,7 +110,7 @@ where
         }
         // Schedule the task to be woken up again.
         // Never called unless Poll::Pending is returned from io objects.
-        cx.waker().wake_by_ref();
+        self.inner.waker.register(cx.waker());
         Poll::Pending
     }
 }
--- a/futures-util/src/io/copy_buf_abortable.rs
+++ b/futures-util/src/io/copy_buf_abortable.rs
@@ -1,0 +1,108 @@
+use crate::abortable::{AbortHandle, AbortInner, Aborted};
+use futures_core::future::Future;
+use futures_core::ready;
+use futures_core::task::{Context, Poll};
+use futures_io::{AsyncBufRead, AsyncWrite};
+use pin_project_lite::pin_project;
+use std::io;
+use std::pin::Pin;
+use std::sync::atomic::Ordering;
+use std::sync::Arc;
+
+/// Creates a future which copies all the bytes from one object to another, with its `AbortHandle`.
+///
+/// The returned future will copy all the bytes read from this `AsyncBufRead` into the
+/// `writer` specified. This future will only complete once abort has been requested or the `reader` has hit
+/// EOF and all bytes have been written to and flushed from the `writer`
+/// provided.
+///
+/// On success the number of bytes is returned. If aborted, `Aborted` is returned. Otherwise, the underlying error is returned.
+///
+/// # Examples
+///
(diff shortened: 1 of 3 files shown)


futures-util/src/io/copy_buf_abortable.rs
```diff
@@ -0,0 +1,109 @@
...
+
+            let i = ready!(Pin::new(&mut this.writer).poll_write(cx, buffer))?;
+            if i == 0 {
+                return Poll::Ready(Err(io::ErrorKind::WriteZero.into()));
+            }
+            *this.amt += i as u64;
+            this.reader.as_mut().consume(i);
+
+            // Register to receive a wakeup if the task is aborted in the future
+            this.inner.waker.register(cx.waker());
```
taiki-e, line 102: I think this is actually needed in the case where `poll_*` returns `Pending`. (See try_poll in `abortable.rs`)

futures-util/src/io/copy_buf_abortable.rs
```diff
@@ -0,0 +1,109 @@
...
+
+            if this.inner.aborted.load(Ordering::Relaxed) {
+                return Poll::Ready(Ok(Err(Aborted)));
+            }
```
taiki-e, line 106: This is not needed because the same check is done at the beginning of the next loop.

futures-util/src/io/copy_buf_abortable.rs
kazuki0824: I think your point is exactly right. I think that instead of wakeup when the task is suspended, we need to call cx.wake_by_ref() when Pending is returned. Thus, I'll remove this wakeup.

futures-util/src/io/copy_buf_abortable.rs
kazuki0824: Your point is right. I'll remove it.

futures-util/src/io/copy_buf_abortable.rs
```diff
@@ -0,0 +1,116 @@
...
+            if i == 0 {
+                return Poll::Ready(Err(io::ErrorKind::WriteZero.into()));
+            }
+            *this.amt += i as u64;
+            this.reader.as_mut().consume(i);
+        }
+        // Schedule the task to be woken up again.
+        // Never called unless Poll::Pending is returned from io objects.
+        cx.waker().wake_by_ref();
+        Poll::Pending
```
taiki-e, line 114: wake+Pending usually means  ["yields execution back to the executor"](https://docs.rs/tokio/latest/tokio/task/fn.yield_now.html), but I think what we want to do here is to register the waker in the register. Like future/stream abortable does:

https://github.com/rust-lang/futures-rs/blob/b48eb2e9a9485ef7388edc2f177094a27e08e28b/futures-util/src/abortable.rs#L137-L138

kazuki0824:
@taiki-e Sorry for the late response. I saw through your review and added another change in this PR.

futures-util/src/io/copy_buf_abortable.rs
```diff
@@ -0,0 +1,116 @@
...
+            if i == 0 {
+                return Poll::Ready(Err(io::ErrorKind::WriteZero.into()));
+            }
+            *this.amt += i as u64;
+            this.reader.as_mut().consume(i);
+        }
+        // Schedule the task to be woken up again.
+        // Never called unless Poll::Pending is returned from io objects.
+        this.inner.waker.register(cx.waker());
+        Poll::Pending
```
taiki-e, line 114: Could you add the check after registration? like abortable does:


https://github.com/rust-lang/futures-rs/blob/b48eb2e9a9485ef7388edc2f177094a27e08e28b/futures-util/src/abortable.rs#L140-L146

taiki-e (commented):
Thanks, looks good to me aside from a nit.

taiki-e (approved):

PR #2513, opened 2021-11-03
add repeat_then function

andrewgazelka:
I am "re-opening" (cannot _actually_ do this because of a force push) this because I ran into a situation where `stream::unfold` is not viable. 
- I am using `async/await` on an embedded system
- the `State` struct happens to be giant (it holds buffers). 
- I am returning a stream from a function that will, in turn, be stored in a struct. It is much easier to take ownership of `State` rather than having to keep a reference to it.
- Because I can't pass in a reference, `stream::unfold` will need to be a memcopy each time the future returns
  - Furthermore, the total memory needed is way higher, because space needs to be reserved for `State`—and its memcopy.
  - This is also quite inefficient.

Here is a concrete example. When I used `stream::unfold`, the stack on the chip I am using overflows.
image

See previous discussion in #2501

### DIFF
--- a/futures-util/src/stream/mod.rs
+++ b/futures-util/src/stream/mod.rs
@@ -105,6 +105,9 @@ pub use self::select_with_strategy::{select_with_strategy, PollNext, SelectWithS
 mod unfold;
 pub use self::unfold::{unfold, Unfold};
 
+mod repeat_then;
+pub use self::repeat_then::{repeat_then, RepeatThen};
+
 #[cfg(not(futures_no_atomic_cas))]
 #[cfg(feature = "alloc")]
 mod futures_ordered;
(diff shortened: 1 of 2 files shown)


taiki-e:
This implementation is unsound as lifetime `'a` is unbounded.

```rust

#[test]
fn test() {
        let a = vec![];
        let mut stream = Box::pin(stream::repeat_then(a, |c: &mut Vec<()>| async { c }));
        let mut_ref1 = stream.as_mut().next().await.unwrap(); // first mutable reference to a
        let mut_ref2 = stream.as_mut().next().await.unwrap(); // second mutable reference to a
        mut_ref1.push(());
        assert_eq!(mut_ref2.len(), 1);
    });
}
```

andrewgazelka:
Ah, I didn't factor this in. @taiki-e  do you think it would be worth making a function `unfold_ref` that could be called like
```rust
let elem: T =  todo!()
unfold_ref(elem, |elem_ref: &mut T| async move {
   Some((..., elem_ref))
}
```
or can this be achieved easily using stream combinators?

PR #2540, opened 2021-12-23
Improve example for TryStreamExt::try_flatten.

kevincox:
The previous example was confusing because it didn't show what happened to `Err(4)`. I also added some more examples to show that entries after `Err(_)` are preserved.


taiki-e (approved):
Thanks!

PR #2566, opened 2022-02-06
`FlattenUnordered`: improve wakers behavior

olegnn:
Fixes a rare case when the same waker (for stream or inner streams) was called twice and more times at the same time, and the stream was polled before the wakers ended their waking process, as this may lead to incorrect access to the `UnsafeCell` from some of the waking wakers.

### DIFF
--- a/futures-util/src/stream/stream/flatten_unordered.rs
+++ b/futures-util/src/stream/stream/flatten_unordered.rs
@@ -92,11 +92,17 @@ impl SharedPollState {
         let value = self
             .state
             .fetch_update(Ordering::SeqCst, Ordering::SeqCst, |value| {
-                let next_value = value | to_poll;
-
+                // Waking process for this waker already started
+                if value & waking != NONE {
+                    return None;
+                }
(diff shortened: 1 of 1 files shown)


taiki-e:
Thanks!

> this may lead to incorrect access to the `UnsafeCell` from some of the waking wakers.

Does this mean it will cause a data race?

olegnn:
> Does this mean it will cause a data race?

Well, the `UnsafeCell` may be potentially modified at the moment while it's being read, so yes.
However, to make it happen, we need the following sequence:
- Some waker (for ex. for inner streams) was called without delay several times from different threads
- We started to poll the stream before the last waker ends its waking process
- <Potentially read some bad waker from unsafe cell>
- Now there will be a problem only if the stream already ended its polling process, and the same waker was called again (the state was changed back to waking), but the waker which did an incorrect read is faster than the waker which changed the state, and now it will call potentially incorrect inner waker.

taiki-e (approved):

PR #2571, opened 2022-02-19
Feat: add Mutex::lock_owned and Mutex::try_lock_owned

Sherlock-Holo:
Add 2 new methods `Mutex::lock_owned` and `Mutex::try_lock_owned`

People who like https://github.com/rust-lang/futures-rs/issues/1972 want a new method such as `poll_lock`, to help them implement `Future` easier and less overhead. But adding `poll_lock` new method is so hard for the currently `Mutex` implementation.

However, if https://github.com/rust-lang/futures-rs/issues/1972 can store the future which returned by `Mutex::lock`, it can `poll` the future when poll its `MyReceiver`. Currently, the `Mutex` only has `lock` and `try_lock` method, which use the `&self` and the `lock` return a future but it has a reference to the `Mutex`. If save the mutex and its `MutexLockFuture` in a struct, will meet the **self-reference** problem.

To solve the **self-reference** problem, add a new way to lock the mutex: `Mutex::lock_owned(self: &Arc<Self>) -> OwnedMutexLockFuture<T>`, the `OwnedMutexLockFuture` doesn't reference any variable, it contains the mutex which wrapped by the `Arc`, so the *self-reference* problem disappears.

with this PR, https://github.com/rust-lang/futures-rs/issues/1972 can rewrite like

```rust
use futures::lock::Mutex;

struct MySender {
    inner: Arc<Mutex<Inner>>,
}

struct MyReceiver {
    inner: Arc<Mutex<Inner>>,
    fut: Option<OwnedMutexLockFuture<Inner>>,
    guard: Option<OwnedMutexGuard<Inner>>,
}

impl Future for MyReceiver {
    type Output = ();

    fn poll(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<()> {
        let this = self.get_mut();

        loop {
            if let Some(_guard) = this.guard.as_mut() {
                todo!()
            } else { 
                match this.fut.as_mut() {
                    None => {
                        this.fut.replace(this.inner.lock_owned());
                    }
                    Some(fut) => {
                        let guard = futures_core::ready!(Pin::new(fut).poll(cx));

                        this.guard.replace(guard);
                    }
                }
            }
        }
    }
}
```

### DIFF
--- a/futures-util/src/lock/mutex.rs
+++ b/futures-util/src/lock/mutex.rs
@@ -133,8 +133,8 @@ impl<T: ?Sized> Mutex<T> {
     ///
     /// This method returns a future that will resolve once the lock has been
     /// successfully acquired.
-    pub fn lock_owned(self: &Arc<Self>) -> OwnedMutexLockFuture<T> {
-        OwnedMutexLockFuture { mutex: Some(self.clone()), wait_key: WAIT_KEY_NONE }
+    pub fn lock_owned(self: Arc<Self>) -> OwnedMutexLockFuture<T> {
+        OwnedMutexLockFuture { mutex: Some(self), wait_key: WAIT_KEY_NONE }
     }
 
     /// Returns a mutable reference to the underlying data.
@@ -514,7 +514,6 @@ impl<T: ?Sized, U: ?Sized> DerefMut for MappedMutexGuard<'_, T, U> {
 // Mutexes can be moved freely between threads and acquired on any thread so long
 // as the inner value can be safely sent between threads.
 unsafe impl<T: ?Sized + Send> Send for Mutex<T> {}
-
 unsafe impl<T: ?Sized + Send> Sync for Mutex<T> {}
 
 // It's safe to switch which thread the acquire is being attempted on so long as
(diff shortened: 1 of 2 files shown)


futures-util/src/lock/mutex.rs
```diff
@@ -381,19 +514,35 @@ impl<T: ?Sized, U: ?Sized> DerefMut for MappedMutexGuard<'_, T, U> {
 // Mutexes can be moved freely between threads and acquired on any thread so long
 // as the inner value can be safely sent between threads.
 unsafe impl<T: ?Sized + Send> Send for Mutex<T> {}
+
```
taiki-e, line 517: (I personally like the previous format where single line trait implementations for the same type are grouped, but I'm not opposed to this change.)

taiki-e (commented):
Thanks!

futures-util/src/lock/mutex.rs
```diff
@@ -115,6 +129,14 @@ impl<T: ?Sized> Mutex<T> {
         MutexLockFuture { mutex: Some(self), wait_key: WAIT_KEY_NONE }
     }
 
+    /// Acquire the lock asynchronously.
+    ///
+    /// This method returns a future that will resolve once the lock has been
+    /// successfully acquired.
+    pub fn lock_owned(self: &Arc<Self>) -> OwnedMutexLockFuture<T> {
```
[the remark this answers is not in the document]
Sherlock-Holo: now it take self by value:)

futures-util/src/lock/mutex.rs
Sherlock-Holo: I have reverted the format, the format of the codes looks the same as the previous version

PR #2574, opened 2022-02-23
FuturesUnordered: fix partial iteration

### LINKED ISSUE #2573: Attempt to subtract with overflow in FuturesUnordered's IntoIter after calling .take(n)
Construct a `FuturesUnordered`, put `n` futures in it, call `.into_iter().take(n-1)` on it, and iterate the result.

Minimal example:
```
let _ = FuturesUnordered::from_iter(vec![ready(1), ready(2), ready(3), ready(4)])
    .into_iter()
    .take(3) // any number less than the number of futures in the collection, except zero
    .count(); // or collect into another FuturesUnordered, or whatever.
```
[playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=ec92405322b8c6d069db559df1b054fd)

---
Separately, while testing this, I found that if you only partially iterate that `IntoIter` type before dropping it, you get a different panic: `future still here when dropping`:

```
let _ = FuturesUnordered::from_iter(vec![ready(1), ready(2), ready(3), ready(4)])
    .into_iter()
    .take(3)
    .next();
```
[playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=6cd3990690b7ac832923edfd7b4cc0b9)

---
I also debugged this a bit and think I see what's going on. It looks like FuturesUnordered's `IntoIter` implementation does some brain surgery on the linked list while iterating which leaves it in a partially invalid state unless the iteration is fully completed. I'm preparing a possible fix.

wfraser:
The IntoIter impl advances the head pointer every iteration, but this breaks the linked list invariant that the head's prev should be null.

If the iteration is not done to completion (which sets head to null), on subsequent drop, FuturesUnordered::unlink relies on this broken invariant and ends up panicking.

The fix is to maintain the `head->prev == null` invariant while iterating.

Fixes #2573

### DIFF
--- a/futures-util/src/stream/futures_unordered/iter.rs
+++ b/futures-util/src/stream/futures_unordered/iter.rs
@@ -2,6 +2,7 @@ use super::task::Task;
 use super::FuturesUnordered;
 use core::marker::PhantomData;
 use core::pin::Pin;
+use core::ptr;
 use core::sync::atomic::Ordering::Relaxed;
 
 /// Mutable iterator over all futures in the unordered set.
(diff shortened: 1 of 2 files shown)


taiki-e (approved):
Thanks!

PR #2577, opened 2022-03-12
`TryFlattenUnordered`

olegnn:
- Basic `TryFlattenUnordered` implementation atop of `FlattenUnordered`.
- ~Improve the `FlattenUnordered` interface by removing the `Unpin` requirement for `Self::Item`~. This's much more convenient and I guess it's a good choice for simplification despite a possibly unneeded `Pin<Box<_>>`.
- `FusedStream` impl for `FlattenUnordered`: remove unneeded `Self::Item: FusedStream` requirement.
- Minor improvement of the `FlattenUnordered` internal logic.
- Export all underlying types.
- Docs improvements.

### DIFF
--- a/futures-util/src/stream/stream/flatten_unordered.rs
+++ b/futures-util/src/stream/stream/flatten_unordered.rs
@@ -1,4 +1,4 @@
-use alloc::sync::Arc;
+use alloc::{boxed::Box, sync::Arc};
 use core::{
     cell::UnsafeCell,
     convert::identity,
--- a/futures-util/src/stream/stream/flatten_unordered.rs
+++ b/futures-util/src/stream/stream/flatten_unordered.rs
@@ -1,4 +1,4 @@
-use alloc::{boxed::Box, sync::Arc};
+use alloc::sync::Arc;
 use core::{
     cell::UnsafeCell,
     convert::identity,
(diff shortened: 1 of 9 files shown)


futures-util/src/stream/stream/flatten_unordered.rs
```diff
@@ -424,15 +400,24 @@ where
...
+                    if stream_waker.is_none() {
+                        // Safety: now state is `POLLING`.
+                        stream_waker
+                            .replace(unsafe { WrappedWaker::replace_waker(this.stream_waker, cx) });
+                    }
+                    let mut cx = Context::from_waker(stream_waker.as_ref().unwrap());
+
+                    match this.stream.as_mut().poll_next(&mut cx) {
                         Poll::Ready(Some(inner_stream)) => {
+                            let next_item_fut = PollStreamFut::new(Box::pin(inner_stream));
```
taiki-e, line 413: Hmm. I feel removing Unpin bound by boxing each stream is worse than requiring Unpin bound; if the stream is already Unpin, it adds an extra cost.

olegnn: I'd say this combinator is much more convenient without `Unpin` requirement, but let's choose the pragmatic performance over the usability)

taiki-e (approved):
Thanks!

PR #2579, opened 2022-03-15
Fix sentence in `try_buffered` docs.

timleslie:
Updates the docs to use the same phrasing as `.buffered()` https://docs.rs/futures/latest/futures/stream/trait.StreamExt.html#method.buffered


taiki-e (approved):
Thanks

PR #2582, opened 2022-03-20
Add `select` benchmark

414owen:
I was trying to optimize `select`, by making it take data that implements a `Strategy` trait, instead of a closure. Didn't seem to work, but it seems like a good function to benchmark.

Next try will be to make it stop calling `.poll_next()` on a stream when it terminates, and removing the `Fuse` layer. Might just be that these details are getting lost in the noise though.


taiki-e (approved):
Thanks!

PR #2583, opened 2022-03-21
Remove `Fuse`s from `select`, and only poll non-terminated streams

414owen:
Benchmarks:

```
// old
test select_streams ... bench:     931,953 ns/iter (+/- 18,919)
// new
test select_streams ... bench:     847,498 ns/iter (+/- 21,188)
```

**Time**: ~-10%
**Benchmark PR**: https://github.com/rust-lang/futures-rs/pull/2582
**Breaking changes**: none

### DIFF
--- a/futures-util/src/stream/select_with_strategy.rs
+++ b/futures-util/src/stream/select_with_strategy.rs
@@ -18,12 +18,7 @@ impl PollNext {
     #[must_use]
     pub fn toggle(&mut self) -> Self {
         let old = *self;
-
-        match self {
-            PollNext::Left => *self = PollNext::Right,
-            PollNext::Right => *self = PollNext::Left,
-        }
-
+        *self = self.other();
         old
     }
 
--- a/futures-util/src/stream/select_with_strategy.rs
+++ b/futures-util/src/stream/select_with_strategy.rs
@@ -1,5 +1,4 @@
 use super::assert_stream;
-use crate::stream::{Fuse, StreamExt};
 use core::{fmt, pin::Pin};
 use futures_core::stream::{FusedStream, Stream};
 use futures_core::task::{Context, Poll};
@@ -8,6 +7,13 @@ use pin_project_lite::pin_project;
 /// Type to tell [`SelectWithStrategy`] which stream to poll next.
 #[derive(Debug, PartialEq, Eq, Copy, Clone, Hash)]
 pub enum PollNext {
+    /// Poll the first stream.
+    Left,
+    /// Poll the second stream.
+    Right,
+}
+
+enum PollSide {
     /// Poll the first stream.
     Left,
     /// Poll the second stream.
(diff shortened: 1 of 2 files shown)


futures-util/src/stream/select_with_strategy.rs
```diff
@@ -14,6 +13,13 @@ pub enum PollNext {
     Right,
 }
 
+enum PollSide {
```
414owen, line 16: This is identical to `PollNext`, but we're using it to represent the last polled side (not the next one).

Would you be open to just renaming `PollNext` to `PollSide`? I think the `Next` suffix only reduces the contexts it can be used in. It would be a breaking change, but that might be acceptable for 0.4.0?

414owen:
Not sure why the build is failing for `-Z minimal-versions`.  
As far as I can see all our `pin*` dependencies are fixed to a patch version.  
@taiki-e any ideas?

futures-util/src/stream/select_with_strategy.rs
```diff
@@ -170,47 +199,88 @@ where
...
-        self.stream1.is_terminated() && self.stream2.is_terminated()
+        match self.internal_state {
+            InternalState::BothFinished => true,
+            _ => false,
+        }
     }
 }
 
-impl<St1, St2, Clos, State> Stream for SelectWithStrategy<St1, St2, Clos, State>
+#[inline(always)]
```
taiki-e, line 209: Why is this #[inline(always)]? AFAIK #[inline] is enough in most cases.

futures-util/src/stream/select_with_strategy.rs
```diff
@@ -170,47 +199,88 @@ where
...
+    match side {
+        PollNext::Left => select.stream1.as_mut().poll_next(cx),
+        PollNext::Right => select.stream2.as_mut().poll_next(cx),
     }
 }
 
-fn poll_inner<St1, St2>(
-    a: Pin<&mut St1>,
-    b: Pin<&mut St2>,
+#[inline(always)]
```
taiki-e, line 225: Same here.

taiki-e (commented):
Thanks! The implementation looks good to me.

taiki-e:
> Doesn't seem to be strictly semver-compliant

Note that if the major version is 0, cargo treats the minor version like the major version and the patch version like the minor version.

https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-dependencies-from-cratesio

futures-util/src/stream/select_with_strategy.rs
414owen: You're right, it looks like the binaries produced are identical.
I'll change it to `#[inline]`.

414owen:
@taiki-e I've implemented your suggestions, looks ready to merge to me.

taiki-e (approved):

PR #2585, opened 2022-03-30
Add try join all chunked

saood06:
Added a chunked version of try join all. 

Drives at most n Futures at a time, and then will return once all are completed, or returns early if any error just like try join all.

### DIFF
--- a/futures-util/src/future/try_join_all_chunked.rs
+++ b/futures-util/src/future/try_join_all_chunked.rs
@@ -102,9 +102,10 @@ where
     fn poll(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
         let mut state = FinalState::AllDone;
         let mut in_progress = 0;
+        let chunk_size = self.chunk_size
 
         for elem in iter_pin_mut(self.elems.as_mut()) {
-            if in_progress < self.chunk_size {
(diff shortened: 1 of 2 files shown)


taiki-e:
Thanks for the PR! What are the advantages of this over try_buffer_unordered and try_buffered?

futures-util/src/future/try_join_all_chunked.rs
```diff
@@ -0,0 +1,140 @@
...
+use alloc::vec::Vec;
+use core::fmt;
+use core::future::Future;
+use core::mem;
+use core::pin::Pin;
+use core::task::{Context, Poll};
+
+use super::{assert_future, TryFuture, TryMaybeDone};
+
+fn iter_pin_mut<T>(slice: Pin<&mut [T]>) -> impl Iterator<Item = Pin<&mut T>> {
```
ibraheemdev, line 14: Can we move this to a util module, it's already used by join_all and try_join_all.

saood06:
> Thanks for the PR! What are the advantages of this over try_buffer_unordered and try_buffered?

Sorry, I did not know they existed, they do seem to handle the use case I thought this covered.

taiki-e:
If it is covered by an existing API, I'm going to close this in favor of that.

Thanks anyway for the PR!

PR #2587, opened 2022-04-15
Refactor: prefer early returns in `futures_util::future::select`

Some-Dood:
Hello there! This PR slightly refactors the `future::select` implementation so that it uses early returns. Although a relatively small change, I believe it makes the code a little bit more readable than the originally nested `match` expressions. It also makes the control flow clearer since it's more "linear", so to speak.

In line with the readability improvements, I've also renamed the `x` variable to `val` so that it's more apparent what `x` represents.

That is all. Thanks! 🎉


taiki-e (approved):
Thanks.

PR #2588, opened 2022-04-15
Add 'join_with' function which joins a future with a value

malthe:
This adds a `join_with` function which splices in a prepared value with the result of the future:

```rust
let f = async { 1 };
let joined = future::join_with(f, 2);
assert_eq!(joined.await, (1, 2));
```
Unlike `join` – the value is known ahead of time, simplifying the type signature and avoiding the use of `ready`.

### DIFF
--- a/futures-util/src/future/join_with.rs
+++ b/futures-util/src/future/join_with.rs
@@ -1,0 +1,74 @@
+use crate::future::assert_future;
+use core::fmt;
+use core::pin::Pin;
+use futures_core::future::Future;
+use futures_core::task::{Context, Poll};
+use pin_project_lite::pin_project;
+
+pin_project! {
+    /// Future for the [`join`](super::FutureExt::join_with) method.
+    #[must_use = "futures do nothing unless you `.await` or poll them"]
+    pub struct JoinWith<F: Future, T> {
+        #[pin]
+        future: F,
+        data: T,
+    }
+}
+
+impl<Fut: Future, T> JoinWith<Fut, T> {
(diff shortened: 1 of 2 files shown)


ibraheemdev:
How is this different than `f.map(|x| (x, 2))`?

malthe:
@ibraheemdev in case you don't want a dynamic type – e.g. when storing a future in a struct without boxing.

ibraheemdev:
I don't see why you would need to store this type, as opposed to just injecting the value when it's needed?

malthe:
> I don't see why you would need to store this type, as opposed to just injecting the value when it's needed?

For example, if you make an HTTP request and you want to return back an identifier together with the response. Or you're storing the time when the future was created to measure the latency. Consider it a building block for a variety of use-cases.

taiki-e:
Thanks for the PR! This seems equivalent to `future::join`+`future::ready`, so I'm not sure if this needs to be added.

---

For the API and implementation:

- future is expected not to be polled after completion, so `T: Copy` bound can be removed by keeping the value as an `Option<T>`.
- \<prefix>_with is a name often used in APIs with closures as arguments. I feel that this is misleading for this API.

(The above are suggestions for this PR, but need not be applied until we decide whether to accept or reject this PR.)

malthe:
As @taiki-e points out, this is easily implemented using already defined `join` and `ready`:
```rust
let f = async { 1 };
let joined = futures::future::join(f, futures::future::ready(2));
assert_eq!(joined.await, (1, 2));
```

PR #2590, opened 2022-04-17
Several tweaks

olegnn:
- One fix is for the requirements of `Sink` for `TryFlattenUnordreded` is to allow using any `TryStream` rather than just `Stream<Result<..>>`
- The other one is for `FlattenUnordredered` to reduce the possible amount of wake calls.


taiki-e (approved):
Thanks.

PR #2591, opened 2022-04-20
Add push_front and push_back to FuturesOrdered

conorbros:
- Renamed `push` to `push_back`
- Added `push_front`, which pushes a future to the front of the queue.

More info [here](https://github.com/rust-lang/futures-rs/issues/2309) (the `push_front` logic is taken from here).

We might want to leave `push` as is and just add `push_front` to avoid renaming it. 

Closes: #2591

### DIFF
--- a/futures-util/src/stream/futures_ordered.rs
+++ b/futures-util/src/stream/futures_ordered.rs
@@ -129,16 +129,33 @@ impl<Fut: Future> FuturesOrdered<Fut> {
         self.in_progress_queue.is_empty() && self.queued_outputs.is_empty()
     }
 
-    /// Push a future into the queue.
+    /// Pushes a future to the back of the queue.
     ///
     /// This function submits the given future to the internal set for managing.
     /// This function will not call `poll` on the submitted future. The caller
     /// must ensure that `FuturesOrdered::poll` is called in order to receive
     /// task notifications.
-    pub fn push(&mut self, future: Fut) {
+    pub fn push_back(&mut self, future: Fut) {
         let wrapped = OrderWrapper { data: future, index: self.next_incoming_index };
         self.next_incoming_index += 1;
         self.in_progress_queue.push(wrapped);
+    }
+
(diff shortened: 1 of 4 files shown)


taiki-e:
Thanks! Can you retain the old `push` method as deprecated? If so, this can be backported to futures 0.3.

futures-util/src/stream/futures_ordered.rs
```diff
@@ -135,11 +135,41 @@ impl<Fut: Future> FuturesOrdered<Fut> {
     /// This function will not call `poll` on the submitted future. The caller
     /// must ensure that `FuturesOrdered::poll` is called in order to receive
     /// task notifications.
+    #[deprecated]
```
taiki-e, line 138:
```suggestion
    #[deprecated(note = "use `push_back` instead")]
```

taiki-e (approved):

rukai:
I noticed this was approved and then stalled.
Is there anything preventing merging?

futures-util/src/stream/futures_ordered.rs
```diff
@@ -135,11 +135,41 @@ impl<Fut: Future> FuturesOrdered<Fut> {
     /// This function will not call `poll` on the submitted future. The caller
     /// must ensure that `FuturesOrdered::poll` is called in order to receive
     /// task notifications.
+    #[deprecated(note = "use `push_back` instead")]
     pub fn push(&mut self, future: Fut) {
         let wrapped = OrderWrapper { data: future, index: self.next_incoming_index };
         self.next_incoming_index += 1;
         self.in_progress_queue.push(wrapped);
```
yjhmelody, line 142: Maybe call `self.push_back` instead of duplicated code?

conorbros:
Opened another PR to fix the clippy lints: https://github.com/rust-lang/futures-rs/pull/2618

taiki-e:
Thanks!

PR #2593, opened 2022-04-21
Make run_until_stalled handle self-waking futures

khollbach:
This fixes a bug in `LocalPool`'s `run_until_stalled` and `try_run_one`, where self-waking futures cause these methods to exit early.

The issue occurs when a future "yields" to the executor by simply waking its waker and returning `Pending`. As of #2551, when an underlying future "yields" in this way, the containing `FuturesUnordered` will do the same. Then, when this happens, the containing `LocalPool` assumes that this `Pending` means that the pool can't make any more progress.

<details>
<summary>repro</summary>

This [playground], where the output is:
```
`Pending(0)` polled!
`Pending(1)` polled!
```
But it should be:
```
`Pending(0)` polled!
`Pending(1)` polled!
`Pending(2)` polled!
`Ready` polled!
```
</details>

I believe this PR should fix it, but I'm new to open source (and async Rust) so feedback is welcome!

CC'ing @gmorenz who helped me track down the cause of the bug -- thanks Greg :)

[playground]: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=b4d238036a95b67260dbfb77b8b83b10

### DIFF
--- a/futures-executor/tests/local_pool.rs
+++ b/futures-executor/tests/local_pool.rs
@@ -437,7 +437,7 @@ fn park_unpark_independence() {
 }
 
 struct SelfWaking {
-    wakeups_remaining: Arc<RefCell<usize>>,
+    wakeups_remaining: Rc<RefCell<usize>>,
 }
 
 impl Future for SelfWaking {
--- a/futures-executor/tests/local_pool.rs
+++ b/futures-executor/tests/local_pool.rs
@@ -1,7 +1,7 @@
 use futures::channel::oneshot;
 use futures::executor::LocalPool;
 use futures::future::{self, lazy, poll_fn, Future};
-use futures::task::{Context, LocalSpawn, Poll, Spawn, Waker};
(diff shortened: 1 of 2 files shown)


taiki-e (commented):
Thanks! Could you add a test case for this?

futures-executor/tests/local_pool.rs
```diff
@@ -435,3 +435,65 @@ fn park_unpark_independence() {
 
     futures::executor::block_on(future)
 }
+
+struct SelfWaking {
+    wakeups_remaining: Arc<RefCell<usize>>,
```
taiki-e, line 440:
```suggestion
    wakeups_remaining: Rc<RefCell<usize>>,
```

khollbach: Good catch -- thank you!

taiki-e (approved):

PR #2596, opened 2022-05-05
Add `WithContext` future that runs a closure with `task::Context`

WaffleLapkin:
This PR adds `WithContext<F>` future and `with_context` function that run a closure with `task::Context`.
This can be used to, for example, [poll futures by hand](https://play.rust-lang.org/?version=nightly&mode=debug&edition=2021&gist=fe1ff8c37a916528d09a212ec37d4f1c).

### DIFF
--- a/futures-util/src/future/mod.rs
+++ b/futures-util/src/future/mod.rs
@@ -68,6 +68,9 @@ pub use self::option::OptionFuture;
 mod poll_fn;
 pub use self::poll_fn::{poll_fn, PollFn};
 
+mod with_context;
+pub use self::with_context::{with_context, WithContext};
+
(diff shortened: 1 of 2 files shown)


ibraheemdev:
How is this different from `poll_fn`?

WaffleLapkin:
`poll_fn` takes `FnMut` that is called until `Ready()` is returned. `with_context` takes `FnOnce` that is called once.

ibraheemdev:
So `lazy`?

WaffleLapkin:
Right. Yeah, that's exactly the same as `lazy`, I didn't notice because of the name I didn't expect. Closing since this is already implemented in futures, thanks!

PR #2598, opened 2022-05-10
Fix new let_unit_value clippy lint

414owen:
Clippy has introduced a new lint, which causes [CI to fail](https://github.com/rust-lang/futures-rs/runs/6373235397?check_suite_focus=true).


taiki-e (approved):
Thanks!

PR #2599, opened 2022-05-11
Remove TryStreamExt::into_async_read Unpin bound

khollbach:
Removed an unnecessary bound on `Unpin` from `TryStreamExt::into_async_read`, and updated the docs.

The only side-effect I'm aware of is that the `IntoAsyncRead` docs now have:
```rust
impl<'__pin, St> Unpin for IntoAsyncRead<St>
where
    St: TryStream<Error = Error>,
    __Origin<'__pin, St>: Unpin,
    <St as TryStream>::Ok: AsRef<[u8]>,
```
instead of simply
```rust
impl<St> Unpin for IntoAsyncRead<St>
where
    St: TryStream<Error = Error> + Unpin,
    <St as TryStream>::Ok: AsRef<[u8]>,
```
since we're using `pin_project!`.

### DIFF
--- a/futures-util/src/stream/try_stream/into_async_read.rs
+++ b/futures-util/src/stream/try_stream/into_async_read.rs
@@ -1,30 +1,26 @@
-use crate::stream::TryStreamExt;
 use core::pin::Pin;
 use futures_core::ready;
 use futures_core::stream::TryStream;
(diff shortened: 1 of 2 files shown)


taiki-e (approved):
Thanks!

PR #2603, opened 2022-05-28
Fix typo

taiki-e:
SeeKRelative -> SeekRelative


taiki-e:
This is a public API, so I cannot backport this fix.
The commit to fix typo itself contained a typo.

😢

PR #2605, opened 2022-05-30
for_each(_concurrent): Return the number of completed stream futures

uklotzde:
Keeping track of the number of completed stream futures comes at almost zero cost. It improves the versatility of these functions as the number of futures yielded by the stream is usually not known in advance. The caller would otherwise need to use an atomic counter to determine it.

Extending the corresponding `try_` variants accordingly seems legit but turns out to be convoluted. The number of completed stream futures would be relevant in case of both an `Ok` and an `Err` result.

### DIFF
--- a/futures-util/src/stream/stream/for_each.rs
+++ b/futures-util/src/stream/stream/for_each.rs
@@ -15,6 +15,7 @@ pin_project! {
         f: F,
         #[pin]
         future: Option<Fut>,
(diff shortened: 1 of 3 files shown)


taiki-e:
Similar functionally is already available as `.map().count()`/`.map().buffer_unordered().count()`, and I think they are more clearer and consistent with the iterator API.

PR #2607, opened 2022-05-31
`TryFlattenUnordered`: propagate base stream error

olegnn:
Instead of allowing polling of the underlying stream after the error, propagate this error first, and only after it will be received by the caller, allow to continue polling.
The caller then will be able to either continue to poll the stream or abort due to the received error. This behavior is similar to the behavior of `try_buffer_unordered`.


taiki-e (approved):
Thanks!

PR #2613, opened 2022-06-13
`doc(alias)` `pending()` to `never`

danielhenrymantilla:
It can be legitimate to look for the future that _never_ resolves using the name `never`. This PR ensures `pending()` shows up when making such queries.

cc @conradludgate

### DIFF
--- a/futures-util/src/future/pending.rs
+++ b/futures-util/src/future/pending.rs
@@ -33,7 +33,7 @@ impl<T> FusedFuture for Pending<T> {
 /// unreachable!();
 /// # });
 /// ```
-#[doc(alias = "never")]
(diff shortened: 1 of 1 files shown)


taiki-e (approved):
Thanks!

futures-util/src/future/pending.rs
```diff
@@ -33,6 +33,7 @@ impl<T> FusedFuture for Pending<T> {
 /// unreachable!();
 /// # });
 /// ```
+#[doc(alias = "never")]
```
taiki-e, line 36:
```suggestion
#[cfg_attr(docsrs, doc(alias = "never"))]
```

doc(alias) is unstable on our MSRV.

PR #2614, opened 2022-06-21
Include crate version for dev-dependencies

j-vanderstoep:
This ensures that test dependencies are included in the Cargo.toml
that is published to crates.io.

As noted in the Cargo book:
When a package is published, only dev-dependencies that specify a
version will be included in the published crate. For most use cases,
dev-dependencies are not needed when published, though some users
(like OS packagers) may want to run tests within a crate, so
providing a version if possible can still be beneficial.

This is helpful for running tests in the Android Open Source Project,
which uses crates.io as its source of truth, and prevents having to
carry local patches.

Test: cargo publish --dry-run
Confirm that dev-dependencies are listed in the resulting
Cargo.toml.


taiki-e:
Thanks for the PR. This is intentional (https://github.com/rust-lang/futures-rs/pull/2305), and cannot be accepted unless https://github.com/rust-lang/cargo/issues/4242 is fixed.

PR #2626, opened 2022-07-22
Inline WakerRef functions

xfix:
Those functions are trivial and non-generic. It's necessary to use `#[inline]` to enable cross-crate inlining for non-generic functions when LTO is disabled.

See also https://internals.rust-lang.org/t/will-it-ever-be-possible-to-optimize-away-trivial-futures/17060/5?u=xfix


taiki-e (approved):
Thanks!

PR #2644, opened 2022-09-14
Fix CI

Xaeroxe:
Fixes the `join_size` test, as the latest nightly brought down the size. Also applies some clippy suggestions about boolean to integer conversion.


taiki-e (approved):
Thanks!

PR #2649, opened 2022-10-04
Fix soundness hole in join macros

Pointerbender:
There is a soundness bug in the join macros where a new exclusive reference is taken out every time a future is polled. Under the Stacked Borrows model, this invalidates self-referential futures. This is then in conflict with the safety requirements for `Pin::new_unchecked`, which say that the pointer may never be invalidated after pinning it - and this leads to undefined behavior in safe Rust code.

I added a miri regression test for completeness (this example was made by @jswrenn), and together with @RalfJung we pin-pointed the problem and created a fix ([Zulip thread](https://rust-lang.zulipchat.com/#narrow/stream/269128-miri/topic/Making.20Miri.20a.20subtree.20and.20toolstate.20always.20green)). There is a minor down-side to the fix, the joined future increased in size, due to having to store an extra pinned reference in it, and I updated the tests to account for the increased sizes.

### DIFF
--- a/futures-macro/src/join.rs
+++ b/futures-macro/src/join.rs
@@ -38,6 +38,7 @@ fn bind_futures(fut_exprs: Vec<Expr>, span: Span) -> (Vec<TokenStream2>, Vec<Ide
                 // Move future into a local so that it is pinned in one place and
                 // is no longer accessible by the end user.
                 let mut #name = __futures_crate::future::maybe_done(#expr);
+                let mut #name = unsafe { __futures_crate::Pin::new_unchecked(&mut #name) };
             });
             name
         })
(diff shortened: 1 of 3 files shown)


RalfJung:
> soundness bug

I should clarify that this is a bit unclear. ;)  If we had proper support for self-referential generators in the language, then I think this code would be fine. But sadly self-referential generators were added in the frontend without first laying the necessary groundwork in the backend (in the form of appropriate primitives to handle the aliasing this generates), and until someone cleans up that mess, we are left with a hack: using the `Unpin` trait to control the aliasing requirements. Due to the fact that `PollFn<F>` is `Unpin` even if `F` is not, the code violates that hacky aliasing model.

See [this IRLO thread](https://internals.rust-lang.org/t/surprising-soundness-trouble-around-pollfn/17484) for more details.

Pointerbender:
p.s. the warnings in the CI seem to be false positives, the `futures_util::stream::futures_unordered::{Iter, IterMut}` items are available to other crates, so I'm not sure why it's complaining about those 2 in particular.

RalfJung:
FWIW comex has demonstrated an [actual miscompilation](https://internals.rust-lang.org/t/surprising-soundness-trouble-around-pollfn/17484/15?u=ralfjung) based on this soundness bug.

taiki-e:
Thanks!

CI failure has been fixed by #2651. Could you rebase?

Pointerbender:
Thank you for the CI fix! I rebased the PR and had to make one small addition to the test cases for `join_size` and `try_join_size` by adding a `#[cfg_attr(not(target_pointer_width = "64"), ignore)]` attribute to them. Due to that the futures now store an extra reference, the size will be different for non-64-bits pointer targets and this makes the tests fail on those targets due to the sizes being hard-coded for 64-bit wide pointers (this is the same fix as that was applied in #2447).

taiki-e (approved):

taiki-e:
Published in 0.3.25.

PR #2664, opened 2022-11-12
fix FuturesOrdered

### LINKED ISSUE #2663: FuturesOrdered push_front does not work at the start
Running this code:
https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=30c5f139539ccf5dd237299c88b56a2e
```rust
use std::future::ready;
use std::collections::VecDeque;
use futures::stream::{FuturesOrdered, StreamExt};

#[tokio::main]
async fn main() {
    let mut set = VecDeque::new();

    set.push_back(2);  // should be the last
    set.push_front(1); // should be the second
    set.push_front(0);  // should be the first
    
    let x: Vec<i32> = set.into_iter().collect();
    assert_eq!(x, [0, 1, 2]);

    let mut set = FuturesOrdered::new();
    
    set.push_back(ready(2));  // should be the last
    set.push_front(ready(1)); // should be the second
    set.push_front(ready(0));  // should be the first
    
    let x: Vec<i32> = set.collect().await;
    assert_eq!(x, [0, 1, 2]);
}
```

You see the assertion on the VecDeque passes, but the same assertion through FuturesOrdered fails.

I noticed this working on my own implementation in https://github.com/conradludgate/futures-buffered - I found that turning the `usize` counters into `isize` fixes it with no noticeable downsides.

conradludgate:
fixes #2663

### DIFF
--- a/futures-util/src/stream/futures_ordered.rs
+++ b/futures-util/src/stream/futures_ordered.rs
@@ -19,7 +19,7 @@ pin_project! {
     struct OrderWrapper<T> {
         #[pin]
         data: T, // A future or a future's output
-        index: usize,
+        index: isize,
     }
 }
 
@@ -95,8 +95,8 @@ where
 pub struct FuturesOrdered<T: Future> {
     in_progress_queue: FuturesUnordered<OrderWrapper<T>>,
     queued_outputs: BinaryHeap<OrderWrapper<T::Output>>,
-    next_incoming_index: usize,
-    next_outgoing_index: usize,
+    next_incoming_index: isize,
+    next_outgoing_index: isize,
 }
 
 impl<T: Future> Unpin for FuturesOrdered<T> {}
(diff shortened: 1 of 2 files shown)


taiki-e:
Thanks for the PR! Considering the overflow of isize (which can actually happen on 32-bit targets), this still seems broken, but this seems more correct than the current one.

conradludgate:
Yeah. I have some ideas how to fix the overflow too, but it's not as trivial a fix

https://github.com/conradludgate/futures-buffered/blob/7db2b4ca822ee95bb53871fee7867b420a04a7fa/src/futures_ordered_bounded.rs#L216-L231

Essentially, if the next_outgoing_index has the MSB set, then it flips the MSB of every task/output. This preserves the order and ensures that the outgoing is < isize::MAX. The length of the queue would have to be another isize::MAX in order to overflow, but you'd OOM before that.

It should be possible to implement here too, but it might not be super efficient with the task linked list

taiki-e (approved):

taiki-e:
Thanks! Opened https://github.com/rust-lang/futures-rs/issues/2689 to track overflow issue.

PR #2666, opened 2022-11-22
Remove `Debug` constraint for `oneshot` types

daxpedda:
This removes the unnecessary `Debug` constraint on `Debug` implementation on `oneshot::Receiver` and `oneshot::Sender`.

I believe this is not a breaking change?

### DIFF
--- a/futures-channel/src/oneshot.rs
+++ b/futures-channel/src/oneshot.rs
@@ -390,7 +390,7 @@ impl<T> Drop for Sender<T> {
     }
 }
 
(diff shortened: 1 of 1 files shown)


taiki-e (approved):
Thanks! These bounds were originally left for future compatibility (https://github.com/rust-lang/futures-rs/pull/2414), but given that the Debug impls for std mpsc and crossbeam-channel also do not have these bounds, it seems unlikely that we will need them in the future.

PR #2667, opened 2022-11-22
Custom `Debug` implementations for `mpsc`

daxpedda:
This is the same PR as #2414 but for `mpsc` instead of `oneshot`. I'm not sure if it is considered okay to do atomic operations in a `Debug` implementation, but I can't think of a reason why not to.

This also removes the `Debug` constraint on `Debug` implementations on all `mpsc` types, similar to #2666.


taiki-e (approved):
Thanks!

PR #2670, opened 2022-12-04
Enable an option for portable-atomic

notgull:
This PR enables an option in `futures-core` that replaces the atomic usage in `AtomicWaker` with `portable-atomic`, and also exposes that flag in `futures-util`. This intends to make it so `futures-core` can be used on platforms without atomics, like AVR.

I look through `futures-util` and didn't see any other usage of atomics that wasn't also locked behind allocation, so I think this is the best I can do here.

### DIFF
--- a/futures-core/Cargo.toml
+++ b/futures-core/Cargo.toml
@@ -16,6 +16,7 @@ std = ["alloc"]
 alloc = []
 
 [dependencies]
+portable-atomic = { version = "0.3.15", default-features = false, optional = true }
 
 [dev-dependencies]
 futures = { path = "../futures" }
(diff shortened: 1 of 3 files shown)


taiki-e:
Thanks!

> I look through `futures-util` and didn't see any other usage of atomics that wasn't also locked behind allocation, so I think this is the best I can do here.

Ah, the `alloc` feature uses `Arc` in the public API, so even if https://github.com/taiki-e/portable-atomic/pull/41 is released it will not be easy to support it. I have a few ~hacks~ ideas to support it, so I will try it after I merge this PR.

(CI failure is probably a problem addressed in https://github.com/rust-lang/miri/pull/2713 I guess it will be fixed after the next Miri update)

taiki-e (approved):

PR #2673, opened 2022-12-13
Add a `ptr_eq` to `Shared` to test if two `Shared`s share the same underlying future.

khuey:
`Shared` already has `Arc`'s `strong_count` and `weak_count`. This will make it easier to implement `PartialEq`/`Eq` for structs containing a `Shared`.

### DIFF
--- a/futures-util/src/future/future/shared.rs
+++ b/futures-util/src/future/future/shared.rs
@@ -4,7 +4,9 @@ use futures_core::task::{Context, Poll, Waker};
 use slab::Slab;
 use std::cell::UnsafeCell;
 use std::fmt;
+use std::hash::Hasher;
 use std::pin::Pin;
+use std::ptr;
 use std::sync::atomic::AtomicUsize;
 use std::sync::atomic::Ordering::{Acquire, SeqCst};
 use std::sync::{Arc, Mutex, Weak};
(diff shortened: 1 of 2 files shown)


khuey:
Hrm, I want to be able to hash it too. Perhaps I should expose a pointer value (e.g. with a `Shared::as_ptr` that returns an opaque pointer value) rather than adding `Shared::ptr_eq`.

khuey:
Or how about this?

futures-util/src/future/future/shared.rs
```diff
@@ -156,6 +158,35 @@ where
...
+            None => {
+                state.write_u8(0);
+            }
+        }
+    }
+
+    /// Returns `true` if the two `Shared`s point to the same future (in a vein similar to
+    /// `Arc::ptr_eq`).
+    ///
+    /// Returns `false` if either `Shared` has terminated.
```
taiki-e, line 178: The implementation that returns false if both are terminated would appear incompatible with ptr_hash above. See also docs: https://doc.rust-lang.org/std/hash/trait.Hash.html#hash-and-eq

That said, it also seems odd that it always returns true when both are terminated.

khuey: I don't understand this comment. This `ptr_eq` returns *false* if one or both `Shared`s are terminated. And the constraint on `Hash` is that if k1 == k2 then hash(k1) == hash(k2), it says nothing about k1 != k2.

taiki-e: Ok, I think it was my misunderstanding.

taiki-e (approved):
Thanks!

PR #2676, opened 2022-12-18
impl FusedStream for Buffered

arcnmx:
it's what it says on the tin

cc @taiki-e for [a prior comment on this](https://github.com/rust-lang/futures-rs/pull/1967#issuecomment-723443537) that doesn't seem to have been acted upon


taiki-e (approved):
Thanks!

PR #2677, opened 2022-12-18
impl Future for stream::{Buffered, BufferUnordered, FuturesOrdered}

arcnmx:
This enables `poll()`ing to fill the buffer without producing or consuming the Stream's outputs.

`Buffered` and `FuturesOrdered` will also poll their inner futures produced by the stream, since they have an output value buffer available.
`BufferUnordered` however will only poll the underlying stream, as it cannot produce outputs without consuming them.

~~Given that the `Output` of the `BufferUnordered` impl isn't particularly useful (because it doesn't indicate anything about the futures produced by the stream), maybe it should be `type Output = Infallible`, and never complete?
For the other two, the completion is more meaningful as it indicates that the stream is empty and that all captured futures have produced outputs (though in the case of `Buffered`, it will stall while the buffer remains full).~~

~~It may also make sense to expose this as a `poll_stream()` function rather than `impl Future` (or otherwise a `IntoFuture for &Self` or something), for the sake of extension trait method resolution.~~

EDIT: moved to inherent `poll_stream()` instead, see latest commit. Final name tbd (`buffered.poll_all()`? `poll_buffer()`? `poll_fill()`?)

### DIFF
--- a/futures-util/src/stream/futures_ordered.rs
+++ b/futures-util/src/stream/futures_ordered.rs
@@ -1,3 +1,4 @@
+use crate::future::FutureExt;
 use crate::stream::{FuturesUnordered, StreamExt};
 use alloc::collections::binary_heap::{BinaryHeap, PeekMut};
 use core::cmp::Ordering;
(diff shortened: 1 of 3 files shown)


taiki-e:
Thanks for the PR. What is the use case for these methods?

PR #2687, opened 2023-01-15
Remove incorrect error behavior from `ready_chunks` docs

jongiddy:
Unlike `TryStream::try_chunks`, which has similar documentation and can handle `Some(Err(_))` specially, `Stream::ready_chunks` does not and cannot know whether the stream item represents an error condition, and does not include any code to return early on an error.


taiki-e (approved):
Thanks.

PR #2690, opened 2023-01-17
add is_some and is_none to OptionFuture

MarinPostma:
This PR adds `is_some` and `is_none` to OptionFuture. This is useful when you only want to poll the `OptionFuture` if it contains a future (i.e treat `None` as `never`).


taiki-e:
Hmm. This could be a bit misleading API, since in the current implementation OptionFuture does not set the inner field to None when the future is complete.

We may want to set the inner field to None on completion and make is_none return true "if the future is absent or completed".

PR #2691, opened 2023-01-17
Add `Either::as_pin_mut` and `Either::as_pin_ref`

thomaseizinger:
Resolves https://github.com/rust-lang/futures-rs/issues/2629.

I added the implementation based on what was done in the `either` crate: https://github.com/bluss/either/pull/77

### DIFF
--- a/futures-util/src/future/either.rs
+++ b/futures-util/src/future/either.rs
@@ -34,12 +34,7 @@ pub enum Either<A, B> {
 
 impl<A, B> Either<A, B> {
     fn project(self: Pin<&mut Self>) -> Either<Pin<&mut A>, Pin<&mut B>> {
-        unsafe {
-            match self.get_unchecked_mut() {
-                Either::Left(a) => Either::Left(Pin::new_unchecked(a)),
-                Either::Right(b) => Either::Right(Pin::new_unchecked(b)),
-            }
-        }
+        self.as_pin_mut()
     }
 
     /// Convert `Pin<&Either<L, R>>` to `Either<Pin<&L>, Pin<&R>>`,
--- a/futures-util/src/future/either.rs
+++ b/futures-util/src/future/either.rs
@@ -33,10 +33,6 @@ pub enum Either<A, B> {
 }
 
 impl<A, B> Either<A, B> {
-    fn project(self: Pin<&mut Self>) -> Either<Pin<&mut A>, Pin<&mut B>> {
-        self.as_pin_mut()
-    }
-
     /// Convert `Pin<&Either<L, R>>` to `Either<Pin<&L>, Pin<&R>>`,
     /// pinned projections of the inner variants.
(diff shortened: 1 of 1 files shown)


futures-util/src/future/either.rs
```diff
@@ -33,11 +33,31 @@ pub enum Either<A, B> {
 }
 
 impl<A, B> Either<A, B> {
-    fn project(self: Pin<&mut Self>) -> Either<Pin<&mut A>, Pin<&mut B>> {
+    /// Convert `Pin<&Either<L, R>>` to `Either<Pin<&L>, Pin<&R>>`,
```
thomaseizinger, line 36:
```suggestion
    /// Convert `Pin<&Either<A, B>>` to `Either<Pin<&A>, Pin<&B>>`,
```

futures-util/src/future/either.rs
```diff
@@ -33,11 +33,31 @@ pub enum Either<A, B> {
...
+    }
+
+    /// Convert `Pin<&mut Either<L, R>>` to `Either<Pin<&mut L>, Pin<&mut R>>`,
+    /// pinned projections of the inner variants.
+    pub fn as_pin_mut(self: Pin<&mut Self>) -> Either<Pin<&mut A>, Pin<&mut B>> {
+        // SAFETY: `get_unchecked_mut` is fine because we don't move anything.
+        // We can use `new_unchecked` because the `inner` parts are guaranteed
+        // to be pinned, as they come from `self` which is pinned, and we never
+        // offer an unpinned `&mut L` or `&mut R` through `Pin<&mut Self>`. We
+        // also don't have an implementation of `Drop`, nor manual `Unpin`.
```
thomaseizinger, line 56:
```suggestion
    /// Convert `Pin<&mut Either<A, B>>` to `Either<Pin<&mut A>, Pin<&mut B>>`,
    /// pinned projections of the inner variants.
    pub fn as_pin_mut(self: Pin<&mut Self>) -> Either<Pin<&mut A>, Pin<&mut B>> {
        // SAFETY: `get_unchecked_mut` is fine because we don't move anything.
        // We can use `new_unchecked` because the `inner` parts are guaranteed
        // to be pinned, as they come from `self` which is pinned, and we never
        // offer an unpinned `&mut A` or `&mut B` through `Pin<&mut Self>`. We
        // also don't have an implementation of `Drop`, nor manual `Unpin`.
```

taiki-e:
Thanks! I'll merge this once docs/comments are updated.

thomaseizinger:
> Thanks! I'll merge this once docs/comments are updated.

Just the two that I marked already or do you need something else too? :)

taiki-e:
> > Thanks! I'll merge this once docs/comments are updated.
> 
> Just the two that I marked already or do you need something else too? :)

I meant the former.

thomaseizinger:
> > > Thanks! I'll merge this once docs/comments are updated.
> > 
> > Just the two that I marked already or do you need something else too? :)
> 
> I meant the former.

Done!

taiki-e (approved):

PR #2693, opened 2023-01-19
Implement `FusedStream` for all streams in `ReadyChunks`

jongiddy:
`ReadyChunks` fuses the inner stream, so `FusedStream` can be implemented for all stream types, not just those that initially implement `FusedStream`.


taiki-e (approved):
Thanks!

PR #2704, opened 2023-02-03
Poll `Select` futures without moving them

EFanZh:
The current implementation of `Select` moves the child futures out of the parent future before polling them, which maybe too costly if the sizes of child futures are too big or the future needs multiple polls before finally ready.

This PR changes the behavior to only move child futures once before ready, which could help improve performance. Here is a benchmark I wrote: <https://gist.github.com/EFanZh/099bb8f6113deab71ed927674005abe7>. With two 800 bytes future (on 64-bit platform) that each requires 100 polls before ready. I got more than 2x speed improvement with the new implementation.

Here is a link for comparing the assembly code of two implementations: <https://godbolt.org/z/hhK1fGE5e>.

This PR adds two `unwraps()`s, but it seems the compiler can optimize them out, or do I need to change them to [`unwrap_unchecked()`](https://doc.rust-lang.org/stable/std/option/enum.Option.html#method.unwrap_unchecked) to make sure no extra cost is introduced?

### DIFF
--- a/futures-util/src/future/select.rs
+++ b/futures-util/src/future/select.rs
@@ -99,17 +99,16 @@ where
     type Output = Either<(A::Output, B), (B::Output, A)>;
 
     fn poll(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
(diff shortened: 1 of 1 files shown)


taiki-e (approved):
Thanks!

PR #2708, opened 2023-02-13
Ensure unreachable branch is eliminated

EFanZh:
With #2704, I introduced two `unwrap()` calls in `Select::poll`, but I found that with `opt-level=z`, the compiler won’t be able to optimize these `unwrap()` calls, so maybe changing them to `unwrap_unchecked()` is a better idea? See <https://godbolt.org/z/cef118zKe> for comparison with `opt-level=z`.

Also, since `unwrap_unchecked()` is introduced in Rust 1.58.0, I have to implement this function manually.

### DIFF
--- a/futures-util/src/future/select.rs
+++ b/futures-util/src/future/select.rs
@@ -99,14 +99,24 @@ where
     type Output = Either<(A::Output, B), (B::Output, A)>;
 
     fn poll(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
+        /// When compiled with `-C opt-level=z`, this function will help the compiler eliminate the `None` branch, where
+        /// `Option::unwrap` does not.
+        #[inline(always)]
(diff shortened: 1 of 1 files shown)


taiki-e:
Hmm, given that opt-level=s works, I feel this is like a compiler bug. https://godbolt.org/z/sPTdMo1Kd

seanmonstar (commented):
Unless the performance gains are critical, I'd tend to err on keeping the safer version, and hoping that the compiler can be made smart enough eventually.

EFanZh:
It seems that the `#[inline]` attribute on `Option::unwrap` is not enough for the compiler to apply inlining, and if I implement `unwrap` manually (panic instead of UB) with `#[inline(always)]`, the compiler can do the optimization: <https://godbolt.org/z/GMq1jaMd5>. Is this change acceptable?

taiki-e (approved):
Thanks! The new approach looks good to me. (It is enough simple and has no unsafe code)

PR #2710, opened 2023-02-18
add AbortHandle::is_aborted()

### LINKED ISSUE #2709: Feature Request: Add `AbortHandle::is_aborted()`
Currently there is already a public [`Abortable::is_aborted()`](https://docs.rs/futures/latest/futures/future/struct.Abortable.html#method.is_aborted) method, and there seems no obstacle in adding a `AbortHandle::is_aborted()`

# Motivation
Sometimes `Abortable` needs to be accompanied by cooperative cancellation mechanism such as a atomic boolean flag. Rather than using another repetitive atomic variable, the `AbortInner` already hosts such atomic variable and the `AbortHandle` seems to be a good way to expose it.

Here is a hypothetical scenario where `Abortable` needs to be accompanied by cooperative cancellation mechanism. Suppose we have a blocking `Mutex` protecting a very tiny shared state, and we wish to stop another task from modifying it by calling its `AbortHandle`

```rust
// Aborter
// ...
let (new_abort_handle, new_abort_reg) = AbortHandle::new_pair();
let cooperative_flag = Asc::new(AtomicBool::new(false)); // new_abort_handle.clone();
{
    let mut guard = mutex.lock();
    // Take out the old task's handle
    let old_abort_handle = std::mem::replace(guard.abort_handle, new_abort_handle);
    old_abort_handle.abort();
    // A modification to the cooperative flag is needed here
    drop(guard);
}
// Spawn a new task using the new handle
tokio::spawn(Abortable::new(some_async_function(args, cooperative_flag), new_abort_reg);
// We wish no modifications to the shared state from the old task from here on

// Abortee
let mut guard = mutex.lock();
// A cooperative checking is needed here
write_into_shared_state(guard.state);
drop(guard);
```
Since we used a blocking `Mutex` to protect our state (which is [warranted](https://tokio.rs/tokio/tutorial/shared-state#on-using-stdsyncmutex) in an async scenario) , it is very likely that our target Future is blocked on the mutex call when the `AbortHandle::abort()` was called. To solve this issue we would need a cooperative checking in our code. Exposing `AbortHandle::is_aborted()` would save the user from another atomic boolean flag.

ZhennanWu:
Closes #2709

### DIFF
--- a/futures-util/src/abortable.rs
+++ b/futures-util/src/abortable.rs
@@ -185,7 +185,11 @@ impl AbortHandle {
 
     /// Checks whether [`AbortHandle::abort`] was *called* on any associated
     /// [`AbortHandle`]s, which includes all the [`AbortHandle`]s linked with
-    /// the same [`AbortRegistration`].
+    /// the same [`AbortRegistration`]. This means that it will return `true`
+    /// even if:
+    /// * `abort` was called after the task had completed.
+    /// * `abort` was called while the task was being polled - the task may still be running and
+    /// will not be stopped until `poll` returns.
     ///
     /// This operation has a Relaxed ordering.
     pub fn is_aborted(&self) -> bool {
--- a/futures-util/src/abortable.rs
+++ b/futures-util/src/abortable.rs
@@ -182,4 +182,13 @@ impl AbortHandle {
         self.inner.aborted.store(true, Ordering::Relaxed);
         self.inner.waker.wake();
     }
+
+    /// Checks whether [`AbortHandle::abort`] was *called* on any associated 
+    /// [`AbortHandle`]s, which includes all the [`AbortHandle`]s linked with 
+    /// the same [`AbortRegistration`].
+    /// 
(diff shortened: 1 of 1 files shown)


futures-util/src/abortable.rs
```diff
@@ -182,4 +182,13 @@ impl AbortHandle {
         self.inner.aborted.store(true, Ordering::Relaxed);
         self.inner.waker.wake();
     }
+
+    /// Checks whether [`AbortHandle::abort`] was *called* on any associated
+    /// [`AbortHandle`]s, which includes all the [`AbortHandle`]s linked with
+    /// the same [`AbortRegistration`].
+    ///
+    /// This operation has a Relaxed ordering.
```
taiki-e, line 190: It may be preferable to mention the caveat mentioned in the Abortable::is_aborted docs ("This means that it will return `true` even if ...").

taiki-e (commented):
Thanks!

taiki-e (approved):

Issue #2541, opened 2021-12-25
Document join is concurrent but not parallel


elbaro:
It should be documeneted in `join`, `join3`, `join4`, ..`join5` that they are concurrently run not run in parallel,
and how to run them in parallel.

https://docs.rs/tokio/latest/tokio/macro.join.html

akauppi:
This entry (from 2021) refers to Tokio, but this repo is futures-rs. Perhaps these were closer at that point, but currently I think this could be closed?

elbaro:
This issue is about futures-rs. The link is an example of how Tokio documented tokio::join!.

PR #2714, opened 2023-03-03
Improve comment for `FuturesOrdered` and `FuturesUnordered`

th4s:
For `FuturesOrdered` there is a comment that says the futures would run in parallel, which is misleading.

This PR makes clear that futures for `FuturesOrdered` and `FuturesUnordered` run to completion concurrently instead of parallel.

### DIFF
--- a/futures-util/src/stream/futures_ordered.rs
+++ b/futures-util/src/stream/futures_ordered.rs
@@ -60,8 +60,9 @@ where
 ///
 /// This "combinator" is similar to [`FuturesUnordered`], but it imposes a FIFO order
 /// on top of the set of futures. While futures in the set will race to
-/// completion in parallel, results will only be returned in the order their
-/// originating futures were added to the queue.
+/// completion, results will only be returned in the order their originating futures
+/// were added to the queue. Note that the futures run to completion concurrently and
+/// not in parallel.
 ///
 /// Futures are pushed into this queue and their realized values are yielded in
 /// order. This structure is optimized to manage a large number of futures.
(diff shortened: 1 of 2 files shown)


taiki-e:
Thanks for the PR! As https://github.com/rust-lang/futures-rs/issues/2541 mentioned, it might be a good idea to mention how to run them in parallel.

Also, could you improve the comments on the APIs mentioned in #2541 as well?

taiki-e:
Hmm, I’m not sure if there is a good way to do that without depending on an executor.

th4s:
So by "mention how to run them in parallel" you mean I should just add the comment that one needs to invoke `spawn` on a runtime, or what do you have in  mind?

vasilakisfil:
I think this PR is helpful. There is some vagueness in existing docs as to what these 2 types are, and I think this PR makes it better.

> As https://github.com/rust-lang/futures-rs/issues/2541 mentioned, it might be a good idea to mention how to run them in parallel.

@taiki-e I am also a bit confused by your comment. In order to run them in parallel you would have to spawn tasks, and this feels a bit out of context.

taiki-e:
> I am also a bit confused by your comment. In order to run them in parallel you would have to spawn tasks, and this feels a bit out of context.

The author of https://github.com/rust-lang/futures-rs/issues/2541 seems to think that a document that states "concurrent but not parallel" should also state "how to run in parallel" and I basically agree with that.

However, if that is difficult, simply stating "concurrent but not parallel" would be fine.

PR #2715, opened 2023-03-10
tests: restore disabled tests and benches for BiLock

LizardWizzard:
Hi! I was curious about BiLock internals (came there from its usage in split) so the easiest way to play with something is to poke into corresponding tests. Found out that tests and benchmarks for BiLock are disabled so I went ahead with reviving them to be up to date with current landscape.

Thank you your great work on the library!


LizardWizzard:
Miri complains:

```
error: unsupported operation: integer-to-pointer casts and `ptr::from_exposed_addr` are not supported with `-Zmiri-strict-provenance`
   --> /home/runner/work/futures-rs/futures-rs/futures-util/src/lock/bilock.rs:178:31
    |
178 |                 Box::from_raw(n as *mut Waker).wake();
    |                               ^^^^^^^^^^^^^^^ integer-to-pointer casts and `ptr::from_exposed_addr` are not supported with `-Zmiri-strict-provenance`
    |
    = help: use Strict Provenance APIs (https://doc.rust-lang.org/nightly/std/ptr/index.html#strict-provenance, https://crates.io/crates/sptr) instead
```

Will it be desirable to go ahead with https://crates.io/crates/sptr as suggested in the error message?

LizardWizzard:
Thanks @taiki-e! I've rebased it on top of current master.

CI / cargo +beta build (pull_request)  failed once with:

```
warning: spurious network error (10 tries remaining): [35] SSL connect error (error:1408F119:SSL routines:ssl3_get_record:decryption failed or bad record mac)
...
error: failed to verify the checksum of `byteorder v1.4.3`
```

futures-util/benches/bilock.rs
```diff
@@ -0,0 +1,70 @@
+#![feature(test)]
+
+extern crate test;
+
+#[cfg(feature = "bilock")]
+mod bench {
```
taiki-e, line 6: Could you change this cfg to a file-level cfg and flatten the module?

futures/tests/bilock.rs
```diff
@@ -0,0 +1,105 @@
+#[cfg(feature = "bilock")]
+mod tests {
```
taiki-e, line 2: Same here.

taiki-e (commented):
Thanks!

futures/tests/bilock.rs
LizardWizzard: Done

LizardWizzard:
Thanks for review! Fixed the comments, please take another look

taiki-e (approved):

PR #2722, opened 2023-03-16
Fix unreachable match arm in FillBuf being reachable

jessa0:
I hit [this `unreachable!()`](https://github.com/rust-lang/futures-rs/blob/8253b784fd45c832ad208d5b8e259f639fc42684/futures-util/src/io/fill_buf.rs#L41) inside `FillBuf` with an `AsyncRead` implementation (wrapped in a `BufReader`) which calls a Javascript callback thru `neon` (which is an async operation) for the actual read, which means every "new" `poll_read` (i.e. either the first call ever or the first call after a `Ready` is returned) will return `Pending` at least once, including after EOF is hit, which is what triggers this bug. Of course, the `AsyncRead` implementation could track when EOF is hit and return `Ready(0)` forever after that, which is indeed what I used as a workaround, but I can't find anything in the `AsyncRead` documentation which actually requires implementors to do this.

Alternatively to this fix, the `AsyncRead` documentation could require implementors to always return `Ready(0)` after an EOF, and the `unreachable!()` changed to a `panic!()`, but that seems like more potential downstream work than merging this fix, and seems easy to screw up.

Another similar alternative would be for the `AsyncBufRead` documentation to require implementors to always return `Ready(&[])` after an EOF.

The added test triggers the panic if run without the bug fix.

### DIFF
--- a/futures-util/src/io/fill_buf.rs
+++ b/futures-util/src/io/fill_buf.rs
@@ -30,6 +30,11 @@ where
         let reader = this.reader.take().expect("Polled FillBuf after completion");
 
         match Pin::new(&mut *reader).poll_fill_buf(cx) {
+            // The reader can try to fill its buffer again if it returned no data, according to the
+            // AsyncBufRead contract. Don't poll again in that case, in order to make the
+            // Poll::Pending match arm below truly unreachable.
+            Poll::Ready(Ok([])) => Poll::Ready(Ok(&[])),
+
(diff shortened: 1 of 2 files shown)


jessa0:
That would be better indeed. If you're willing to merge that tokio commit, want me to bring it into this PR for you?

jessa0:
Fixed by #2801

PR #2723, opened 2023-03-18
Use `Waker::will_wake()` to avoid a cloning op

daxpedda:
> This uses [`Waker::will_wake()`](https://doc.rust-lang.org/stable/std/task/struct.Waker.html#method.will_wake) to check if the old waker will awaken the same task as the new one, which could save a clone operation.

This is analogous to https://github.com/smol-rs/atomic-waker/pull/12.


taiki-e (approved):
Thanks!

PR #2726, opened 2023-03-20
fu: always replace inner wakers

### LINKED ISSUE #2725: nested flatten_unordered hangs
I've (also) run into [this issue that was reported on users.rust-lang](https://users.rust-lang.org/t/nesting-streamext-flatten-unordered-producing-unexpected-results/90551/1). 

I suspect this might be the right forum to contact the maintainers of the `futures` crate, so here's an issue as well and a cleaned up repro. Namely, this hangs:

```rust
#[tokio::main]
async fn main() {
    use futures::future;
    use futures::stream;
    use futures::StreamExt;
    use tokio::time::interval;
    use tokio_stream::wrappers::IntervalStream;

    let inner = IntervalStream::new(interval(std::time::Duration::from_millis(500)))
        .map(|_| stream::once(future::ready("VALUE")))
        .flatten_unordered(None);

    let mut stream = stream::once(future::ready(inner)).flatten_unordered(None);

    while let Some(value) = stream.next().await {
        dbg!(&value);
    }
}
```

This seems like it should evidently just yield values forever (by creating a new stream-of-1-value every 500ms), but as-is it does not and it just hangs after the first value.

olegnn:
Fixes #2725.
As nested `FlattenUnordered` will be placed into a new future inside of `FuturesUnordered` every time its inner stream emits an item, its base stream's waker will point to an outdated task, and this will result in an inability to [propagate the call](https://github.com/rust-lang/futures-rs/blob/master/futures-util/src/stream/futures_unordered/task.rs#L71) properly.

### DIFF
--- a/futures-util/src/stream/stream/flatten_unordered.rs
+++ b/futures-util/src/stream/stream/flatten_unordered.rs
@@ -414,8 +414,13 @@ where
             }
         };
 
+        // Safety: now state is `POLLING`.
+        let stream_waker = unsafe { WrappedWaker::replace_waker(this.stream_waker, cx) };
+        let inner_streams_waker =
+            unsafe { WrappedWaker::replace_waker(this.inner_streams_waker, cx) };
+
         if poll_state_value & NEED_TO_POLL_STREAM != NONE {
-            let mut stream_waker = None;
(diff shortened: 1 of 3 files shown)


taiki-e (approved):
Thanks!

PR #2728, opened 2023-03-22
Don't ignore empty state polling

olegnn:
To solve the same problem as #2726: if the stream was moved to another task and polled while its state is empty, wakers must be replaced.


taiki-e:
Is it possible to add a test case for this?

taiki-e (approved):
Thanks!

PR #2730, opened 2023-03-23
Update to syn-2

maurer:
Updates futures-rs to use syn-2.x


taiki-e (approved):
Thanks!

taiki-e:
Published in 0.3.28.

PR #2742, opened 2023-05-08
futures-util: Update Future combinators to preserve Clone

zjijz:
This updates most `Future` combinators to preserve `Clone` when available on the base `Future`.

For motivation, imagine you have some complicated `Future` that is not `Clone` and requires `.shared()` to properly share it. Then imagine you have a library function that is meant to bundle together a bunch of combinators to fulfill some semantic purpose. That library function will have to call `.shared()` if it wants to try to guarantee the return `Future` is `Clone`, but this might be suboptimal if the input `Future` was already `Clone`, plus it obfuscates the `.shared()` allocation. With this change, you can instead require `Future + Clone` on the input `Future` and have a guarantee the output will be `Clone` as well.

The hold-out `Future` implementations are:
- `Remote` / `RemoteHandle` due to their use of `futures_channel::oneshot::{Sender, Receiver}`. This seems like it is by design that these cannot be `Clone`.
- `JoinAll` / `TryJoinAll` due to their use of `Stream` combinators, but also that it seems unlikely that people would expect them to offer `Clone` since they are used to perform a potentially costly sync barrier that would probably be desired to happen only once.

For the hold-outs, the existing pattern of using `.shared()` allows for `Clone`, and follows the intended semantics of those combinators.

Some combinators that might not make the most sense to preserve `Clone`:
- `IntoStream`
- `TryFlattenStream`

If these changes make sense, I think it would also make sense to apply them to `Stream` combinators as well (although I don't see myself utilizing this property as much with them). If that is the case, these `Future` -> `Stream` combinators make sense to preserve `Clone`.

Tested:
- `cargo doc`.
- `cargo fmt`.
- `cargo test --all-features`.

### DIFF
--- a/futures-util/src/future/future/catch_unwind.rs
+++ b/futures-util/src/future/future/catch_unwind.rs
@@ -9,7 +9,7 @@ use pin_project_lite::pin_project;
 
 pin_project! {
     /// Future for the [`catch_unwind`](super::FutureExt::catch_unwind) method.
-    #[derive(Debug)]
+    #[derive(Clone, Debug)]
     #[must_use = "futures do nothing unless you `.await` or poll them"]
     pub struct CatchUnwind<Fut> {
         #[pin]
--- a/futures-util/src/future/future/flatten.rs
+++ b/futures-util/src/future/future/flatten.rs
@@ -9,7 +9,7 @@ use pin_project_lite::pin_project;
 
 pin_project! {
     #[project = FlattenProj]
-    #[derive(Debug)]
+    #[derive(Clone, Debug)]
     pub enum Flatten<Fut1, Fut2> {
         First { #[pin] f: Fut1 },
         Second { #[pin] f: Fut2 },
--- a/futures-util/src/future/future/fuse.rs
+++ b/futures-util/src/future/future/fuse.rs
@@ -5,7 +5,7 @@ use pin_project_lite::pin_project;
 
 pin_project! {
     /// Future for the [`fuse`](super::FutureExt::fuse) method.
-    #[derive(Debug)]
+    #[derive(Clone, Debug)]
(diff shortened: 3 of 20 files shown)


zjijz:
#2743 is a separate fix for the missing `#[derive(Debug)]` warning.

zjijz:
@taiki-e Any thoughts on this?

This doesn't seem to impose any oddities for me when using the `await` path to consume a `Future` (possible wrapped in combinators) since all cloning would have to happen before the `Future` was ever polled once.

However, with the ability to easily poll with things like `FutureExt::poll_unpin` or `Box::pin` + `Future::poll`, I'm worried that some of the `Future` implementations weren't written with the possibility of cloning in mind and might clone a `Future`'s intermediate state instead of fully recreating the initial version (which might be what people expect, but seems to be an edge case when I think about it). `Lazy` is one example of this that if cloned after a call to `poll_unpin` would return a `Lazy` that always panics on the next poll.

taiki-e:
Thanks for the PR. Which future combinator is actually needed for your use case?

zjijz:
`Map` is the big one for my use case. I'm not sure the current implementation lends itself to `Clone` being thrown on it to where it can potentially panic if there was a `clone` thrown in between or after manual polls. I can also write my own `Map` that is less panicky about polling after completion if you don't want to expose the potential footgun here.

afeistel:
stateful `Future`s/`Stream`s combinators really shouldn't be `Clone` since that means deciding whether or not `clone`ing them means both `clone`s receive the value (decided by either cloning or dropping the state)

and I believe we want, for example, both multicast and MPMC channels to be valid `Clone` `Stream`s, and choosing one of the ways to do `Clone` will invalidate the other

PR #2746, opened 2023-05-22
Only return `Poll::Ready` from `Sender::poll_flush` if the channel is empty

### LINKED ISSUE #2504: Question about Sink's poll_flush and channels: mpsc senders don't fully flush the buffer before returning Poll::Ready
Maybe this is just a misunderstanding of what `poll_flush` actually means, but from the documentation I would expect `poll_flush` of the `Sink` trait to only return `Poll::Ready` once all buffered messages have been received by the `Receiver` of the channel, meaning the buffer is empty.

https://github.com/rust-lang/futures-rs/blob/939614bb6fc5de9c186247a18615b19babcddc29/futures-sink/src/lib.rs#L93-L95

Currently the `Sink` implementation for `Sender` and `UnboundedSender` don't do that though, making `poll_flush` almost equivalent to `poll_ready`. See:
https://github.com/rust-lang/futures-rs/blob/939614bb6fc5de9c186247a18615b19babcddc29/futures-channel/src/mpsc/sink_impl.rs#L17-L25

In the Unbounded case, it doesn't flush anything at all:
https://github.com/rust-lang/futures-rs/blob/939614bb6fc5de9c186247a18615b19babcddc29/futures-channel/src/mpsc/sink_impl.rs#L44-L46

Is this intentional? And if so, how does this relate to the documentation of the `Sink` trait? Maybe the documentation can be updated to clarify this?

thomaseizinger:
I've found it to be very surprising behaviour that `poll_flush` on a `Sender` returns `Poll::Ready` despite the channel not being empty. This PR attempts to fix this by checking the number of messages in the channel and parking the task if is > 0.

Resolves: https://github.com/rust-lang/futures-rs/issues/2504.

### DIFF
--- a/futures-channel/src/mpsc/mod.rs
+++ b/futures-channel/src/mpsc/mod.rs
@@ -632,7 +632,7 @@ impl<T> BoundedSenderInner<T> {
 
         // If there are messages in the channel, we must park the task unconditionally.
         self.sender_task.lock().unwrap().task = Some(cx.waker().clone());
-        return Poll::Pending;
+        Poll::Pending
     }
 
     /// Returns whether the senders send to the same receiver.
--- a/futures-channel/src/mpsc/mod.rs
+++ b/futures-channel/src/mpsc/mod.rs
@@ -616,6 +616,25 @@ impl<T> BoundedSenderInner<T> {
         self.poll_unparked(Some(cx)).map(Ok)
     }
 
+    /// Polls the channel to determine if it is empty.
+    ///
+    /// # Return value
+    ///
+    /// This method returns:
+    ///
+    /// - `Poll::Ready(()` if there are no messages in the channel;
+    /// - `Poll::Pending` if there are messages in the channel.
+    fn poll_is_empty(&mut self, cx: &mut Context<'_>) -> Poll<()> {
+        let state = decode_state(self.inner.state.load(SeqCst));
+        if state.num_messages == 0 {
+            return Poll::Ready(());
+        }
+
+        // If there are messages in the channel, we must park the task unconditionally.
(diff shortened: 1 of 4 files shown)


thomaseizinger:
Miri is detecting a deadlock but I am not quite sure why. Help is much appreciated.

thomaseizinger:
Okay, I think I found the problem. `SinkExt::send` implies `flush` which caused a deadlock if we have a task that suspends on `send` and doesn't at the same time read from the receiver.

I've replaced that with `feed` which makes it work again.

This makes me wonder whether this might perhaps be a fairly intrusive change for the ecosystem. I still think that this is the more correct behaviour but I wonder, how many people are relying on `SinkExt::send` essentially being equivalent to `SinkExt::feed`.

taiki-e:
Thanks for the PR. 

> Not sure why https://github.com/rust-lang/futures-rs/actions/runs/5046514587/jobs/9052136457?pr=2746#step:5:326 is failing.

armv7 failure is unrelated to this PR (should be fixed by rebase).

> This makes me wonder whether this might perhaps be a fairly intrusive change for the ecosystem.

Yeah, I'm not open to introducing patches that may introduce hard-to-debug deadlocks.

thomaseizinger:
> > This makes me wonder whether this might perhaps be a fairly intrusive change for the ecosystem.
> 
> Yeah, I'm not open to introducing a patch that may introduce hard-to-debug deadlocks. (even if it is in a breaking release)

To be fair, it only occurs if both sender and receiver are polled by the _same_ task. I think that is a rather unlikely situation because you could also just push and pop a `Vec` then instead of using a channel.

0x7CFE:
Any chance this will be merged someday?

PR #2749, opened 2023-06-08
Add `num_messages` method for `UnboundedSender`

thement:
Add method to inspect how many messages are enqueued in the message queue.


taiki-e:
Thanks for the PR. I think such a method would usually be called `len`. (e.g., [crossbeam](https://docs.rs/crossbeam/latest/crossbeam/channel/struct.Sender.html#method.len))

Also, please submit PR to master, not 0.3, except for barkports.

thement:
Sorry! I'll redo the PR.

PR #2750, opened 2023-06-09
Add `len` method for `UnboundedSender`

thement:
Add `len` method to inspect how many messages are enqueued in the message queue.


taiki-e:
Thanks for the PR. Could you fix a clippy warning? (Other CI failures should be fixed by rebase)

thement:
Fixed.

taiki-e (approved):

PR #2751, opened 2023-06-11
add mapped_futures

StoicDeveloper:
EDIT: This implementation of this PR has been completely reworked, though the interface is the same. In short, the new version makes one very minor change to the existing FuturesUnordered implementation. All of the mapping features are instead introduced by wrapping FuturesUnordered in a struct that also includes a `HashSet< HashTask<K, HashFut<K, Fut>>>`, where `HashTask { inner: *const Task, key: K }` and HashFut exists to know the associated key of a completed future, so that the corresponding task can be removed from the set. See comment below for details. 

Adds the functionality asked for in this issue: https://github.com/rust-lang/futures-rs/issues/2329. I wanted this feature, searched for it, found only people wishing that the feature existed, and so decided to write it myself. A previous attempt that wrapped FuturesUnordered didn't work; I wanted to avoid anything `unsafe` which I haven't used before. Having worked with it now though, I can see that unsafe rust is really just C with some guardrails.

The motivation is to have a HashMap that can run futures. Adds the MappedFutures struct, which supports inserting, `get_mut`ing, cancelling, removing, and replacing futures by key, and running them in an identical style as FuturesUnordered.

Internally it uses a HashSet and adds a hashable wrapper around the Task struct from FuturesUnordered.
In terms of ergonomics, the API is perhaps a bit confused, including both the `insert(key, fut) -> bool` and `replace(key, fut) -> Option<Fut>` methods which are more similar to the methods of HashSet than HashMap. The source of this confusion is that some methods can only work if the Future is Unpin; ie. anything that would return an owned future or future reference.

Some justifications:
- I set the output for the stream as `(key, fut_output)` to be similar to a [LinkedHashMap](https://docs.rs/linked-hash-map/latest/linked_hash_map/), which in some ways this library is similar to (but with `next().await` in place of `pop_front()`. I think it will be typical for a consumer to want to know the key of whichever future has completed, though of course any future could be mapped as such prior to insertion.
- an earlier version used HashMap internally, which was the obvious solution for a struct which is "A hashmap that runs futures", but returning the key with the output required having a copy of the key in both the HashMap and in the Task, and so the Key had to be Clone. I added a wrapper HashTask for `Arc<Task>` which is placed in a HashSet to get around this requirement. Its possible that a different solution would have worked (like pointers to and from a Key wrapper and Task), but this method seemed the most expedient.

Potential improvements:
- Allow selecting which hasher to use, like how HashSet does. This should be a pretty easy change.

I did already publish these changes in a crate which I will yank if this gets merged. https://crates.io/crates/mapped_futures


StoicDeveloper:
I'm also considering further additions for more complex data structure support, such as: 
- MultiMapFutures which allows a key to index for one or more futures
- BiMapFutures which allows futures  to be accessed from either of two key types (really the BiMap is between Key-Value pairs, neither of which is a future, and each pair is associated with a future) where each key and value must be unique
- BiMultiMapFutures which is like BiMapFutures, but allows duplicates among keys and values, so that each key or value may index to one or more futures, though each key-value pair will index to only zero or one futures.

taiki-e:
Thanks for the PR! It looks like a lot of the code added here is a copy from FutureUnordered. Is it possible to replace some of them with re-exports or wrappers for the types and functions used in FutureUnordered?

StoicDeveloper:
> Thanks for the PR! It looks like a lot of the code added here is a copy from FutureUnordered. Is it possible to replace some of them with re-exports or wrappers for the types and functions used in FutureUnordered?

This was my first approach, but it didn't work out. This PR required several small internal changes to FuturesUnordered, such that I don't think it would be possible to implement this as a simple wrapper. However, it may be possible to do so by making changes within the actual FuturesUnordered module that would make it easier to reuse. I'll look into that.

StoicDeveloper:
@taiki-e 

I've implemented a new module futures_keyed which contains most of the code that was in futures_unordered. futures_unordered and mapped_futures now consume the API of futures_keyed. There is more work to do, and there are some things about this implementation that are not ideal

Work remaining:
- implement more mapping types, like mapped_streams, bi_multi_map_futures, and bi_multi_map_streams, which I have implemented in the mapped_futures crate, but I wanted to check whether what I have so far is acceptable to this crate's maintainers before going further
- other housekeeping, remove commented code, add more comments and documentation, implement Debug on some structs to silence compiler warnings

Things that aren't ideal
- The API for futures_keyed adds the ReleasesTask trait, and consumers that contain KeyedFutures must provide an implementor of this trait. I personally find this to be very unergonomic, but it was the only way I could think of to make this change. The API for futures_unordered is of course unchanged, but internally it now creates a DummyStruct that implements ReleasesTask using logic that does nothing, since the existing FuturesKeyed::release_task() contains all necessary logic. This trait was necessary so that a consumer can remove residual elements from its more-complex data structure if release_task is called. Ex. MappedFutures uses a HashSet of Tasks, but calling release_task on the Task will not remove the corresponding struct from the HashSet; now release_task() will handle that by calling code from the trait.
- I guess that's all

Please let me know of any problems with this implementation; if you find none then I will proceed with implementing more consumers of futures_keyed

taiki-e:
> I've implemented a new module futures_keyed which contains most of the code that was in futures_unordered.

Thanks.

> * implement more mapping types, like mapped_streams, bi_multi_map_futures, and bi_multi_map_streams, which I have implemented in the mapped_futures crate, but I wanted to check whether what I have so far is acceptable to this crate's maintainers before going further

For now, I would like to add a minimum (mapped_futures and mapped_streams; other ones are implemented on top of them, right?).

> I couldn't think of a better name than futures_keyed; perhaps futures_unordered_internal. It might be better to not say "keyed" since there is the potential to use the new `key` field of `Task` for things other than keys.

Yeah, futures_unordered_internal would be better.

futures-util/src/stream/mapped_futures/mod.rs
```diff
@@ -0,0 +1,473 @@
...
+            }
+        }
+        None
+    }
+
+    /// Get a shared reference to the mapped future.
+    pub fn get(&mut self, key: &K) -> Option<&Fut>
+    where
+        Fut: Unpin,
+    {
```
taiki-e, line 239: `Pin` always implements `Deref` (i.e., we can always safely get `&T` from `Pin<&T>`/`Pin<&mut T>`), so I think we can remove the `Unpin` bound here.

StoicDeveloper:
@taiki-e 
I've addressed your comments and some merge conflicts resulting from work over the past few months. Let me know if there's anything else; I should be able to address comments more promptly in the near future.

futures-util/src/stream/mapped_futures/mod.rs
```diff
@@ -0,0 +1,464 @@
...
+    }
+
+    /// Insert a future into the set and return the displaced future, if there was one.
+    ///
+    /// This method adds the given future to the set. This method will not
+    /// call [`poll`](core::future::Future::poll) on the submitted future. The caller must
+    /// ensure that [`MappedFutures::poll_next`](Stream::poll_next) is called
+    /// in order to receive wake-up notifications for the given future.
+    /// Returns true if another future was ma
+    pub fn replace(&mut self, key: K, future: Fut) -> Option<Fut>
```
awused, line 151: Having both `replace` and `insert` (and remove + cancel below) seems redundant with the only difference seems to be if the value is dropped before the function returns or not. std's maps only have `insert`, which behaves as `replace` here, have you considered just having `insert` with `replace`'s signature?

StoicDeveloper: The reasoning for having the separate methods is covered in the PR, nonetheless I'm open to ideas for how these functions could be renamed to make their intention more clear. Ideally each should be a single word, and the result, and difference between the methods should be clear at a glance, without reading the full signature. For remove & cancel it is clear, since the mental model of removing something includes that you have a thing after doing so, and not so with cancel. But is the difference as clear with replace & insert?

StoicDeveloper:
Also, the crate I published evidently has unsound causes this bug in tokio: https://github.com/StoicDeveloper/mapped_futures/issues/4
That issue will obviously have to be resolved here before this PR can proceed.

StoicDeveloper:
I found unsoundness by running this PR's tests through miri. The unsoundness comes from the potential for a dropped task to still have a pointer in the ReadyToRunQueue, which will get dereferenced during polling. Solutions may include:

 -  add to Task Option and Option fields that can be used to set remove an item from the queue if it gets cancelled
 -  drop the future on cancellation, but don't drop the task. Instead check during ReadyToRunQueue dequeue() whether the future was dropped. If so, dequeue again and drop the current task.

I'm leaning towards the latter as the simpler option, and its runtime cost should be negligible, but I'm open to suggestions.

awused:
This probably needs to be rebased after the breaking changes in 0.3.31. I'd really like to see this get merged into futures-rs, as I think it addresses a big gap in functionality, but there hasn't been a comment from an owner in over a year now.

StoicDeveloper:
@taiki-e I've completely reworked this PR, I suspect you'll be more amenable to a merge now; it hardly touches the internals of FuturesUnordered, instead it adds a wrapper type. An overview of the changes

1) Adds dev dependency on tokio 1.25. The timer library I was using was causing problems with the Miri actions step; tokio 1.25 is the most recent tokio version that has `tokio::time::sleep()` and also support edition 2018, which another actions step requires
2) adds mapped_futures module (if std feature enabled) nested within futures_unordered (since it requires access to the new `FuturesUnordered::push_inner(fut: Fut) -> *const Task<Fut>` function. The former `FuturesUnordered::push(fut)` function now just calls `push_inner()` and drops the resulting points. The function was added because MappedFutures needs to also keep track of tasks.
3) MappedFutures uses Arc<Key> in HashTask, which wraps Task, and HashFut, which wraps the generic future type (its possible that could be done with futures::future::Map instead). The key is taken from HashFut when a future completes and is used to remove the task pointer from the set. If a future is removed or cancelled by key, then they key is used to get the task, which is then unliked and released.
4) Unsafety in these additions has to do with handling `Pin` or using `Arc::from_raw(*const task)` in order to access the contained future; in the latter case `Arc::into_raw(task)` is used in order to avoid use-after-free.

Let me know if there are any changes required for merge.

awused:
Please remove me as a reviewer, it looks like I got pulled into this because of old comments and I can't remove myself.

StoicDeveloper:
@taiki-e sorry to ping you again. Is there any obstacle with this PR? Lmk if it can't be merged, at the moment its just dangling, cluttering the open PR list.

futures-util/src/stream/futures_unordered/mapped_futures/mod.rs
```diff
@@ -0,0 +1,569 @@
...
+    pub fn remove(&mut self, key: &K) -> Option<Fut>
+    where
+        Fut: Unpin,
+    {
+        self.task_set
+            .take(key)
+            .and_then(|task| unsafe {
+                // SAFETY:
+                // - If removing the future from task, must do so before releasing task
+                // - Derefernce must be safe; if the task had been released then it would have been
```
Hywan, line 233:
```suggestion
                // - Dereference must be safe; if the task had been released then it would have been
```

PR #2756, opened 2023-06-28
docs: futures-util::StreamExt: cite map and then in docs

marcospb19:


taiki-e:
Thanks for the PR.

I don't think "synchronous alternative"/"asynchronous alternative" is a very clear explanation.

How about something like "if you want to use a closure that returns a value instead of a future"?

futures-util/src/stream/stream/mod.rs
```diff
@@ -337,6 +337,9 @@ pub trait StreamExt: Stream {
     /// wrapped version of it, similar to the existing `map` methods in the
     /// standard library.
     ///
+    /// See [`StreamExt::then`](Self::then) if you want to use a closure that
+    /// returns a future instead of a value.
```
marcospb19, line 341: I force-pushed accordingly, WDYT?

taiki-e (approved):

marcospb19:
I believe CI is failing for unrelated reasons.

PR #2757, opened 2023-07-05
Add `TryStreamExt::try_ready_chunks` as failable version of `StreamExt::ready_chunks`

DzenanJupic:
I noticed, that there's no try version of `ready_chunks`, which would be quite handy.


taiki-e (approved):
Thanks!

PR #2763, opened 2023-07-19
Constrain `Future::Output` in supertrait of `TryFuture`

hlbarber:
The obstruction to

```rust
trait TryFuture: Future<Output = Result<Self::Ok, Self::Error>> {
    type Ok;
    type Error;
}
```

originally stated in https://github.com/rust-lang/futures-rs/issues/1776#issuecomment-518307092 no longer exists.

This PR provides an insight into what taking such a change now would look like now.

In addition to the minimal change to the `TryFuture` trait seen in a88e11011e33df8e6345b1a8527447f3dd5399c5, we additionally:

- Remove `IntoFuture` from the trait bounds as it's no longer needed 60bcaed15e76c7a1b585cb4b39ae0a5fc671a6e7
- Remove `try_poll` as it's now redundant 75e53ff6a9891471714330207fde9c02986f6cba
- Remove `IntoFuture` entirely 4587897773ed4792eaaa656ceb3589ebe2667f14
- Remove sealing from `TryFuture` as any attempted implementation would be met with a conflicting implementation from the blanket implementation 9e15e25d1cab668a0ae03cdd92fccfee90d9b2ad


hlbarber:
A corresponding change can be made to `TryStream`.

taiki-e:
See https://github.com/rust-lang/futures-rs/issues/2508 for the previous discussion.

hlbarber:
This would require an MSRV bump to `rustc` version ??.

taiki-e:
As mentioned in https://github.com/rust-lang/futures-rs/issues/2508#issuecomment-942407374, https://github.com/rust-lang/futures-rs/pull/2344 will unblock this.

Issue #2211, opened 2020-09-09
Should types like `AtomicWaker` and `AbortHandle` be declared `UnwindSafe`


najamelan:
I run into issues with a type I would like to be `UnwindSafe`. It contains an `AbortHandle` which contains a `AtomicWaker` and these types aren't unwindsafe because `AtomicWaker` contains an `UnsafeCell`. 

In the current state of affairs it is up to a user in such situation to go figure out from the futures source code to go figure out whether these are `UnwindSafe` but it was forgotten to mark them as such, or whether they are not `UnwindSafe` but that was not mentioned in the docs.

From a quick look over the code for `Abortable` I can't immediately see how it could be in an invalid state because of a panic in an inopportune moment, but guaranteeing that with certainty for code one hasn't written is quite some work.

So should these types be marked `UnwindSafe` and has anyone enough understanding of their internals to quickly verify that?

taiki-e:
Seems AtomicWaker is not actually UnwindSafe. e.g., If the following clone panicked, AtomicWaker will probably not work well.

https://github.com/rust-lang/futures-rs/blob/c77cd3c9ba19b65db3d1931d2087032752fe9c7d/futures-core/src/task/__internal/atomic_waker.rs#L267

see also https://github.com/tokio-rs/tokio/pull/3689

najamelan:
I suppose that depends on whether the clone can panic? 

https://doc.rust-lang.org/src/core/task/wake.rs.html#190-192

It's all unsafe, and I have never looked into this code, so I have no idea.

taiki-e:
clone can cause panic because [you can pass a function that can cause panic when creating a waker](https://doc.rust-lang.org/stable/core/task/struct.RawWakerVTable.html#clone).

PR #2772, opened 2023-09-09
Implement the `UnwindSafe`/`RefUnwindSafe` traits on channels

tomaka:
To my surprise, this trait isn't implemented on channels, even though channels are kind of a critical component to unwind safety.

See the code comment for an explanation of the reasoning.

I'm not 100% confident in what I'm doing, so feel free to summon unwind-safety experts, if there is any.


taiki-e:
Thanks for the PR. I think we should make AtomicWaker UnwindSafe (#2211) first.

PR #2774, opened 2023-09-25
Remove an outdated comment from docs

WaffleLapkin:
The comment in the example is wrong, since the example is actually wrapped in `async{}` (and this is used, note the `.await` at the end).


taiki-e (approved):
Thanks.

PR #2778, opened 2023-10-09
Update `FuturesOrdered` docs

zachs18:
* `poll` -> `poll_next`
* `push` -> `push_back` (and `push_front`)
* Add intra-doc links.

In futures 0.2.0, `Stream::poll` was renamed to `Stream::poll_next`, but the docs for `FuturesOrdered` weren't updated, so they continue to refer to (now nonexistent) `FuturesOrdered::poll`.

Similarly, `FuturesOrdered::push` was deprecated at some point, but the docs for `FuturesOrdered` still mention it ("Futures are submitted to the queue using `push`").

I also added intra-doc links, of similar frequency as `FuturesUnordered` has.

### DIFF
--- a/futures-util/src/stream/futures_ordered.rs
+++ b/futures-util/src/stream/futures_ordered.rs
@@ -70,18 +70,18 @@ where
 /// large numbers of futures.
 ///
 /// When a `FuturesOrdered` is first created, it does not contain any futures.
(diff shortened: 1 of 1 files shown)


taiki-e (approved):
Thanks.

PR #2782, opened 2023-10-22
Minor improvements to `All` and `Any` adapters

cstyles:
I found a few things in the implementation for these adapters that I thought could be improved while adding `TryAll` and `TryAny`.


taiki-e (approved):
Thanks!

PR #2783, opened 2023-10-22
Add `TryAll` and `TryAny` adapters

cstyles:
Closes #2771.

PR #2794, opened 2023-10-28
Fix docs for `TryAll` and `TryAny` adapters

cstyles:
Just a tiny follow-up to #2783 to fix some issues with the documentation.


taiki-e (approved):
Thanks.

PR #2797, opened 2023-11-07
provide a non-destructive mechanism to determine if a sink/stream are paired

tones111:
The only mechanism I could find to determine if a sink and stream are paired is ```reunite``` which takes ownership of the values.  The mechanism proposed here allows the caller to determine if the values are paired without mutation.

This behavior is useful in the case where a service receives messages from multiple connections and would like to send messages to the non-originating sinks.  Currently there appears to be no way for the server to determine which sink is paired with the connection that originated the message.

I'm not tied to the function names and welcome any suggestions.

### DIFF
--- a/futures-util/src/lock/bilock.rs
+++ b/futures-util/src/lock/bilock.rs
@@ -149,6 +149,11 @@ impl<T> BiLock<T> {
         BiLockAcquire { bilock: self }
     }
 
+    /// Returns `true` only if the other `BiLock<T>` originated from the same call to `BiLock::new`.
+    pub fn is_pair_of(&self, other: &Self) -> bool {
+        Arc::ptr_eq(&self.arc, &other.arc)
+    }
+
     /// Attempts to put the two "halves" of a `BiLock<T>` back together and
     /// recover the original value. Succeeds only if the two `BiLock<T>`s
     /// originated from the same call to `BiLock::new`.
@@ -156,7 +161,7 @@ impl<T> BiLock<T> {
     where
         T: Unpin,
     {
-        if Arc::ptr_eq(&self.arc, &other.arc) {
+        if self.is_pair_of(&other) {
             drop(other);
             let inner = Arc::try_unwrap(self.arc)
                 .ok()
--- a/futures-util/src/stream/stream/split.rs
+++ b/futures-util/src/stream/stream/split.rs
@@ -15,6 +15,13 @@ pub struct SplitStream<S>(BiLock<S>);
 
 impl<S> Unpin for SplitStream<S> {}
 
+impl<S> SplitStream<S> {
+    /// Returns `true` if the `SplitStream<S>` and `SplitSink<S>` originate from the same call to `StreamExt::split`.
+    pub fn is_pair_of<Item>(&self, other: &SplitSink<S, Item>) -> bool {
+        other.is_pair_of(&self)
+    }
+}
+
 impl<S: Unpin> SplitStream<S> {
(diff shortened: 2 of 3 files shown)


futures-util/src/lock/bilock.rs
```diff
@@ -69,6 +69,11 @@ impl<T> BiLock<T> {
         (Self { arc: arc.clone() }, Self { arc })
     }
 
+    /// Returns `true` only if the two `BiLock<T>`s originated from the same call to `BiLock::new`.
+    pub fn are_paired(first: &Self, second: &Self) -> bool {
```
taiki-e, line 73: Why is this different than others? (are_paired(_, _) vs is_pair(&self, _)).

taiki-e (commented):
Thanks, this seems reasonable. Channels also have [similar APIs](https://github.com/rust-lang/futures-rs/blob/6e1857a69b5e2ce195f8d3d8c382cec0236a91a9/futures-channel/src/mpsc/mod.rs#L812-L826).

futures-util/src/lock/bilock.rs
tones111: There's no technical reason.  I went with the non-method implementation for BiLock because I felt like the instances are more independent than the sink/stream pair, but the primary purpose of the abstraction is to only be coupled to one other instance so the method syntax is still reasonable.  The other benefit of using the method signature is that it's more consistent with the sink/stream api, which would be more consistent with ```reunite```.

I've made that change and rebased to hopefully get a clean CI run.  Thanks for taking a look.

futures-util/src/stream/stream/split.rs
```diff
@@ -142,3 +156,70 @@ impl<T, Item> fmt::Display for ReuniteError<T, Item> {
...
+        fn poll_close(
+            self: Pin<&mut Self>,
+            _cx: &mut Context<'_>,
+        ) -> Poll<Result<(), Self::Error>> {
+            todo!()
+        }
+    }
+
+    #[test]
+    #[cfg_attr(docsrs, doc(cfg(feature = "sink")))]
```
taiki-e, line 208:
```suggestion
```

Test code will not be displayed in docs.

taiki-e (commented):
Thanks. 

I noticed a similar function in tokio called [is_pair_of](https://docs.rs/tokio/latest/tokio/io/struct.ReadHalf.html#method.is_pair_of). It would make sense to give it the same name to make it easier for users to find.

Also, we have [`split` in `io` module](https://github.com/rust-lang/futures-rs/blob/6e1857a69b5e2ce195f8d3d8c382cec0236a91a9/futures-util/src/io/split.rs) too. Could you add the same methods to it?

futures-util/src/stream/stream/split.rs
```diff
@@ -142,3 +156,69 @@ impl<T, Item> fmt::Display for ReuniteError<T, Item> {
...
 #[cfg(feature = "std")]
 impl<T: core::any::Any, Item> std::error::Error for ReuniteError<T, Item> {}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::{sink::Sink, stream::StreamExt};
+    use core::marker::PhantomData;
+
+    struct NopStream<Item> {
```
tones111, line 166: This struct is a hack because I couldn't find a type to use without introducing a dependency.  I'm open to suggestions if there's something available that I'm missing.

taiki-e (approved):

taiki-e:
Published in 0.3.30.

PR #2799, opened 2023-11-15
Fix missing wake() in PendingOnce

nickd-airo:
PendingOnce should wake() before returning Pending, otherwise executors may not poll() again.

Without this change, the following code blocks forever:
```rust
#[test]
fn test_block_on() {
    async fn test_fut() {
        futures::pending!();
    }
    futures::executor::block_on(test_fut())
}
```

### DIFF
--- a/futures-util/src/async_await/pending.rs
+++ b/futures-util/src/async_await/pending.rs
@@ -32,11 +32,12 @@ pub struct PendingOnce {
 
 impl Future for PendingOnce {
     type Output = ();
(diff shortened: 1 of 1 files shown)


taiki-e:
The current one is the correct behavior. Your change makes the executor immediately re-poll the future, which is not the intended behavior of this macro. See also docs.

> Similarly, when using this macro, it must be ensured that [`wake`](std::task::Waker::wake) is called somewhere *when further progress can be made*.

So I'm going to close this, but thanks anyway for the PR.

nickd-airo:
Sorry about that, I didn't read the documentation properly. Thanks for clarifying!

PR #2803, opened 2023-11-21
Add `BlockAwait` trait

hwittenborn:
This change adds in a trait called `BlockAwait`, which functions like `futures::executor::block_on`, but allows for a chaining style similar to `.await` syntax.

I wasn't positive if this would be wanted, but considering how simple it was I decided I'd just go ahead and make it. 

I also wasn't completely sure on how I should limit the trait to only be implemented on `Future` types (and possibly how to prevent it's implementation from being overridden), but I'm open to feedback in the area if my implementation isn't the best.

### DIFF
--- a/futures-executor/src/lib.rs
+++ b/futures-executor/src/lib.rs
@@ -56,7 +56,9 @@
 #[cfg(feature = "std")]
 mod local_pool;
 #[cfg(feature = "std")]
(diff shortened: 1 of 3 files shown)


taiki-e:
Users are free to define such helpers themselves, but I would prefer that this not be provided as part of the futures crate.
See https://github.com/smol-rs/async-task/issues/16#issuecomment-1066674511 and subsequent comments for more.

Thanks anyway for the PR!

PR #2804, opened 2023-11-25
Fix incorrect feature flag documented on join! macro

RAnders00:
The join! and try_join! macros are behind the async-await-macro feature flag, not async-await.

### DIFF
--- a/futures-util/src/async_await/join_mod.rs
+++ b/futures-util/src/async_await/join_mod.rs
@@ -9,7 +9,7 @@ macro_rules! document_join_macro {
         /// `join!` polls both futures concurrently and therefore is more efficient.
         ///
         /// This macro is only usable inside of async functions, closures, and blocks.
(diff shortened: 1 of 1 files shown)


taiki-e:
Hmm. The current one is correct for the `futures` documentation, but indeed incorrect for the `futures-util` documentation.

https://github.com/rust-lang/futures-rs/blob/ae3297f25d4bc26e000990cc1dfd16505ae9998a/futures-util/Cargo.toml#L17-L18
https://github.com/rust-lang/futures-rs/blob/ae3297f25d4bc26e000990cc1dfd16505ae9998a/futures/Cargo.toml#L39

taiki-e:
Closing -- as said above, 0.3 requires a more complex fix and 0.4 does not need to do anything. In any case, this PR will not work as a corrective to any of these as it is.

Thanks anyway for the PR!

PR #2805, opened 2023-12-01
Add esp32s2 to no CAS list

MabezDev:
Without this entry, esp32s2 users with `futures-*` dependencies in their tree, indirectly or not must somehow enable the portable atomic feature. This change makes it seamless like it is on RISCV etc.

### DIFF
--- a/ci/no_atomic_cas.sh
+++ b/ci/no_atomic_cas.sh
@@ -26,6 +26,8 @@ EOF
 for target in "${no_atomic_cas[@]}"; do
     echo "    \"${target}\"," >>"${file}"
 done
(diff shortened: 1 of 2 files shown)


taiki-e:
Thanks for the PR. I would prefer not to accept a way to manually manage such a list of platforms. I've said this many times in other repos like https://github.com/crossbeam-rs/crossbeam/pull/751#pullrequestreview-796739843, https://github.com/crossbeam-rs/crossbeam/pull/937#issuecomment-1338885510.

Instead, I would accept a patch to use `target_has_atomic = "ptr"`, like done in crossbeam https://github.com/crossbeam-rs/crossbeam/issues/1033.

PR #2808, opened 2023-12-06
Implement Deref{Mut} for AllowStdIo.

Frostie314159:
This PR adds `Deref{Mut}` which internally just call `get_{ref|mut}`, which will stay, as to not break compatibility.


Frostie314159:
The clippy lint as well as the rustdoc crash appear to be unrelated to this PR, does anyone know why they are happening. I tested it against the latest nightly version of clippy and rustdoc and both succeed with out warnings.

Frostie314159:
The bug causing the CI failure was patched in the latest release of rustdoc.

taiki-e:
I think this could cause breakage if a different method with the same name as a method in std::io traits is added to futures::io traits.

taiki-e:
Closing this due to the above concern.

Thanks anyway for the PR!

PR #2811, opened 2023-12-07
Use cfg(target_has_atomic) on no-std targets

taiki-e:
Use `cfg(target_has_atomic)` on `cfg(target_os = "none")` targets.

`cfg(target_has_atomic)` is only actually needed on `cfg(target_os = "none")` targets since CAS is one of the requirements of the std, and no-std targets other than `cfg(target_os = "none")` targets provide CAS.

This increases the MSRV of the `cfg(target_os = "none")` targets to Rust 1.60, but since they are all tier2/tier3 targets in rustc and those use cases usually require at least 1.59 for inline-asm, it would be acceptable for them to have a slightly higher MSRV than the other targets.

See also https://github.com/rust-lang/futures-rs/pull/2805#issuecomment-1836422561 and linked issues.


taiki-e:
FYI @MabezDev

taiki-e:
Published in 0.3.30.

PR #2823, opened 2024-01-18
Ignore newly added clippy lint

taiki-e:


taiki-e:
Hmm... it seems that someone has taken the liberty of changing the configuration of this repository.

gh

It is problematic to set up such branch protection in a project where there is effectively only one maintainer but they do not have the permissions to bypass the branch protection.

taiki-e:
Ok, I believe that is due to https://github.com/rust-lang/team/pull/1187.

cc @Kobzol

I apologize for not being able to comment on that issue sooner, but I can't find an explanation about the possibility of such problems.
Could you tell me what needs to be done to fix this? Is it just a removal of the field regarding branch protection? Or can it be adjusted by adding a specific field?

taiki-e:
> Could you tell me what needs to be done to fix this? Is it just a removal of the [table regarding branch protection](https://github.com/rust-lang/team/pull/1187/files#diff-be9764df9ef7713bfccc89720914fa42e9160b542b7cb23012bdadbd5ac5e707R9-R10)? Or can it be adjusted by adding a specific field?

It would be ideal if it would give me the (admin?) permission to bypass the protections if necessary...

Kobzol:
Sorry for that, I missed that approvals weren't required before! https://github.com/rust-lang/team/pull/1225 should fix it.

Kobzol:
I'll investigate why it didn't work.

taiki-e:
The branch protection issue seems to be fixed now (by https://github.com/rust-lang/sync-team/pull/62?). Thanks for fixing it @Kobzol.

Kobzol:
Sorry, I forgot to post here :facepalm:

PR #2825, opened 2024-01-30
Add a helper for always ready futures

GnomedDev:
Currently, there is no easy way to wrap a given value in a zst future, so this PR adds one. The `always_ready` function has to take a closure to allow it to yield non-zsts without having to carry around it's size.

This is particularly useful in avoiding a heap allocation when a library needs `Pin<Box<dyn Future<Output<T>>>`, as a simple empty async block still has to allocate a boolean in this example for "has been polled".

### DIFF
--- a/futures-util/src/future/always_ready.rs
+++ b/futures-util/src/future/always_ready.rs
@@ -1,0 +1,50 @@
+use super::assert_future;
+use core::pin::Pin;
+use futures_core::future::{FusedFuture, Future};
+use futures_core::task::{Context, Poll};
+
+/// Future for the [`always_ready`](always_ready()) function.
+#[derive(Debug, Clone, Copy)]
+#[must_use = "futures do nothing unless you `.await` or poll them"]
+pub struct AlwaysReady<T, F: Fn() -> T>(F);
+
+impl<T, F: Fn() -> T> Unpin for AlwaysReady<T, F> {}
+
+impl<T, F: Fn() -> T> FusedFuture for AlwaysReady<T, F> {
+    fn is_terminated(&self) -> bool {
+        false
+    }
+}
+
+impl<T, F: Fn() -> T> Future for AlwaysReady<T, F> {
+    type Output = T;
+
+    #[inline]
+    fn poll(self: Pin<&mut Self>, _cx: &mut Context<'_>) -> Poll<T> {
+        Poll::Ready(self.0())
+    }
+}
+
+/// Creates a future that is always immediately ready with a value.
+///
+/// # Examples
+///
+/// ```
+/// # futures::executor::block_on(async {
+/// use std::mem::size_of_val;
+///
+/// use futures::future;
+///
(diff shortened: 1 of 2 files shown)


futures-util/src/future/always_ready.rs
```diff
@@ -0,0 +1,50 @@
...
+/// let a = future::always_ready(|| 1);
+/// assert_eq!(size_of_val(&a), 0);
+/// assert_eq!(a.await, 1);
+/// assert_eq!(a.await, 1);
+/// # });
+/// ```
+pub fn always_ready<'a, T, F>(prod: F) -> AlwaysReady<T, F>
+where
+    T: Copy + 'a,
+    F: Fn() -> T + 'a,
```
taiki-e, line 47: Why are lifetime and Copy bounds needed here?

GnomedDev: They aren't, I just removed them, guess they were a leftover from previous designs.

GnomedDev:
Hey, @taiki-e, just letting you know I responded to your review comment.

futures-util/src/future/always_ready.rs
```diff
@@ -0,0 +1,46 @@
+use super::assert_future;
+use core::pin::Pin;
+use futures_core::future::{FusedFuture, Future};
+use futures_core::task::{Context, Poll};
+
+/// Future for the [`always_ready`](always_ready()) function.
+#[derive(Debug, Clone, Copy)]
```
taiki-e, line 7: derive creates trait impl that have trait bounds for all generics. In this case, they are `T: Copy, F: Copy` for `impl Copy for AlwaysReady`, `T: Debug, F: Debug` for `impl Debug for AlwaysReady`, but `T: Copy` bound and all `Debug` bounds are useless.

Could you replace these derives with manual impls?

taiki-e:
> This is particularly useful in avoiding a heap allocation when a library needs `Pin<Box<dyn Future<Output<T>>>`

I think adding this statement to the document would be helpful.

futures-util/src/future/always_ready.rs
GnomedDev: Done, I required `Debug` for `F`, but if you give the desired debug output I can relax that bound aswell.

taiki-e: > I required `Debug` for `F`, but if you give the desired debug output I can relax that bound aswell.

Other our futures/streams don't print closure (https://github.com/rust-lang/futures-rs/pull/1677), so I would like to adopt the same behavior (like [PollFn](https://github.com/rust-lang/futures-rs/pull/1677/files#diff-72ac786dc163d40e11613729e088985b6a3def2c2182270c11303b77273be0b1R45) does) here.

GnomedDev: Okay, done

taiki-e (approved):
Thanks!

PR #2826, opened 2024-02-10
Add dependabot

oriontvv:
This PR adds dependabot config

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,15 @@
+version: 2
+updates:
+  - package-ecosystem: "cargo"
+    directory: "/"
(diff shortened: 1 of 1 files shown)


taiki-e:
Thanks for the PR. It seems that what you are trying to add is a template that does not care about this project's actual status. For example we do not use docker.

Instead, I would like to use a copy of [my project's dependabot config](https://github.com/taiki-e/pin-project/blob/d42e9cc402793697cfb7928dbe9721b0606de66b/.github/dependabot.yml) (but for this repo, I would like to change the update interval to weekly).

oriontvv:
Yep, done

PR #2829, opened 2024-02-25
Fix CI failure

taiki-e:


taiki-e:
Hmm. https://github.com/rust-lang/rust/pull/119863 seems to have caused regression.

```
---- ref_wake_same stdout ----
thread 'ref_wake_same' panicked at futures/tests/task_arc_wake.rs:56:5:
assertion failed: w2.will_wake(&w3)
stack backtrace:
   0: rust_begin_unwind
   1: core::panicking::panic_fmt
   2: core::panicking::panic
   3: core::ops::function::FnOnce::call_once
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.


failures:
    ref_wake_same
```

https://github.com/rust-lang/futures-rs/blob/0fa6a1af9089e587a8fde1882284d08436b0c05c/futures/tests/task_arc_wake.rs#L47-L57

cc @tmiasko: Is this the regression you intended?

PR #2832, opened 2024-02-28
Parse rhs of `select!` arms using match-arm rules

### LINKED ISSUE #2824: parse error in `select!`/`select_biased!` macro
I noticed that in the `select!`/`select_biased!` macros, the unit type (`()`) only works for the pattern-side of a branch in the first position.

Consider the following minimal example:
```rust
use futures::{future::pending, select, FutureExt};

async fn foo() {
    select! {
        () = pending::<()>().fuse() => {}
        () = pending::<()>().fuse() => {}
    }
}
```

It produces the following error (note: line 6 is the _second_ branch):
```
error: expected `,`
 --> src/lib.rs:6:37
  |
6 |         () = ready::<()>(()).fuse() => {}
  |                                     ^
```

This version compiles without issues:
```rust
use futures::{future::ready, select, FutureExt};

async fn foo() {
    select! {
        () = ready::<()>(()).fuse() => {}
        _ = ready::<()>(()).fuse() => {}
    }
}
```

dtolnay:
Fixes #2824.

This is a breaking change because, for example, the following program would break.

```rust
use futures::{future, select};

fn main() {
    let _ = async {
        select! {
            () = future::ready(()) => {}()
        }
    };
}
```

### DIFF
--- a/futures-macro/Cargo.toml
+++ b/futures-macro/Cargo.toml
@@ -18,7 +18,7 @@ proc-macro = true
 [dependencies]
 proc-macro2 = "1.0.60"
 quote = "1.0"
-syn = { version = "2.0.8", features = ["full"] }
+syn = { version = "2.0.52", features = ["full"] }
 
 [lints]
 workspace = true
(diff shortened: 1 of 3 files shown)


taiki-e (approved):
Thanks!

PR #2842, opened 2024-03-16
Added future::on_poll, future::on_poll_pending and future::on_poll_ready

kulst:
This pull request adds the possibility to create wrapped futures that additionally invoke a function when they get polled. The function has mutable access to the Context and the returning Poll of the wrapped future. 

I can imagine several possible use cases for that:
- Call the waker of a future that would not be woken otherwise
- Let the application know each time a future's poll returns (e.g. for logging)
- Abort a future after a specific number of pending polls

The futures are created by three free functions. Examples are given in the code comments.

Comments would be really appreciated as 
- this is my first pull request and I am not entirely sure if a feature like this is in scope of the project
- the futures could also be created by extending the FutureExt trait with additional methods
- the OnPoll, OnPollPending and OnPollReady structs could also be realized as generic structs

### DIFF
--- a/futures-util/src/future/mod.rs
+++ b/futures-util/src/future/mod.rs
@@ -71,6 +71,11 @@ pub use self::poll_fn::{poll_fn, PollFn};
 mod poll_immediate;
 pub use self::poll_immediate::{poll_immediate, PollImmediate};
 
+mod on_poll;
+pub use self::on_poll::{on_poll, OnPoll};
+pub use self::on_poll::{on_poll_pending, OnPollPending};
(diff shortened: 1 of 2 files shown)


taiki-e:
Thanks for the PR. We recently received another PR similar to this one (#2840 by @coolreader18). (Although this PR is 3 futures with a post-called function in different conditions, and #2840 is a future with a pre-polled future, so it's not like either one covers the other.)

To be honest, I don't want to accept all of these kinds of similar and slightly different APIs, none of them seem that common, and they seem easy to implement locally.

If any particular one is widely used, I would prefer to accept it, but if not, I would tend to close both for now.

PR #2848, opened 2024-03-31
Make access inner of futures::io::{BufReader,BufWriter} not require inner trait bound

ethe:
Currently, accessing the inner of `futures::io::{BufReader, BufWriter}` requires that the inner type implements `AsyncRead` or `AsyncWrite`. This constraint is much stricter than the semantics of inner access require. Removing the trait bound would be helpful in cases where pass-through functions simply obtain a reference to the inner I/O object for further use, without having to specify the inner trait bound explicitly.

### DIFF
--- a/futures-util/src/io/buf_reader.rs
+++ b/futures-util/src/io/buf_reader.rs
@@ -52,7 +52,9 @@ impl<R: AsyncRead> BufReader<R> {
         let buffer = vec![0; capacity];
         Self { inner, buffer: buffer.into_boxed_slice(), pos: 0, cap: 0 }
     }
+}
 
+impl<R> BufReader<R> {
     delegate_access_inner!(inner, R, ());
 
     /// Returns a reference to the internally buffered data.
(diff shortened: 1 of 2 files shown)


taiki-e:
Thanks for the PR. I'm in favor of accepting this as std::io::BufReader does the same thing.
https://doc.rust-lang.org/nightly/std/io/struct.BufReader.html

That said, I think it would be good to remove the bound from not only these accessors, but also from non constructor methods such as `buffer`, as std::io::BufReader does.

ethe:
I totally agree, I done the same thing on other non custructor methods just now.

futures-util/src/io/buf_reader.rs
```diff
@@ -71,6 +58,21 @@ impl<R: AsyncRead> BufReader<R> {
     }
 }
 
+impl<R: AsyncRead> BufReader<R> {
```
taiki-e, line 61: I think adding the `} impl<R> BufReader<R> {` after the constructor definitions, rather than moving the constructor definitions, would make the diff smaller and make the document more readable. (I think the order constructor->accessor is usually more readable because it is the same as the actual order in the program.)

taiki-e (approved):
Thanks!

PR #2865, opened 2024-06-14
`#[inline(always)]` on `clone_arc_raw`

timotheyca:
Similar patch (by, if I understand correctly, forcing these to be in the same CGUs) to https://github.com/rust-lang/rust/pull/121622


taiki-e (approved):
Thanks!

PR #2868, opened 2024-07-02
Support shared futures on no_std

adavis628:
Currently, `Shared` futures are only available when the `std` feature is enabled because they use `std::sync::Mutex` internally. This PR changes this so that `Shared` futures will fall back to using a spinlock when `std` is not enabled.

Alternatively, `Shared` could be changed to be generic over a mutex trait (e.g. `lock_api::RawMutex`, though it isn't implemented for `std::sync::Mutex`) to allow arbitrary user implementations.

### DIFF
--- a/futures-util/Cargo.toml
+++ b/futures-util/Cargo.toml
@@ -43,6 +43,7 @@ futures_01 = { version = "0.1.25", optional = true, package = "futures" }
 tokio-io = { version = "0.1.9", optional = true }
 pin-utils = "0.1.0"
 pin-project-lite = "0.2.6"
+spin = "0.9.8"
 
 [dev-dependencies]
 futures = { path = "../futures", features = ["async-await", "thread-pool"] }
--- a/futures-util/Cargo.toml
+++ b/futures-util/Cargo.toml
@@ -22,7 +22,6 @@ sink = ["futures-sink"]
 io = ["std", "futures-io", "memchr"]
 channel = ["std", "futures-channel"]
 portable-atomic = ["futures-core/portable-atomic"]
-spin = ["dep:spin"]
 
 # Unstable features
 # These features are outside of the normal semver guarantees and require the
--- a/futures-util/Cargo.toml
+++ b/futures-util/Cargo.toml
@@ -12,8 +12,8 @@ Common utilities and extension traits for the futures-rs library.
 
 [features]
 default = ["std", "async-await", "async-await-macro"]
-std = ["alloc", "futures-core/std", "futures-task/std", "slab"]
-alloc = ["futures-core/alloc", "futures-task/alloc"]
+std = ["alloc", "futures-core/std", "futures-task/std", "slab/std"]
+alloc = ["futures-core/alloc", "futures-task/alloc", "slab"]
 async-await = []
 async-await-macro = ["async-await", "futures-macro"]
 compat = ["std", "futures_01"]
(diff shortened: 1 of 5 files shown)


taiki-e:
Thanks for the PR. I would prefer not to use spinlock by default around operations that would involve allocations even if it is no_std-only.

That said, it might be fine if this is an optional feature and spinlock is used only when std is disabled and the feature is explicitly enabled.

adavis628:
I moved the spinlock behind an optional `spin` feature that will use a spinlock when std is disabled.

futures-util/Cargo.toml
```diff
@@ -22,6 +22,7 @@ sink = ["futures-sink"]
 io = ["std", "futures-io", "memchr"]
 channel = ["std", "futures-channel"]
 portable-atomic = ["futures-core/portable-atomic"]
+spin = ["dep:spin"]
```
taiki-e, line 25:
```suggestion
```

(This is unneeded and causing CI issue.)

futures-util/src/future/future/mod.rs
```diff
@@ -107,9 +107,9 @@ mod remote_handle;
 #[cfg(feature = "std")]
 pub use self::remote_handle::{Remote, RemoteHandle};
 
-#[cfg(feature = "std")]
+#[cfg(any(feature = "std", feature = "spin"))]
```
taiki-e, line 110: This needs to be `#[cfg(any(feature = "std", all(feature = "alloc", feature = "spin")))]`

https://github.com/rust-lang/futures-rs/actions/runs/12735209519/job/35493701034?pr=2868#step:5:199

```
running `cargo check --no-default-features --features unstable,spin` on futures (24/193)
      Checking futures-util v0.4.0-alpha.0 (/home/runner/work/futures-rs/futures-rs/futures-util)
  error[E0433]: failed to resolve: use of undeclared crate or module `alloc`
   --> futures-util/src/future/future/shared.rs:2:5
    |
  2 | use alloc::sync::{Arc, Weak};
    |     ^^^^^ use of undeclared crate or module `alloc`
    |
    = help: add `extern crate alloc` to use the `alloc` crate
  
  error[E0432]: unresolved imports `crate::task::waker_ref`, `crate::task::ArcWake`
    --> futures-util/src/future/future/shared.rs:1:19
     |
  1  | use crate::task::{waker_ref, ArcWake};
     |                   ^^^^^^^^^  ^^^^^^^ no `ArcWake` in `task`
     |                   |
     |                   no `waker_ref` in `task`
     |
  note: found an item that was configured out
    --> futures-util/src/task/mod.rs:31:24
     |
  31 | pub use futures_task::{waker_ref, WakerRef};
     |                        ^^^^^^^^^
  note: the item is gated behind the `alloc` feature
    --> futures-util/src/task/mod.rs:30:7
     |
  30 | #[cfg(feature = "alloc")]
     |       ^^^^^^^^^^^^^^^^^
  note: found an item that was configured out
    --> futures-util/src/task/mod.rs:23:23
     |
  23 | pub use futures_task::ArcWake;
     |                       ^^^^^^^
  note: the item is gated behind the `alloc` feature
    --> futures-util/src/task/mod.rs:22:7
     |
  22 | #[cfg(feature = "alloc")]
     |       ^^^^^^^^^^^^^^^^^
  
  error[E0432]: unresolved import `slab`
    --> futures-util/src/future/future/shared.rs:12:5
     |
  12 | use slab::Slab;
     |     ^^^^ use of undeclared crate or module `slab`
 
```

adavis628: Fixed!

taiki-e (approved):
Thanks!

PR #2875, opened 2024-08-02
Add accessors for the inner of stream::Iter

wfraser:
This PR adds analogues to the methods of all the stream/sink combinators which allow accessing the thing being wrapped. Here it lets you access the iterator, which is currently not possible.

I'm intentionally not using the `delegate_access_inner` macro, because the doc comments describe the inner member as being a stream or sink, which it is not for `Iter` (it also requires using `pin_project`, which `Iter` does not do or need).

### DIFF
--- a/futures-util/src/stream/iter.rs
+++ b/futures-util/src/stream/iter.rs
@@ -10,6 +10,23 @@ pub struct Iter<I> {
     iter: I,
 }
 
+impl<I> Iter<I> {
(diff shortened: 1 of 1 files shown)


taiki-e (approved):
Thanks!

> the doc comments describe the inner member as being a stream or sink

This is a mistake in the existing code and needs to be corrected to something like " the underlying type".
https://github.com/rust-lang/futures-rs/issues/2681

PR #2878, opened 2024-08-19
refactor: `TryFuture`s are `Future`s now, `into_future` is no longer need

yhx-12243:
Reduce a wrapper for `IntoFuture` struct, reduce code complexity.


yhx-12243:
AFAIK there are almost no crates use these `IntoFuture`s and `IntoStream`s. But if one want the backward compatibility, one can define `type IntoFuture<T> = T` and `type IntoStream<T> = T` or just bump to v0.4.0, both is OK.

taiki-e:
This appears to be a duplicate of https://github.com/rust-lang/futures-rs/pull/2763?

yhx-12243:
> This appears to be a duplicate of #2763?

Maybe to say it is an **extension** better.

PR #2885, opened 2024-09-14
Various fixes too make the CI green

hkratz:
Split into semantically separate commits for easier reviewing.

compiler warnings:
* Fix unexpected `cfg` condition name: ... warnings introduced in Rust 1.80 (see https://blog.rust-lang.org/2024/05/06/check-cfg.html)
* io_slice_advance feature is now stable

clippy lints:
* `missing_const_for_thread_local` lint was renamed to `missing_const_for_thread_local`, but disabling it is no longer necessary since https://github.com/rust-lang/rust-clippy/issues/12404 has been fixed

clippy doc lints:
* fixes for "doc list item without indentation" lint
* ignore incorrect "first doc comment paragraph is too long" warning (see https://github.com/rust-lang/rust-clippy/issues/13315)
* allow long first paragraphs in `select`... fn doc comments

### DIFF
--- a/futures/build.rs
+++ b/futures/build.rs
@@ -1,3 +1,0 @@
-fn main() {
-    println!("cargo:rustc-check-cfg=cfg(futures_sanitizer)");
-}
--- a/futures/build.rs
+++ b/futures/build.rs
@@ -1,0 +1,3 @@
+fn main() {
+    println!("cargo:rustc-check-cfg=cfg(futures_sanitizer)");
+}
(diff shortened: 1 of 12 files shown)


futures/build.rs
```diff
@@ -0,0 +1,3 @@
+fn main() {
+    println!("cargo:rustc-check-cfg=cfg(futures_sanitizer)");
```
taiki-e, line 2: This is a cfg that is set globally by CI, so I would like to allow it across the workspace using the `[workspace.lints]` table (like [this](https://github.com/crossbeam-rs/crossbeam/blob/abf24e1f31a76ef2590688d0c2bb55f82c7410a9/Cargo.toml#L66-L69)) instead of only allowing it in a specific crate using a build script.

taiki-e (commented):
Thanks! LGTM aside from a nit.

taiki-e (approved):

PR #2886, opened 2024-10-01
Fix use after free of task in FuturesUnordered when dropped future panics

Imberflur:
Fixes https://github.com/rust-lang/futures-rs/issues/2863

### DIFF
--- a/futures-util/src/stream/futures_unordered/mod.rs
+++ b/futures-util/src/stream/futures_unordered/mod.rs
@@ -255,16 +255,6 @@ impl<Fut> FuturesUnordered<Fut> {
         // `wake` from doing any work in the future
         let prev = task.queued.swap(true, SeqCst);
 
-        // Drop the future, even if it hasn't finished yet. This is safe
-        // because we're dropping the future on the thread that owns
-        // `FuturesUnordered`, which correctly tracks `Fut`'s lifetimes and
-        // such.
-        unsafe {
-            // Set to `None` rather than `take()`ing to prevent moving the
-            // future.
-            *task.future.get() = None;
-        }
-
         // If the queued flag was previously set, then it means that this task
         // is still in our internal ready to run queue. We then transfer
         // ownership of our reference count to the ready to run queue, and it'll
@@ -276,8 +266,25 @@ impl<Fut> FuturesUnordered<Fut> {
         // enqueue the task, so our task will never see the ready to run queue
         // again. The task itself will be deallocated once all reference counts
         // have been dropped elsewhere by the various wakers that contain it.
-        if prev {
-            mem::forget(task);
+        //
+        // Use ManuallyDrop to transfer the reference count ownership before
+        // dropping the future so unwinding won't release the reference count.
+        let md_slot;
+        let task = if prev {
+            md_slot = mem::ManuallyDrop::new(task);
+            &*md_slot
+        } else {
+            &task
+        };
+
+        // Drop the future, even if it hasn't finished yet. This is safe
+        // because we're dropping the future on the thread that owns
+        // `FuturesUnordered`, which correctly tracks `Fut`'s lifetimes and
+        // such.
+        unsafe {
+            // Set to `None` rather than `take()`ing to prevent moving the
+            // future.
+            *task.future.get() = None;
         }
     }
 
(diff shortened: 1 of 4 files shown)


futures-util/src/stream/futures_unordered/mod.rs
```diff
@@ -276,8 +266,25 @@ impl<Fut> FuturesUnordered<Fut> {
...
+        //
+        // Use ManuallyDrop to transfer the reference count ownership before
+        // dropping the future so unwinding won't release the reference count.
+        let md_slot;
+        let task = if prev {
+            md_slot = mem::ManuallyDrop::new(task);
+            &*md_slot
+        } else {
+            &task
+        };
```
Imberflur, line 278: This is the main fix

futures-util/src/stream/futures_unordered/mod.rs
```diff
@@ -566,15 +573,36 @@ impl<Fut> FuturesUnordered<Fut> {
...
+            }
+        }
+        impl<Fut> Drop for DropGuard<'_, Fut> {
+            fn drop(&mut self) {
+                self.drop_futures();
+            }
         }
+        let mut guard = DropGuard(self);
+        guard.drop_futures();
+        mem::forget(guard); // no need to check head_all again
```
Imberflur, line 605: This was added to make sure all futures are dropped if one panics. Although it looks like tasks have a check for this in their drop that aborts. At least with this, leaking and aborts may be avoided if only one future panics.

futures/tests/stream_futures_unordered.rs
```diff
@@ -406,3 +406,26 @@ fn clear_in_loop() {
         }
     });
 }
+
+// https://github.com/rust-lang/futures-rs/issues/2863#issuecomment-2219441515
+#[test]
+#[should_panic]
+fn panic_on_drop_fut() {
```
Imberflur, line 413: Minimal example taken from issue. This failed under miri before the changes to fix it.

Imberflur (commented):
Notes for reviewers

futures-util/src/stream/futures_unordered/mod.rs
```diff
@@ -566,15 +573,36 @@ impl<Fut> FuturesUnordered<Fut> {
...
-            let task = unsafe { self.unlink(head) };
-            self.release_task(task);
+        // Before the strong reference to the queue is dropped we need all
+        // futures to be dropped. See note at the bottom of this method.
+        //
+        // This ensures we drop all futures even in the case of one drop
+        // panicking and unwinding.
+        //
+        // If a second future panics, that will trigger an abort from panicking
+        // while panicking.
```
taiki-e, line 583: If possible, I would prefer to leak the remaining elements after the first panic has occurred. That is the same behavior as in the standard library and crossbeam's queues/channels (https://github.com/rust-lang/rust/pull/108164#discussion_r1134233602).

taiki-e (changes requested):
Thanks! The fix looks good to me.

futures-util/src/stream/futures_unordered/mod.rs
Imberflur: Hmm miri is complaining about the leak.

With
```sh
MIRIFLAGS="-Zmiri-strict-provenance -Zmiri-symbolic-alignment-check -Zmiri-disable-isolation" cargo +nightly miri nextest run --workspace --all-features -- panic_on_drop_fut
```

Imberflur: I added a separate miri invocation for the test that ignores leaks

Imberflur: Btw there might be some potential to make the miri CI faster using nextest

taiki-e (approved):
Thanks!

taiki-e:
Hmm. I see AddressSanitizer also complains about the same leak.

Given that it would complicate CI, and how Vec and VecDeque behave, trying to avoid the risk of potential abort for bad future here may not actually be a worthwhile goal...

Imberflur:
I could just ignore this test there? The main thing it tests for is caught by miri.

It feels kind of bad to change this due to testing issues....

TBH if there was a test for hitting the abort case that would probably have issues too since iirc there isn't a good way to test `should_abort` :(

taiki-e (approved):

PR #2887, opened 2024-10-03
Document how `BoxFuture`s / `BoxStream`s are often made

alamb:
# Rationale

Unless you are already familiar with the Rust futures ecosystem and `FuturesExt::boxed` it is not always obvious how to create a `BoxFuture`

Specifically, we have [APIs for reading parquet files](https://docs.rs/parquet/latest/parquet/arrow/async_reader/trait.MetadataFetch.html), which involve `BoxFuture`s that when a user clicks on the return type, 

Screenshot 2024-10-03 at 9 22 37 AM

The documentation doesn't lead them to `FutureExt::boxed`:

Screenshot 2024-10-03 at 9 21 43 AM

# Proposal

While a better improvement for our users is a full featured example (added in https://github.com/apache/arrow-rs/pull/6505), I think this to propose improvements in this crate too to help the broader community

If the maintainers like this pattern, I am happy to add it for other similar types in Futures and Streams but I won't bother if you are not interested in this type of change

### DIFF
--- a/futures-core/src/future.rs
+++ b/futures-core/src/future.rs
@@ -9,6 +9,11 @@ pub use core::future::Future;
 
 /// An owned dynamically typed [`Future`] for use in cases where you can't
 /// statically type your result or need to add some indirection.
+///
+/// This type is often created by the [`boxed`] method on [`FutureExt`]. See its documentation for more.
+///
+/// [`boxed`]: https://docs.rs/futures/latest/futures/future/trait.FutureExt.html#method.boxed
+/// [`FutureExt`]: https://docs.rs/futures/latest/futures/future/trait.FutureExt.html
 #[cfg(feature = "alloc")]
 pub type BoxFuture<'a, T> = Pin<alloc::boxed::Box<dyn Future<Output = T> + Send + 'a>>;
 
--- a/futures-core/src/future.rs
+++ b/futures-core/src/future.rs
@@ -18,6 +18,11 @@ pub use core::future::Future;
 pub type BoxFuture<'a, T> = Pin<alloc::boxed::Box<dyn Future<Output = T> + Send + 'a>>;
 
 /// `BoxFuture`, but without the `Send` requirement.
+///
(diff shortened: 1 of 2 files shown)


futures-core/src/future.rs
```diff
@@ -9,6 +9,11 @@ pub use core::future::Future;
 
 /// An owned dynamically typed [`Future`] for use in cases where you can't
 /// statically type your result or need to add some indirection.
+///
```
alamb, line 12: This follows the wording and pattern in std, for example https://doc.rust-lang.org/std/iter/struct.Cloned.html

> This struct is created by the [cloned](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.cloned) method on [Iterator](https://doc.rust-lang.org/std/iter/trait.Iterator.html). See its documentation for more.

Here is how it looks rendered: 
Screenshot 2024-10-03 at 9 34 52 AM

futures-core/src/future.rs
```diff
@@ -9,6 +9,11 @@ pub use core::future::Future;
...
+/// This type is often created by the [`boxed`] method on [`FutureExt`]. See its documentation for more.
+///
+/// [`boxed`]: https://docs.rs/futures/latest/futures/future/trait.FutureExt.html#method.boxed
+/// [`FutureExt`]: https://docs.rs/futures/latest/futures/future/trait.FutureExt.html
```
alamb, line 16: this links back to docs.rs because futures-core does not depened on futures-util, where `FutureExt` is defined

taiki-e:

taiki-e:
Could you add similar docs to other boxed types?

alamb:
> Could you add similar docs to other boxed types?

Yes, I will do so

alamb:
> > Could you add similar docs to other boxed types?
> 
> Yes, I will do so

Done

taiki-e (approved):

alamb:
Thank you again @taiki-e

PR #2891, opened 2024-10-07
Update changelog and clarify about yanking

taiki-e:
This adds the following note to old 0.3 releases:

> **Note:** This release of futures-util has been yanked due to a bug fixed in 0.3.31 and a compile failure with futures-task 0.3.31 and later.

And, adds the following to header to clarify when the yank will be done:

> Releases may yanked if there is a security bug, a soundness bug, or a regression.

cc @shepmaster


shepmaster:
Much appreciated, thank you.

PR #2896, opened 2024-11-13
Document cancel-safety of StreamExt::next

### LINKED ISSUE #2701: Document cancel safety of `StreamExt::next`
Here: https://docs.rs/futures/latest/futures/stream/trait.StreamExt.html#method.next

And the equivalent tokio docs: https://docs.rs/tokio-stream/latest/tokio_stream/trait.StreamExt.html#method.next

I think it's critical that users know if stream items can be accidentally dropped if the future is cancelled.

akonradi-signal:
Fixes #2701

### DIFF
--- a/futures-util/src/stream/stream/mod.rs
+++ b/futures-util/src/stream/stream/mod.rs
@@ -232,6 +232,12 @@ pub trait StreamExt: Stream {
     /// pinning it to the stack using the `pin_mut!` macro from the `pin_utils`
     /// crate.
     ///
+    /// # Cancel safety
+    ///
+    /// The returned future only holds a reference to the underlying stream, and
+    /// so is cancel safe (assuming the stream is also cancel-safe). If it is
+    /// dropped before it resolves, no item from the stream will be dropped.
+    ///
     /// # Examples
     ///
     /// ```


futures-util/src/stream/stream/mod.rs
```diff
@@ -232,6 +232,12 @@ pub trait StreamExt: Stream {
     /// pinning it to the stack using the `pin_mut!` macro from the `pin_utils`
     /// crate.
     ///
+    /// # Cancel safety
+    ///
+    /// The returned future only holds a reference to the underlying stream, and
+    /// so is cancel safe (assuming the stream is also cancel-safe). If it is
```
taiki-e, line 238: > assuming the stream is also cancel-safe

AFAIK, in the term "cancel-safe" used in tokio, cancel-safe is a property of `Future` and `StreamExt::next` is always cancel-safe. What is the definition of cancel-safe you are referring to?

https://docs.rs/tokio-stream/latest/tokio_stream/trait.StreamExt.html#method.next

taiki-e (changes requested):
Thanks for the PR.

futures-util/src/stream/stream/mod.rs
akonradi-signal: Good question. I'm drawing an analogy between `Future::poll` and `Stream::poll_next`, where calling the latter once but not until it returns `Poll::Ready` could lead to bad behavior.

The example I'm imagining is the result of calling (`StreamExt::chunks_timeout`)[https://docs.rs/tokio-stream/latest/tokio_stream/trait.StreamExt.html#method.chunks_timeout] on a `&mut impl Stream`. Calling `poll_once` on that `ChunksTimeout`, then dropping it, then polling the wrapped stream could result in values from the input stream being dropped without ever being yielded. Given that possibility, I'd label `ChunksTimeout` not "cancel-safe".

PR #2898, opened 2024-11-20
Can we haz const new Mutex?

danii:
All code in `futures_util::lock::Mutex::new` is `const` compatible, and I don't see a reason why it shouldn't be marked as a `const` fn. Can we haz?


danii:
Guess not. My bad.

PR #2899, opened 2024-11-27
Shared: fix shared futures missing wake up

### LINKED ISSUE #2706: Missed wake call using `.shared()`
This is the reproduction:
```rs
use std::{
    io::Write,
    pin::Pin,
    task::{Context, Poll},
};

use futures::prelude::*;

pub async fn yield_now() {
    /// Yield implementation
    struct YieldNow {
        yielded: bool,
    }

    impl Future for YieldNow {
        type Output = ();

        fn poll(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<()> {
            if self.yielded {
                return Poll::Ready(());
            }

            self.yielded = true;
            cx.waker().wake_by_ref();
            Poll::Pending
        }
    }

    YieldNow { yielded: false }.await
}

#[tokio::main]
async fn main() {
    for _ in 0..200 {
        print!(".");
        std::io::stdout().flush().unwrap();
        for _ in 0..10000 {
            test().await;
        }
    }
    println!();
}

async fn test() {
    let f1 = yield_now().shared();
    let f2 = f1.clone();
    let x1 = tokio::spawn(async move {
        f1.now_or_never();
    });
    let x2 = tokio::spawn(async move {
        f2.await;
    });
    x1.await.ok();
    x2.await.ok();
}
```

It manages to hang up eventually on most runs, both in debug and release mode.

The example is crafted to expose a weakness spotted during code review of `.shared()`. The problematic part is here:

https://github.com/rust-lang/futures-rs/blob/aa1f5c74251bb494d1bc7a016381c76f909316bc/futures-util/src/future/future/shared.rs#L288-L293

This code is simply insufficient to ensure the current `context`’s waker is *actually* ever woken up. The current context’s waker was stored into the inner slab previously in this line

https://github.com/rust-lang/futures-rs/blob/aa1f5c74251bb494d1bc7a016381c76f909316bc/futures-util/src/future/future/shared.rs#L277

However, a concurrently running `poll` could have already been within the call to `poll`, i.e. here:

https://github.com/rust-lang/futures-rs/blob/aa1f5c74251bb494d1bc7a016381c76f909316bc/futures-util/src/future/future/shared.rs#L328

at that point. As soon as the call to `poll` on the other future starts, it could already be calling the waker, thus doing the waking ***before*** the new waker that we care about was put into the slab in the first place. This is because it is *okay* for a `.poll` call to do the waking immediately, before it returns, as the `yield_now` function in the example does (copied from the old `yield_now` implementation of `tokio`).

zekun000:
an attempt to fix https://github.com/rust-lang/futures-rs/issues/2706

the waker mutex is used to protect the state transition that involves wakers (add, wake, re-wake), if a waker is called during the polling, it sets the state to `WOKEN_DURING_POLLING` then the polling thread re-checks for potential new wakers and set state back to IDLE under the same mutex to avoid the race condition.

an unit test is adopted from the issue and added, without the fix, it'll be stuck within 5 loops on my laptop locally.

### DIFF
--- a/futures-util/src/future/future/shared.rs
+++ b/futures-util/src/future/future/shared.rs
@@ -9,7 +9,7 @@ use std::pin::Pin;
 use std::ptr;
 use std::sync::atomic::AtomicUsize;
 use std::sync::atomic::Ordering::{Acquire, SeqCst};
-use std::sync::{Arc, Mutex, Weak};
+use std::sync::{Arc, Mutex, MutexGuard, Weak};
 
 /// Future for the [`shared`](super::FutureExt::shared) method.
 #[must_use = "futures do nothing unless you `.await` or poll them"]
@@ -81,6 +81,7 @@ const IDLE: usize = 0;
 const POLLING: usize = 1;
 const COMPLETE: usize = 2;
 const POISONED: usize = 3;
+const AWAKEN_DURING_POLLING: usize = 4;
 
 const NULL_WAKER_KEY: usize = usize::MAX;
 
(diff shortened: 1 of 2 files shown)


zekun000:
doesn't look like I can add reviewers or labels, manually tag here @taiki-e , would appreciate feedbacks or loop in proper reviewers, thanks

steffahn:
Thanks for the ping, I think I even saw this PR before and forgot about it again.

I guess I'll have to re-read my own issue first to understand what exactly needed fixing.

PR #2903, opened 2024-12-25
feat: `select!` macro for no-std environments 

ytakano:
Currently, the `select!` macro works only in the std environment.
This PR fixes the implementation so that the `select!` macro also works in no-std environments for 32- and 64-bit CPUs.

The behavior in the std environment remains unchanged.

### DIFF
--- a/futures-util/src/async_await/random.rs
+++ b/futures-util/src/async_await/random.rs
@@ -1,11 +1,3 @@
-use std::{
-    cell::Cell,
-    collections::hash_map::DefaultHasher,
-    hash::Hasher,
-    num::Wrapping,
-    sync::atomic::{AtomicUsize, Ordering},
-};
-
 // Based on [Fisher–Yates shuffle].
 //
 // [Fisher–Yates shuffle]: https://en.wikipedia.org/wiki/Fisher–Yates_shuffle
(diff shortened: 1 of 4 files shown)


futures-util/src/async_await/random.rs
```diff
@@ -52,3 +53,26 @@ fn random() -> u64 {
...
+
+    if core::mem::size_of::<usize>() == 4 {
+        x ^= x << 13;
+        x ^= x >> 17;
+        x ^= x << 5;
+    } else if core::mem::size_of::<usize>() == 8 {
+        x ^= x >> 12;
+        x ^= x << 25;
+        x ^= x >> 27;
+        x = x.wrapping_mul(0x2545_f491_4f6c_dd1d);
```
brodycj, line 73: This looks like repeated code, would be nice if we could refactor this somehow.

ytakano: Thank you for your review.
I prepared `xorshift64star` and `xorshift32` functions.

brodycj: @ytakano thanks, I should have mentioned that I am just an outside contributor, hope we can get a review from @taiki-e

PR #2904, opened 2025-01-02
resolve unnecessary map_or Clippy issue

brodycj:
(as suggested by Clippy nightly)

PR #2905, opened 2025-01-02
quick workaround for CI issue with libc & MSRV (Rust 1.56)

brodycj:
Should resolve a build issue in PR #2904 & other recent PRs

---

NOTE: CI issue with Clippy is resolved in other PR #2904 - hopeful we can get both of these PRs reviewed & merged soon :)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -109,6 +109,9 @@ jobs:
     timeout-minutes: 60
     steps:
       - uses: taiki-e/checkout-action@v1
+      # QUICK WORKAROUND FOR CI ISSUE with libc & MSRV
+      - run: cargo add libc@=0.2.153 --no-default-features -p futures-executor
+      - run: git clean -dfX
       - name: Install Rust
         run: rustup update ${{ matrix.rust }} && rustup default ${{ matrix.rust }}
       - name: Install cargo-hack


.github/workflows/ci.yml
```diff
@@ -109,6 +109,9 @@ jobs:
     timeout-minutes: 60
     steps:
       - uses: taiki-e/checkout-action@v1
+      # QUICK WORKAROUND FOR CI ISSUE with libc & MSRV
+      - run: cargo add libc@=0.2.153 --no-default-features -p futures-executor
+      - run: git clean -dfX
```
brodycj, line 114: Should update something like this:

```suggestion
      - name: Workaround for CI issue in futures-executor between libc & MSRV
        run: |
          cargo add libc@=0.2.153 --no-default-features -p futures-executor
          git clean -dfX
```

similar to this:
- https://github.com/seanmonstar/num_cpus/blob/815551cfc7f9393410de47e4bd1fe77d11f6ace5/.github/workflows/ci.yml#L39-L40

It may be possible for this to work with some newer version of `libc`, may need some time for trial-and-error.

I am actually not 100% happy about this kind of workaround. IMHO it should be possible to simply install & work with MSRV.

taiki-e:
libc is not dev-dep, so we have to bump MSRV of utility crates, like https://github.com/rust-lang/futures-rs/pull/2733 did.

PR #2906, opened 2025-01-07
Bump MSRV of core crates to 1.41.0

brodycj:
This should allow us to use `core::panic::AssertUnwindSafe` rather than depending on `std` for this.


brodycj (commented):
I think we should bump MSRV for all crates to `1.63` to enable Rust edition 2021, as I proposed in PR #2909.

taiki-e:
Closing as https://github.com/rust-lang/futures-rs/pull/2908 and https://github.com/rust-lang/futures-rs/pull/2909 are closed. 

Thanks anyway for the PR.

PR #2907, opened 2025-01-07
Bump general MSRV to 1.63

brodycj:
as discussed in: https://github.com/rust-lang/futures-rs/pull/2905#issuecomment-2568660258

This should avoid MSRV issues that show up in CI testing with transitive `libc` dependency & possibly other dependencies in the future.

---

__CI TESTING TODO(S):__

- expected clippy failure - already resolved by PR [resolve unnecessary map_or Clippy issue #2904](https://github.com/rust-lang/futures-rs/pull/2904)

__OTHER TODO(S):__

- [x] ~~ update packages to use Rust edition 2021 - now proposed in PR #2909~~

PR #2908, opened 2025-01-07
core: no-std panic::AssertUnwindSafe (etc.)

brodycj:
__STATUS:__

I think the following PRs should be merged first to resolve the CI build issues:

- [ ] #2904 - resolve clippy failure
- [ ] #2907 - bump main / general MSRV to resolve MSRV failure
- [ ] #2906 - bump core MSRV as needed for this PR

---

__WHAT:__

- use panic::AssertUnwindSafe from Rust core lib in `futures-core`
- remove `extern crate std` not needed from `futures-core` & `futures-sink`

I would be happy to split these updates into separate PRs, if needed. I figured we may as well do these updates together.

---

__CI testing TODO(s):__

- core MSRV failure (expected) - requires core MSRV bump to 1.41 - already resolved by PR: #2906
- clippy failure - already resolved by PR: #2904
- general msrv failure - as resolved by PR: #2907

---

__other TODO(s):__

- [ ] remove `std` option from `futures-core` & `futures-sink` - this would be another breaking change - I would be happy to do this here or in a separate PR

### DIFF
--- a/futures-core/src/lib.rs
+++ b/futures-core/src/lib.rs
@@ -12,8 +12,6 @@
 
 #[cfg(feature = "alloc")]
 extern crate alloc;
-#[cfg(feature = "std")]
-extern crate std;
 
 pub mod future;
 #[doc(no_inline)]
--- a/futures-core/src/future.rs
+++ b/futures-core/src/future.rs
@@ -104,8 +104,7 @@ mod if_alloc {
         }
     }
 
-    #[cfg(feature = "std")]
-    impl<F: FusedFuture> FusedFuture for std::panic::AssertUnwindSafe<F> {
+    impl<F: FusedFuture> FusedFuture for core::panic::AssertUnwindSafe<F> {
         fn is_terminated(&self) -> bool {
             <F as FusedFuture>::is_terminated(&**self)
         }
--- a/futures-core/src/stream.rs
+++ b/futures-core/src/stream.rs
@@ -224,8 +224,7 @@ mod if_alloc {
         }
     }
 
-    #[cfg(feature = "std")]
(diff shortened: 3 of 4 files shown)


futures-core/src/lib.rs
```diff
@@ -12,8 +12,6 @@
 
 #[cfg(feature = "alloc")]
 extern crate alloc;
-#[cfg(feature = "std")]
-extern crate std;
```
taiki-e, line 16: This is necessary to generate the appropriate error if the user accidentally enables the std feature.

Otherwise, we may run into trouble with users who accidentally rely on the fact that currently we are not using std and get angry that their builds are corrupted during future updates.

taiki-e:
Thanks for the PR. 

Is this PR based on a specific use case? Or is this a “do it because it's possible” style of PR?
catch_unwind is not in core, so I don't think using this in core is that common.

(You can use catch_unwind from third-party no-std crate such as unwinding crate, but it [doesn't need UnwindSafe](https://docs.rs/unwinding/latest/unwinding/panic/fn.catch_unwind.html).)

brodycj:
> Is this PR based on a specific use case? Or is this a “do it because it's possible” style of PR?

I was trying to get as much functionality as possible working with `no-std`.

> catch_unwind is not in core, so I don't think using this in core is that common

Makes sense to me, closing then.

> (You can use catch_unwind from third-party no-std crate such as unwinding crate, but it [doesn't need UnwindSafe](https://docs.rs/unwinding/latest/unwinding/panic/fn.catch_unwind.html).)

https://github.com/nbdd0121/unwinding

Thanks!

PR #2925, opened 2025-02-17
chore: recommend to use core lib's ready macro

tisonkun:
Ref - https://rust-lang.zulipchat.com/#narrow/channel/187312-wg-async/topic/Move.20the.20.60ready!.60.20macro.20to.20libcore.3F/near/497123105

Weak Ref - https://github.com/rust-lang/futures-rs/issues/2924

### DIFF
--- a/futures-core/src/task/poll.rs
+++ b/futures-core/src/task/poll.rs
@@ -4,10 +4,6 @@
 ///
 /// **Note:** Since Rust 1.64, this macro is soft-deprecated in favor of
 /// [`ready!`](core::task::ready) macro in the standard library.
-#[deprecated(
-    since = "0.4.0",
-    note = "use core::task::ready macro in the standard library instead"
-)]
 #[macro_export]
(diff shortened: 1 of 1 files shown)


tisonkun:
Mark with a `deprecated` attribution seems failed all the CI tasks since we forbid using deprecated methods.

We can:

* Delete the `deprecated` attribution, retain only the note comment;
* Replace futures_core::ready with core::task::ready everywhere and bump MSRV to 1.64.

I'd prefer the second option but wait a bit for any reviewer's advice.

taiki-e:
As said in zulip, I would prefer *soft*-deprecate it as well as the pin_utils::pin_mut (https://github.com/rust-lang/pin-utils/pull/39). i.e., add a note to docs, but don't emit deprecated warning.

tisonkun:
Thanks for your information. Updated.

taiki-e (approved):
Thanks.

PR #2926, opened 2025-02-25
Make assert in select_biased documentation clearer

TobiasBengtsson:


TobiasBengtsson:
> Thanks for the PR. This is select__biased_, so `res == 12` is indeed redundant and only `res == 63` is fine. I think using `res == 13` like this PR is not a good idea because it makes it impossible to tell from the example the difference from a normal select.

Good point, I changed it to `assert_eq!(res, 63)` to make it clearer

PR #2927, opened 2025-02-25
futures ordered: add `clear` fn

Slixe:
Add a `clear` function in `FuturesOrdered` struct similar to `FuturesUnordered` to drop both queues (outputs and pending futures).


futures-util/src/stream/futures_ordered.rs
```diff
@@ -169,6 +169,14 @@ impl<Fut: Future> FuturesOrdered<Fut> {
         self.next_outgoing_index -= 1;
         self.in_progress_queue.push(wrapped);
     }
+
+    /// Clear the whole `FuturesOrdered` queue.
+    ///
+    /// This function clears the pending futures and the queued outputs
+    /// to make it fully empty.
+    pub fn clear(&mut self) {
+        *self = Self::new();
```
taiki-e, line 178: Unlike FuturesUnordered, which cannot reuse allocations, BinaryHeap (queued_outputs field) uses Vec internally, so the following would be more efficient since BinaryHeap allocations can be reused.

```rs
self.in_progress_queue.clear();
self.queued_outputs.clear();
self.next_incoming_index = 0;
self.next_outgoing_index = 0;
```

taiki-e (changes requested):
Thanks for the PR.

futures-util/src/stream/futures_ordered.rs
Slixe: Fine! I wasn't sure if you preferred similar to `FuturesUnordered` or clear everything by hand.
Will update.

taiki-e (approved):

Issue #2901, opened 2024-12-15
Use AsyncFn family of traits (async closures)


taiki-e:
async closures have been stabilized in https://github.com/rust-lang/rust/pull/132706 (Rust 1.85).

tisonkun:
This looks interesting and I'd like to investigate it in futrues-rs.

One prerequisite is bumping the MSRV and I wonder what our MSRV strategy is.

That said, I'm trying to replace the dependency to `pin-utils` (only used for `pin_utils::pin_mut`) with `core::pin::pin` but noticed that `futures-utils`' MSRV is 1.63 < 1.68.

This issue requires an MSRV >= 1.85 and I'm unsure when we can bump the MSRV to unblock the actual migration work.

taiki-e:
The next MSRV bump will probably be to 1.85 or 1.86 when Debian 13 is released. It will probably be 4 to 8 months later.

However, you probably don't need to worry about that since the master branch is not scheduled for immediate release.

---

Btw, this is actually already supported at least for some simple cases:

```rs
use futures::stream::{self, StreamExt};

let stream = stream::iter(1..=3);
let stream = stream.then(async move |x| x + 3);

assert_eq!(vec![4, 5, 6], stream.collect::<Vec<_>>().await);
```

[playground](https://play.rust-lang.org/?version=beta&mode=debug&edition=2021&gist=809490dd14066be23516413c640470c1)

However, doesn't seem to work if the closure bound is `FnMut`/`Fn` and argument of the closure contains a reference.

```rs
use futures::stream::{self, StreamExt};

let stream = stream::iter(1..=3);
// ok
let stream = stream.filter(|_x| futures::future::ready(true));
// error
let stream = stream.filter(async |_x| true);
```

```text
error[E0308]: mismatched types
   --> src/main.rs:9:18
    |
9   |     let stream = stream.filter(async |_x| true);
    |                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ one type is more general than the other
    |
    = note: expected `async` closure body `{async closure body@src/main.rs:9:43: 9:47}`
               found `async` closure body `{async closure body@src/main.rs:9:43: 9:47}`
    = note: no two async blocks, even if identical, have the same type
    = help: consider pinning your async block and casting it to a trait object
note: the lifetime requirement is introduced here
[...]

PR #2929, opened 2025-02-26
chore: reduce dependency to pin-utils

tisonkun:
This refers to https://github.com/rust-lang/futures-rs/issues/2924.

cc @taiki-e as for "it is reasonable to remove it in v0.4", IIUC I can open a follow-up PR to remove this macro and update the corresponding related usage, only we don't mark that PR as 0.3-backport?

BTW, if there is anything I can help in backporting, please let me know. In my experience, it means opening a PR against branch `0.3` and do the same change.

### DIFF
--- a/futures-util/src/future/future/mod.rs
+++ b/futures-util/src/future/future/mod.rs
@@ -10,6 +10,7 @@ use core::pin::Pin;
 
 use crate::fns::{inspect_fn, into_fn, ok_fn, InspectFn, IntoFn, OkFn};
 use crate::future::{assert_future, Either};
+use crate::pin_mut;
 use crate::stream::assert_stream;
 #[cfg(feature = "alloc")]
 use futures_core::future::{BoxFuture, LocalBoxFuture};
@@ -110,36 +111,6 @@ pub use self::remote_handle::{Remote, RemoteHandle};
 mod shared;
 #[cfg(any(feature = "std", all(feature = "alloc", feature = "spin")))]
 pub use self::shared::{Shared, WeakShared};
-
-/// Pins a value on the stack.
-///
-/// **Note:** Since Rust 1.68, this macro is soft-deprecated in favor of
-/// [`pin!`](https://doc.rust-lang.org/std/pin/macro.pin.html) macro
-/// in the standard library.
-///
-/// # Example
-///
-/// ```rust
-/// # use futures_util::pin_mut;
-/// # use core::pin::Pin;
-/// # struct Foo {}
-/// let foo = Foo { /* ... */ };
-/// pin_mut!(foo);
-/// let _: Pin<&mut Foo> = foo;
-/// ```
-#[macro_export]
-macro_rules! pin_mut {
-    ($($x:ident),* $(,)?) => { $(
-        // Move the value to ensure that it is owned
-        let mut $x = $x;
-        // Shadow the original binding so that it can't be directly accessed
-        // ever again.
-        #[allow(unused_mut)]
-        let mut $x = unsafe {
-            ::core::pin::Pin::new_unchecked(&mut $x)
-        };
-    )* }
-}
 
 impl<T: ?Sized> FutureExt for T where T: Future {}
 
--- a/futures-util/src/future/future/mod.rs
+++ b/futures-util/src/future/future/mod.rs
@@ -18,7 +18,6 @@ use futures_core::{
     stream::Stream,
     task::{Context, Poll},
 };
-use pin_utils::pin_mut;
 
 // Combinators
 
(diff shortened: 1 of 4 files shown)


futures-util/src/future/future/mod.rs
```diff
@@ -112,6 +111,36 @@ mod shared;
 #[cfg(any(feature = "std", all(feature = "alloc", feature = "spin")))]
 pub use self::shared::{Shared, WeakShared};
 
+/// Pins a value on the stack.
+///
+/// **Note:** Since Rust 1.68, this macro is soft-deprecated in favor of
+/// [`pin!`](https://doc.rust-lang.org/std/pin/macro.pin.html) macro
+/// in the standard library.
```
taiki-e, line 118: Why dropped [this line](https://github.com/rust-lang/pin-utils/blob/c635c0ca87cd9f9173152f367ef1d84df45943ab/src/stack_pin.rs#L3) from the original doc?

futures-util/src/future/future/mod.rs
```diff
@@ -112,6 +111,36 @@ mod shared;
...
+#[macro_export]
+macro_rules! pin_mut {
+    ($($x:ident),* $(,)?) => { $(
+        // Move the value to ensure that it is owned
+        let mut $x = $x;
+        // Shadow the original binding so that it can't be directly accessed
+        // ever again.
+        #[allow(unused_mut)]
+        let mut $x = unsafe {
+            ::core::pin::Pin::new_unchecked(&mut $x)
```
taiki-e, line 139: This should refer `$crate::__private::Pin` instead of `::core::pin::Pin`, [as pin-utils does](https://github.com/rust-lang/pin-utils/blob/c635c0ca87cd9f9173152f367ef1d84df45943ab/src/stack_pin.rs#L27).

taiki-e (changes requested):
Thanks for the PR!

`futures-util/src/future/future/mod.rs` (which is a module for `Future` trait related things) is not the proper place to put this macro, so perhaps it is better to create `futures-util/src/macros.rs` and put it there.

futures-util/src/future/future/mod.rs
tisonkun: Any rationale here?

I'm afraid that, as `futures_util::__private` is guarded by `async-await` flag, it would create an extra flag gate to use this name.

futures-util/src/future/future/mod.rs
tisonkun: Should be by accident. Will add it back.

tisonkun:
@taiki-e Thanks for your review! Comments addressed.

... except the one about `$crate::__private::Pin` that I found the `__private` mod is guarded by "async-await" flag and being a bit uncertain to use.

tisonkun:
```
error: unused attribute `<cfg_attr>`
   --> futures-util/src/stream/stream/mod.rs:181:1
    |
181 | #[cfg_attr(target_os = "none", cfg(target_has_atomic = "ptr"))]
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
note: the built-in attribute `<cfg_attr>` will be ignored, since it's applied to the macro invocation `delegate_all`
   --> futures-util/src/stream/stream/mod.rs:183:1
    |
183 | delegate_all!(
    | ^^^^^^^^^^^^
    = note: `-D unused-attributes` implied by `-D warnings`
    = help: to override `-D warnings` add `#[allow(unused_attributes)]`
```

Doesn't seem related .. Unsure what happen.

futures-util/src/future/future/mod.rs
taiki-e: > Any rationale here?


See https://github.com/rust-lang/pin-utils/pull/26#discussion_r344491597.



> I'm afraid that, as `futures_util::__private` is guarded by `async-await` flag, it would create an extra flag gate to use this name.

There is no problem with removing the cfg from `__private` module.
That cfg just indicates who the current user is, nothing special.

taiki-e:
> ```
> error: unused attribute `<cfg_attr>`
>    --> futures-util/src/stream/stream/mod.rs:181:1
>     |
> 181 | #[cfg_attr(target_os = "none", cfg(target_has_atomic = "ptr"))]
>     | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
>     |
> note: the built-in attribute `<cfg_attr>` will be ignored, since it's applied to the macro invocation `delegate_all`
>    --> futures-util/src/stream/stream/mod.rs:183:1
>     |
> 183 | delegate_all!(
>     | ^^^^^^^^^^^^
>     = note: `-D unused-attributes` implied by `-D warnings`
>     = help: to override `-D warnings` add `#[allow(unused_attributes)]`
> ```
> 
> Doesn't seem related .. Unsure what happen.

This is https://github.com/rust-lang/rust/issues/138779.

tisonkun:
@taiki-e Looks good now. Please give another review. And see my comment for the following step:

> cc @taiki-e as for "it is reasonable to remove it in v0.4", IIUC I can open a follow-up PR to remove this macro and update the corresponding related usage, only we don't mark that PR as 0.3-backport?
>
> BTW, if there is anything I can help in backporting, please let me know. In my experience, it means opening a PR against branch 0.3 and do the same change.

taiki-e (approved):

tisonkun:
@taiki-e any comment on my following plan? I may create a new PR to remove `pin_mut!` macro and use the core lib macro instead.

That PR would not backport to 0.3 so that we really remove the dependency to pin-utils in 0.4.

taiki-e:
I think the steps to remove pin_mut in v0.4 are:
1. Replace pin_mut with pin in docs (other than docs of pin_mut itself). This can be backported to v0.3.
1. Bump MSRV to 1.68 (for now only for v0.4).
1. Replace pin_mut uses in our code base.
1. Remove pin_mut.

(2 and 3 can be done at the same time, but it would be better to separate 1 and 4 due to backporting and potential reverting.)

tisonkun:
Thanks for your guide @taiki-e! Let me try to handle it in the following days.

BTW, to implement https://github.com/rust-lang/futures-rs/issues/2901, we may end up with MSRV >= 1.85 in futures 0.4?

taiki-e:
> BTW, to implement #2901, we may end up with MSRV >= 1.85 in futures 0.4?

Yes. The next debian stable will use at least 1.85, so 1.85 would be an acceptable and reasonable.

tisonkun:
> Replace pin_mut with pin in docs (other than docs of pin_mut itself).

I'm trying to replace `pin_mut` with `pin` ..

@taiki-e But the macro `core::pin::pin` is only stabilized after 1.68. Do you mean that our docs (tests?) should be `no_run` now? Or I don't know how to search those no-code references. I may start with MSRV bump and then replace all the internal usage of pin_mut.

PR #2930, opened 2025-02-27
Add `merge_round_robin` for StreamExt

RustyNova016:
This function allows merging two streams using a custom "round robin" policy.

This is an useful alternative to `chain` as the second stream doesn't wait for the first to finish, and both can be interweaved toghether

## Example:
```rust
let stream1 = stream::iter(vec!["a", "a"]);
let stream2 = stream::iter(vec!["b", "b", "b", "b", "c"]);
let stream = stream1.merge_round_robin(stream2, 1, 2);
let result: Vec<_> = stream.collect().await;
assert_eq!(result, vec![
    "a",
    "b",
    "b",
    "a",
    "b",
    "b",
    "c"
]);
```

I am quite new to contributing to the official channels, so if there's anything wrong, please tell!

### DIFF
--- a/futures-util/src/stream/mod.rs
+++ b/futures-util/src/stream/mod.rs
@@ -19,9 +19,9 @@ pub use futures_core::stream::{FusedStream, Stream, TryStream};
 mod stream;
 pub use self::stream::{
     All, Any, Chain, Collect, Concat, Count, Cycle, Enumerate, Filter, FilterMap, FlatMap, Flatten,
-    Fold, ForEach, Fuse, Inspect, Map, Next, NextIf, NextIfEq, Peek, PeekMut, Peekable, Scan,
-    SelectNextSome, Skip, SkipWhile, StreamExt, StreamFuture, Take, TakeUntil, TakeWhile, Then,
-    TryFold, TryForEach, Unzip, Zip,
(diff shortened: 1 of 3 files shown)


taiki-e:
Thanks for the PR. What is the difference compared to stream::select and stream::select_with_strategy?

https://docs.rs/futures/latest/futures/stream/fn.select.html
https://docs.rs/futures/latest/futures/stream/fn.select_with_strategy.html#round-robin

RustyNova016:
Well... I admit I didn't see that `stream::select` existed.

But there's a couple changes:
- It's not a standalone function, and is part of `StreamExt`. So it can be inlined
- The strategy is easier to implement. If we want to pass 2 items from stream A and 1 item from stream B, we only have to set the numbers. Doing it with `select_with_strategy` would require either storing state in a mutex or use `enumerate` and modulos

But I understand if it feels too specific when we already have something else implemented

taiki-e:
> * It's not a standalone function, and is part of `StreamExt`. So it can be inlined

It is entirely intentional that it be provided as a standalone function. See https://github.com/rust-lang/futures-rs/issues/1215.

Users who wish to use it as a method are also free to write a few lines of extended trait with a method that calls it, so I prefer not to add a new API that does the same thing but with a completely different name for it.

Therefore, I'm closing this PR, but thanks anyway for the PR!

PR #2940, opened 2025-04-05
chore: replace all doctest usage of pin_mut with pin 

tisonkun:
This follows up https://github.com/rust-lang/futures-rs/pull/2929#issuecomment-2746232982 step 1.

cc @taiki-e

### DIFF
--- a/futures-util/src/macros.rs
+++ b/futures-util/src/macros.rs
@@ -9,11 +9,11 @@
 /// # Example
 ///
 /// ```rust
-/// # use futures_util::pin_mut;
+/// # use core::pin::pin;
 /// # use core::pin::Pin;
 /// # struct Foo {}
 /// let foo = Foo { /* ... */ };
-/// pin_mut!(foo);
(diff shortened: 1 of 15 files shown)


futures-util/src/macros.rs
```diff
@@ -9,11 +9,11 @@
 /// # Example
 ///
 /// ```rust
-/// # use futures_util::pin_mut;
+/// # use core::pin::pin;
 /// # use core::pin::Pin;
 /// # struct Foo {}
 /// let foo = Foo { /* ... */ };
-/// pin_mut!(foo);
+/// let foo = pin!(foo);
```
taiki-e, line 16: Please do not change this example because this is an example for pin_mut macro itself.

tisonkun: Yes .. I should retain this the first time but seems some global editions touch this ><

Reverted.

taiki-e (approved):
Thank!

PR #2941, opened 2025-04-05
chore: replace all internal usage of pin_mut

tisonkun:
This follows up https://github.com/rust-lang/futures-rs/pull/2929#issuecomment-2746232982 step 2.

cc @taiki-e 

We don't use a unified MSRV. So I'm understanding which versions should be updated.

### DIFF
--- a/futures-util/src/future/future/mod.rs
+++ b/futures-util/src/future/future/mod.rs
@@ -3,14 +3,14 @@
 //! This module contains a number of functions for working with `Future`s,
 //! including the `FutureExt` trait which adds methods to `Future` types.
 
-#[cfg(feature = "alloc")]
-use alloc::boxed::Box;
-use core::convert::Infallible;
-use core::pin::Pin;
-use std::pin::pin;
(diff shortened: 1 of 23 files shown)


tisonkun:
We have some MSRV as `rust-version = "1.36"`, not sure why they are different.

taiki-e:
Thanks!

Considering that all of the affected code except futures-executor/src/local_pool.rs and futures-util/src/future/future/mod.rs are tests, it might be a good idea to split those two into a separate PR. In that case, the changes to the tests would not affect the MSRV, so we can backport these right in.

futures-util/src/future/future/mod.rs
```diff
@@ -27,7 +26,7 @@ mod fuse;
 mod map;
 
 delegate_all!(
-    /// Future for the [`flatten`](super::FutureExt::flatten) method.
```
taiki-e, line 30: Please split unrelated changes into separate PRs as they will be difficult to review.

tisonkun:
OK. I'll submit two separated PRs.

PR #2942, opened 2025-04-05
chore: replace all test usage of pin_mut

tisonkun:
cc @taiki-e 

This splits from https://github.com/rust-lang/futures-rs/pull/2941


taiki-e (approved):
Thanks!

PR #2943, opened 2025-04-05
chore: replace all usage of pin_mut

tisonkun:
cc @taiki-e 

This splits from https://github.com/rust-lang/futures-rs/pull/2941


tisonkun:
@taiki-e if you have spare time now, we can finish the whole deprecation and removal today :D

taiki-e (approved):
Thanks!

PR #2944, opened 2025-04-07
channel: Add try_recv and deprecate try_next

### LINKED ISSUE #2197: rename `mpsc::Receiver::try_next`
I think [`mpsc::Receiver::try_next`] should be renamed, since it conflicts with [`TryStreamExt::try_next`]. Since receiver is actually a `Stream` itself it's definitely confusing that there's two `try_next()`s that could apply that do distinctly different things.

I'd argue that the `Receiver` one is more niche, since ([docs][`mpsc::Receiver::try_next`]):

> Tries to receive the next message without notifying a context if empty.
>
> It is not recommended to call this function from inside of a future, only when you've otherwise arranged to be notified when the channel is no longer empty.
>
> This function will panic if called after try_next or poll_next has returned None.

Those caveats make it so you have to be careful in how you use it. IMO we should save the "good name" `try_next` for the one that returns a future, and pick a less nice name for this case.

I'm not sure what it should be called though. `next_nonblock`?

Also side note it might be worth checking to see if there's any cases where currently compiling code using `Receiver::try_next` might erroneously _keep_ compiling after that method is renamed because TryStreamExt is still in scope.

[`mpsc::Receiver::try_next`]: https://docs.rs/futures/0.3.5/futures/channel/mpsc/struct.Receiver.html#method.try_next
[`TryStreamExt::try_next`]: https://docs.rs/futures/0.3.5/futures/stream/trait.TryStreamExt.html#method.try_next

### LINKED ISSUE #2936: api: consider changing  `mpsc::Receiver::try_next` for future_channels v0.4.
This is just a minor API gripe, feel free to close as "working as intended" if you disagree.

My hope is just to see if people agree so we could maybe improve a minor pain point,
since the library in question is still not at 1.0.

The awesome `future_channels` crate contains an mpsc Receiver who has a [`try_next`](https://docs.rs/futures-channel/latest/futures_channel/mpsc/struct.Receiver.html#method.try_next) function
with the following API:

```rust
pub fn try_next(&mut self) -> Result<Option<T>, TryRecvError>
```

Exercise: Try to guess from the api what this function will return if you don't happen to remember.

In reality,  this function will return `None` if the channel is closed, 
but return an `TryRecvError`, if there's currently no message available.

This is completely backwards from what the user expectation would be in my opinion.
The channel being closed feels like the error, whereas currently no messages being queued 
is just a valid state, since the whole point is that this method is not `async`.

As a compromise I can also get behind the way that e.g. the
`async_channel` crate [does it](
https://docs.rs/async-channel/latest/async_channel/struct.Receiver.html#method.try_recv):

```rust
pub fn try_recv(&self) -> Result<T, TryRecvError> {
```

where the method does not return an option, and both special cases are grouped in an error enum.
Changing it this way would also have the advantage of causing a compiler error in users of the previous
version instead of silently changing behavior.

cmrschwarz:
Adds `try_recv` and deprecates `try_next` as discussed in #2936.
  
  It does so by changing the previously private `TryRecvError` internals from a struct to an an enum.
  
  Note that this is *technically* not semver compatible with 0.3 as the following is legal (albeit nonsensical) Rust:
  
  ```Rust
  let TryRecvError {..} = my_err;
  ```
  This would no longer compile after this change.
  I'm happy to introduce some slightly hacky `TryReceiveError` or similar instead if that is preferred.

---

 I've also moved most most tests over to `try_recv` but kept around [these two](https://github.com/rust-lang/futures-rs/compare/master...cmrschwarz:try_recv?expand=1#diff-dd03b534a872e5823ae44f0ff4de693c8eab5e87b94cf82ebcfde65121b643b0L280) using `try_next` just to make sure it isn't somehow broken.

--- 

Adding `recv` aswell proved harder than expected because I can't add a dependency on 
  `futures-util`, which would be cyclic. I see two options:
  1. Take out the `futures-channel` specific code from `futures-util` and move it into `futures-channel`.   
  2. Essentially reimplement [`Next`](https://docs.rs/futures-util/latest/src/futures_util/stream/stream/next.rs.html#10-12) as `Recv`. Might be reasonable as `Next` is pretty small.

Which would you prefer? I'm happy do do either as a follow up if that still seems good, or of course do a third option if there's a better Idea.

---

Closes #2936
Closes #2197

### DIFF
--- a/futures-channel/src/mpsc/mod.rs
+++ b/futures-channel/src/mpsc/mod.rs
@@ -167,9 +167,14 @@ enum SendErrorKind {
     Disconnected,
 }
 
-/// The error type returned from [`try_next`](Receiver::try_next).
-pub struct TryRecvError {
-    _priv: (),
+/// The error type returned from [`try_recv`](Receiver::try_recv).
+#[derive(PartialEq, Eq, Clone, Copy, Debug)]
+pub enum TryRecvError {
+    /// The channel is empty but not closed.
+    Empty,
+
+    /// The channel is empty and closed.
+    Closed,
 }
 
 impl fmt::Display for SendError {
@@ -193,6 +198,18 @@ impl SendError {
     /// Returns `true` if this error is a result of the receiver being dropped.
     pub fn is_disconnected(&self) -> bool {
         matches!(self.kind, SendErrorKind::Disconnected)
+    }
+}
+
+impl TryRecvError {
+    /// Returns `true` if the channel is empty but not closed.
+    pub fn is_empty(&self) -> bool {
+        matches!(self, TryRecvError::Empty)
+    }
+
+    /// Returns `true` if the channel is empty and closed.
+    pub fn is_closed(&self) -> bool {
+        matches!(self, TryRecvError::Closed)
     }
 }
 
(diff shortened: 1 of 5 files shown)


cmrschwarz:
Sorry for the double CI, I accidentally left a comment that doesn't make sense anymore. Fixed.

PR #2945, opened 2025-04-08
chore: remove deprecated pin_mut macro in 0.4

tisonkun:
This follows up https://github.com/rust-lang/futures-rs/pull/2929#issuecomment-2746232982 step 4.

cc @taiki-e 

Also, since our MSRV has been bumped to 1.68, we have `core::task::ready` macro in the namespace. Shall we do the same on `ready!` so that encourage downstream projects to converge to the core lib?

Besides, shall we add an unreleased changelog for this patch on migration/upgrade notes?


taiki-e (approved):
Thanks!

PR #2946, opened 2025-04-08
chore: replace num_cpus with available_parallelism

tisonkun:
Related discussion: https://github.com/tokio-rs/tokio/issues/6211 and https://github.com/tokio-rs/tokio/pull/6709

cc @taiki-e

### DIFF
--- a/futures-executor/Cargo.toml
+++ b/futures-executor/Cargo.toml
@@ -13,13 +13,12 @@ Executors for asynchronous tasks based on the futures-rs library.
 [features]
 default = ["std"]
 std = ["futures-core/std", "futures-task/std", "futures-util/std"]
-thread-pool = ["std", "num_cpus"]
+thread-pool = ["std"]
 
 [dependencies]
 futures-core = { path = "../futures-core", version = "=1.0.0-alpha.0", default-features = false }
 futures-task = { path = "../futures-task", version = "=0.4.0-alpha.0", default-features = false }
 futures-util = { path = "../futures-util", version = "=0.4.0-alpha.0", default-features = false }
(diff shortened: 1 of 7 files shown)


tisonkun:
https://github.com/rust-lang/futures-rs/actions/runs/14322140475/job/40140977417

Failed to compile iovec 0.1.0 - doesn't seem to be related. It should be resolved to 0.1.4 and previously runs all passed. This patch doesn't touch it.

taiki-e:
iovec crate is used by old dependencies that are enabled via the compat/compat-io feature, but given that our traits are planned to remain basically unchanged between futures 0.3 and 0.4, it is not a problem to remove compat and compat-io features in 0.4. (From a maintenance standpoint, their removal is also welcome because they were not easy to maintain (e.g., https://github.com/rust-lang/futures-rs/issues/2514).)

tisonkun:
Thanks for your information! Seems I can take a look at dropping dependency to iovec in another PR. For this one, we'd on hold until then or I can rebase on master and ping you once CI is happy.

tisonkun:
Seems libc's version causes this issue.

Previously, we have:

>       Adding libc v0.2.26 (available: v0.2.172)

Now, it becomes:

>       Adding libc v0.2.0 (available: v0.2.172)

Maybe num_cpus add a requirement to help indicates the version. I'll try to add a specific libc version and see.

tisonkun:
@taiki-e This patch should be ready for merge now.

The clippy related changes are attached because we have a clippy CI check that runs on latest compiler. I suppose it's intended.

taiki-e (approved):
Thanks!

PR #2947, opened 2025-04-09
channel: Add recv

### LINKED ISSUE #2184: Add recv method to mpsc::Receiver
Add recv method to `Receiver` how was done in std and tokio.

cmrschwarz:
Implements `recv()` for `Receiver` and `UnboundedReceiver`, as discussed in #2944.

Since we're reimplementing `Next` anyways we might aswell get the benefits so this returns a new `RecvError`
instead of an `Option` for better symmetry with `try_recv` and other crates like [async_channel](https://docs.rs/async-channel/latest/async_channel/struct.Receiver.html#method.recv).

To be fair tokio [uses an option](https://docs.rs/tokio/latest/tokio/sync/mpsc/struct.Receiver.html#method.recv), but if we want to do comparisons then the style of [std::mpsc](https://doc.rust-lang.org/std/sync/mpsc/struct.Receiver.html#method.recv) is probably the one to immitate, and they are using a dedicated error. 
  
---
Closes #2184

### DIFF
--- a/futures-channel/src/mpsc/mod.rs
+++ b/futures-channel/src/mpsc/mod.rs
@@ -78,9 +78,11 @@
 // happens-before semantics required for the acquire / release semantics used
 // by the queue structure.
 
+use core::future::Future;
 use futures_core::stream::{FusedStream, Stream};
 use futures_core::task::__internal::AtomicWaker;
 use futures_core::task::{Context, Poll, Waker};
+use futures_core::FusedFuture;
 use std::fmt;
 use std::pin::Pin;
 use std::sync::atomic::AtomicUsize;
@@ -177,6 +179,18 @@ pub enum TryRecvError {
     Closed,
 }
 
+/// Error returned by `Receiver::recv()`[Receiver::recv].
+/// Received when the channel is empty and closed.
+#[derive(PartialEq, Eq, Clone, Copy, Debug)]
+pub struct RecvError;
+
+/// Future returned by `Receiver::recv()`[Receiver::recv].
+#[derive(Debug)]
+#[must_use = "futures do nothing unless you `.await` or poll them"]
+pub struct Recv<'a, St: ?Sized> {
+    stream: &'a mut St,
+}
+
 impl fmt::Display for SendError {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         if self.is_full() {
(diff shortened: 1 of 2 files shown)


futures-channel/src/mpsc/mod.rs
```diff
@@ -189,6 +203,14 @@ impl fmt::Display for SendError {
 
 impl std::error::Error for SendError {}
 
+impl fmt::Display for RecvError {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "receive failed because channel is closed")
```
taiki-e, line 208:
```suggestion
        write!(f, "receive failed because channel is empty and closed")
```

futures-channel/src/mpsc/mod.rs
```diff
@@ -177,6 +179,18 @@ pub enum TryRecvError {
     Closed,
 }
 
+/// Error returned by `Receiver::recv()`[Receiver::recv].
+/// Received when the channel is empty and closed.
+#[derive(PartialEq, Eq, Clone, Copy, Debug)]
+pub struct RecvError;
+
+/// Future returned by `Receiver::recv()`[Receiver::recv].
```
taiki-e, line 187:
```suggestion
/// Future returned by [`Receiver::recv()`] or [`UnboundedReceiver::recv()`].
```

futures-channel/src/mpsc/mod.rs
```diff
@@ -1121,6 +1149,31 @@ impl<T> Stream for Receiver<T> {
     }
 }
 
+impl<St: ?Sized + Unpin> Unpin for Recv<'_, St> {}
+impl<'a, St: ?Sized + Stream + Unpin> Recv<'a, St> {
+    pub(super) fn new(stream: &'a mut St) -> Self {
```
taiki-e, line 1154:
```suggestion
    fn new(stream: &'a mut St) -> Self {
```

futures-channel/src/mpsc/mod.rs
```diff
@@ -177,6 +179,18 @@ pub enum TryRecvError {
     Closed,
 }
 
+/// Error returned by `Receiver::recv()`[Receiver::recv].
```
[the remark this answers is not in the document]
cmrschwarz: addressed by [b3205f7](https://github.com/rust-lang/futures-rs/pull/2947/commits/b3205f77f7df2255e28b927357c459c6d6cdc997)

futures-channel/src/mpsc/mod.rs
cmrschwarz: Addressed by  [58804ce](https://github.com/rust-lang/futures-rs/pull/2947/commits/58804ce091bf42aecf7028b58453489cd04df29f).

taiki-e (approved):

cmrschwarz:
Thanks for the review, this should address the mentioned concerns.
I just realized that my commit message in [58804ce](https://github.com/rust-lang/futures-rs/pull/2947/commits/58804ce091bf42aecf7028b58453489cd04df29f) is the wrong way around, I somehow misread the diff. I assume we'll squash anyways?

PR #2949, opened 2025-04-12
refactor: remove leftover from `read_initializer` feature

paolobarbolini:
While auditing the use of unsafe, I've discovered a strange way of growing and initialising the extended part of a `Vec`, which resulted in a `Vec::set_len` call that technically violates the safety invariants - the new length isn't fully init yet, and also uses unsafe operations and a `read_initializer` style API for something so simple.

This replaces it with a much simpler implementation.


taiki-e (approved):
Thanks!

PR #2966, opened 2025-08-26
fix: docs and cfg std for select macro

tisonkun:
This closes https://github.com/rust-lang/futures-rs/issues/2932

cc @taiki-e @petekubiak

### DIFF
--- a/futures-util/src/async_await/mod.rs
+++ b/futures-util/src/async_await/mod.rs
@@ -27,10 +27,8 @@ mod select_mod;
 pub use self::select_mod::*;
 
 // Primary export is a macro
-#[cfg(feature = "std")]
 #[cfg(feature = "async-await-macro")]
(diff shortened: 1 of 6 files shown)


tisonkun:
@taiki-e Thanks for your information! I'll take a look later.

> Thanks for the PR, but this is not correct. select! refers random module which gated under std feature, so cannot be used without std feature.

I tried to run:

```
cargo build --no-default-features --features async-await
cargo build --no-default-features --features async-await-macro
```

locally. And both are compiled. But yes now I see: `__futures_crate::async_await::shuffle(&mut __select_arr);` that depends on std.

tisonkun:
Cross ref:

* https://github.com/rust-lang/futures-rs/pull/1912
* https://github.com/rust-lang/futures-rs/pull/2903

PR #2982, opened 2025-12-17
feat: implement `last` helper method on `StreamExt`

### LINKED ISSUE #2822: Analogue of .last() method
I'd like to run `FuturesOrdered::from_iter(...)` and retrieve the result of the last element, which is `Option<...>`. The same function [exists](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.last) in std.

elsirion:
The `last` function consumes the stream and returns the last element.

Fixes #2822


elsirion:
I'm a bit lost why clippy and docs are failing for files I haven't even touched. What am I doing wrong?

PR #2989, opened 2026-02-15
Bump MSRV of utility crates to 1.71

taiki-e:
```
  error: package `syn v2.0.115` cannot be built because it requires rustc 1.71 or newer, while the currently active rustc version is 1.68.2
```

### DIFF
--- a/futures-util/src/stream/stream/flatten_unordered.rs
+++ b/futures-util/src/stream/stream/flatten_unordered.rs
@@ -371,7 +371,7 @@ where
 {
     /// Checks if current `inner_streams` bucket size is greater than optional limit.
     fn is_exceeded_limit(&self) -> bool {
-        self.limit.map_or(false, |limit| self.inner_streams.len() >= limit.get())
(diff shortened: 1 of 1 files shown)


matklad:
As a user of the futures crate, I was negatively affected by this change, and, at a first glance, this change doesn't make sense, given the syn motivation:

* Crates like `futures-channel` do not transitively depend on the syn ecosystem, and were perfectly usable with older rustc. Now however, the build is actively broken:

```
error: package `futures-channel v0.3.32` cannot be built because it requires rustc 1.71 or newer, while the currently active rustc version is 1.68.2
```

* Even crates like `futures-macros`, which do depend transitively on syn, are still usable with older `rustc`, if the user explicitly pins a combination of older syn and newer `futures-macros`. 

In other words, my understanding is that it is an intentional feature that a library MSRV can be lower (more relaxed), than MSRV of its transitive dependencies. 

Of course, if the motivation here is not to just do what syn does, but rather to directly take advantage of newer language features _directly_ in futures code, then that's totally fine and I'll adjust my code!

PR #2997, opened 2026-03-05
feat: Add `StreamExt::switch`

Hywan:
This patch introduces the `Switch` combinator stream: it flattens a higher-order stream into a first-order stream.

This combinator flattens a stream of streams, i.e. an outer stream yielding inner streams. This combinator always keeps the most recently yielded inner stream, and yields items from it, until the outer stream produces a new inner stream, at which point the inner stream to yield items from is switched to the new one.

### Examples

An empty (outer) stream can be switched:

```rust
use futures::stream::{self, Empty, StreamExt, Switch};

let stream: Switch<Empty<Empty<()>>> = stream::empty().switch();

assert!(stream.collect::<Vec<_>>().await.is_empty());
```

An outer stream can produce several inner streams — once switched, the last one is immediately selected:

```rust

let stream = stream::iter([
    stream::iter([1, 2, 3]),
    stream::iter([4, 5, 6]),
    stream::iter([7, 8, 9]),
])
.switch();

assert_eq!(vec![7, 8, 9], stream.collect::<Vec<_>>().await);
```

One of the most interesting usecase is when an outer stream produces new inner  streams dynamically. Let's imagine the outer stream produces inner streams yielding a sequence of integers starting from a particular value. For example:

- outer stream yields 7:
  - inner stream yields 7, 8, 9, 10, 11…
- outer stream yields 42:
  - inner stream yields 42, 43, 44, 45, 46…

```rust
use futures::channel::mpsc;
use futures::task::Poll;
use std::pin::pin;

// `receiver` is the outer stream: it yields integers received by
// `sender`.
let (sender, receiver) = mpsc::unbounded::<u8>();

let mut stream = pin!(receiver
    // For every new received value from `sender`…
    .map(|init_value| {
        // … let's create a new inner stream.
        //
        // First off, the inner stream starts with `init_value`. Then,
        // it continues with incremented integers.
        let mut next_value = init_value;

        stream::poll_fn(move |_| {
            let current_value = next_value;
            next_value += 1;

            Poll::Ready(Some(current_value))
        })
    })
    .switch());

// `stream` is pending until the outer stream yields something.
sender.unbounded_send(7).unwrap();

// `stream` has switched to the inner stream.
assert_eq!(vec![7, 8, 9, 10, 11], stream.by_ref().take(5).collect::<Vec<_>>().await);
assert_eq!(vec![12, 13, 14, 15, 16], stream.by_ref().take(5).collect::<Vec<_>>().await);

// The outer stream will yield a new value, which will create a new
// inner stream.
sender.unbounded_send(42).unwrap();

// `stream` has been “reset” and will produce a new inner stream.
assert_eq!(vec![42, 43, 44, 45, 46], stream.take(5).collect::<Vec<_>>().await);
```

This is a port of [`async_rx::Switch`](https://github.com/jplatte/async-rx), written by @jplatte, and maintained by @jplatte and I. The test suite has been improved compared to the original repository.

I believe this combinator is useful when one wants to generate streams dynamically (the inner streams) based on another input (managed by the outer stream).

### DIFF
--- a/futures-util/src/stream/stream/switch.rs
+++ b/futures-util/src/stream/stream/switch.rs
@@ -1,0 +1,90 @@
+use core::pin::Pin;
+
+use futures_core::stream::Stream;
+use futures_core::task::{Context, Poll};
+use futures_core::FusedStream;
+use pin_project_lite::pin_project;
+
+pin_project! {
+    /// Stream for the [`switch`](super::StreamExt::switch) method.
+    #[derive(Debug)]
+    #[must_use = "futures do nothing unless you `.await` or poll them"]
+    pub struct Switch<Outer, Inner> {
+        #[pin]
+        outer_stream: Outer,
+        outer_stream_is_closed: bool,
+        #[pin]
+        inner_stream: Option<Inner>,
+    }
+}
+
+impl<Outer: Stream> Switch<Outer, Outer::Item> {
+    /// Creates a new [`Switch`] stream.
+    pub(super) fn new(outer_stream: Outer) -> Self {
+        Self { outer_stream, outer_stream_is_closed: false, inner_stream: None }
+    }
+
+    delegate_access_inner!(outer_stream, Outer, ());
+}
+
+impl<Outer> FusedStream for Switch<Outer, Outer::Item>
+where
+    Outer: FusedStream,
+    Outer::Item: Stream,
+{
+    fn is_terminated(&self) -> bool {
+        self.inner_stream.is_none() && self.outer_stream.is_terminated()
+    }
+}
+
+impl<Outer> Stream for Switch<Outer, Outer::Item>
+where
+    Outer: Stream,
+    Outer::Item: Stream,
+{
+    type Item = <Outer::Item as Stream>::Item;
+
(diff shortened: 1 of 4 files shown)


futures-util/src/stream/stream/switch.rs
```diff
@@ -0,0 +1,90 @@
...
+use pin_project_lite::pin_project;
+
+pin_project! {
+    /// Stream for the [`switch`](super::StreamExt::switch) method.
+    #[derive(Debug)]
+    #[must_use = "futures do nothing unless you `.await` or poll them"]
+    pub struct Switch<Outer, Inner> {
+        #[pin]
+        outer_stream: Outer,
+        outer_stream_is_closed: bool,
```
Hywan, line 15: Can be removed by `Outer: FusedStream`. This is debatable and I'm happy to change it.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/futures-rs/2997/d8836438bc686972f8359ec20fe6bad78304730a..71a9245e1a112c216427ab954b7479a333424872)*

PR #3000, opened 2026-04-06
Hash-pin some action references, add dependency cooldowns

woodruffw:
This uses `pinact run -v` to hash-pin some (but not all) of futures-rs's actions dependencies; some of the actions used by the CI are not trivial to pin automatically and require manual consideration (like `taiki-e/github-actions/install-rust@nightly`).

Separately, I've added [cooldown](https://docs.github.com/en/code-security/reference/supply-chain-security/dependabot-options-reference#cooldown-) settings to each of the configured Dependabot groups; this won't slow the actual update cadence but _will_ prevent updates until each dependency is at least a week old, which gives security practitioners and scanners a bit of time to discover malware before it can propagate through the ecosystem.

(I've split these into separate commits to hopefully make them easier to review!)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -52,10 +52,10 @@ jobs:
     runs-on: ${{ matrix.os }}
     timeout-minutes: 60
     steps:
-      - uses: taiki-e/checkout-action@v1
+      - uses: taiki-e/checkout-action@7d1e50e93dc4fb3bba58f85018fadf77898aee8b # v1.4.2
       - name: Install Rust
         uses: taiki-e/github-actions/install-rust@nightly
-      - uses: taiki-e/setup-cross-toolchain-action@v1
+      - uses: taiki-e/setup-cross-toolchain-action@d62f33b587e73b0004731caebc7d2d46b18a7567 # v1.39.2
         with:
           target: ${{ matrix.target }}
         if: matrix.target != ''
@@ -73,7 +73,7 @@ jobs:
     runs-on: ubuntu-latest
     timeout-minutes: 60
     steps:
-      - uses: taiki-e/checkout-action@v1
+      - uses: taiki-e/checkout-action@7d1e50e93dc4fb3bba58f85018fadf77898aee8b # v1.4.2
       - name: Install Rust
         uses: taiki-e/github-actions/install-rust@main
         with:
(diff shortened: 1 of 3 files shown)


taiki-e:
Thanks for the PR! 

I’ve added zizmor to CI in #3001, so could you rebase to master and remove the unpinned-uses configuration?

I think we can use hash-pin for ones that currently using branch in the same way as https://github.com/taiki-e/github-actions/commit/90dc768e86204d39674d33877056f56209184956.

Also, since Dependabot gets a bit noisy when hash-pin is used,  I'd appreciate it if grouped updates could be enabled as well like https://github.com/taiki-e/github-actions/commit/f3fa0898d7188b104fe42b4c808aed8ce4755203.

woodruffw:
Thanks for the review! Yeah, I'll deconflict and follow your suggestions for the other bits 🙂

.github/workflows/ci.yml
```diff
@@ -52,10 +52,10 @@ jobs:
     runs-on: ${{ matrix.os }}
     timeout-minutes: 60
     steps:
-      - uses: taiki-e/checkout-action@v1
+      - uses: taiki-e/checkout-action@7d1e50e93dc4fb3bba58f85018fadf77898aee8b # v1.4.2
       - name: Install Rust
-        uses: taiki-e/github-actions/install-rust@nightly
-      - uses: taiki-e/setup-cross-toolchain-action@v1
+        uses: taiki-e/github-actions/install-rust@4defe325132c540b152e731162b9067fab71ec21 # nightly # zizmor: ignore[stale-action-refs]
```
taiki-e, line 57: Handling of branches other than `main` is similar to taiki-e/install-action's `@<tool_name>` tags and dtolnay/rust-toolchain's `@<rust_version>` branches, so here I would like to use `@<hash> # main` and `toolchain: nightly`.

FYI, I recently documented in the install-action readme why I don’t recommend using hash-pin for this kind of tags/branches:

https://github.com/taiki-e/install-action?tab=readme-ov-file#security

> Pinning `@<tool_name>` tags by hash is strongly discouraged, as it causes the workflow to reference a [commit that is not present on the repository](https://docs.zizmor.sh/audits/#impostor-commit) when a new version is released.

woodruffw:
Sorry, I haven't had time to get back to this. I'm going to close to avoid confusing people about its state, but IMO it's still worth doing 🙂

taiki-e:
(This has been addressed in https://github.com/rust-lang/futures-rs/pull/3022.)

PR #3003, opened 2026-04-11
Fix unsound `Send` impl for `IterPinRef` and `Iter`

### LINKED ISSUE #3002: `IterPinRef` and `Iter` unsoundly implement `Send` when `Fut: Send` instead of `Fut: Sync`
AI disclosure: this issue was found via Claude Code running Claude Opus 4.6.

## Summary

`IterPinRef` (and the `Iter` wrapper) implement `Send` when `Fut: Send`, but they yield `Pin<&Fut>` (shared references). Sending the iterator to another thread while retaining shared access on the original thread creates `&Fut` on two threads simultaneously, which requires `Fut: Sync`. When `Fut` is `Send` but not `Sync` (e.g. `Cell<i32>`), this is a data race.

The `unsafe impl` at `futures-util/src/stream/futures_unordered/iter.rs:165` overrides auto-trait derivation entirely, so the `PhantomData<&'a FuturesUnordered<Fut>>` field does not contribute its normal bounds.

`IterPinMut` is not affected because it yields `Pin<&mut Fut>`, so exclusive access prevents aliasing.

## Minimal reproducer

```rust
use futures::stream::FuturesUnordered;
use std::cell::Cell;
use std::pin::Pin;

fn main() {
    let fu: FuturesUnordered<Cell<i32>> = FuturesUnordered::new();
    fu.push(Cell::new(0));

    std::thread::scope(|s| {
        let pinned = Pin::new(&fu);
        let iter1 = pinned.iter_pin_ref();
        let iter2 = pinned.iter_pin_ref();

        s.spawn(move || {
            for cell_ref in iter1 {
                cell_ref.set(cell_ref.get() + 1);
            }
        });

        for cell_ref in iter2 {
            cell_ref.set(cell_ref.get() + 1);
        }
    });
}
```

```
cargo +nightly miri run
```

The result:

```
error: Undefined Behavior: Data race detected between (1) non-atomic write on thread `main` and (2) non-atomic read on thread `unnamed-1` at alloc475+0x1c
   --> /playground/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/cell.rs:555:18
    |
555 |         unsafe { *self.value.get() }
    |                  ^^^^^^^^^^^^^^^^^ (2) just happened here
    |
help: and (1) occurred earlier here
   --> src/main.rs:21:13
    |
 21 |             cell_ref.set(cell_ref.get() + 1);
    |             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    = help: this indicates a bug in the program: it performed an invalid operation, and caused Undefined Behavior
    = help: see https://doc.rust-lang.org/nightly/reference/behavior-considered-undefined.html for further information
    = note: this is on thread `unnamed-1`
    = note: stack backtrace:
            0: std::cell::Cell::<i32>::get
                at /playground/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/cell.rs:555:18: 555:35
            1: main::{closure#0}::{closure#0}
                at src/main.rs:16:30: 16:44
note: the last function in that backtrace got called indirectly due to this code
   --> src/main.rs:14:9
    |
 14 | /         s.spawn(move || {
 15 | |             for cell_ref in iter1 {
 16 | |                 cell_ref.set(cell_ref.get() + 1);
 17 | |             }
 18 | |         });
    | |__________^

note: some details are omitted, run with `MIRIFLAGS=-Zmiri-backtrace=full` for a verbose backtrace

error: aborting due to 1 previous error
```

paolobarbolini:
Fixes #3002

### DIFF
--- a/futures-util/src/stream/futures_unordered/iter.rs
+++ b/futures-util/src/stream/futures_unordered/iter.rs
@@ -160,11 +160,14 @@ impl<'a, Fut: Unpin> Iterator for Iter<'a, Fut> {
 
 impl<Fut: Unpin> ExactSizeIterator for Iter<'_, Fut> {}
 
-// SAFETY: we do nothing thread-local and there is no interior mutability,
-// so the usual structural `Send`/`Sync` apply.
-unsafe impl<Fut: Send> Send for IterPinRef<'_, Fut> {}
+// SAFETY: `IterPinRef` yields `Pin<&Fut>` shared references. Sending these
+// to another thread requires `Fut: Sync` since both the sender and receiver
+// can concurrently access the same `Fut` through their shared references.
+unsafe impl<Fut: Sync> Send for IterPinRef<'_, Fut> {}
 unsafe impl<Fut: Sync> Sync for IterPinRef<'_, Fut> {}
 
+// SAFETY: we do nothing thread-local and there is no interior mutability,
+// so the usual structural `Send`/`Sync` apply.
 unsafe impl<Fut: Send> Send for IterPinMut<'_, Fut> {}
 unsafe impl<Fut: Sync> Sync for IterPinMut<'_, Fut> {}
 
--- a/futures/tests/auto_traits.rs
+++ b/futures/tests/auto_traits.rs
@@ -10,6 +10,7 @@ use futures::{
     task::{Context, Poll},
 };
 use static_assertions::{assert_impl_all as assert_impl, assert_not_impl_all as assert_not_impl};
+use std::cell::Cell;
 use std::marker::PhantomPinned;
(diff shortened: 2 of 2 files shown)


taiki-e (approved):
Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/futures-rs/3003/2fe9302dbbfe3f9553163ed586e9ef408190b77e..b031294152b705c4434d32b5de6e0f16853ba5ac)*

taiki-e:
Published in 0.3.33.

PR #3020, opened 2026-07-05
Fix ReadLine's soundness issue regarding to exception safety

### LINKED ISSUE #3019: `futures-util`'s `Readline` is not exception safe, leading to unsound `Readline::drop`
Hello, we are security researchers targeting Rust security. By running our tool in this repository, we found a soundness issue. **This report is written by 100% human**. We promise that all you read will never be generated by LLM.

The problem happens here:

https://github.com/rust-lang/futures-rs/blob/f68806c1205a6495d5c381bc9180d162e791b010/futures-util/src/io/read_until.rs#L27-L51

`ReadLine::read_until_internal` is called by `ReadLine::poll`. In this method at line 45, `reader.as_mut().consume(used)`, the `AsyncBufRead::consume` is called, which is a generic trait method that can be implemented by users. If the user deliberately panics at `consume` function, then the `*read += used` will never be executed at line 46, while `buf` is extended at line 38 or 41.

However, in `ReadLine::drop`:

https://github.com/rust-lang/futures-rs/blob/f68806c1205a6495d5c381bc9180d162e791b010/futures-util/src/io/read_line.rs#L67-L77

The safety comment says "`bytes` is valid UTF-8 because it was taken from a String and the newly read bytes have been removed", but `self.read` is incorrect due to the panic above, leading to `self.bytes.truncate` not guarantee "the newly read bytes have been removed", violating the safety contract.

To fix it, I think we can just make the `*read += used` executed before `reader.as_mut().consume(used)`.

Evian-Zhang:
Fix #3019

### DIFF
--- a/futures-util/src/io/read_until.rs
+++ b/futures-util/src/io/read_until.rs
@@ -42,8 +42,8 @@ pub(super) fn read_until_internal<R: AsyncBufRead + ?Sized>(
                 (false, available.len())
             }
         };
-        reader.as_mut().consume(used);
         *read += used;
+        reader.as_mut().consume(used);
         if done || used == 0 {
             return Poll::Ready(Ok(*read));
         }


taiki-e (approved):
Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/futures-rs/3020/f68806c1205a6495d5c381bc9180d162e791b010..7aeb38a70ad2bc1a389df36666d11dca77bf6907)*

taiki-e:
Published in 0.3.33.

PR #3028, opened 2026-07-26
chore(macro): bump `syn` to v3

AudaciousAxiom:
Following the [release of a new major version of `syn`](https://github.com/dtolnay/syn/releases/tag/3.0.0), this bumps the version used by `futures-macro`, with the goal of eventually removing the duplicated dependency in binaries (and therefore reducing compile times).

`syn` v3 [advertises an MSRV of 1.71](https://crates.io/crates/syn/3.0.0/code/Cargo.toml#L14), so this shouldn't affect the MSRV of this crate.

`syn` v3 of course contains breaking changes: I've been through the changelog but didn't find anything relevant (but there are quite a few of them so I could have missed some). I've run `cargo test --all-features` but didn't do any other kind of testing.

Closes #3034

### DIFF
--- a/futures-macro/Cargo.toml
+++ b/futures-macro/Cargo.toml
@@ -19,7 +19,7 @@ proc-macro = true
 [dependencies]
 proc-macro2 = "1.0.60"
 quote = "1.0"
-syn = { version = "2.0.52", features = ["full"] }
+syn = { version = "3.0", features = ["full"] }
 
 [lints]
 workspace = true


futures-macro/Cargo.toml
```diff
@@ -19,7 +19,7 @@ proc-macro = true
 [dependencies]
 proc-macro2 = "1.0.60"
 quote = "1.0"
-syn = { version = "2.0.52", features = ["full"] }
+syn = { version = "3.0", features = ["full"] }
```
AudaciousAxiom, line 22: Alternatively, it should also be possible to allow both, to avoid bringing in the v3 in dependents that still depend on the v2 and don't (yet) require the v3:
```suggestion
syn = { version = ">= 2.0.52, < 4.0", features = ["full"] }
```

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/futures-rs/3028/4fe6c2070e57596bc0dcdd144797a716885e9596..8e47aa799704b51324543967f1e89e6b76b09b43)*

taiki-e (approved):
Thanks!

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/futures-rs/3028/4fe6c2070e57596bc0dcdd144797a716885e9596..8e47aa799704b51324543967f1e89e6b76b09b43)*

futures-macro/Cargo.toml
taiki-e: I prefer to avoid that because Cargo's semver-incompat versions handling sometimes behave oddly.

taiki-e:
Published in 0.3.34.

PR #3033, opened 2026-08-04
`futures-io`: Use nightly-only `core::io`

bushrat011899:
# Description

Implements `no_std` support behind a new rustflag `futures_unstable_nightly_core_io` by relying on the unstable nightly `core_io` feature. This allows targets such as `wasm32v1-none` to use the async traits in this crate. Please see [rust-lang/rust#154046](https://github.com/rust-lang/rust/issues/154046) for details on tracking issue.

---

## Notes

* No AI tooling of any kind was used during the making of this PR.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -56,8 +56,12 @@ jobs:
             target: i686-unknown-linux-gnu
           - os: ubuntu-latest
             target: s390x-unknown-linux-gnu
+          - os: ubuntu-latest
+            rustflags: "--cfg futures_unstable_core_io"
     runs-on: ${{ matrix.os }}
     timeout-minutes: 30
+    env:
+      RUSTFLAGS: ${{ env.RUSTFLAGS }} ${{ matrix.rustflags }}
     steps:
       - uses: taiki-e/checkout-action@7d1e50e93dc4fb3bba58f85018fadf77898aee8b # v1.4.2
       - name: Install Rust
@@ -171,7 +175,7 @@ jobs:
       - run: cargo minimal-versions build --workspace --ignore-private --all-features
 
   no-std:
-    name: cargo build --target ${{ matrix.target }}
+    name: RUSTFLAGS=${{ matrix.core_io }} cargo build --target ${{ matrix.target }}
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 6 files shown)


bushrat011899:
I'm _pretty_ sure the Miri failure is unrelated to this PR?

futures-io/src/lib.rs
```diff
@@ -18,20 +18,27 @@
     clippy::std_instead_of_core
 )]
 #![cfg_attr(docsrs, feature(doc_cfg))]
+#![cfg_attr(futures_unstable_nightly_core_io, feature(core_io))]
```
taiki-e: I feel that "unstable" and "nightly" overlap. I think "futures_unstable_core_io" would be fine.

taiki-e (changes requested):
Thanks for the PR! Could you add a build with this cfg enabled to CI?

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/futures-rs/3033/c738ddedae1a6a8ec74d54d80d3c3367fd43fccb..798aaba2b98548f3455a1ff5593b4635ab9d22af)*

futures-io/src/lib.rs
bushrat011899: Yeah that's reasonable, I've swapped to that flag name instead.

bushrat011899:
> Thanks for the PR! Could you add a build with this cfg enabled to CI?

Can do! I've added it in two jobs:

1. As an extra matrix entry to the `test` job to ensure the code still *works* the same way.
2. As an extra dimension to the `no_std` job matrix, so we know it's actually `no_std` compatible with and without the flag.

For the latter, I've added `futures-io` to the `no_std` example crate.

.github/workflows/ci.yml
```diff
@@ -56,8 +56,12 @@ jobs:
             target: i686-unknown-linux-gnu
           - os: ubuntu-latest
             target: s390x-unknown-linux-gnu
+          - os: ubuntu-latest
+            rustflags: "--cfg futures_unstable_core_io"
     runs-on: ${{ matrix.os }}
     timeout-minutes: 30
+    env:
+      RUSTFLAGS: ${{ matrix.rustflags }}
```
taiki-e: This overrides the top-level RUSTFLAGS.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/futures-rs/3033/fbcc281f361d2d39b6d287a11e85676d50c2e9b4..cbbd853471ca0706a6d5186aecbb1c67b602a18a)*

.github/workflows/ci.yml
```diff
@@ -180,8 +184,13 @@ jobs:
         target:
           - thumbv7m-none-eabi
           - thumbv6m-none-eabi
+        core_io:
+          - "--cfg futures_unstable_core_io"
+          - ""
     runs-on: ubuntu-latest
     timeout-minutes: 30
+    env:
+      RUSTFLAGS: ${{ matrix.core_io }}
```
taiki-e: Same here.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/futures-rs/3033/fbcc281f361d2d39b6d287a11e85676d50c2e9b4..cbbd853471ca0706a6d5186aecbb1c67b602a18a)*

taiki-e (changes requested):

